PR #41, opened 2014-10-27
An alternative Lock Implementation

aamirr:
Implement a complete alternative to the OntoLock using the same interface
Fix the problem of releasing lock
Unit testing the new implementation
https://oat-tao.atlassian.net/browse/TP-281


jbout:
Deprecated due to new lock interface

PR #93, opened 2015-01-06
don't use index to remove a property

antoinerobin:


jbout:
deprecated by https://github.com/oat-sa/tao-core/pull/124

PR #107, opened 2015-01-20
Fix index issues

jbout:


jbout:
Visually confirmed by Lionel

PR #111, opened 2015-01-22
highlight the right root in filebrowser

antoinerobin:
allow subtree close only if the root is already selected


views/js/ui/resourcemgr/fileBrowser.js
```diff
@@ -26,6 +26,8 @@ define(['jquery', 'lodash'], function($, _) {
             }
             updateFolders(content, $innerList);
             //internal event to set the file-selector content
+            $('.file-browser').find('li.active').removeClass('active');
```
krampstudio: `$('.file-browser')` is not scoped. Don't you have a container to do a `$('.file-browser', $container)` to prevent side effects (of there are multiple instances of the component or a name collision)

PR #123, opened 2015-02-04
Created Lock abstraction

jbout:


models/classes/lock/LockManager.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * 
+ */
+class LockManager {
+    
+    const CONFIG_ID = 'lock';
+    
+    private static $implementation = null;
+
+    static public function setImplementation($implementation) {
```
llecaque: Shall we assert that the params type as the Lock Interface + phpdoc to know if it is string or the object 

static public function setImplementation(tao_models_classes_lock_Lock $implementation) {

PR #124, opened 2015-02-05
Class properties refactoring

antoinerobin:


views/js/form/property.js
```diff
@@ -38,10 +38,7 @@ define(['jquery', 'lodash', 'form/post-render-props'], function ($, _, postRende
                 success: function (response) {
                     response = $(response);
```
krampstudio: Could you inline that, to prevent side effects:

```javascript
postRenderProps.init($(response).last());
```

views/js/uiForm.js
```diff
@@ -130,10 +130,101 @@ define([
...
             var that = this;
-            $(".form-submitter").off('click').on('click', function () {
+            $(".form-submitter").off('click').on('click', function (e) {
+                e.preventDefault();
                 var myForm = $(this).parents("form");
-                if (that.submitForm(myForm)) {
-                    myForm.submit();
+
+                if($('[id="tao.forms.class"]').length !== 0){
+                    var formData = {};
```
krampstudio: javascript doesn't have block scope variable, they are scoped to the function (and hoisted).
You should declare them before the use : 
```javascript
$(".form-submitter").off('click').on('click', function (e) {
    var myForm, 
          formData, 
          clazz;

     e.preventDefault();
     myForm = $(this).parents("form");
     if($('[id="tao.forms.class"]').length !== 0){
           formData = {};
           clazz = {};

...
});

File: views/js/uiForm.js [hunk 3]
```diff
@@ -130,10 +130,101 @@ define([
...
+                        });
+                        //get data for each index
+                        var indexes = [];
+                        $(':input.index', this).each(function(){
+
+                            var i;
+                            var found = false;
+                            var name = '';
+                            for(i in indexes){
+                                if(indexes[i] && $(this).attr('data-related-index') === indexes[i].uri){
```
krampstudio: assign `$(this)` to a variable or chain the calls. Wrapping an element in jquery does a lot of calculations, so each time you do a `$(this)` you repeat unnecessary calculions.

llecaque:
save range broken in advance mode

krampstudio:
client side syntax seems correct.

llecaque:
still issue on add properties in Advance Mode, when changing properties label

helpers/form/elements/xhtml/class.Treeview.php
```diff
@@ -151,8 +151,10 @@ public function render()
     				callback:{
 	    				onload: function(TREE_OBJ) {
 	    					checkedElements = '.json_encode($this->values).';
+                            var tree = $("#'. $widgetTreeName.'");
 	    					$.each(checkedElements, function(i, elt){
-								NODE = $("li[id=\'"+elt+"\']");
+								NODE = $("li[id=\'"+elt+"\']", tree);
+								console.log(NODE);
```
llecaque: console.log ?

llecaque:
Now we could not save in simple mode

PR #126, opened 2015-02-06
debounce the keyup callback to give user a chance to rectify

ssipasseuth:
debounce the keyup callback to give user a chance to rectify an invalid input + allow negative value input

### DIFF
--- a/views/js/ui/incrementer.js
+++ b/views/js/ui/incrementer.js
@@ -111,7 +111,7 @@ define(['jquery', 'lodash', 'core/pluginifier'], function($, _, Pluginifier){
                                     }
                                 }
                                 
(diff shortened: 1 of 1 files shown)


views/js/ui/incrementer.js
```diff
@@ -98,7 +111,7 @@ define(['jquery', 'lodash', 'core/pluginifier'], function($, _, Pluginifier){
                                     }
                                 }
                                 
-                            })
+                            }, 1000))
```
krampstudio: This delay seems to long

ssipasseuth: I reduce it to 600 ms. What do you think of that ?

PR #131, opened 2015-02-11
Index properties

antoinerobin:


jbout:
Reopened to determine the cause of recreated .css files. Was linked to the scss preprocessing launched by the file-watcher in the middle of the merge process.

PR #143, opened 2015-02-24
TAO-351 : Add the new style for the extended text incrementer

clawfire:
Fix all the bugs for TAO-351.


krampstudio:
this should be part instead of the qti creator css

clawfire:
Right. Let's drop this one

PR #151, opened 2015-02-27
TAO-392: Display lock on read-only resources in tree

hutnikau:
tree

Readonly nodes labeled with yellow `warning` icon and locked nodes labeled with red `lock` icon.


krampstudio:
I'm fine with the code except for the use ES5 features. Even though they are supported in our targeted browser our approach is to use lodash whenever possible. The reason is the ES5 implementations may have little difference in the behavior (particularly when something goes wrong, like with null, undefined or empty values). Also lodash is performance oriented, so it uses the best approach on each browser and provide a consistent behavior.
Please replace the occurences of `Array.indexOf` and `Array.filter` by `_.indexOf` or `_.contains` and `_.filter`

krampstudio:
You should also change the actionManager `tao/views/js/layout/actions.js`. The actions has to be hidden or disabled reagrding the permission (see `updateContext` or `updateState`).

hutnikau:
@krampstudio  I checked updating state of actions (hiding and showing controls depending on the permissions). It think works fine. Could you tell me please what needs to change in the `updateContext` and `updateState`?

krampstudio:
Selet a locked resource (no read nor write permission), then you've got the the "Properties" action available and if you click on it, then you're kicked off the platform.

hutnikau:
Yes, you're right. To prevent this behavior I've added permission rules (`@requiresRight` tag) into [taoTest](https://github.com/oat-sa/extension-tao-test) extension (`taoTests_actions_Tests::editTest()` and `taoTests_actions_Tests::authoring()` methods). Also I added hiding permission error message after the allowed item has been selected (sometimes it was bit confusing).

krampstudio:
Seems ok to me. I'll let @jbout to check the features.

PR #152, opened 2015-03-02
Lock component

antoinerobin:
TAO-464


views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
+/**
```
krampstudio: THE GPL header is missing

views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
...
+ * @author Antoine Robin <antoine.robin@vesperiagroup.com>
+ */
+define([
+    'jquery', 
+    'lodash',
+    'tpl!ui/lock/lock'
+], function($, _, tpl){
+
```
krampstudio: missing0 `'use strict';` statement

views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
...
+    //keep a reference to alive lock
+    var current = null;
+
+    //contains the reference to the main lock box. We expect other containers to be only edge cases.
+    var $lockBox;
```
krampstudio: these 2 lines makes unable to have multiple lock containers and multiple lock on the same page (until you refresh your browser or change extension). Is it the required behavior ?

views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
...
+        return receiver;
+    }
+
+    /**
+     * It provides the lock behavior
+     * @typedef lockApi
+     */
+    var lockApi = {
+
+        level : null,
```
krampstudio: Document the API using jsdoc. Particularly when you use `options` like in the `message` method.

views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
...
+        display : function display(){
+            var self = this;
+            if(this.content){
+                this.setState(states.displayed);
+
+                $(this.content)
+                    .attr('id', this.id)
+                    .appendTo(this._container);
+
+                this._trigger();
```
krampstudio: trigger what ?

views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
...
+            };
+        });
+        return receiver;
+    }
+
+    /**
+     * It provides the lock behavior
+     * @typedef lockApi
+     */
+    var lockApi = {
```
krampstudio: Are you sure you want to change the display level of the messages ?

views/js/test/ui/lock/test.js
```diff
@@ -0,0 +1,197 @@
+define(['jquery', 'ui/lock'], function($, lock){
```
krampstudio: We prefer now the long notation for unit test : 

```javascript
QUnit.module('lock');

QUnit.test('module',  1, function(assert){
    assert.ok(typeof lock === 'function', 'The module expose a function');
});
```

views/js/ui/lock.js
```diff
@@ -0,0 +1,244 @@
...
+                    .appendTo(this._container);
+
+                this._trigger();
+
+            }
+            return this;
+        },
+
+        release : function release(){
+            var self = this;
```
krampstudio: You should check the value of the URL. I'm wondering if the request will be sent to the page URL (ie. `window.location`) with an empty URL.

views/js/ui/lock.js
krampstudio: document events when they are infered: 

```javascript
/**
 * ...
 * @fires display.lock
 */
```

krampstudio:
I've just realized by running the test that `caller` is deprecated in strict mode and may raise an error on some browser. Could you remove it from `_trigger` and add the eventName instead. (We will have the same issue in the feedback component)

PR #155, opened 2015-03-03
form wrapper now automagically

dietertaotesting:

PR #156, opened 2015-03-03
Ck underline

dietertaotesting:
most of this is removing legacy copies of ckConfigurator etc and generated css. Mind you that this contains a file from upload, that should have been another commit


krampstudio:
Seems OK. I'll wait your modifications on #155  to merge this one to avoid conflicts.

PR #157, opened 2015-03-03
Layout customization

dietertaotesting:


dietertaotesting:
a config file should look about like this:
<pre>// array _must_ have the following keys
return array(
    'version-type' => '',
    'is-unstable'  => false,
    'is-sandbox'   => false,
    'logo'         => 'tao-logo.png',
    'link'         => 'http://taotesting.com',
    'msg'          => 'Tao Home'
);</pre>

llecaque:
first 3 commits should not be part of these pull request I suppose

PR #159, opened 2015-03-04
close the lock before opening another

antoinerobin:


views/js/ui/lock.js
```diff
@@ -184,17 +180,22 @@ define([
...
-                    .attr('id', this.id)
-                    .appendTo(this._container);
+                $(self.content)
+                    .attr('id', self.id)
+                    .appendTo(self._container);
 
-                this._trigger('display');
+                self._trigger('display');
+
+                $('#release').on('click',function(){
```
krampstudio: this is dangereous to use an id. What happens if there is another `#release` element. A best practice is to use a class and to target it from it's container:

```javascript
$('.release', self._container).on('click', function(){ ...
```

PR #165, opened 2015-03-05
readded ckeditor with tao custom floatingspace plugin

ssipasseuth:


ssipasseuth:
I removed a lot of unsued localization files so there may be a lot of deleted files.
The biggest change is the replacement of the minified ckeditor, that is built from https://github.com/oat-sa/ckeditor-dev/tree/tao-custom

PR #172, opened 2015-03-11
Added register for default lock behaviour

jbout:


views/js/ui/lock.js
```diff
@@ -233,6 +236,26 @@ define([
             return this;
 
         },
+        
+        /**
+         * Default behaviour
+         */
+        register : function() {
+        	var msg = this._container.data('msg');
+        	var id = this._container.data('id');
```
antoinerobin: you can use `var id = this._container.data('id') || defaultOptions.uri;` instead to ensure that id and msg are not undefined

and add a default options for msg
 ```
var defaultOptions = {
        msg : __('This resource is locked'),
```

PR #174, opened 2015-03-11
removed hard coded unlock msgs

jbout:


views/js/ui/lock.js
```diff
@@ -222,13 +222,13 @@ define([
...
-                            self._trigger('released');
+                            self._trigger('released', response);
                         }
                         else{
-                            self._trigger('failed');
+                            self._trigger('failed', response);
                         }
                     },
-                    error : function(){
+                    error : function(what){
```
antoinerobin: are you sure of the name `what` ? you can keep an empty args list

PR #178, opened 2015-03-12
Feature/unlock manager

jbout:


llecaque:
@jbout could you merge conflict actions/class.Lock.php

clawfire:
✅ for me but man ... stop making html like my 👵. Don't use `<table>` that much next time ;)

PR #179, opened 2015-03-12
encode messages in footer

krampstudio:
Fix messages with a quotes crashing the app.


ssipasseuth:
Can you please tell me an easy way to reproduce the original bug "craching the app" so I can test it ? Thanks!

PR #182, opened 2015-03-12
TAO-532 : Import the new 4.4.7 build of ckeditor

clawfire:
Plugins not included in this update event if the `config-build` file is require them ( Please @ssipasseuth come talk w/ me to check this ) : 

- codesnippet
- codesnippetgeshi
- image
- table
- tableresize
- tabletools


ssipasseuth:
About the no included plugins, this is strange ... can you tell me which tag/branch of tao/ckeditor-dev you are building from ?
By the way, are they in use ? The last thing I remember from a plugin like a image and table is they generate such messy html that we (@llecaque and I) preferred to deactivate (unless in the qti creator).

clawfire:
I agree and we ( you :) ) built your own qtiImage plugins that are indeed included so ... I build from tag 4.4.7 . My advice is : it's looks like working without them, so we can remove them from the build config and include them maybe later if we need them.

ssipasseuth:
I rebuild the ckeditor and removed the taomediamanaer from extendedTextInteraction

ssipasseuth:
this is related to https://github.com/oat-sa/extension-tao-itemqti/pull/158

PR #189, opened 2015-03-17
don't restrict access if no permissions are defined

jbout:


views/js/layout/tree.js
```diff
@@ -530,7 +522,7 @@ define([
                     addClassToNode(node, 'permissions-full');
                 } else if (containsTrue && containsFalse) {
                     addClassToNode(node, 'permissions-partial');
-                } else {
+                } else if (containsFalse) {
```
krampstudio: i would rather do somthing like
```javavscript
        var computeSelectionAccess = function(node){
            
            if(_.isArray(node)){
                _.forEach(node, computeSelectionAccess);
                return;
            } 
            if(node.type && node.permissions){
                var actions = _.pluck(_.filter(options.actions, function (val) {
                    return val.context === node.type || val.context === 'resource';
                }), 'id'),
                    keys = _.intersection(_.keys(node.permissions), actions),
                    values = _.filter(node.permissions, function (val, key) {
                        return _.contains(keys, key);
                    }),
                    containsTrue = _.contains(values, true),
                    containsFalse = _.contains(values, false);
            
                if (containsTrue && !containsFalse) {
                    addClassToNode(node, 'permissions-full');
                } else if (containsTrue && containsFalse) {
                    addClassToNode(node, 'permissions-partial');
                } else {
                    addClassToNode(node, 'permissions-none');
                }
                
                if (!hasAccessTo('moveInstance', node)) {
                    addClassToNode(node, 'node-undraggable');
                }
            }
            if(node.children){
                _.forEach(node.children, computeSelectionAccess);
            }
        };
```

views/js/layout/tree.js
```diff
@@ -543,6 +521,24 @@ define([
             }
         };
 
+		function getPermissionClass(node){
```
krampstudio: could you just add the jsdoc:

```javascript
/**
 * Get the CSS class to apply to the node regarding the computed persmissions
 * @param {Object} node - the tree node
 * @returns {String} the CSS class
 */

krampstudio:
Test done on result server :
- before, I got an error saying insufficient permissions and the red lock on the tree
- after, the tree is back to normal

PR #190, opened 2015-03-18
TAO-550: Context Menu does not take access rights into consideration

hutnikau:


actions/class.Main.php
```diff
@@ -268,41 +265,35 @@ private function getNavigationElementsByGroup($groupId)
...
-     * @return boolean true if the user is allowed
+     * @return array menu elements list
      */
-    private function hasAccessToMenuElement(Perspective $menuElement)
+    private function getMenuElementChilds(Perspective $menuElement)
     {
-        $binding = $menuElement->getBinding();
-        if (!is_null($binding) && !empty($binding)) {
-            return true;
+        $сhilds = array();
```
jbout: that character in $childs between '$' and 'hilds' is not a 'c' but the cyrilic es https://en.wikipedia.org/wiki/Es_(Cyrillic)

PR #191, opened 2015-03-19
TAO-435: QTI Item with PCI cannot export

hutnikau:
Before sending archive to the end user should be cleaned all levels of output buffering. Otherwise user may get corrupted file.


bugalood:
This will not be the entire solution to the problem, but it might help not having broken .zip archives.

PR #195, opened 2015-03-20
Fix wrong version of jquery in build config

krampstudio:


krampstudio:
It will require rebundling

PR #213, opened 2015-03-27
Gap match interaction

dietertaotesting:


views/js/ui/mediasizer.js
```diff
@@ -242,6 +334,11 @@ define([
                 return;
             }
 
+            if(!options){
+                console.log($elt, $elt.data(dataNs), dataNs, arguments)
```
krampstudio: forgotten console.log

views/js/ui/mediasizer.js
```diff
@@ -115,19 +168,54 @@ define([
...
             var options = $elt.data(dataNs),
                 $mediaSizer = $elt.find('.media-sizer'),
-                self = this;
-            $elt.find('.media-sizer-link').on('click', function () {
+                self = this,
+                $syncBtn = $elt.find('.media-sizer-sync');
+
+            if(!options.showSync) {
+              //  $syncBtn.hide();
+                $elt.find('.media-sizer').addClass('media-sizer-sync-off');
```
krampstudio: your should reuse `$mediaSizer`

krampstudio:
There is an issue on adding more than one gap img. The gap images are replaced, duplicated and the number increase even though we remove them.

PR #219, opened 2015-03-31
Fix/tree action urlencoding

jbout:


krampstudio:
I'm fine with this PR but I'll look before merging it if there are no better place to decode the actionContext data, otherwise we will have to decode in each new action.

PR #220, opened 2015-03-31
Tao 576 feedback overflow

dietertaotesting:
tao-576, for testing refer to readme in JIRA comment.
Also added license. Could not make JSHint compliant, see inline comments

### DIFF
--- a/views/js/ui/feedback.js
+++ b/views/js/ui/feedback.js
@@ -23,8 +23,9 @@
 define([
     'jquery', 
     'lodash',
+    'util/wrapLongWords',
     'tpl!ui/feedback/feedback'
-], function($, _, tpl){
+], function($, _, wrapLongWords, tpl){
 
     //'use strict';
     // @todo cannot be used here because _trigger() relies on arguments.caller!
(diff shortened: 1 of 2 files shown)


views/js/ui/feedback.js
```diff
@@ -87,7 +141,7 @@ define([
 
             this.content  = tpl({
                 level : level,
-                msg : msg
+                msg : options.wrapLongWords ? warpLongWords(msg) : msg
```
krampstudio: a typo : warpLongWords

dietertaotesting: How can you tell I don't want to warp the words too? U+1F596!

krampstudio:
Shouldn't you use the CSS3 property `word-break` ? It has a good support for the `break-all` value and seems to work out of the box (tried with your test on firefox and chrome). It would save us from  style related JS.

dietertaotesting:
I played the likes of &lt;wbr&gt;, word-break, &shy; and so on and nothing did exactly what I needed in all my test scenarios. The downside of [word-break](https://css-tricks.com/almanac/properties/w/word-break/) is that it can break of *any* word - not just long ones. The particular case I was fixing is so super-edgy-edgy that in 90% of all cases nothing at all will happen.

krampstudio:
Ok, but could you then move the function `wrapLongWords` to a util module `tao/views/js/util/`

PR #221, opened 2015-04-01
TAO-479 vertical scrollbar on tree now independent from horizontal one

dietertaotesting:
Is the CSS format alright like that?


krampstudio:
The CSS format seems to be the same.

PR #224, opened 2015-04-02
TAO-577 Pressing the enter button within a form ...

dietertaotesting:
...no longer submits the form under any circumstances. Beyond this task I removed some legacy code pieces and upgraded the format. Also confirmed that the removed function initNav() is not used anywhere else in TAO. this.initFormPattern was defined but never used


krampstudio:
When you hit enter in a form, you're now redirected to the content page. The submit event should be intercepted.

views/js/uiForm.js
```diff
@@ -24,19 +43,26 @@ define([
         encode
         ) {
 
-    function getUrl(action) {
-        var conf = module.config();
-        return context.root_url + conf.extension + '/' + conf.module + '/' + action;
-    }
+    'use strict';
+        /*jshint camelcase: false */
```
krampstudio: Please do not hide lint issues. I prefer to have it referenced.

views/js/uiForm.js
```diff
@@ -128,144 +144,145 @@ define([
...
-                        else{
-                            formData[name] = $global.val();
-                        }
-                    });
-                    if(clazz.length !==0){
-                        formData.class = clazz;
-                    }
+            //for backward compatibility
+            if (!$('[id="tao.forms.class"]').length) {
+                return undefined;
```
krampstudio: use `return;` instead

views/js/uiForm.js
```diff
@@ -128,144 +144,145 @@ define([
                                                                     .find('[id*="ns_filter"]')
                                                                     .addClass('btn-default small');
             }
+
+            $('body').off('submit').on('submit', '.xhtml_form form', function (e) {
```
krampstudio: You should do `off` on the exact same element, otherwise you'll unbind all submit events on body (it may have side effects):
```javascript
            $('body')
              .off('submit', '.xhtml_form form')
              .on('submit', '.xhtml_form form', function (e) {
                e.preventDefault();
                var $form = $(this),
                    formData = self.getFormData($form);
                return self.submitForm($form, formData);
            });
```

krampstudio:
except the style issue the fix appears to be working on FF and Chrome.

boajer:
PR is ready for review

PR #235, opened 2015-04-07
Fix/class label encoding

jbout:


bugalot:
We found an issue with this one. Create a list with funky chars (eg &, <, >...) and values with funky chars for list values. Go in the class editor, and create a new property. Select this new list with as the possible values. You will see &quote, &amp, ... instead of &, < ...

PR #236, opened 2015-04-07
add initTree action

hutnikau:
I need to execute action after tree initialized.
`structures.xml` example:
```
<trees>
    <tree name="My library"
        initTree="init-tree"
    />
    </trees>
        <actions>
            <action id="init-tree" name="init" url="/myExt/MyController/index">
            </action>
            ...
```


krampstudio:
I've merged it, but consider also updating the structures.xml DTD `tao/doc/structures.dtd` 
I would also have call the attribute `init` instead of `initTree` (it can be part of another PR).

PR #248, opened 2015-04-09
Escapes -\/\\^$*+?.()|[\]{} inside a string

dietertaotesting:
Use when concatenating a string for usage in new RegExp().


jbout:
Comparison of alternative approaches: http://jsperf.com/wraplongwords

krampstudio:
Ideally this kind of util should land with their own unit test.

PR #256, opened 2015-04-10
Init tree action

hutnikau:


hutnikau:
Bertrand, thank you for accepting previous PR.
Here I've made changes which you described here: https://github.com/oat-sa/tao-core/pull/236#issuecomment-91479696

krampstudio:
Is the init action already implemented in an extension ?

PR #261, opened 2015-04-10
Refactor/property index authoring

antoinerobin:
will generate conflicts but you just have to get files from this branch ... normally


jbout:
merge of devel into zendsaerch is not recognised as a merge, so this branch cannot be merged

PR #280, opened 2015-04-20
Tao 684

llecaque:


pplichart:
fixed 
-  Impossible to  edit else content part of a modal dialog
-  The edition dialog for the then part is greyed down, looks disabled --> css
-  None of the dialogs are shown in the preview (if >20 ... sort of condition )

PR #281, opened 2015-04-21
TAO-668	-media element wrapped in AMD-style

boajer:
related to https://github.com/oat-sa/extension-tao-itemqti/pull/204


krampstudio:
If `views/js/lib/mediaelement/mediaelement-and-player.min.js` isn't used anymore could you please remove it.

PR #284, opened 2015-04-21
TAO-479 horizontal scroll failed at random on Chrome. 

dietertaotesting:
Rewrote whole height calculation, also no longer scroll whole page but panel only


views/js/layout/section-height.js
```diff
@@ -1,95 +1,124 @@
...
+    function getTreeActionIdealHeight() {
+        // we need at least four actions to have a two-row ul
+        var $treeActions = $('.tree-action-bar-box'),
+            $treeActionUl = $treeActions.find('ul'),
+            liNum = $treeActions.find('li:visible').length || 0,
+            idealHeight;
+
+        while (liNum < 5) {
+            $treeActionUl.append($('<li class="dummy"><a/></li>'));
+            liNum++;
```
ssipasseuth: Isn't there a better way to do this ? like take an existing li outerwidth and multiply by five.
Here you are actually adding (so update + render) dummy DOM elements....

views/js/layout/section-height.js
```diff
@@ -1,95 +1,124 @@
...
-                $treeActionUl.append($('<li class="dummy"><a/></li>'));
-                liNum++;
-            }
-            idealHeight = $treeActions.outerHeight() + parseInt($treeActions.css('margin-bottom'));
-            $treeActionUl.find('li.dummy').remove();
-            return idealHeight;
+
+    var $versionWarning = $('.version-warning'),
+        $window = $(window),
+        $footer = $('body > footer');
```
ssipasseuth: I find it dangerous to assign reference to jquery container here : are we sure that those elements will stay after another page loading. Isn't it safer to initialize them in some sort of init function ? Just wondering because we had many times issues when doing something like that.

ssipasseuth:
Sorry I don't see how this is related to the issue https://oat-sa.atlassian.net/browse/TAO-479.
Actually I cannot reproduce it and cannot see any difference when using this code. (I test on chrome)
I would need you assistance to understand this.

PR #295, opened 2015-05-05
add an xml template to preview them

antoinerobin:


views/js/ui/previewer.js
```diff
@@ -30,7 +31,7 @@ define(['jquery', 'lodash', 'i18n', 'core/mimetype', 'core/pluginifier', 'mediaE
          */
         generate: function generate(type, data) {
             var tmpl = this[type + 'Template'];
-            data.jsonurl = JSON.stringify(data.url)
+            data.jsonurl = JSON.stringify(data.url);
```
krampstudio: Why do we use `JSON.stringify(data.url);` instead of using `<video src='${url}'` directly. It looks a little bit weird to have some attributes that are quoted and others that aren't. 
This is also dangerous because if `data.url` is `undefined` you won't have the quotes.

antoinerobin: The `JSON.stringify` is used to deal with url containing ' or ". The final dom will look like ``so the file will not be correctly loaded. If template can handle this issue please let me know, it will certainly be better

krampstudio:
XML content should have html entities encoded (`<`, `>`, `&`) in order to display the tags in a `<pre>`, otherwise. 
See :

views/js/ui/previewer.js
krampstudio: For this issue, we will move at some point to handlebars (external) templates. Thoses one were there before we integrate it. I think handlebars manage escaping so we will have a cleaner implementation

PR #299, opened 2015-05-06
Delete node callback

hutnikau:


krampstudio:
Why did you delete this callback ?

hutnikau:
I meant create `ondelete` callback. I need be able track deleting nodes from the tree to remove it from calendar in case when delivery was deleted by pressing button under the tree.

views/js/layout/tree.js
```diff
@@ -120,7 +120,13 @@ define([
...
             //lifecycle callbacks
             callback: {
-
+                /**
+                 * Delete node callback.
+                 * @returns {undefined}
+                 */
+                ondelete: function () {
+                    $elt.trigger('delete.taotree', arguments);
+                },
```
krampstudio: Just some styling considerations : 
```javascript

 
                 /**
                 * Delete node callback.
                 * @fires layout/tree#ready.taotree
                 */
               ondelete: function ondelete () {
                    $elt.trigger('delete.taotree', [].slice.call(arguments));
                },
```

 - We prefer named function over anonymous functions.  
 - Events have to be documented
 - arguments isn't an array, so you need to cast it to avoid side effects.

hutnikau:
Done.

PR #303, opened 2015-05-08
Tao 727/shared stimulus import

antoinerobin:


views/js/ui/resourcemgr/fileSelector.js
```diff
@@ -45,9 +45,9 @@ define([
     return function(options){
 
         var root            = options.root || '/';
+        var disallowUpload  = options.disallowUpload || false;
```
ssipasseuth: maybe a more suitable name for this option : disableUpload ?

PR #304, opened 2015-05-08
Feature/check in button

antoinerobin:
depends on : https://github.com/oat-sa/extension-tao-revision/pull/22

### DIFF
--- a/views/js/ui/lock.js
+++ b/views/js/ui/lock.js
@@ -256,9 +256,9 @@ define([
         commit : function commit(){
             var self = this;
             if(self.options.commitUrl !== ''){
-                $('#message-container').slideToggle();
-                $('.commit').on('click',function(){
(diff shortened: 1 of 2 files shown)


views/js/ui/lock.js
```diff
@@ -240,6 +246,50 @@ define([
...
+        /**
+         * ask a message and call the url to commit the resource
+         * @example lock().commit();
+         * @fires commit.lock
+         * @returns {lockApi}
+         */
+        commit : function commit(){
+            var self = this;
+            if(self.options.commitUrl !== ''){
+                $('#message-container').slideToggle();
```
krampstudio: prefer classes to id or you need to be sure the element won't be in the same page (even in an hidden section) and scope every selector using the container, so it's safe. 
Apply for all selectors in this function.

PR #307, opened 2015-05-12
Loop icon

hutnikau:
Add loop icon:
loop2
For recurring events


dietertaotesting:
Please name the icons more verbose, spinner22 should be spinner for example. As for coloring schemes as used in the calendar please use in the future tao/views/scss/inc/_colors.scss as a reference.

PR #331, opened 2015-05-20
Fix/sourcemap urls

krampstudio:
Requires https://github.com/oat-sa/extension-tao-wfauthoring/pull/8
Related to https://github.com/oat-sa/extension-tao-itemqti/pull/265

It fixes https://oat-sa.atlassian.net/browse/TAO-716

The paths of source maps  of the javascript bundles in production mode were not corrects. (it can be seen in chrome's console).


krampstudio:
to test the bundles 
 - `cd tao/views/build`
 - `grunt bundleall` 
 - take a coffee

Test the pltaform with DEBUG_MODE to false in `config/generis.conf.php`

PR #335, opened 2015-05-22
All liststyler.js related scss, this essentially allows to apply all …

dietertaotesting:
…ol/ul list-styles to any other block element. The one in _header-footer.scss is a tiny unrelated bugfix.
Moved existing similar code from _base.scss to the new _list-style.scss. The deleted part in _select2.scss removes an earlier, but abandoned implementation of similar code.


krampstudio:
Could you separate the fix for the loader in another PR. This is something that may be part of a patch (cc @llecaque)

PR #337, opened 2015-05-22
Fixed TAO-742

dietertaotesting:


jbout:
Causes double scroll bars, but allows items to be added again

dietertaotesting:
I'll check the double scroll - did you re-compile the CSS?

PR #343, opened 2015-05-28
Refactor/cleanup getservice

jbout:


jbout:
wrong destination branch

PR #349, opened 2015-05-29
add a method to get MediaManagement sources

antoinerobin:


antoinerobin:
enable the item import (or anything else) to have a list of possible source into which write

PR #354, opened 2015-06-01
Fix TU for lib/history

jsconan:
Fix issue in URL matching when non standard port is used


krampstudio:
There's still an issue when you're on default ports : 

```
Expected: "https://tao.localdomain:/Main/index?structure=items&ext=taoItems&section=authoring"
Result:   "https://tao.localdomain/Main/index?structure=items&ext=taoItems&section=authoring"
```

jsconan:
Default ports issue fixed

PR #361, opened 2015-06-05
bad var call

llecaque:


jbout:
Missed this request and created a second pull request for it, so this branch is redundant

PR #362, opened 2015-06-05
Tao 796 contextual popup

ssipasseuth:
It has been initially added for https://oat-sa.atlassian.net/browse/TAO-796
But it is also required for https://oat-sa.atlassian.net/browse/TAO-874
And also for https://oat-sa.atlassian.net/browse/TAO-874


views/js/test/ui/contextualPopup/test.js
```diff
@@ -0,0 +1,161 @@
...
+    'use strict';
+    
+    QUnit.module('init popup');
+    
+    QUnit.test('with content string', function(){
+
+        var $container = $('#main-container');
+        var popup1 = contextualPopup($('#center1'), $container, {content : 'content 1'});
+        
+        QUnit.assert.equal(popup1.getPopup().length, 1, 'popup created');
```
krampstudio: The notation to use is 

```javascript
QUnit.test('with content string', function( assert ){
  assert.equal(popup1.getPopup().length, 1, 'popup created');
```

because it becomes the main notation in the next version of QUnit (so it will be easier for us to upgrade)

views/js/test/ui/contextualPopup/test.html
```diff
@@ -0,0 +1,41 @@
...
+                require(['test/ui/contextualPopup/test'], function(){
+                    QUnit.config.reorder = false;
+                    QUnit.start();
+                });
+            });
+        </script>
+    </head>
+    <body>
+        <div id="qunit"></div>
+        <main id="main-container">
```
krampstudio: Why isn't main container in `#qunit-fixture` ? 
I suggest you to run your test in the fixture and you can have a container outside for your visual test

views/js/ui/contextualPopup.js
```diff
@@ -0,0 +1,265 @@
...
+             */
+            isVisible : function(){
+                return $element.is(':visible');
+            }
+        };
+        //need to reposition the popup after this has been attached to the dom
+        popup.reposition();
+        return popup;
+    }
+
```
krampstudio: Why those functions have been externalised from the `popup` object ?

views/js/test/ui/contextualPopup/test.html
ssipasseuth: The visual test actual uses the same dom elements, should I duplicate it in the fixture ?

views/js/test/ui/contextualPopup/test.js
ssipasseuth: ok

views/js/ui/contextualPopup.js
ssipasseuth: This provides a sort of api to interact with your newly created popup object.

views/js/test/ui/contextualPopup/test.html
krampstudio: yes

ssipasseuth:
I have updated the code from the unit test, please check it when you have some time

PR #368, opened 2015-06-10
Imporove CVS mapping form class for make selected matched labels duri…

Alroniks:
### What does it do ?
I added some logic to CSVMappingForm, that check labels from csv file (imported) and existing properties and then try make selected same.

### Why is it needed ?
When user load file with labels it would be awesome set checked those which coincide.

### Related issue(s)/PR(s)
Task: https://oat-sa.atlassian.net/browse/TAO-835


test/form/CsvMappingFormTest.php
```diff
@@ -0,0 +1,47 @@
...
+            'class_properties' => $properties,
+            'ranged_properties' => array(),
+            'csv_column' => $csv_column,
+            tao_helpers_data_CsvFile::FIRST_ROW_COLUMN_NAMES => true,
+        );
+
+        $formContainer = new tao_models_classes_import_CSVMappingForm($data, $options);
+        $form = $formContainer->getForm();
+
+        $this->assertEquals('csv_select', $form->getElement('http_2_www_0_w3_0_org_1_2000_1_01_1_rdf-schema_3_label')->getEvaluatedValue());
```
jbout: test-case fails as abelabel matches label

Alroniks:
Rebased branch and fix mapping and test

Alroniks:
Created new pr, here branch is broken. https://github.com/oat-sa/tao-core/pull/423

PR #370, opened 2015-06-11
Media management

antoinerobin:


antoinerobin:
https://github.com/oat-sa/extension-tao-item/pull/102

jbout:
change to interface not absolutely nescessary

PR #378, opened 2015-06-17
Add method for HTML attribute value encoding

jsconan:


dietertaotesting:
can you please add a unit test?

jsconan:
TU just added!

PR #387, opened 2015-06-22
Fix format of DateType object

jbout:
Fix in dev: https://github.com/oat-sa/tao-core/pull/386


llecaque:
Will be merged at the end  of the sprint

PR #393, opened 2015-06-24
$_FILE should be empty so throw exception if it is the case (large fi…

antoinerobin:
…les)

### DIFF
--- a/helpers/class.Http.php
+++ b/helpers/class.Http.php
@@ -105,7 +105,9 @@ class tao_helpers_Http
     {
 
         // for large file, the $_FILES may be empty so see this before checking for other updates
-        if(intval($_SERVER['CONTENT_LENGTH'])>0 && count(self::getFiles())===0){
+        $limit = tao_helpers_Environment::getFileUploadLimit();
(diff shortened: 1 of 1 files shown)


helpers/class.Http.php
```diff
@@ -104,6 +104,11 @@ public static function getFiles()
     public static function getUploadedFile($name)
     {
 
+        // for large file, the $_FILES may be empty so see this before checking for other updates
+        if(intval($_SERVER['CONTENT_LENGTH'])>0 && count(self::getFiles())===0){
```
krampstudio: Is `$_SERVER['CONTENT_LENGTH']` the uploaded file size ? If yes, shouldn't you check :
```php
$limit = tao_helpers_Environment::getFileUploadLimit();
$contentLength = intval($_SERVER['CONTENT_LENGTH'])
if( $limit > 0 && $contentLength > $limit && count(self::getFiles())===0){
```
I assume limit and contentLength are using the same unit. 

I want to be sure this case is handled only for exceeded size upload.

PR #398, opened 2015-06-25
Remove problematic htmlentities function for accented characters

chazotte:
This fix removes the display issue that is shown with accented characters when warning about access denied.

Before: `"Acc&egrave;s refus&eacute;. Veuillez renouveler votre authentification !"`
After: `"Accès refusé. Veuillez renouveler votre authentification !"`

I don't know if such change couldn't led to a security issue.


krampstudio:
I guess this may also break some error messages that XML output... (Yes I know).

@llecaque @jbout It becomes urgent that we work on a real error and feedback management API between the server and the client, that we distinguish system errors, and user feedback, etc.

krampstudio:
I'll merge it or now, but the messaging system MUST get our attention at some point...

PR #409, opened 2015-07-02
Fix client lib registry test case

bursonic:


jbout:
While working, changing the constant of an extension is dangerous, as a failed test-case would prevent the reset of the constant

PR #415, opened 2015-07-03
Error messaging improved in DEBUG mode for searching

boajer:


jbout:
Debug mode specific error handling is dangerous as it cannot be tested using unit tests

PR #418, opened 2015-07-07
fix unit tests

krampstudio:
- Fix the failing tests (except `ui/autocomplete` and `lib/history`)
 - refactor the tests to use the object notation instead of global functions (in order to prepare the migration to QUnit 2)

Test it using 

```
grunt connect taotest
```

You should have 4 assertions failing (5 if develop isn't yet merged).

### DIFF
--- a/views/js/test/ui/lock/test.html
+++ b/views/js/test/ui/lock/test.html
@@ -3,23 +3,22 @@
     <head>
         <meta charset="utf-8">
         <title>Ui Test - Lock</title>
-        <base href="../../../../" />
-        <link rel="stylesheet" type="text/css" href="js/lib/qunit/qunit.css">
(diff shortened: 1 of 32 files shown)


views/js/test/ui/lock/test.html
```diff
@@ -3,23 +3,22 @@
     <head>
         <meta charset="utf-8">
         <title>Ui Test - Lock</title>
-        <base href="../../../../" />
```
ssipasseuth: is it not a good practice to use <base> ?

krampstudio: I discard it when the the tests that needs to mock XHR calls this test simulate JSON responses by calling files on the same folder.

PR #425, opened 2015-07-09
Convert values to strings

hutnikau:


ssipasseuth:
Hi, sorry for the late response. It seems that that modification now breaks the unit test http://tao.localdomain/tao/views/js/test/core/validator/validate/test.html could check it out please ?

ssipasseuth:
sorry my mistake. I didn't merge develop. Please forget what I just said !

PR #447, opened 2015-07-22
Add recursive checking

hutnikau:


jbout:
Does not address the actual errorhandling

PR #450, opened 2015-07-24
Format date iso8601 with microseconds

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1106

https://github.com/oat-sa/extension-tao-act/pull/20
https://github.com/oat-sa/extension-tao-outcomeui/pull/33


jbout:
Please 
* Add the new Format to IntlFormatter
* Add testcases for the new format (for both formatters)
* Add testcases for the use of microseconds with the old formats

PR #455, opened 2015-08-04
Fix/tao 1170 improper unicode

bugalot:
This also prevents the platform to crash on PHP 5.3 :roller_coaster:


bugalot:
Reassigned to @antoinerobin  because @jbout is ultra busy.

PR #457, opened 2015-08-04
TAO-1169 It's now possible to attach resources using CSV import. It s…

bugalot:
…hould be possible in the past but it was (simply) buggy.


bugalot:
Please see issue "TAO-1169" in JIRA.

PR #461, opened 2015-08-06
Add checkboxes on datatable

jsconan:
Add option: 'selectable'
Add option: 'data'
Add events: query.datatable, beforeload.datatable
Provide context parameters on events: query.datatable, beforeload.datatable, load.datatable

You can run the unit test using either your browser or the CLI:
- `/tao/views/js/test/ui/datatable/test.html`
- `grunt connect qunit:single --test /tao/views/js/test/ui/datatable/test.html`


views/js/ui/datatable/tpl/layout.tpl
```diff
@@ -1,50 +1,96 @@
...
+                                            {{#with ../../this}}
+                            <button class="btn-info small {{id}}"{{#if title}} title="{{title}}"{{/if}}><span class="icon-{{#if icon}}{{icon}}{{else}}{{id}}{{/if}}"></span> {{label}}</button>
+                                            {{/with}}
+                                        {{/unless}}
+                                    {{/with}}
+                                {{else}}
+                            <button class="btn-info small {{@key}}"><span class="icon-{{@key}}"></span> {{@key}}</button>
+                                {{/if}}
+
+                            {{/each}}
```
ssipasseuth: Open-ended question : this loop is quite complex, are we not going beyond what handlebars is supposed to do ? Won't it be cleaner (so for later code maintenance) to move the logic to the controller and to have more simpler template here ?

views/js/ui/datatable/tpl/layout.tpl
```diff
@@ -1,50 +1,96 @@
...
-        <button class="btn-info small datatable-forward">{{__ 'Next'}}<span class="icon-forward r"></span></button>
+            </thead>
+            <tbody>
+                {{#data}}
+                    <tr data-item-identifier="{{id}}">
+                        {{#if ../selectable}}
+                        <td class="checkboxes"><input type="checkbox" name="cb[{{id}}]" value="1" /></td>
+                        {{/if}}
+                        {{#each ../model}}
+                        <td class="{{id}}">{{property id ../this}}</td>
```
ssipasseuth: Do we have to add a new helper "property" ? Isn't it the same as {{#with ../this}}{{id}}{{/with}} ?

views/js/ui/datatable.js
```diff
@@ -245,47 +443,124 @@ define([
...
-                    // If I never sort by this element or
-                    // I sort by this element & the order was desc
-                    options.sortorder = 'asc';
+
+            if (typeof asc !== 'undefined') {
+                switch (asc) {
+                    case 'asc': break;
+                    case 'desc': break;
+                    default: asc = (!!asc) ? 'asc' : 'desc';
                 }
```
ssipasseuth: why using a switch here and not a simple if ?

views/js/ui/datatable/tpl/layout.tpl
jsconan: The reason why is because I did not want to redevelop the full component, with the risk of breaking existing code. I know this code is very complex, but this is the only way to access an external property with this version of Handlebars.

views/js/ui/datatable/tpl/layout.tpl
jsconan: Unfortunately, this is not the same.

The `property` block lookups the value of a particular field through another context by using the value of an in-context field.

A simpler explication can be made by code comparison.

Assume we have this data:

```javascript
var model = [{id: 'value'}, {id: 'name'}] ;
var data = [{id: 1, value: 10, name: 'x'}]
```

The code `{{#with ../this}}{{id}}{{/with}}`, in this context can be:

```javascript
return data[i].id;
```

While `{{property id ../this}}` can be:

```javascript
var key = model[j].id;
return data[i][key];
```

views/js/ui/datatable.js
jsconan: Ok, my mistake. While developing, this code was more complex before, and when I simplified it I did not transform the `switch` to `if`.

views/js/ui/datatable/tpl/layout.tpl
ssipasseuth: Ok, I see. Thanks for the explanation.

views/js/ui/datatable/tpl/layout.tpl
ssipasseuth: Thanks for the example and the explanation.

PR #464, opened 2015-08-07
fixed TAO-1163 underline styling

ssipasseuth:


ssipasseuth:
requires https://github.com/oat-sa/ckeditor-dev/pull/4

PR #465, opened 2015-08-07
New install Ext script

jbout:


antoinerobin:
related to https://github.com/oat-sa/generis/pull/100

PR #475, opened 2015-08-12
Keep the set state to avoid multi ajax calls

jsconan:
This patch avoid the test runner to send several requests to store the same state each time an item is validated.

### DIFF
--- a/views/js/serviceApi/StateStorage.js
+++ b/views/js/serviceApi/StateStorage.js
@@ -1,4 +1,22 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


views/js/serviceApi/StateStorage.js
```diff
@@ -19,6 +37,7 @@ define(['jquery'], function($){
                     callback();
             }
         } else {
+            this.state = state;
```
ssipasseuth: But won't that mean that new state won't be saved to the server ? Is it the expected behaviour ?
Maybe I am missing something but I don't even understand the use of the "if" statement from line 35 to 38. It seems that it has 2 modes (one when the state is given in the constructor then it will never change) another when it is not given (so undefined) then get will always return undefined, right ? Isn't it weird ??

jsconan: In my understanding, when the state is set, the callback function is always called. But if the set state is different from the inner one, an AJAX request is also made to store it server side.

The state is always provided as a stringified value. So, a difference is immediately seen.

The purpose of my patch is to avoid the client sends an AJAX request each time the state is set if there is no change.

Currently each time an item is validated, there is several identical requests sent to store the state, as the setState method is called from many locations inside the code (indirectly?). But maybe it is a wanted behavior.

ssipasseuth: Yes, I thing it makes sense.
Thanks again for the clarification. :)

PR #478, opened 2015-08-12
Wrong translation when the key contains parameters

jsconan:
When a translation contains parameters, the `translate`method was using the source as a translation instead of the localized version...


krampstudio:
Merged but  a unit test should be made

PR #487, opened 2015-08-14
neutral prefix, test moved

dietertaotesting:


krampstudio:
Could you link the PR that removes the test from taoQtiItem

dietertaotesting:
connected to https://github.com/oat-sa/extension-tao-itemqti/pull/400

PR #488, opened 2015-08-14
fixed even binding for elements used as submitters

boajer:


krampstudio:
it works but the data are sent multiple times

krampstudio:
You should add `e.prenventDefault()`

PR #490, opened 2015-08-17
Tao 1009 hierarchy as metadata

boajer:


boajer:
depended on https://github.com/oat-sa/extension-tao-backoffice/pull/17

PR #495, opened 2015-08-17
prevented ambiguity interpretation during file names transcoding

boajer:


jbout:
Changes the name of uploaded files everywhere in Tao, but does not fix the rendering issue on imported items.

PR #502, opened 2015-08-20
Fix modal animate callback when not visible

jsconan:
The close and open actions were not called when animate is enabled and the modal is hidden.

Now, when the modal is hidden, the animate is not used.

### DIFF
--- a/views/js/ui/modal.js
+++ b/views/js/ui/modal.js
@@ -71,7 +71,7 @@ define(['jquery', 'core/pluginifier', 'core/dataattrhandler'], function ($, Plug
                 if ('number' !== typeof options.animate) {
                     options.animate = defaults.animate;
                 } else {
-                    options.animate = Math.min(animateDiff, options.animate);
+                    options.animate = Math.max(animateDiff, options.animate);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


krampstudio:
I have the feeling that the open animation isn't called anymore. You can see that by clicking the _?_ on the backoffice toolbar (it's not that clear)

views/js/ui/modal.js
```diff
@@ -229,7 +229,7 @@ define(['jquery', 'core/pluginifier', 'core/dataattrhandler'], function ($, Plug
 
                 $overlay = $('#' + options.modalOverlay);
 
-                if (options.animate) {
+                if (options.animate && $element.is(':visible')) {
```
krampstudio: the modal might never be visible when we want to open it

jsconan: Yes, but when the element is hidden the animate does not operate, as see in the jQuery doc:

https://api.jquery.com/animate/
`Note: Unlike shorthand animation methods such as .slideDown() and .fadeIn(), the .animate() method does not make hidden elements visible as part of the effect. For example, given $( "someElement" ).hide().animate({height: "20px"}, 500), the animation will run, but the element will remain hidden.`

What I can do is to ensure the element is not hidden by showing it before starting the animate, so I can remove the check.

jsconan:
Yes, there was an issue with the default animate timing which was forced to 100ms instead of the provided value. This is now fixed.

PR #504, opened 2015-08-21
Tao 1272/svgz aac support

antoinerobin:


views/js/ui/resourcemgr/fileSelector.js
```diff
@@ -178,7 +178,15 @@ define([
 
                     //check the mime-type
                     if(options.params.filters){
-                        var filters = options.params.filters.split(',');
+                        var filters = [];
+
+                        if (!_.isString(options.params.filters)) {
```
krampstudio: better written
```javascript
if(_.isPlainObject(options.params.filters)){
    filters = _.pluck(options.params.filters, 'mime');
} else if (_.isString(options.params.filters)){
    filters = options.params.filters.split(',');
}
```

krampstudio:
I'll merge it because it's needed, but could you please add a case in the unit test with the new configuration.

PR #507, opened 2015-08-25
Add custom event triggering tool

jsconan:
Tool that allows to trigger natively custom events without use of jQuery


krampstudio:
That's great but the lib folder is reserved for external libraries. This one has to be part of `tao/views/js/core`

PR #511, opened 2015-08-28
redundant reference to legacy theme settings

dietertaotesting:


krampstudio:
Is there a way to test that ?

PR #514, opened 2015-08-31
moved taoQtiItem/qtiCreator/helper/adaptSize to tao

dietertaotesting:
Also some re-factoring, added unit test

### DIFF
--- a/views/js/util/adaptSize.js
+++ b/views/js/util/adaptSize.js
@@ -1,0 +1,77 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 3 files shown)


views/js/util/adaptSize.js
```diff
@@ -0,0 +1,77 @@
...
+         * @private
+         */
+        var _resize = function ($elements, dimensions) {
+
+            // This whole function is based on calculating the largest height/width.
+            // Therefor the elements need to have style: height/width to be removed
+            // since otherwise we could never track when something is actually getting smaller than before.
+            $elements.each(function () {
+                for (var dimension in dimensions) {
+                    if (dimensions.hasOwnProperty(dimension)) {
```
ssipasseuth: Why using hasOwnProperty() simply to detect a feature ? I find using a boolean more intuitive because {width: 0} looks like the width can set to a given integer whereas {width: true/false} clearly show that the option is on or off.

views/js/util/adaptSize.js
```diff
@@ -0,0 +1,77 @@
...
+     * Adapts the size of several elements
+     */
+    var adaptSize = (function () {
+
+
+        /**
+         * The actual resize function
+         *
+         * @param {jQueryElements} elements
+         * @param dimensions
```
ssipasseuth: the param type is missing

views/js/util/adaptSize.js
ssipasseuth: Then you may use the faster _.forIn() to loop over

PR #518, opened 2015-09-03
Fixed 'close button' behavior, removed redundant event listeners that…

boajer:
… caused performance leaks


views/js/ui/deleter.js
```diff
@@ -304,30 +304,5 @@ define([
...
-                options.undo = false;
-                if(confirm.length > 0){
-                    options.confirmMessage = confirm;
-                }
-            }
-            $elt.deleter(options);
-        }).trigger(function($elt) {
-            $elt.deleter('delete');
-        });
     };
```
krampstudio: Why did you remove that part ? 
The deleter is used for 2 cases : delete (with or without undo) and close (close is some kind of delete with a confirm ~ the name  `delete-confirm` would be more appropriate), but did you check there's no `data-close` that use the feature?

PR #521, opened 2015-09-07
Introduced pagination for generis search

boajer:


antoinerobin:
@jbout this PR works well for me but if I merge it, it will break the solarium implementation. I will let you merge both PR

PR #524, opened 2015-09-07
Tao 1391 theming custom template

ssipasseuth:


krampstudio:
@bugalot could you look at the PHP styling and if nothing hurts you

Alroniks:
Do we use PSR-1/2? I think we should start if not.

krampstudio:
@jbout would like to make some improvements in top of this work. I'll wait for his work to merge the PR.

PR #527, opened 2015-09-09
open the right item on tree refresh

antoinerobin:


bugalot:
Reviewing this PR makes me feel there is no propper way to fix this and that a refactoring must occur. We must address this ASAP with the team.

krampstudio:
@bugalot yes!!! this is a recurring issue and we continue building on top of broken stuffs

bugalot:
As far as I remember, we ended up with a final solution by "contextualizing" the tree with the search results. It was some months ago. With the new search bar, this solution disappeared. We have at least, some hints from the past.

jbout:
While this pull request does not address the server side issues of opening the tree correctly, the JavaScript part already takes into account the file that should be selected and submits it to the server.

A fix for the server-sided response is to be expected from https://github.com/oat-sa/tao-core/pull/545

PR #528, opened 2015-09-09
Unbind event before binding

hutnikau:


krampstudio:
Your fix seems to be working but the unit test seems broken. Could you also update it (and ensure it tests correctly)

hutnikau:
I have fixed the tests. Please review PR again.

PR #532, opened 2015-09-10
Fileexists validator now aware of encoding

boajer:


views/js/util/url.js
```diff
@@ -126,6 +126,15 @@ define([], function(){
...
                 return parsers.base64.test(url);
             }
+        },
+
+        /**
+         * Determine whether encoding is required to match XML standards for attributes
+         * @param {String }url
+         * @returns {String}
+         */
+        encodeAsXmlAttr : function encodeAsXmlAttr(uri) {
```
krampstudio: could you update the unit test in order to test this method

PR #539, opened 2015-09-15
Improved linear-gradient mixin

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1243 and https://github.com/oat-sa/extension-tao-testqti/pull/195

This fix allow use list of color with positions.
In past: 
```sass
@include linear-gradient((color1, color2, color3)) // generates gradient with 33,333% per color
```

Now you can define color and position:
```sass
@include linear-gradient((color1 0px, color2 30px, color3 50px))
```

### DIFF
--- a/views/scss/inc/_functions.scss
+++ b/views/scss/inc/_functions.scss
@@ -11,6 +11,13 @@
     }
 }
 
+/*
+Usage:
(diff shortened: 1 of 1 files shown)


krampstudio:
cool!
Could you also add some comments on top of the mixin in order to describe the possible parameters

Alroniks:
@krampstudio Added comment with usage instactions

PR #541, opened 2015-09-16
validate method return true to avoid constraints on password

antoinerobin:


jbout:
While this fix works around the validation issue, the validation should be moved to Tao instead as a service that can be overriden properly.

PR #542, opened 2015-09-16
Fixed displaying focusing on search result in ontology tree if requir…

boajer:
…ed instance don't fit in first offset window


jbout:
usage of parameter $openNodes makes code hard to understand and error prone

PR #543, opened 2015-09-17
added partial class to select2-search-choice

ssipasseuth:


views/scss/inc/_select2.scss
```diff
@@ -510,6 +510,10 @@ disabled look for disabled choices in the results dropdown
     user-select: none;
 
     background-color: #e4e4e4;
+    
+    &.partial {
+        background-color: #fefefe;
```
krampstudio: Shouldn't we use a color variable to be able to theme it more easily ?

PR #547, opened 2015-09-21
function overloading

hutnikau:
This functionality needed for developing the form validator (though can be helpful in other parts of tao).

### DIFF
--- a/views/js/core/functionOverload.js
+++ b/views/js/core/functionOverload.js
@@ -1,0 +1,64 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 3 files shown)


ssipasseuth:
Since I am not sure about this one, I would delegate this to @krampstudio

krampstudio:
I'll reject this PR because it's not a good idea to have some kind of polymorphism based on arguments. 
In JS polymorphism is usually achieved using the Prototype chain, but for TAO we choose to use Prototyping only when it's really required (I would say hardly ever). We decide to use the patterns that are closest to the JS lang :
 - factory
 - mixins
 - delegation
 - forwarding
 - provider/consumer
 - observer/observable
 - ...

some of those concepts are explained in http://raganwald.com/2014/04/10/mixins-forwarding-delegation.html .

Also our choice to put aside the prototype is to avoid the traps of classical inheritance model and pretending a class based OO programming when it's based on prototype... 

For the validator, you can describe me what you want to achieve (ie. having different behavior using the same validator?) and we can draft the patterns you'll need.

hutnikau:
Function overload it's quite common pattern which is used in a lot of libraries (jQuery for example) and can be really useful.

For example this functionality can help in this part of code:
https://github.com/oat-sa/tao-core/blob/master/views/js/ui/validator.js#L4-L42

This function is very messy. It expects to receive different number of arguments and arguments of different type. It's very difficult to document properly such functions (impossible to describe parameters) and difficult to understand, debug and maintain.

We can rectify this situation using function overloader:

```
/**
 * jsDoc here 
 */
functionOverload.addMethod($.fn. 'validator', function (options) {
     //initialize
});

/**
 * jsDoc here 
 */
functionOverload.addMethod($.fn. 'validator', function (method, methodOptions) {
     //call method
});
```
And so on.

Such approach looks much better and more understandable.

PR #549, opened 2015-09-22
Tao 849 form field validation mechanism

hutnikau:
A have added one rule to the spoken APIP form and created the pull request:
https://github.com/oat-sa/extension-tao-itemqti/pull/435

You can checkout to this branch to test it.

I have not assigned this pull request yet because validation rules should be added to the other APIP forms.


views/js/test/ui/formValidator/test.html
```diff
@@ -0,0 +1,36 @@
...
+        <div id="qunit"></div>
+        <div id="qunit-fixture">
+            <form id="form_1">
+                <input name="field_1" id="field_1" type="text" data-validate="$numeric"/>
+                <input name="field_2" id="field_2" type="text" data-validate="$notEmpty"/>
+                <input name="field_3" id="field_3" type="text" data-validate="$pattern(pattern=\w{3}\s\w{3}\sbaz, modifier=i)"/>
+                <input name="field_5" id="field_5" type="text" data-validate="$length(min=4, max=10)"/>
+
+                <!-- this field should be ignored -->
+                <input name="field_6" id="field_6" class="ignore" type="text" data-validate="$length(min=4, max=10)"/>
```
ssipasseuth: Just wondering, should ignore not be a data attribute that is related to validation, such as data-validate-ignore ?
Unless is there any styling issue attached to that css class ?
I am asking because, seeing that class ignore may not be very explicit....

goodnickoff: This is only for unit test! I used it to test that `fieldSelector` option works fine, and only satisfying fields will be validated:
https://github.com/oat-sa/tao-core/pull/549/files#diff-cfbf5abc66f4785365dae9fb42c8b5eeR20

hutnikau: (sorry, previously commented from another account).
This is only for unit test! I used it to test that `fieldSelector` option works fine, and only satisfying fields will be validated:
https://github.com/oat-sa/tao-core/pull/549/files#diff-cfbf5abc66f4785365dae9fb42c8b5eeR20

views/js/ui/formValidator/highlighters/message.js
```diff
@@ -0,0 +1,67 @@
...
+     * @param {string} [options.errorMessageClass] - error message class
+     * @constructor
+     */
+    function Highlighter(options) {
+        var self = this;
+
+        self.options = $.extend(true, {
+            errorClass : 'error',
+            errorMessageClass : 'validate-error'
+        }, self.options);
```
ssipasseuth: using self.options = _.default(options, defaultOptions); may be more suitable when there is no multilevel objects

views/js/test/ui/formValidator/test.html
ssipasseuth: that is what I thought, sorry forget about

views/js/ui/formValidator/highlighters/message.js
hutnikau: But i guess there could be multilevel objects in the future.
I think _.defaultsDeep will be more suitable here if you prefer lodash instead of jQuery.

PR #554, opened 2015-09-23
Add callbacks

hutnikau:


views/js/ui/contextualPopup.js
```diff
@@ -64,9 +67,13 @@ define([
         $element.css('width', options.style.popupWidth);
         $anchor.append($element);
         $element.off(_ns).on('click' + _ns, '.done', function(){
-            _done($element);
+            if (runCallback('done')) {
```
ssipasseuth: it is more "beforedone" and "beforecancel", isn't it ?

PR #556, opened 2015-09-23
Forward on login instead of redirect

jbout:
backport of https://github.com/oat-sa/tao-core/pull/555


llecaque:
close not needed anymore

PR #568, opened 2015-09-29
need the autoload script in order to install tao correctly by the web…

antoinerobin:
… interface


jbout:
There should be no scenario where the vendor autoload has not yet been called

antoinerobin:
I know but without this line the web interface doesn't install anything for me

jbout:
unable to reproduce

PR #565, opened 2015-09-29
configure the text plugin to use always xhr

krampstudio:
Please cherry-pick it also to develop

PR #570, opened 2015-09-29
configure the text plugin to use always xhr

krampstudio:
It makes client side templates to load even in a Cross Origin context
(Cherry pick from https://github.com/oat-sa/tao-core/pull/565)


krampstudio:
https://oat-sa.atlassian.net/browse/TAO-1524

PR #571, opened 2015-09-30
Fixed styles for modal media manager

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1495


views/js/ui/modal.js
```diff
@@ -219,7 +219,7 @@ define(['jquery', 'core/pluginifier', 'core/dataattrhandler'], function ($, Plug
                 };
 
                 //Calculate the top offset
-                topOffset = (options.vCenter || modalHeight > windowHeight) ? 40 : (windowHeight - modalHeight) / 2;
+                topOffset = (!options.vCenter || modalHeight > windowHeight) ? 40 : (windowHeight - modalHeight) / 2;
```
Alroniks: Also fixed conditions, when vCenter defined as true (by default), modal window should appear on center, now always on top.

dietertaotesting:
The scroll on the right hand side should not make the closer disappear, there should be no scroll area if not needed (in particular overflow-x on the parent element. The left bar does not scroll, it should in fact scroll x and y. It could also be wider, maybe 260 or even 300px

PR #575, opened 2015-10-06
Now OPcache settings and presence are checked during installation

boajer:


bugalot:
Great thanks Michael !

PR #579, opened 2015-10-12
Fixed styles for modal window in media manager

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1495
Previous disscusion here https://github.com/oat-sa/tao-core/pull/571


Alroniks:
Hi @dietertaotesting I've followed your recommendations and tried do task as we discussed. Please, check it.

PR #581, opened 2015-10-13
TAO-1563 Selectable datatable component

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1563


Alroniks:
Added unit tests. @krampstudio Please, look it when you'll have time.

views/js/ui/datatable.js
```diff
@@ -13,13 +13,13 @@ define([
     var dataNs = 'ui.' + ns;
```
krampstudio: could you add the file header with the license

views/js/ui/datatable.js
```diff
@@ -49,6 +49,7 @@ define([
          * @param {Boolean} options.selectable - enables the selection of rows using checkboxes.
          * @param {Object} options.data - inject predefined data to avoid the first query.
          * @param {Object} options.tools - a list of tool buttons to display above the table
+         *
          * @param {Object|Boolean} options.status - allow to display a status bar
          * @param {Object|Boolean} options.filter - allow to display a filter bar
          * @fires dataTable#create.datatable
```
krampstudio: add missing events and complete the documentation, for example 
```
@param {Object|Boolean} [options.filter] - allow to display a filter bar
@param {String[]} [options.filter.columns] - bla bla
```

views/js/ui/datatable.js
```diff
@@ -185,11 +196,31 @@ define([
                 options = this._sortOptions($elt, dataset.sortby, dataset.sortorder);
             }
 
+            // default value for filtering
+            options.column = undefined;
+            _(dataset.model).forEach(function (model) {
+                model.filter = undefined;
+            });
```
krampstudio: why do you do that ?

views/js/ui/datatable.js
```diff
@@ -265,14 +297,32 @@ define([
...
+            $row = $rendering.find('tbody tr');
             $forwardBtn = $rendering.find('.datatable-forward');
             $backwardBtn = $rendering.find('.datatable-backward');
-            $sortBy = $rendering.find('th[data-sort-by]');
+            $sortBy = $rendering.find('th [data-sort-by]');
             $sortElement = $rendering.find('[data-sort-by="'+ options.sortby +'"]');
             $checkAll = $rendering.find('th.checkboxes input');
             $checkboxes = $rendering.find('td.checkboxes input');
 
+            $row.find('td:not(.checkboxes,.actions)').click(function () {
```
krampstudio: Use event delegation instead adding an event handler by cell :+1: 

```
$rendering.on('click', 'tbody tr', function(e){
 //then check if e.target isn't in .checkboxes or .actions
});
```
or even better delegation

views/js/ui/datatable.js
```diff
@@ -461,8 +518,9 @@ define([
          * @fires dataTable#filter.datatable
          * @private
          */
-        _filter: function($elt, query) {
+        _filter: function($elt, filter, columns) {
```
krampstudio: use named function expressions `_filter: function _filter`

krampstudio:
That's good, but could you please : 
 - enhance the documentation, especially what the queries to the server will look like
 - selectable row  and checkbox should configurable
 - check the other places where the datatable is used  (taoDacSimple, taoProctoring, etc.) and ensure it will continue to work otherwise perform the needed changes
 - adjust the row style when hovering for selection 
 - configure the code coverage in your test

Alroniks:
Hi @krampstudio 
I've finished work, I need your attention. 
I've rewriten part of filtering a little bit, now each filter form is independent. I've remove dirty hacks with underfined values and changed implementation of filter input rendering. Also I've changed name of fields for filtering, especailly for columns[], because maybe in future we need columns name for load only needed columns from database.
I have to check other extensions that use datatable and I'll do it tomorrow.

views/js/test/ui/datatable/data.json
```diff
@@ -23,8 +23,8 @@
             "id": "3",
             "login": "jfoo",
             "name": "Jack Foo",
-            "mail": "jack@foo.me",
-            "roles": "Back Office, Global Manager, TAO Manager",
+            "email": "jack@foo.me",
+            "role": "Back Office, Global Manager, TAO Manager",
```
krampstudio: The column was named `roles` because there are multiple roles. Also, I am expecting to be able to have an array, like `roles : ["Back Office", "Global Manager", "TAO Manager"]` and the component join them together. Having a `multiple` option in the model would be helpful. What do you think ?

views/js/ui/datatable.js
```diff
@@ -162,6 +207,7 @@ define([
             var $checkAll;
             var $checkboxes;
             var $massActionBtns = $();
+            var $row;
```
krampstudio: could you rename it `$rows`

views/js/test/ui/datatable/data.json
Alroniks: Yep, should be `roles`. It was error in tests. 
I agree with auto join from array, but could you explain more about `multiple`?

krampstudio:
We're very close! That's good.

I have still a  pending concern. Correct me if I'm wrong, but everything is rendered again when you query the server, like the pagination buttons, and the filtering. What would be the effort to render only the table rows ? (don't do it but do you think it will be difficult? )

views/js/test/ui/datatable/data.json
krampstudio: I see different ways to achieve that:
1. by guessing the type (if type is array then auto join)
2. by closed configuration : `model : [{ id : 'roles',  multiple: true }, ...`
3. by opened configuration : 
```js
model : [{
    id : 'roles', 
    transform : function(values){
         values = values.length || [values];
         return values.map(function(val){ 
            return val.toLowerCase();
        }).join(' :: ');
    }
```

Alroniks:
Yep, now whole component is refreshing after updates from server. In real would be better refresh only rows, because now you need restore states and bindings for all elements on each refresh. I think it should be improved in future. Also I had some issues with handlebars, that can not get root element in 1.3 version (only from 2.0), so it also could be upgraded.

krampstudio:
Ok, so let's enhance it progressively

Alroniks:
Hi @krampstudio 
Please, see commit https://github.com/oat-sa/tao-core/commit/f8555598fd9f07112e00b0b0cd2082a726cb598e
I've added ability turn on custom tranform handlers to fields. By default it not used, because for process all rows to need run two nested cycles, so it's less operations, that are not needed.
For turn on just to need define option `transformable` and define custom handler in model config. 
```javascript
$userList.datatable({
    url: helpers._url('data', 'Users', 'tao'),
    transformable: true,
    model: [
        {
            id : 'login',
            label : __('Login'),
            sortable : true
        },{
            id : 'roles',
            label : __('Roles'),
            sortable : false,
            transform: function(values) {
                if (typeof values === 'string') {
                    return values;
                }

                return _(values).map(function (val) {
                    return val.toLowerCase();
                }).join(' :: ');
            }
        }
    ]
});
```
Arrays are processing to strings by handlebars (but it set comma without spaces), so all will works with arrays too. But I've added custom join with space when transformable set as true.

krampstudio:
Hum there's still 2 points (sorry to be so 	pernickety, but this component is really important).

1. I saw model , options and dataset are merged for the rendering. In order to prevent side effect I would keep them separated and create an object for the template : 

```js
template({ options : options, data : data, model: model});
```
or something similar
	
2. For the transform, I prefer to not have the transformable option and to have it defined at the level of the model (as we do for sortable) and use something like `multiple` for automated transforms. Look at the following example that should stay effective even with large datasets:

http://jsbin.com/vohavazehi/edit?html,js,console

Alroniks:
I've checked search controls in items/test/test takers parts, also check outcomeUI (search results) and current versions works fine. taoDacSimple not use datatable (search of usage in phpstorm not gives results). 

For taoProctoring need fix this settings of data for datatable, because data not should be in options.
https://github.com/oat-sa/extension-tao-proctoring/blob/develop/views/js/controller/ProctorDelivery/index.js#L121
https://github.com/oat-sa/extension-tao-proctoring/blob/develop/views/js/controller/ProctorDelivery/testTakers.js#L95

Data should be passed by seconds param after options, like this:
```javascript
$elt.datatable({
    url : 'js/test/ui/datatable/data.json',
    model: model
}, dataset);
```

views/js/controller/users/index.js
```diff
@@ -56,12 +56,13 @@ define(['jquery', 'i18n', 'helpers', 'layout/section', 'ui/feedback', 'ui/datata
 
             //initialize the user manager component
             $userList.datatable({
-                'url': helpers._url('data', 'Users', 'tao'),
-                'actions' : {
+                url: helpers._url('data', 'Users', 'tao'),
+                filter: true,
```
krampstudio: remove please the `filter` property unless the backend supports it.

krampstudio:
Yeah! :+1:

views/js/ui/datatable.js
```diff
@@ -260,19 +313,43 @@ define([
                         var $btn = $(this);
                         e.preventDefault();
                         if (!$btn.hasClass('disabled')) {
-                            action.apply($btn, [self._selection($elt)]);
+                            action.apply($btn, self._selection($elt));
```
jsconan: Why did you change this?
The goal of the selection is to provide an array of selected items, not to explode them as separate arguments...

Alroniks: _selection already returns array of selected elements https://github.com/oat-sa/tao-core/pull/581/files#diff-b33337cbc7c8f8770970eb2141a5c542L551

jsconan: Of course it does, and the `.apply` method also needs an array. So now, instead of providing an array in one argument, this provide plenty of separated arguments...

So, you can do:

```javascript
action.call($btn, self._selection($elt));
```
or
```javascript
action.apply($btn, [self._selection($elt)]);
```

Both of these codes are similar. And produce this:

```javascript
theAction(theSelection);
```


But if you make this:
```javascript
action.apply($btn, self._selection($elt));
```

this is not the same, and this will produce:

```javascript
theAction(theSelection[0], theSelection[1], ..., theSelection[n]);
```

krampstudio: @Alroniks please include this fix in the same PR than the other fixes

Alroniks:
@krampstudio @jsconan new pr here https://github.com/oat-sa/tao-core/pull/614

PR #587, opened 2015-10-14
Fixed issue with wrong calculations of milliseconds when it has leadi…

Alroniks:
…ng 0

Before:
123456.045 -> 123456.450 

After:
123456.045 -> 123456.045


Alroniks:
@hutnikau Please, check it again

PR #590, opened 2015-10-15
Fix/order options by level

antoinerobin:


bugalot:
Excellent job! Thanks!

PR #595, opened 2015-10-19
Feature/eventifier before

ssipasseuth:


krampstudio:
For the API, I don't really like the syntax. What about something like : 

```js
var app = eventifier();

app.on('leave', function(reason){
    console.log("I'm leaving this app because " + reason);
});

//prevent leaving or continue 
app.before('leave', function(event, reason){
  if(!reason){
       return event.preventNow();
   }
     if(reason === 'dont-like-it'){
         return event.prevent();
    }
});

//same but in async
app.before('leave', function(event, reason){
   var done = event.done();
    fetch('someUrl?reason=' + reason).then(fucntion(resp){
         done();
    }).catch(function(){
        event.prevent();
    })
});

app.trigger('leave',  'dont-like-it');
```

jsconan:
What is the difference between `prevent()` and `preventNow()`?

krampstudio:
`preventNow` will prevent other before to execute : everything stops now

ssipasseuth:
There are advantage and disadvantage for each one of them :

* old one (ok(), reject() as arguments of the "before" callback) 
    * advantage : when set as arguments of the callback, it reveals that those functions are important and we cannot proceed to event without one of them being called
    * disadvantage : hide the event data in the event.data object, the params sent by trigger becauses part of the 

* new one (ok(), reject() as methods of the event object)
   * advantage : it provides a more similar form for the parameters for the event
   * disadvantage : it hides the important methods ok(), reject()n rejectNow() in the event object, it looks like they are optional just as the preventDefault() or stopPropagation(), but one of them need to be called.

So the old one emphicizes the importance of the functions ok or reject and underline the exceptional function that before callbacks represent whereas the new one preserves the "form" of the event parameters better. The question is what would we want to favour here ? Please let me know then I can make the changes if need be.

A site note for reject(true). I actually find little use of having an immediate interruption. Because immediate interruption means that you somehow knows that there may be other "before" callback being executed after you. But that is not sure since the order of execution for the three groups of callback (before(), the event itself and after()) is never guaranted so we don't want random situation when the immediate interruption another one in a situation but not in other.

Site note 2 : I don't mind changing ok for done and reject for prevent.

krampstudio:
There's one subtle thing.  You don't need to call `done/ok` unless you declare it using :-1: 

```js
var done = event.done();
```

It based on the same model than the [QUnit.async](https://api.qunitjs.com/async/) or [grunt](http://gruntjs.com/api/inside-tasks#this.async) 
In most of the case, you won't use anything unless:
 - you're in an async context and you want it to be finished to proceed to the event handler (then you use this form)
 - you want to reject/interrupt the event
 - you want to stop everything right now

jsconan:
We can also use this syntax to activate the async mode:

```javascript
var async = event.async();

if (somethingWrong) {
    async.reject();
} else {
    async.resolve();
}
```

And we can also extend this behavior to all groups: before, event, after

ssipasseuth:
That is interesting. I would like to have this behaviour actually. By any change, if this an existing design pattern, do you have name for such a pattern or an exemple of implementation ?

Also, after looking at your synchronous exemple more closely : is the return value of event.prevent()  useful ?
I mean, are we expecting a return value for the callback in the synchronous case (but not in the async case)?

ssipasseuth:
I find the latest suggestion of @jsconan interesting.

krampstudio:
There's an example implementation here : https://github.com/cowboy/javascript-sync-async-foreach/blob/master/lib/foreach.js 

@sam I won't use return values in any cases. It returns only to break exit the function.
And yes, we could use it also to controll the event execution in any cases

krampstudio:
I had a look at the [QUnit implementation](https://github.com/jquery/qunit/blob/master/src/assert.js#L20) but it sucks : they use a semaphore to stop the processing ... it works for unit testing but it's not efficient for real code, especially because we expect to have a lot of events

ssipasseuth:
OK, thanks for your comments on sample implementation. I get the principle now, thanks.
So do we agree on the following solution which mixes suggestions from @krampstudio and @jsconan ?

```js
var app = eventifier();

app.on('leave', function(reason){
    console.log("I'm leaving this app because " + reason);
});

//prevent leaving or continue 
app.before('leave', function(event, reason){

  if(!reason){
       return event.preventNow();   //interrupt immediate so all other before are killed
   }

   if(reason === 'dont-like-it'){
         return event.prevent();  //interrupt
    }
});
//prevent leaving or continue 
app.before('leave', function(event, reason){

  if(!reason){
       return event.preventNow();   //interrupt immediate so all other before are killed
   }

   if(reason === 'dont-like-it'){
         return event.prevent();  //interrupt
    }
});

//same but in async
app.before('leave', function(event, reason){

    //this says "I'm in async mode"
     var async = event.sync();

    //do something async
    fetch('someUrl?reason=' + reason).then(fucntion(resp){
         async.done();
    }).catch(function(){
        async.prevent();
    })
});

app.trigger('leave',  'dont-like-it');
```

krampstudio:
I don't really see why we should have 2 ways to prevent, because `event.prevent` can be used in async mode also.

ssipasseuth:
Hey guys, I have updated the before api considering all our discussion points. Please take a look when you have a little time. Thanks !

PR #607, opened 2015-10-29
add a polling manager with async mode

jsconan:
Polling manager: execute action at interval with async mode.

Each iteration wait for the previous one to be executed.

### DIFF
--- a/views/js/core/polling.js
+++ b/views/js/core/polling.js
@@ -69,9 +69,8 @@ define([
             /**
              * Notifies the action is about to be called
              * @event polling#call
-             * @param {polling} polling
              */
-            polling.trigger('call', polling);
+            polling.trigger('call');
 
             action.call(context, polling);
 
--- a/views/js/core/polling.js
+++ b/views/js/core/polling.js
@@ -40,7 +40,6 @@ define([
      * @param {Number|String} [config.interval] - The minimal time between two iterations
      * @param {Boolean} [config.autoStart] - Whether or not the polling should start immediately
      * @param {Object} [config.context] - An optional context to apply on each action call
-     * @param {Function} [config.oneventname] - Bind an event handler to the event eventname
      * @returns {polling}
      */
     var pollingFactory = function pollingFactory(config) {
@@ -83,9 +82,8 @@ define([
                     /**
                      * Notifies the polling continues
                      * @event polling#resolved
-                     * @param {polling} polling
                      */
-                    polling.trigger('resolved', polling);
+                    polling.trigger('resolved');
 
                     // next iteration
                     startTimer();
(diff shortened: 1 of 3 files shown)


views/js/core/polling.js
```diff
@@ -0,0 +1,324 @@
...
+    var _defaultInterval = 60 * 1000;
+
+    /**
+     * Create a polling manager for a particular action
+     * @param {Object|Function} [config] - A config object, or the action called on each iteration
+     * @param {Function} [config.action] - The callback action called on each iteration
+     * @param {Number|String} [config.interval] - The minimal time between two iterations
+     * @param {Boolean} [config.autoStart] - Whether or not the polling should start immediately
+     * @param {Object} [config.context] - An optional context to apply on each action call
+     * @param {Function} [config.oneventname] - Bind an event handler to the event eventname
```
krampstudio: Please do not attach listeners using the config

views/js/core/polling.js
```diff
@@ -0,0 +1,324 @@
...
+        /**
+         * Runs an iteration of the polling loop
+         */
+        var iteration = function iteration() {
+            /**
+             * Notifies the action is about to be called
+             * @event polling#call
+             * @param {polling} polling
+             */
+            polling.trigger('call', polling);
```
krampstudio: Events doesn't need the context in parameter.

views/js/core/polling.js
```diff
@@ -0,0 +1,324 @@
...
+             */
+            async : function async() {
+                var cb = {};
+
+                promise = new Promise(function(resolve, reject) {
+                    cb.resolve = resolve;
+                    cb.reject = reject;
+                });
+
+                _.assign(promise, cb);
```
krampstudio: I think  you can just return the promise instead : 

```js
async : function async() {
   promise = new Promise();
   this.trigger('async', promise);
   return promise;
}
```

Or did I miss something ?

views/js/core/polling.js
```diff
@@ -0,0 +1,324 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2015 (original work) Open Assessment Technologies SA ;
+ */
+/**
+ * @author Jean-Sébastien Conan <jean-sebastien.conan@vesperiagroup.com>
```
krampstudio: a little `@example` would be welcomed

views/js/core/polling.js
jsconan: Ok, maybe it is better to get access to the Promise object to add some actions. I agree for that, and it is already done.

However, the Promise object does not provide access to `resolve()`/`reject()` methods. So I still need to assign them into the final object. Natively, the only provided methods are `then()` and `catch()`

jsconan:
Changes done.
Thanks for your review.

views/js/core/polling.js
krampstudio: Yes, my bad, the only way to resolve a promise is `Promise.resolve(promise)` which cannot be used here

PR #608, opened 2015-10-30
create an history router

krampstudio:
Added history router.
There's no test because mocking up the browser history doesn't really work (the actual test fails and will be removed)

### DIFF
--- a/views/js/core/historyRouter.js
+++ b/views/js/core/historyRouter.js
@@ -30,6 +30,7 @@ define([
     'use strict';
 
     //start the history polyfill, see https://github.com/devote/HTML5-History-API (a getter should trigger it)
+    var historyRouter;
     var location = window.history.location || window.location;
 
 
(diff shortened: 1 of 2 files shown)


views/js/core/historyRouter.js
```diff
@@ -0,0 +1,112 @@
...
+                var self = this;
+                if(_.isString(state)){
+                    state = { url : state };
+                }
+                if(!state || !state.url){
+                    return;
+                }
+
+                if(replace === true){
+                    history.replaceState(state, '', state.url);
```
ssipasseuth: history does not seem to have a method replaceState(), does it ?

ssipasseuth: I am actually confused, because the history seems to be using methods and properties that are not declared in https://github.com/oat-sa/tao-core/blob/feature/history-router/views/js/core/history.js 
Am I missing something here ?

views/js/core/historyRouter.js
```diff
@@ -0,0 +1,112 @@
...
+                if(_.isString(state)){
+                    state = { url : state };
+                }
+                history.pushState(state, '', state.url);
+                this.dispatch(state);
+            }
+        });
+
+        //back & forward button, and push state
+        $(window).off('popstate').on('popstate', function (event) {
```
ssipasseuth: Does this not break any subsequent historyRouter ? 
```
var routerA = historyRouter();
var routerB = historyRouter();
//routerA can no longer dispatch based on the window popstate event ...
```

views/js/core/historyRouter.js
krampstudio: core/history.js !== lib/history.js
core/history is misspelled and the other is a polyfill

views/js/core/historyRouter.js
krampstudio: oh yeah will fix it

views/js/core/historyRouter.js
ssipasseuth: I knew I missed something stupid ...

PR #609, opened 2015-10-30
Feature/components

jsconan:
Add a few components:
- listBox
- breadcrumbs

Add a new CSS style: flex-grid

Add a base class for similar components: component.js


views/scss/inc/_breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+  font-weight: normal;
+  font-family: $headingFont;
+  @include font-size(11);
+  padding: 8px;
+  margin: 0;
+  list-style: none;
+
+  li {
+    display: inline-block;
+    positio: relative;
```
Alroniks: typo in property name, should be `position` I guess

jsconan: You are right. Thank you!

krampstudio:
I have a structural problem with `ui/component` : 
 -  use delegation instead of assignment 
 - dom -> container
 - I would not use setup/teardown but rather trigger events you can plug on
 - separate the state management in another module (and to not attach the state directly to the object, to enfore using the method or setState to modify it)  and then use the separated state module
 - the template shouldn't be mandatory

jsconan:
Ok Bertrand, I already made some changes regarding your comment:
- use delegation instead of assignment
- use events instead of template methods (setup/teardown)
- the template is not mandatory anymore

For the separate state management, we still need to discuss about, as I did not clearly understand the concern.

Last but not least, what do you mean about this: dom -> container ?

krampstudio:
rename getDom methods to getContainer
And for the state, you can start by not attaching the flags to `this.config.is` but to a variable in the closure. It will prevent someone to access it directly (within or outside the component) but ensure the setState/is method are called

jsconan:
Ok, thank you for the explanation
I made the changes

krampstudio:
Ok cool, over to you @ssipasseuth

views/js/ui/component.js
```diff
@@ -0,0 +1,242 @@
...
+             * Sets the template used to render this component
+             * @param {Function} template
+             * @returns {componentApi}
+             */
+            setTemplate : function setTemplate(template) {
+                var tpl = template || defaultTpl;
+                componentTpl = tpl;
+
+                // ensure the template is defined as a function
+                if (!_.isFunction(componentTpl)) {
```
ssipasseuth: should we allow a template as a string at all? In this case, why not handlebars.compile() it ?

views/js/ui/component.js
```diff
@@ -0,0 +1,242 @@
...
+
+                return this.$component;
+            },
+
+            /**
+             * Shows the component
+             * @returns {component}
+             */
+            show : function show() {
+                return this.setState('hidden', false);
```
ssipasseuth: What about triggering an event on each action call (show, hide, enable, disable) ? it may be useful.

views/js/ui/component.js
```diff
@@ -0,0 +1,242 @@
...
+                }
+
+                return this;
+            },
+
+            /**
+             * Gets the underlying DOM element
+             * @returns {jQuery}
+             */
+            getContainer : function getDom() {
```
ssipasseuth: not very important but still... getDom -> getContainer

jsconan: Oups! fixed just now

views/js/ui/component.js
jsconan: Yes, you are right! I forgot to generalise the use of events when I added eventifier...

views/js/ui/component.js
jsconan: It can be useful in two situations:
- for very simple components, like buttons or static text, with no need of complex template with data merging
- for test units

ssipasseuth:
I forgot to mention : it would be helpful to add the event(s) fired in the jsdoc.

views/js/ui/component.js
```diff
@@ -0,0 +1,288 @@
...
+                var self = this;
+                var initConfig = config || {};
+                this.config = _.omit(initConfig, function(value, name) {
+                    var omit = value === undefined || value === null;
+                    if (!omit && !name.indexOf('on')) {
+                        self.on(name.substr(2), value);
+                        omit = true;
+                    }
+                    return omit;
+                });
```
ssipasseuth: I would rather prefer the format
```
var config = {
    on : {
        init : function(){
        },
        render : function(){
        },
        destroy : function(){
        }
    }
}
```
But that is matter of personal taste ...

views/js/ui/component.js
```diff
@@ -0,0 +1,288 @@
...
+                this.trigger('template', componentTpl, this);
+
+                return this;
+            }
+        };
+
+        // let's extend the instance with extra methods
+        if (specs) {
+            _(specs).functions().forEach(function(method){
+                componentApi[method] = function delegate(){
```
ssipasseuth: Here you are not really taking advantage of the delegation, are you ? Since componentApi is in the same scope, it will be create along with its functions each time the component factory is being called, isn't it ?

views/js/test/ui/listbox/test.js
```diff
@@ -0,0 +1,329 @@
...
+        assert.equal(instance.config.textEmpty, config.textEmpty, 'The listBox instance has its own empty list message');
+        assert.equal(instance.config.textNumber, config.textNumber, 'The listBox instance has its own number label');
+        assert.equal(instance.config.textLoading, config.textLoading, 'The listBox instance has its own loading message');
+
+        assert.equal(instance.getContainer().find('h1').text(), config.title, 'The listBox instance has rendered a title with its own content');
+        assert.equal(instance.getContainer().find('.empty-list').text(), config.textEmpty, 'The listBox instance has rendered a message to display when the list is empty, and set its own content');
+        assert.equal(instance.getContainer().find('.available-list .label').text(), config.textNumber, 'The listBox instance has rendered a message to show the number of boxes, and set its own content');
+        assert.equal(instance.getContainer().find('.loading').text(), config.textLoading + '...', 'The listBox instance has rendered a message to show when the component is in loading state, and set its own content');
+    });
```
ssipasseuth: it would be nice to have a visual test in the end, outside of the fixture, so we can control the rendering in a simple glimpse (this is applicable also for other visual component)

PR #614, opened 2015-11-04
Fix some minor issues after review

Alroniks:


Alroniks:
Ups, I've broken tests. Fixing now

Alroniks:
@krampstudio Sorry for my rush, fixed tests.

jsconan:
Cool, the mass actions work again (multi selection).

actions/class.Users.php
```diff
@@ -256,11 +256,14 @@ public function checkLogin(){
...
-        if(!$this->hasRequestParameter('uri')){
+        if (!$this->hasRequestParameter('uri')) {
             throw new Exception('Please set the user uri in request parameter');
         }
 
-        $user = new core_kernel_classes_Resource(tao_helpers_Uri::decode($this->getRequestParameter('uri')));
+        $uri = $this->getRequestParameter('uri');
+        $uri = is_array($uri) ? array_shift($uri) : $uri;
+
+        $user = new core_kernel_classes_Resource(tao_helpers_Uri::decode($uri));
```
jsconan: Why change the webservice? Isn't be better to change the javascript action to extract the array before requesting the server?
I think it is preferable to keep webservices consistent, and adapt the clients to work with.
@krampstudio what do you think about that?

Alroniks: It's a feature of mass actions. When using selection and mass actions, component pass to controller array of fields. In this case array of URIs (uri[] = something). And actions should have ability process this cases, IMO.

krampstudio: For me we must differentiate the 2 cases:
 - using an action or a "selection by row" we should get only the selected item
 - using the selection by checkboxes we always have an array of item

Then in the user table, the edit action is always on one user, so the controller shouldn't take care of the arrays (this controller expects an uri as a string, if you send him something else, then it's not valid).

krampstudio:
By the way, could you change  the class `selectablebyclick` by `hoverable` or `selectable` 
and also the property name `selectbyclick` by `rowSelection` or something similar

Alroniks:
Fixed names of classes in styles, and also was reverted changes in users controller. Summarizing previous discussions, now it works as expected and the developer responsible for proper use actions (especially mass actions).

PR #617, opened 2015-11-05
datatable: add context arguments to model transform methods

jsconan:


krampstudio:
could you add a test case please

jsconan:
I just added a unit test for model transform.
It test the arguments provided as well as the rendering.

PR #619, opened 2015-11-05
fix and test validator registration

krampstudio:
Reapply hotfix from oat-sa/tao-core#618

### DIFF
--- a/views/js/test/core/validator/validators/test.js
+++ b/views/js/test/core/validator/validators/test.js
@@ -1,0 +1,134 @@
+define([
+    'core/validator/validators'
+], function(validators){
+    'use strict';
+
+    QUnit.module('API');
+
+    QUnit.test('module', function(assert){
+        QUnit.expect(3);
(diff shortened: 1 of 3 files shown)


views/js/test/core/validator/validators/test.js
```diff
@@ -0,0 +1,134 @@
...
+        assert.ok(typeof validators === 'object', 'The validators module exports an object');
+        assert.ok(typeof validators.register === 'function', 'The validators object has a register method');
+        assert.ok(typeof validators.validators === 'object', 'The validators object has a validators property');
+    });
+
+
+    QUnit.module('register');
+
+
+    QUnit.test('validtor validation', function(assert){
```
ssipasseuth: not important: typo "validtor"

krampstudio: If this is the only one, then I'll fix it only on develop and after to prevent me redoing the cherry picking

PR #620, opened 2015-11-05
minor components API refactoring and simplifications

krampstudio:
changes : 
 - `getContainer` returns the component container (= renderTo) 
 - `getElement` (previously `getContainer`) returns the component element
 - `render` chains to the API and does not return the container anymore
 - events can only be attached using the eventifier API  (`on`) and there is no   `config.onSomething` anymore
 - events doesn't use the context in arguments as the eventifier delegates already the context on this.

### DIFF
--- a/views/js/ui/component.js
+++ b/views/js/ui/component.js
@@ -35,10 +35,15 @@ define([
      * @returns {component}
      */
     var component = function component(specs, defaults) {
+
         // the template is a private property
         var componentTpl = defaultTpl;
 
-        var componentState;
+        //contains the states of the components
+        var componentState = {};
+
+        //where the component is added
+        var $container;
 
         // base skeleton
         var componentApi = {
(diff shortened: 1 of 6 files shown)


views/js/ui/component.js
```diff
@@ -47,40 +52,23 @@ define([
...
-                    var omit = value === undefined || value === null;
-                    if (!omit && !name.indexOf('on')) {
-                        self.on(name.substr(2), value);
-                        omit = true;
-                    }
-                    return omit;
-                });
-                _.defaults(this.config, defaults || {});
+                this.config = _.defaults(_.clone(config) || {}, defaults || {});
+
```
jsconan: I have a concern about this: you do not filter null values anymore.

This filter has been added to take care of config provided by the server, in which sometimes null values have been added to remove previously set properties (once a property has been set in the registry, it is impossible to remove it, unless setting it to null). In that case, null values must be replaced by the default ones.

Examples can be found in QTI Tools: previously the label was set in the config, sever side.
Now, those labels have been moved in the code, but there is still the possibility to set them from the config. However, the previously set values have been disabled by null assignment.

views/js/ui/component.js
```diff
@@ -47,40 +52,23 @@ define([
...
                 componentState = {};
 
                 /**
                  * Executes extra init tasks
                  * @event component#init
-                 * @param {component} component
                  */
-                this.trigger('init', this);
+                this.trigger('init');
```
jsconan: I have a concern about this.

Even if the context is bound to `this` inside the event handler, it is also possible to use proxied functions, in which the only way to get access to the component is to get it as argument, like in this example:

```javascript
other = {
    display: function(msg) {
         // display something
    },
    close: function() {
        // do something
    },
    action: function(cmp) {
        if (cmp.is('enabled')) {
            this.display('component is enabled');
        } else {
            this.close();
        }
    }
}
mycomponent.on('myevent', other.action.bind(other)); 
```

krampstudio:
@jsconan here we go

PR #622, opened 2015-11-06
Refactored and fix users controller: added support of filtering

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1710


Alroniks:
@krampstudio Take a look please, here some minor changes for datatable. I've added abilty define placeholder for column filter and fixed css styles.

antoinerobin:
The fact that we display ```$firstname . ' ' . $lastname```in the name column but filter only on lastname is not really clear. You could either split this field in two part or search on the combination of both

krampstudio:
Agree with @antoinerobin split the field
And from a GUI point of view, just let the global filter as it works on any filterable field.

Alroniks:
@krampstudio Can you explain more about "let the global filter as it works on any filterable field"? Now it works the same, but if columns not defined, will be used all columns for filtering with 'or' chaining.

krampstudio:
For the users having only the global filter is sufficient

Alroniks:
Updated @antoinerobin @krampstudio

antoinerobin:
I think that you should replace the column Name by 2 columns, firstname and lastname.
It will simplify your php code and the user experience. Otherwise your changes seem ok to me.

Alroniks:
Agree @antoinerobin In latest commit split name into two columns and simplify user controller.

PR #625, opened 2015-11-10
add milliseconds to formater

antoinerobin:
TAO-1789

### DIFF
--- a/helpers/class.Date.php
+++ b/helpers/class.Date.php
@@ -60,14 +60,15 @@ class tao_helpers_Date
     }
 
     /**
-     * Dispalys a date/time
-     * Should in theorie be dependant on the users locale and timezone
-     *
-     * @param mixed $timestamp            
-     * @param int $format  The date format. See tao_helpers_Date's constants.
+     * Displays a date/time
+     * Should in theory be dependant on the users locale and timezone
+     * @param mixed $timestamp
+     * @param int $format The date format. See tao_helpers_Date's constants.
+     * @param string $timeZone user timezone
      * @return string The formatted date.
+     * @throws common_Exception when timestamp is not recognized
      */
-    static public function displayeDate($timestamp, $format = self::FORMAT_LONG)
+    static public function displayeDate($timestamp, $format = self::FORMAT_LONG, $timeZone = null)
     {
         if (is_object($timestamp) && $timestamp instanceof core_kernel_classes_Literal) {
             $ts = $timestamp->__toString();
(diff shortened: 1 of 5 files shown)


helpers/dateFormatter/EuropeanFormatter.php
```diff
@@ -53,7 +56,8 @@ public function format($timestamp, $format)
         	    $formatString = 'F j, Y, g:i:s a';
         	    break;
 			case \tao_helpers_Date::FORMAT_ISO8601:
-				$formatString = 'Y-m-d\TH:i:sO';
+				$milliseconds = str_replace('0.', '', sprintf('%0.3f', fmod($timestamp, 1)));
+				$formatString = 'Y-m-d\TH:i:s.'.$milliseconds;
```
hutnikau: May be time zone should be represented in the formatted value?
I guess we have to include timezone here:
`$formatString = 'Y-m-d\TH:i:s.' . $milliseconds . 'O';`

helpers/class.Date.php
```diff
@@ -60,14 +60,15 @@ static protected function getDateFormatter()
...
+     * Displays a date/time
+     * Should in theory be dependant on the users locale and timezone
+     * @param mixed $timestamp
+     * @param int $format The date format. See tao_helpers_Date's constants.
+     * @param string $timeZone user timezone
      * @return string The formatted date.
+     * @throws common_Exception when timestamp is not recognized
      */
-    static public function displayeDate($timestamp, $format = self::FORMAT_LONG)
+    static public function displayeDate($timestamp, $format = self::FORMAT_LONG, $timeZone = null)
```
hutnikau: Where it supposed to be used? Maybe pass instance of `\DateTimeZone` instead of time zone name can be better solution.

helpers/dateFormatter/EuropeanFormatter.php
antoinerobin: As seen with Tom this should be separated for simplicity

helpers/class.Date.php
antoinerobin: You're right I change it right now. It is used in the related PR (that I forgot to mention before)

helpers/dateFormatter/EuropeanFormatter.php
hutnikau: And the last thing. Please update phpdoc and add type of `$timeZone` parameter:
```php
public function format($timestamp, $format, DateTimeZone $timeZone = null)
```

PR #627, opened 2015-11-11
polling.js: improved version

jsconan:
- promise is better handled as the post actions are now set directly at create time
- the next() method, when called, will wait for the promise if any
- added better state handling
- added missing tests

### DIFF
--- a/views/js/core/polling.js
+++ b/views/js/core/polling.js
@@ -139,14 +139,16 @@ define([
      * @returns {polling}
      */
     var pollingFactory = function pollingFactory(config) {
-        var stopped, timer, promise, interval, max, iter, action, context, autoStart;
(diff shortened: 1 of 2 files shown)


ssipasseuth:
It is not the object of the PR but just a personal comment : I would personally prefered to have the callback in the async mode to be continue() and stop() instead of resolve() and reject().
continue and stop are closer to this usage context whereas resolve/reject seem too general a concept. So as a developer, I wonder what resolve() and reject() would mean in the polling context although continue and stop are straight forward.
Food for thoughts...

jsconan:
Yes you get a point. But to keep consistency across API using promises, I think it is better to always have the same names, whatever the context is. Also, there is already a stop method in the polling API, having a different function with the same name can be confusing.

PR #634, opened 2015-11-18
add value to the model if not exist

hutnikau:
related to:
https://github.com/oat-sa/qti-sdk/pull/22

### DIFF
--- a/views/js/core/databinder.js
+++ b/views/js/core/databinder.js
@@ -253,7 +253,7 @@ function($, _, Handlebars, Encoders, Filters){
      */
     DataBinder.prototype._bindNode = function _bindNode($node, path, model, domFirst) {
         if(!$node.data('bound')){
-            if(domFirst === true){
+            if(domFirst === true || locate(model, path) === undefined){
                   update(model, path, this._getNodeValue($node));
             }
 
--- a/views/js/test/core/databinder/test.js
+++ b/views/js/test/core/databinder/test.js
@@ -195,7 +195,7 @@ define(['jquery', 'core/databinder'], function($, DataBinder){
 
         $sectionParts.append($newSection);
 
-        databinder = new DataBinder($container, model);
+        var databinder = new DataBinder($container, model);
         databinder.bind();
 
         strictEqual(model.testParts[0].assessmentSections[0].sectionParts.length, 4, 'model length has been updated');
--- a/views/js/test/core/databinder/test.js
+++ b/views/js/test/core/databinder/test.js
@@ -183,7 +183,7 @@ define(['jquery', 'core/databinder'], function($, DataBinder){
     });
 
     QUnit.test('Add value which is not represented in the model', function(assert){
-        QUnit.expect(3);
+        QUnit.expect(2);
 
         var $container = $('#container-4');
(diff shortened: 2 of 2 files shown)


views/js/core/databinder.js
```diff
@@ -253,7 +253,7 @@ function($, _, Handlebars, Encoders, Filters){
      */
     DataBinder.prototype._bindNode = function _bindNode($node, path, model, domFirst) {
         if(!$node.data('bound')){
-            if(domFirst === true){
+            if(domFirst === true || locate(model, path) === undefined){
```
ssipasseuth: please add a new unit test to illustrate this bug fix

hutnikau: I created unit test. Please review it again.

views/js/test/core/databinder/test.js
```diff
@@ -182,6 +182,26 @@ define(['jquery', 'core/databinder'], function($, DataBinder){
...
+        var $sectionParts = $('ul', $container);
+        assert.ok($container.length === 1, 'Test the fixture is available');
+
+        var $newSection = $("<li>" +
+            "<input checked='checked' type='checkbox' value='true' data-bind-encoder='boolean' data-bind='visible'>" +
+            "</li>");
+
+        $sectionParts.append($newSection);
+
+        databinder = new DataBinder($container, model);
```
ssipasseuth: that is a global var ...

hutnikau: Oops! Fixed.

views/js/test/core/databinder/test.js
```diff
@@ -182,6 +182,26 @@ define(['jquery', 'core/databinder'], function($, DataBinder){
...
+        var $newSection = $("<li>" +
+            "<input checked='checked' type='checkbox' value='true' data-bind-encoder='boolean' data-bind='visible'>" +
+            "</li>");
+
+        $sectionParts.append($newSection);
+
+        var databinder = new DataBinder($container, model);
+        databinder.bind();
+
+        strictEqual(model.testParts[0].assessmentSections[0].sectionParts.length, 4, 'model length has been updated');
```
ssipasseuth: ok, now I am still trying to understand.... but it seems that the model lenght has not been update : the length was initially 4, wasn't it ?
If you remove the change you made on  databinder.js, this condition still evaluate to true, is that expected ?

ssipasseuth: also, I am wondering, should you not bind using domFirst=true first to solve this issue ?

hutnikau: You are right. This assertion is redundant.
I removed it.

hutnikau: `domFirst=true` will not solve this issue. At the moment when property panel will be added to the DOM model (for example `identifier` property) of this panel is already exists, but all inputs in added panel is empty. So, if we set `domFirst=true` then all existing properties in the model will be erased. You can try it by setting `databinder.bind(null, null, null, true);` here:
https://github.com/oat-sa/extension-tao-testqti/blob/develop/views/js/controller/creator/views/property.js#L58

PR #637, opened 2015-11-18
Feature/dialog extended

jsconan:
Add factory methods to quickly display an alert or a confirm message using the modal dialog.

### DIFF
--- a/views/js/ui/dialog.js
+++ b/views/js/ui/dialog.js
@@ -415,5 +415,57 @@ define([
         return instance;
     };
 
+    /**
+     * Displays an alert message
+     * @param {String} message - The displayed message
+     * @param {Function} action - An action called when the alert is closed
+     * @returns {dialog} - Returns the dialog instance
+     */
+    dialogFactory.alert = function dialogAlert(message, action) {
+        var dlg = dialogFactory({
+            message: message,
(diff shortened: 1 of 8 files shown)


views/js/ui/dialog.js
```diff
@@ -415,5 +415,57 @@ define([
         return instance;
     };
 
+    /**
+     * Displays an alert message
+     * @param {String} message - The displayed message
+     * @param {Function} action - An action called when the alert is closed
+     * @returns {dialog} - Returns the dialog instance
+     */
+    dialogFactory.alert = function dialogAlert(message, action) {
```
ssipasseuth: This is a strange way to add a new behaviour to the factory. Would it not make sense to simply create another component out of the original one and preconfigure it the way you are doing in here ?
It goes the same for "dialogFactory.confirm"

jsconan: It is a special case of use for the dialog. So, I think this the right place to put this behavior, as this is a common pattern for this sort of API. Now, if this pattern is not acceptable in TAO, I can change it to use separate components.

jsconan:
I changed the pattern for the extensions. Now they use composition, with two new components:
- ui/dialog/alert
- ui/dialog/confirm

PR #653, opened 2015-11-26
Tao 1855 multiple value delimiter has no effect

zShurick:
https://oat-sa.atlassian.net/browse/TAO-1855


krampstudio:
@antoinerobin will validate your PR. FYI you don't need to merge develop into the branch before submitting. The conflict resolution has to be handled by the reviewer.

PR #656, opened 2015-11-26
render properly checkboxes instead of putting radio buttons

antoinerobin:
in csv import when you have a checkbox list it was displaying radio buttons but you could select multiple.
related to : TAO-1826

### DIFF
--- a/views/js/form/post-render-props.js
+++ b/views/js/form/post-render-props.js
@@ -24,16 +24,23 @@ define([
         /**
          * Reposition the radio buttons or checkboxes of a property and make them look nice.
          *
+         * @param $container the container in which to search and upgrade buttons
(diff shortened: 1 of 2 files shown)


views/js/form/post-render-props.js
```diff
@@ -22,16 +22,22 @@ define([
         }
 
         /**
-         * Reposition the radio buttons of a property and make them look nice.
+         * Reposition the radio buttons or checkboxes of a property and make them look nice.
          *
          * @private
          */
-        function _upgradeRadioButtons($container) {
+        function _upgradeButtons($container, type){
```
krampstudio: when you add a parameter please document it (what are the possible values to expect ?)

PR #657, opened 2015-11-27
Fix/csv import options

krampstudio:
- Fix a notice in case of multi value import
 - Enable to add a predefined static Map (a side mapping, for example to add default values for excluded properties)

### DIFF
--- a/models/classes/import/CsvAbstractImporter.php
+++ b/models/classes/import/CsvAbstractImporter.php
@@ -134,6 +134,18 @@ abstract class CsvAbstractImporter
 	}
 
 
+        /**
+         * Additional mapping values but that comes from another source than the CSV file.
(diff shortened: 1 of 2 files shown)


models/classes/import/CsvAbstractImporter.php
```diff
@@ -150,8 +162,11 @@ public function setValidators($validators)
 	 */
 	public function importFile($class, $options) {
 
-
-		$options['staticMap'] = $this->getStaticData();
+                if(!isset($options['staticMap']) && !is_array($options['staticMap'])){
```
antoinerobin: shouldn't it be OR instead of AND ? to avoid array_merge possible error ?

krampstudio: yes, sure

krampstudio:
You can check again

PR #659, opened 2015-12-01
Redirect instead of forward final page

jbout:
Temporary fix until history rewriting is available in back office


krampstudio:
I won't have the time to review it this Sprint, @llecaque could you take it ?

PR #661, opened 2015-12-03
Add aynchronous MIME type getter for remote resources

jsconan:


krampstudio:
That's great, but could you look if it makes sens to group it with `core/mimetype` ?

jsconan:
Ok, I can move the getter to the existent module. The reason why I did not already do it is that `core/mimetype` seems to be very specific to a particular use case.

jsconan:
I moved the MIME type getter into the existing module, and removed the new one.

krampstudio:
Why do you use both a Promise and a callback ?

jsconan:
The callback allows the use of Node.js style tools like async, while the promise allows to chain other promises. I know there is two ways of doing things in this case, but this pattern (promise + callback) cover many more use cases and bring more compatibility possibilities.

krampstudio:
I have an error while running the test in the cli : `grunt connect qunit:single --test /tao/views/js/test/core/mimetype/test.html`

jsconan:
My bad, I didn't see the test failed in CLI. (The error status code is not properly returned in this mode)

krampstudio:
Also, could you please use only the promise, not the callback (or the inverse if the promises are more restrictive), it could be a source of confusion.

jsconan:
Ok, I removed the promise. The callback allows a more concise syntax, and is compatible with async.

krampstudio:
Did you manage to make the test to work on CLI ?

jsconan:
Yes the issue with CLI was also fixed.

PR #664, opened 2015-12-09
Feature/user helper

jsconan:
Add helpers to:
- get user from URI
- get user full name
- get user first name
- get user last name
- get user label

### DIFF
--- a/helpers/UserHelper.php
+++ b/helpers/UserHelper.php
@@ -21,7 +21,9 @@
 
 namespace oat\tao\helpers;
 
+use oat\oatbox\user\User;
 use core_kernel_classes_Resource;
+use core_kernel_users_GenerisUser;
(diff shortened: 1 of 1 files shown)


helpers/UserHelper.php
```diff
@@ -47,4 +49,101 @@ static public function renderHtmlUser($userId)
...
+     * @param string $userId
+     * @return User
+     */
+    public static function getUser($userId)
+    {
+        if (is_string($userId)) {
+            $userId = new core_kernel_classes_Resource($userId);
+        }
+
+        if (!($userId instanceof core_kernel_users_GenerisUser)) {
```
antoinerobin: You should check if the userId is a Resource at this point (it could be a boolean or a NULL) in order to prevent failure.

jsconan:
I updated the getUser method

PR #666, opened 2015-12-10
create a logger api and a console logger

krampstudio:


krampstudio:
I'll add the registration of the logger

PR #670, opened 2015-12-15
Introduced possibility to customize behavior for select boxes

boajer:


views/js/ui/bulkActionPopup/cascadingComboBox.js
```diff
@@ -0,0 +1,119 @@
...
+            var $comboBox;
+
+            if(_.isArray(options.categoriesDefinitions) && _.isArray(options.categories)){
+                $comboBox = createCombobox(0, options.categoriesDefinitions, options.categories);
+                $container.append($comboBox);
+            }
+        } ;
+    };
+
+});
```
ssipasseuth: Please wrap it using ui/component in order to garantee a consistent api as the other recent components (e.g. ui/bulkActionPopup)

ssipasseuth:
Also, is it good to isolate the cascading combobox into a separate component, but new component means new unit tests !

views/js/ui/bulkActionPopup/independentComboBox.js
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    /**
+     * @param {object} options
+     * @param {Array} [options.categoriesDefinitions] - the array that defines the number and config for each level of combobox
+     * @param {Array} [options.categories] - the array that contains array of categories
+     * @returns {function}
+     */
+    return function independentComboBoxFactory(options) {
```
ssipasseuth: I would rather call this component groupedComboBox instead because "independent" combobox is only understandable in the context of the comparison againt the cascading combobox component, whereas "grouped" combobox is understandable on its own.

ssipasseuth:
since the combobox may be reused outside of the bulk action popup, please also move it outside of the bulkActionPopup and directly in ui/cascadingCOmbobox and ui/groupedCombobox

PR #671, opened 2015-12-15
Tao 1936 required actions

hutnikau:


actions/class.requiredAction.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
bugalot: This class file name should have an uppercase first letter (RequiredAction).

actions/class.Main.php
```diff
@@ -133,7 +135,10 @@ public function login()
 			    $success = LoginService::login($myForm->getValue('login'), $myForm->getValue('password'));
 				if($success){
 				    \common_Logger::i("Successful login of user '" . $myForm->getValue('login') . "'.");
-				    
+
+                    $eventManager = $this->getServiceManager()->get(EventManager::CONFIG_ID);
+                    $eventManager->trigger(new LoginEvent());
```
bugalot: Good use of events !

bugalot:
Hi @hutnikau !

I really like the concept! It's a very good start, but also a good use of events and services. In order to continue in the same way, I would like you to think about, but certainly change the following aspects of your Pull Request:

* The Code of Conduct feature is a dedicated/specific feature for ACT. I'm sure it was a nice "Required Action" use case for testing your new feature, but we don't need it at  the level of the TAO core distribution. Could you please put it for integration in the taoAct extension? You might want to create another _dummy_ implementation of "Required Action" for testing purpose... Why not a simple "Hello World" at login time that we can configure for testing only.

* At the moment, registered "Required Actions" will be attempted to be executed on every single request, if their ::mustBeExecuted() method returns true. I think the "must be executed" approach is good, but we need more. Indeed, if you look at the Code of Conduct use case, we call ::mustBeExecuted() for every single request, which might be a performance issue on the long term. I think that we should be able to configure that a particular "Required Action" must be executed only for one, multiple, or all actions. So we would just need an extra optional parameter to the "Required Action" parameters. Here's an example of Code of Conduct, required only for action /tao/main/login.

```php
return  new RequiredActionService([
    RequiredActionService::OPTION_REQUIRED_ACTIONS => [
        new RequiredActionRedirect(
            'codeOfConduct',
            [
                new CodeOfConductRule(new \DateInterval('P1Y')),
            ],
           _url('codeofconduct', 'RequiredAction', 'tao'),
            // If given, will restrict the use of the required action only for
            // ext TAO, module main, action login
           array('tao', 'main', 'login')
       ),
    ]
]);
```

Does that make sense @hutnikau ? Do not hesitate to reply if you have any question.

Bests,
Jérôme

hutnikau:
@bugalot  Thank you for review !

We can't restrict checking of required action only for one action. This checking definitely **must** be executed for every request. This is the part of task:
_... can not access system functionality until they have accepted the Code of Conduct_

It means that user can't make request to any action (not only to main/login) until he accept code of conduct.

If we restrict checking to only login action then user which must accept Code of Conduct will have access to any action until he log out and log in again. It is contrary to the concept of _Required_ action !

bugalot:
@hutnikau , thanks for your prompt response.

I just discussed with @jbout what you describe. You have convincing arguments ;)! However, we discussed the ratio issue-coverage/performance-cost, and we feel it will be better to restrict the scope of the code of conduct to "just after login" of Proctors and Test Site administrators. As it is written in the acceptance criteria: _The ACT configurable Code of Conduct is presented to every Proctor and Test Site Administrator who has a valid username and password when they initially logon_. In addition, we have to present that to Proctor and Test Site administrators only so we have to be careful.

So if the entry point is the one dedicated to Proctors or Test Site administrators, this is where the check should be performed. In such a context, we should be able to restrict this Required Action to a particular MVC action or module, or extension as in my previous comment. However, that does not mean there is no place for global Required Action, it's a matter of configuration.

jbout:
While we need to be able to restrict it to specific roles in the future, presenting it to all users might be an acceptable solution for now.

hutnikau:
To which roles action will be executed depends on rules, specified for this action. I implemented special rule for "code of conduct" action (`CodeOfConductRule` class):
https://github.com/oat-sa/tao-core/pull/671/files#diff-9cf9dac1d63402f84e696cd5045c70a1R28
And as you can see there specified roles, for which this action will be executed.

bugalot:
That's a good point!

bugalot:
So if we 

* move codeOfConduct rule the ACT project;
* add the capability to restrict required action to specific MVC extension/module/action

We should be all good for a first iteration!

hutnikau:
I see. I will push new commits soon.
Thank you for review!

hutnikau:
Please ignore the user account of previous commit ! I'm first time working from home and by accident used my old account.

bugalot:
No problem @hutnikau

bugalot:
Hi @hutnikau,
I can see dependencies to CodeOfConductRule in tao-core. Shouldn't they be only in taoACT? This makes me feel that tao-core depends on taoACT.

hutnikau:
@bugalot i have just removed these dependencies. Please review it again.

actions/class.requiredAction.php
```diff
@@ -0,0 +1,52 @@
...
+
+use oat\oatbox\service\ServiceManager;
+use oat\tao\model\requiredAction\implementation\RequiredActionService;
+
+/**
+ * @author Aleh Hutnikau <hutniaku@1pt.com>
+ * @package tao
+ *
+ */
+class tao_actions_RequiredAction extends \tao_actions_Main
```
bugalot: @hutnikau 

This Module should also be part of taoACT, as it is the specific Module enabling the code of conduct behaviour to be called by the client.

In taoACT, this module could have a class name of CodeOfConduct, with an index() action containing the main controller logic. You will just have to change your ::attach() call in the taoAct's Updater.

Sorry for all these round trips and thanks for your work.

models/classes/requiredAction/RequiredActionInterface.php
```diff
@@ -0,0 +1,73 @@
...
+ * Interface RequiredActionInterface
+ *
+ * RequiredAction is action which should be executed by user before performing any activities in the TAO
+ *
+ * @package oat\tao\model\requiredAction
+ * @author Aleh Hutnilau <hutnikau@1pt.com>
+ */
+interface RequiredActionInterface
+{
+    const CLASS_URI = 'http://www.tao.lu/Ontologies/TAO.rdf#RequiredAction';
```
bugalot: @hutnikau 

I wonder if these constants should not be placed into TimeRule class, indeed, as far as I understand, this is the only place where they are used.

By doing so, RequiredActionInterface becomes perfectly generic, and clarifies that a RequiredAction implementations are not "forced" to be database dependent (e.g. cookie based implementation).

bugalot:
@hutnikau I added some comments but don't worry, we're getting closer!

hutnikau:
@bugalot I have just made changes according to your comments. Please review it one more time.

bugalot:
@hutnikau  If we push the last remaining files related to codeOfConduct in taoACT, I'm fine with it ;) !

* config/default/codeOfConduct.conf.php
* config/header/codeOfConduct.conf.php
* views/templates/requiredAction/codeofconduct.tpl

Thanks for the continuous effort. This is how we will get perfect quality :+1:  !

hutnikau:
@bugalot Done!

scripts/update/Updater.php
```diff
@@ -426,7 +428,24 @@ public function update($initialVersion) {
...
+                // all good, already configured
+            } catch (ServiceNotFoundException $error) {
+                $requiredActionService = new RequiredActionService();
+                $this->getServiceManager()->register(RequiredActionService::CONFIG_ID, $requiredActionService);
+            }
+
+            OntologyUpdater::syncModels();
+
+            $ext = \common_ext_ExtensionsManager::singleton()->getExtensionById('tao');
+            $ext->setConfig(\tao_actions_RequiredAction::CONF_CODE_OF_CONDUCT, '');
```
bugalot: It seems that this constant is now in taoAct, so it might make fail the update script.

hutnikau:
I have merged develop branch into this one and resolved conflicts.

PR #673, opened 2015-12-16
filter unique objects from array

jbout:


helpers/class.Array.php
```diff
@@ -61,7 +61,27 @@ public static function sortByField($input, $field, $descending = false)
...
+    
+    /**
+     * Sorts an array of objects implementing the __equal() function
+     * 
+     * @param array $array
+     */
+    public static function array_unique($array) {
+        $keys = array_keys($array);
+        $toDrop = array();
+        for ($i = count($keys) - 1; $i >=0 ; $i-- ) {
```
antoinerobin: Could we use the ```count``` on $array instead of doing an ```array_keys``` and then a ```count```
Here we assume that keys are numerical, start at 0 and are continuous. Is it ok ?

PR #688, opened 2015-12-29
Feature/return stream

jbout:


helpers/class.Http.php
```diff
@@ -286,7 +287,23 @@ public static function returnFile($filename, $contenttype = true, $svgzSupport =
...
     }
+    
+    public static function returnStream(StreamInterface $stream, $fileSize, $mimeType = null)
+    {
+        if (!is_null($mimeType)) {
+            header('Content-Type: ' . $mimeType);
+        }
+        header('HTTP/1.1 200 OK');
+        header("Content-Length: " . $fileSize);
+        if (ob_get_level() > 0) {
```
hutnikau: I think that here should be `while(ob_get_level() > 0){...}` because `ob_end_flush()` send content only of the last buffer.

PR #697, opened 2016-01-06
Add promiseFactory

jsconan:
A factory that produces promises with nested resolve/reject methods


jsconan:
Cherry-picked into: https://github.com/oat-sa/tao-core/pull/817

PR #698, opened 2016-01-12
Tao 1726 localised parse numbers

Alroniks:
Related to https://github.com/oat-sa/extension-tao-itemqti/pull/500
Related to https://oat-sa.atlassian.net/browse/TAO-1726
Comments by task here: https://oat-sa.atlassian.net/browse/TAO-1723

### DIFF
--- a/views/js/util/locale.js
+++ b/views/js/util/locale.js
@@ -23,20 +23,36 @@
 define(['module'], function (module) {
     'use strict';
 
+    var configuration = module.config();
+
     /**
      * Util object for manipulate locale dependent data
      * @exports util/locale
      */
     return {
 
-        config: module.config(),
+        /**
+         * Returns config of component
+         * @returns {*}
+         */
+        getConfig: function getConfig() {
+            return configuration;
+        },
+
+        /**
+         * Sets config of component
+         * @param config
+         */
+        setConfig: function setConfig(config) {
+            configuration = config || {};
+        },
 
         /**
          * Returns current system decimal separator
          * @returns {string}
          */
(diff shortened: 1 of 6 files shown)


views/js/util/locale.js
```diff
@@ -0,0 +1,86 @@
...
+define(['module'], function (module) {
+    'use strict';
+
+    /**
+     * Util object for manipulate locale dependent data
+     * @exports util/locale
+     */
+    return {
+
+        config: module.config(),
```
ssipasseuth: I would suggest to set the variable config as private and add a setter and getter instead. This way we can better ace and trcontrol the modification on the config.

Alroniks: I agree. @ssipasseuth Can you share examples or any article about how correct create setter and getter on javascript?

ssipasseuth:
```javascript
define(function(){
       //private var, accessible only in the scope
       var privateVar = null;
       
       //return object with public methods
       return {
            getPrivateVar : function(){
                return privateVar;
            },
            setPrivateVar : function(newValue){
                privateVar = newValue
            }
       };
    });

ssipasseuth:
This is a bug fix for project PARCC. And I just checked with @jbout, the fix needs to be done to the master branch for extensions that are not parcc related. So could you please create a branch from master, cherry pick your changes and create a PR to master.

Alroniks:
@ssipasseuth Fixed your notises and created PR for master branch https://github.com/oat-sa/tao-core/pull/711

ssipasseuth:
Replaced with https://github.com/oat-sa/tao-core/pull/714

PR #708, opened 2016-01-18
Add file to form data

hutnikau:


views/js/ui/filesender.js
```diff
@@ -51,7 +51,7 @@ define(['jquery', 'lodash'], function($, _) {
 
                 //post the full form that contains the file
                 fd = new FormData(this[0]);
-
+                fd.append('content', options.file);
```
krampstudio: could you add it conditionally and if you can pass the name also in `options` it would be perfect:
Something like : 
```js
if(options.file && options.fileParamName && options.file instanceof File){ //Blob might be also accurate
   fd.append( options.fileParamName, options.file);
}
```

hutnikau:
The branch has been updated. Please review it again.

PR #711, opened 2016-01-19
Tao 1726 localised number parser core

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1726


scripts/update/Updater.php
```diff
@@ -431,6 +432,15 @@ public function update($initialVersion) {
         if($this->isVersion('2.15.1')){
             $this->setVersion('2.15.2');
         }
+
+        if ($currentVersion === '2.15.2') {
```
Alroniks: @ssipasseuth Not sure that is correct, but for proper work of changes this code should be run during update.

PR #729, opened 2016-02-03
Remove __construct method

hutnikau:
I decided that declare `__construct` method in the interface is not very good idea.


ssipasseuth:
ok, sounds reasonable but should you not update the manifest version number to reflect this change, so the another extension that is using this updated interface can target a corrrect minimum version ?

hutnikau:
I guess that this change does not impact on classes which implements this interface, so i think there is no sense to change manifest version.

ssipasseuth:
Humm ok. I mean if you have a new class that implement this interface with no __construct method, then its extension should have this version of tao-core right ? That is why I thought it might make sense to update the extension number just in case.
Now in practive, I believe that it is not the case so we could potentially ignore that. It's your call.

hutnikau:
I increased version and resolved conflicts.

PR #737, opened 2016-02-15
hide search icon when advanced mode is active for custom properties

no-chris:
this was not a bug, as the "expected behaviour" wasn't implemented on purpose.

However, I think this is misleading having a button with no behaviour implemented, so I decided to hide it when in advanced mode.

### DIFF
--- a/views/js/form/post-render-props.js
+++ b/views/js/form/post-render-props.js
@@ -100,6 +100,8 @@ define([
                 var $property = $(this);
                 if($property.attr !== undefined){
                     var type = (function() {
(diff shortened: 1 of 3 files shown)


views/js/form/post-render-props.js
```diff
@@ -117,6 +117,13 @@ define([
                                 _hideProperties($editContainer);
                                 _hideIndexes($editContainer);
 
+                                var $propertyMode = $('.property-mode');
```
krampstudio: Even though the function doesn't respect this rule, we declare variable on top of their scope in order to reflect hoisting.

PR #745, opened 2016-02-24
Feature/rendering interface

antoinerobin:


antoinerobin:
preview and authoring does not display images yet

PR #749, opened 2016-03-04
ui/hider

krampstudio:


views/js/ui/hider.js
```diff
@@ -0,0 +1,91 @@
...
+        hide : function hide(element){
+            return jqWrap(element).addClass(hiddenClass);
+        },
+
+        /**
+         * Toggle (show if hidden / hide if shown) the given element
+         * @param {jQueryElement|HTMLElement|String} element - the element/node/selector
+         * @returns {jQueryElement} the element
+         */
+        toggle : function toggle(element){
```
jsconan: Not very important, but I think the optional parameter that bring a condition would be very helpful.

See: http://api.jquery.com/toggleclass/

PR #759, opened 2016-03-14
better support of dropped files

krampstudio:
I was not able to reproduce the drag 'n drop issue, but while debugging, sometime the `dataTransfer`  from 'e.originalEvent` is null and then an error is thrown.

This is only a safety fix.

Test in different browsers if you can:
 - import a test or an item using drag and drop.
 - download one or multiple images from the item creator

### DIFF
--- a/views/js/ui/uploader.js
+++ b/views/js/ui/uploader.js
@@ -215,9 +215,20 @@ define([
                             .on('dragend', dragOutHandler)
                             .on('dragleave', dragOutHandler)
                             .on('drop', function(e){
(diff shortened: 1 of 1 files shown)


no-chris:
I don't know to which point this is related to this issue but while trying to import multiple items by drag & drop, only the last one is actually imported. Works fine for multiple images in an item.

krampstudio:
Yes, for import the `multiple` option is false while it allow multiple download in the resource manager (item creator). So it's the expected behavior

PR #762, opened 2016-03-16
PHP Fatal error in TAO GUI installation

zagovorichev:
https://github.com/oat-sa/generis/pull/165
(for directories method isReadable should not be called, when directory doesn't exists)


jbout:
No longer required as error-handling is covered in generis

PR #766, opened 2016-03-21
At the opening of the dialog, sets focus on a modal window

zagovorichev:
reason: https://oat-sa.atlassian.net/browse/TAO-2427

### DIFF
--- a/views/js/ui/dialog.js
+++ b/views/js/ui/dialog.js
@@ -379,6 +379,20 @@ define([
             }
         },
 
+        _setFocusOnModal: function _setFocusOnModal(){
+            // default OK button (for enter key)
+            var $ok = $('button.ok', this.$buttons);
+            if ($ok.length) {
+                $ok.focus();
+            } else {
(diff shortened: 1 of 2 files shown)


views/js/ui/dialog.js
```diff
@@ -379,6 +379,20 @@ define([
...
         },
 
+        _setFocusOnModal: function _setFocusOnModal(){
+            // default OK button (for enter key)
+            var $ok = $('button.ok', this.$buttons);
+            if ($ok.length) {
+                $ok.focus();
+            } else {
+                // other button
+                var $btn = $('button', this.$buttons).first();
```
krampstudio: Please declare your variables in a way that reflects hoisting

krampstudio:
Agree with the feature, however you need to add a test case into https://github.com/oat-sa/tao-core/blob/master/views/js/test/ui/dialog/test.js

zagovorichev:
@krampstudio Fixed

PR #771, opened 2016-03-23
Tao 2327 map getfile to flysystem

hutnikau:


getFile.php
```diff
@@ -21,6 +21,8 @@
  *
  */
 
+require_once __DIR__ . '/../vendor/autoload.php';
```
jbout: Is this already required?

hutnikau: No, autoloader got there by accident. Removed.

PR #797, opened 2016-04-12
Remove mkdir

siwane:


ssipasseuth:
Will be waiting for https://github.com/oat-sa/extension-tao-testqti/pull/405 to be completed.

PR #803, opened 2016-04-20
Move the plugin model and pluginLoader to core

krampstudio:
I've moved the plugin factory from taoTests to tao in order to use it in the item creator.
All references to the test runner should have been removed.
The plugins have now an access to it's "host" using the `getHost` method. For convenience it's possible to alias this method if the `hostName` is provided in the config.

### DIFF
--- a/views/js/core/plugin.js
+++ b/views/js/core/plugin.js
@@ -58,14 +58,15 @@ define([
      * @param {Object} provider - the plugin provider
      * @param {String} provider.name - the plugin name
      * @param {Function} provider.init - the plugin initialization method
-     * @param {Function} [provider.render] - plugin render behaviorV
-     * @param {Function} [provider.finish] - plugin render behaviorV
+     * @param {Function} [provider.render] - plugin rendering behavior
+     * @param {Function} [provider.finish] - plugin finish behavior
      * @param {Function} [provider.destroy] - plugin destroy behavior
      * @param {Function} [provider.show] - plugin show behavior
      * @param {Function} [provider.hide] - plugin hide behavior
      * @param {Function} [provider.enable] - plugin enable behavior
      * @param {Function} [provider.disable] - plugin disable behavior
      * @param {Object} defaults - default configuration to be assigned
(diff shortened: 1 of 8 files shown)


views/js/core/plugin.js
```diff
@@ -0,0 +1,337 @@
...
+    'use strict';
+
+    /**
+     * Meta factory for plugins. Let's you create a plugin definition.
+     *
+     * @param {Object} provider - the plugin provider
+     * @param {String} provider.name - the plugin name
+     * @param {Function} provider.init - the plugin initialization method
+     * @param {Function} [provider.render] - plugin render behaviorV
+     * @param {Function} [provider.finish] - plugin render behaviorV
```
jsconan: Wrong text ? What does it mean "behaviorV" ?

views/js/core/plugin.js
```diff
@@ -0,0 +1,337 @@
...
+                 */
+                getHost : function getHost(){
+                    return host;
+                },
+
+                /**
+                 * Get the host testRunner
+                 * @returns {host} the plugins's host
+                 */
+                getAreaBroker : function getAreaBroker(){
```
jsconan: Seems to be a wrong annotation

views/js/core/plugin.js
```diff
@@ -0,0 +1,337 @@
...
+     * @param {String} provider.name - the plugin name
+     * @param {Function} provider.init - the plugin initialization method
+     * @param {Function} [provider.render] - plugin render behaviorV
+     * @param {Function} [provider.finish] - plugin render behaviorV
+     * @param {Function} [provider.destroy] - plugin destroy behavior
+     * @param {Function} [provider.show] - plugin show behavior
+     * @param {Function} [provider.hide] - plugin hide behavior
+     * @param {Function} [provider.enable] - plugin enable behavior
+     * @param {Function} [provider.disable] - plugin disable behavior
+     * @param {Object} defaults - default configuration to be assigned
```
jsconan: Not very essential, but maybe can you add a description for the hostName default config ?

PR #816, opened 2016-04-29
TAO-2573 replaced background images by font symbols

dietertaotesting:


krampstudio:
The icons are changed, but the position of the buttons is still odd...

PR #817, opened 2016-05-02
Feature/tao 2582 communication interface

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2582

Moved components from taoTests: 
- core/providerRegistry
- core/tokenHandler

New component:
- core/delegator
- core/communicator
- core/promiseFactory


views/js/core/communicator.js
```diff
@@ -0,0 +1,273 @@
...
+    var defaults = {
+        timeout: 30 * 1000
+    };
+
+    /**
+     * Creates a communicator implementation.
+     * The communicator relies on a provider to execute the actions.
+     * Most of the delegated methods must return promises.
+     *
+     * @param {Object} providerName - The name of the provider instance,
```
krampstudio: isn't it a `{String}` ?

jsconan: Yes it is! My bad.

views/js/core/communicator.js
```diff
@@ -0,0 +1,273 @@
...
+        // creates the implementation by setting an API an delegating calls to the provider
+        communicator = eventifier({
+            /**
+             * Initializes the communication implementation.
+             * Sets the `ready` state.
+             * @returns {Promise} The delegated provider's method must return a promise
+             * @fires init
+             * @fires ready
+             */
+            init: function init() {
```
krampstudio: Just a hint but it seems it's not needed to wrap inside a new promise and manage the different levels of catch.

```js
function init() {
  if(this.getState('ready')){
      return Promise.resolve();
  }    
  return delegate('init').then(function(){
         self.setState('ready').trigger('ready');
  });
}
```

jsconan: Yes I saw this while I was working on the provider. Shall I update this before you merge?

views/js/core/delegator.js
```diff
@@ -0,0 +1,89 @@
...
+                    // delegate the call to the provider
+                    response = (provider[fnName] || extendedConfig.default).apply(context, args);
+
+                    // if supported fire the method related event
+                    if (eventifier) {
+                        // the response has to be provided as first argument in all events
+                        api.trigger.apply(api, [fnName, response].concat(args));
+                    }
+                } else {
+                    throw new Error('There is no method called ' + fnName + ' in the ' + extendedConfig.name + ' provider!');
```
krampstudio: I don't think we should fail if the provider hasn't one of the method. At least this should be configurable.

views/js/core/delegator.js
```diff
@@ -0,0 +1,89 @@
...
+         * @private
+         * @throws Error
+         */
+        function delegate(fnName, args) {
+            var response;
+
+            if (provider) {
+                if (_.isFunction(provider[fnName]) || _.isFunction(extendedConfig.default)) {
+                    // need real array of params, even if empty
+                    args = args ? _slice.call(args) : [];
```
krampstudio: It's only a preference, but I'm more in favor of rest parameters instead of an array of arguments.

views/js/core/delegator.js
jsconan: In fact it is configurable: you can provide a default delegated method

views/js/core/communicator.js
krampstudio: yes, it would be more readable

views/js/core/delegator.js
krampstudio: in terms of API, it means you have to define an empty function to as a default value. I was thinking to something like : `config.required`

views/js/core/promiseFactory.js
```diff
@@ -0,0 +1,44 @@
...
+ * @author Jean-Sébastien Conan <jean-sebastien.conan@vesperiagroup.com>
+ */
+define(['lodash', 'core/promise'], function (_, Promise) {
+    'use strict';
+
+    /**
+     * Gets a promise with nested resolve/reject methods
+     * @returns {Promise}
+     */
+    function promiseFactory() {
```
krampstudio: This is something I wouldn't see generalized. Do you use it somewhere else than the polling ?

jsconan: Yes, in the provider I am working on. This is an edge case, but it sometimes occurs, when the promise is reversed.

views/js/core/delegator.js
jsconan: Ok, I am changing that.

views/js/core/delegator.js
jsconan: Ok, I am going to add such a config

views/js/core/promiseFactory.js
krampstudio: This might lead to a misuses. Could you revert that part so we can work on a better alternative later.

PR #821, opened 2016-05-05
Tree structure for delivery results

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-2559


zagovorichev:
https://github.com/oat-sa/extension-tao-act/pull/217
https://github.com/oat-sa/extension-tao-outcomeui/pull/66

PR #823, opened 2016-05-09
Add polling communictor implementation

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2583

Add a communicator implementation that use polling to communicate with a remote service

### DIFF
--- a/views/js/core/communicator/poll.js
+++ b/views/js/core/communicator/poll.js
@@ -206,7 +206,7 @@ define([
                     self.polling.start();
                     resolve();
                 } else {
-                    reject();
+                    reject('The communicator has not been properly initialized');
                 }
             });
         },
@@ -222,7 +222,7 @@ define([
                     self.polling.stop();
                     resolve();
                 } else {
-                    reject();
+                    reject('The communicator has not been properly initialized');
                 }
             });
         },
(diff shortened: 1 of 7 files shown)


views/js/core/communicator/poll.js
```diff
@@ -0,0 +1,296 @@
...
+         */
+        init: function init() {
+            var self = this;
+            var config = _.defaults(this.getConfig(), defaults);
+            var tokenHandler = tokenHandlerFactory(config.token);
+
+            // validate the config
+            if (!config.service) {
+                // a remote service is needed to build a long poll communication
+                return Promise.reject('You must provide a service URL');
```
krampstudio: I prefer that we reject Errors. It's then easier to get the stack trace especially because we chain/compose Promises

views/js/core/communicator/poll.js
```diff
@@ -0,0 +1,296 @@
...
+            }
+
+            return stopped;
+        },
+
+        /**
+         * Sends an messages through the communication implementation
+         * @param {String} channel - The name of the communication channel to use
+         * @param {Object} message - The message to send
+         * @returns {Promise}
```
krampstudio: could you just explain the promise has to be resolved/rejected

PR #834, opened 2016-05-23
move translate calc to gpu

krampstudio:
Optimize transform and some doc fixes

### DIFF
--- a/views/js/ui/dynamicComponent.js
+++ b/views/js/ui/dynamicComponent.js
@@ -75,7 +75,7 @@ define([
             return this;
         }
     };
-    
+
     var _defaults = {
         title : '',
         resizable : true,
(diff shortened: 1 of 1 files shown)


views/js/ui/dynamicComponent.js
```diff
@@ -178,7 +163,7 @@ define([
                     var $target = $(e.target),
                         x = (parseFloat($target.attr('data-x')) || 0) + e.dx,
                         y = (parseFloat($target.attr('data-y')) || 0) + e.dy,
-                        transform = 'translate(' + x + 'px, ' + y + 'px)';
+                        transform = 'translate(' + x + 'px, ' + y + 'px) translateZ(0)';
```
ssipasseuth: we could actually use translate3d(x,y,z) to be more explicit ?

krampstudio: not sure. `translateZ(0)` will move the calcul of the transform to the gpu, but it's a simple 2d translation. 
`translate3d` might involve a different way to calculate the translation.  
Also if the browser doesn't have accelerated rendering setup , `translateZ` will be ignored but `translate3d` will be heaviest were a simple translate will be sufficient.

PR #837, opened 2016-05-24
Add core/persistence to handle a data accessor with persistence

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2585

Add a data accessor with persistence. The data is maintained in memory as well as using the browser storage. Although the updates are using promises and are asynchronous, the read is synchronous and allows to immediately access to the value.


krampstudio:
I'm not really in favor of this kind of implementation, it adds too many layers of complexity for a simple feature. I don't really set why we have 2 levels of promises and an eventifier.

I see 2 solutions : 

1. You should simply implement the persistence using the store object within the TestRunner provider.

```js
        loadPersistentStates : function loadPersistentStates() {
           var self = this;
           var config = this.getConfig();
            this.stateBuffer = {};
           return store('test-states-' + config.serviceCallId).then(function(state){
                self.stateBuffer = state;
           });
        },

        getPersistentState : function getPersistentState(name) {
                return !!this.stateBuffer[name];
        },

        setPersistentState : function setPersistentState(name, active) {
             this.stateBuffer[name] = !!active;
             return this.stateStorage.setItem('state',  this.stateBuffer);
        }
```

2. Clean up and simplify your persitence implementation (remove the eventifier, etc.) : 

```js
var key = 'foo';
function bufferdStore(name){
 return store(name).then(function(store){
       var buffer = {};
       return store.getItem(key).then(fuction(value){
         buffer = value;
         return {
            getItem : function(name) { return buffer[name] };
            //...
        };
 });
}
```
```

**Comment by jsconan** [comment]
The eventifier have been added especially to centralize error handling.

Nevermind, I removed it and also updated the test runner

File: views/js/core/persistence.js [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+                    // just provide a data accessor that:
+                    // - immediately gets the values
+                    // - stores the changes through a promise.
+                    return {
+                        /**
+                         * Gets a value from the data
+                         * @param {String} name
+                         * @returns {Object}
+                         */
+                        get : function getPersistenceValue(name) {
```
krampstudio: could you synchronize the property name with the function name.

krampstudio:
Could you also rename the module, `persistence` is too generic. Something like `cachedStore`, `bufferedStore`, `accessorStorage`, etc

jsconan:
I update the API to be more consistent with the other stores and also renamed the module.

PR #838, opened 2016-05-24
Feature/filestorage delete dir

siwane:


test/service/FileStorageTest.php
```diff
@@ -0,0 +1,139 @@
...
+                    'options' => array(
+                        'root' => $this->privateDir
+                    )
+                )
+            )
+        );
+
+        $fileSystemService = new FileSystemService($adaptersFixture);
+
+        $smProphecy = $this->prophesize(ServiceLocatorInterface::class);
```
hutnikau: `class` constant can be used only on php 5.5+
As far as I know TAO is still compatible with php 5.4

siwane: It's official, we no longer support the version 5.4
Take a look at ticket [2492,](https://oat-sa.atlassian.net/browse/TAO-2492)

PR #843, opened 2016-05-30
added layer when dragging

ssipasseuth:
* fixed dragging issue when the content contains an iframe
* set dragging handle


views/js/ui/dynamicComponent.js
```diff
@@ -132,15 +133,45 @@ define([
                         draggableContainer = draggableContainer[0];
                     }
                     if(_.isElement(draggableContainer) || _.isString(draggableContainer)){
+                        
+                        //add dragging layer on top of the content area to fix issue when dragging over an iframe
+                        $draggingLayer = $('<div class="dynamic-component-layer">').appendTo($content);
```
krampstudio: Why don't you add it directly to the template ? 

```hbs
<div class="dynamic-component-container">
    <div class="dynamic-component-title-bar">
        <h3>{{title}}</h3>
        <a title="{{__ "close"}}" class="closer" href="#"></a>
    </div>
    <div class="dynamic-component-content">
  {{#if draggableContainer}}
        <div class="dynamic-component-layer"></div>
  {{/if}}
    </div>
</div>

```

ssipasseuth:
right, moved the dynamic-component-layer into the tpl.

PR #845, opened 2016-06-02
update grunt and plugins

krampstudio:
I've updated the dependency to grunt (to grunt 1.0.0) and most of the grunt plugins. 
On my computer it reduces the compilation time from 27min to 3min.

Even if the requirejs, uglify and sass version are backward compatible this could have an impact everywhere.

Please reinstall the dependencies, compile all full build (`npm run build`) and test the platform in production mode (a configuration with the new TR for example because it involves a lot compiled bundles).

It requires https://github.com/oat-sa/extension-tao-delivery/pull/174 because the new version of `grunt-contrib-sass` is more strict and fails when a file does not exists.

Because this is a more or less risky PR, don't hesitate to involve the test team.

### DIFF
--- a/views/build/package.json
+++ b/views/build/package.json
@@ -2,23 +2,23 @@
   "name": "tao-build",
   "description": "Client side build tool chain for TAO",
   "scripts": {
-    "build": "grunt build --no-color",
-    "bundle": "grunt bundleall --no-color",
-    "sass": "grunt sassall --no-color",
-    "test": "grunt testall --no-color --force"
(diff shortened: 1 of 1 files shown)


llecaque:
Previous comment shuold be ignored, I am not sure how I have done that

krampstudio:
@llecaque in my mailbox the user appears as "Leeroy Jenkins" :)

llecaque:
ok to test

llecaque:
retest this please

PR #853, opened 2016-06-09
Breaking/tao 2653 requirements checks

krampstudio:
This PR is used for 2 tasks : 
 - TAO-2662 because it removes Modernizr
 - TAO-2653 because it fixes the js detection

It also provides minimal browser requirement checks (basic ES5 and DOM features that are mandatory for TAO to work), previously there was modernizr loaded (~50Kb) loaded on every page and the [checks](https://github.com/oat-sa/tao-core/compare/develop...breaking/TAO-2653_requirements-checks?expand=1#diff-9617d0ce36f45f333725f99ae03b1f8eL29) wasn't made at all...


krampstudio:
@oat-devops are you a human being ?

jsconan:
Works fine in admin, but breaks in delivery.

krampstudio:
requires https://github.com/oat-sa/extension-tao-testqti/pull/480

krampstudio:
@jsconan should be fixed

PR #855, opened 2016-06-14
Feature/tao 2641 clean browser db

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2641

Add the ability to remove store from the browser.
Set an activity timestamp when updating the store (write or delete).

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -27,7 +27,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '4.0.0',
+    'version' => '4.1.0',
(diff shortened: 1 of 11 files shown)


krampstudio:
I need more time to review this one. I'd like profile the implementation and look at the transactions workflow.

krampstudio:
I know this is additional work, but is it possible to separate the PR in two parts : 
1. remove the databases
2. last activity 

I'm asking because I don't think we will use the last activity removal. By thinking, there're lot of situations where this doesn't make sense at all.

jsconan:
This PR is splitted in two other PR. The first is already under review and the second will come just after.

PR #859, opened 2016-06-16
Extend guzzle StreamWrapper

hutnikau:


hutnikau:
I removed custom streamWrapper and it works fine, but provided that these both PR will be accepted:
https://github.com/oat-sa/extension-tao-testqti/pull/518
https://github.com/oat-sa/extension-tao-itemqti/pull/741

As you said before we have to use `write()` instead of `writeStream()`

PR #867, opened 2016-06-23
in bulkActionPopup was added option reasonRequired

zagovorichev:
all fields in the form are required

https://oat-sa.atlassian.net/browse/TAO-2696


ssipasseuth:
Hi, the logic looks good but TAO already has a common validation lib, using it whenever possible offers more consistent result across all the application. Could you please check if you can benefit from it in this case ?

zagovorichev:
But I don't know how many and what kind of the fields will be in the form. How can I use validation lib for all fields in the form?

ssipasseuth:
Hi,
A sample usage can be found in https://github.com/oat-sa/extension-tao-itemqti/blob/master/views/js/qtiCreator/widgets/helpers/formElement.js#L100-L103

But it is true that this plugin currently only allows event based triggered... so it may require a little bit more work. Next time you have some form validation requirement, please remember about the ui/groupvalidator. ;)

So I test it with https://github.com/oat-sa/extension-tao-proctoring/pull/158 and I would suggest the following small improvements
1. maybe try set the size of the notification box to class .small (it seems currently too big)
2. move it bellow the comment textarea (this way the feedback will be closer to the action trigger)

Then we are ready to go, thanks !

PR #868, opened 2016-06-23
Tao 2709/error exposing

boajer:


bugalot:
@boajer , seems good to me at a first glance. However, do you know any existing code extending the RestController class that I could check so that I can extend my testing?

boajer:
@bugalot as an example you may use such endpoints as http://<taointanse>/taoGroups/Api or http://<taointanse>/taoItems/RestItems
So corresponding code is here https://github.com/oat-sa/extension-tao-group/blob/develop/controller/Api.php and 
https://github.com/oat-sa/extension-tao-item/blob/develop/actions/class.RestItems.php

bugalot:
Thanks @boajer, will proceed then !

PR #869, opened 2016-06-24
Feature/tao 2663 new amd loader

krampstudio:
- Refactor the AMD loaders
 - update the `history` lib name to `html5-history-api` in order to package it
 - added a version cache buster
 - wrap the `login` and the `backoffice` entry point into controllers


krampstudio:
I'm creating a testing env fr this PR and dependencies

krampstudio:
Could be tested from http://spielplatz.taocloud.org/testOptim

PR #884, opened 2016-07-04
added z-index on buttons on the action bar

zagovorichev:
title attribute for title on item authoring (for large titles), and put title on the right side of the action bar

https://oat-sa.atlassian.net/browse/TAO-2726


krampstudio:
Closing for now, see https://github.com/oat-sa/extension-tao-itemqti/pull/695

PR #890, opened 2016-07-08
Add method getId() to Section and Perspective.

dietertaotesting:
They are not used in a regular TAO environment but on creation of id tree for the help files.


dietertaotesting:
redundant code

PR #892, opened 2016-07-11
Fixnon installed extension called

chrisphpgarcia:


models/classes/mvc/error/ExceptionInterpretor.php
```diff
@@ -0,0 +1,117 @@
...
+    protected function interpretError() {
+        $this->trace = $this->exception->getMessage();
+        switch (get_class($this->exception)) {
+            case 'tao_models_classes_AccessDeniedException':
+            case 'ResolverException':   
+                $this->returnHttpCode    = 403;
+                $this->trace = 'Extension or controller not found';
+                $this->responseClassName = 'RedirectResponse';
+            break;
+            case 'tao_models_classes_UserException':
```
bugalot: User exceptions were designed as errors that can be understood by the end user. I'm not sure that 403 is the right way to go, as the client side (TAO JS)  will certainly interpret this as a session timeout. Maybe a 5XX is better there ?

helpers/class.Request.php
```diff
@@ -47,18 +48,31 @@ public static function isAjax()
...
         return (bool) $returnValue;
     }
     
     /**
+     * return false if client is expecting for html (is a real web browser)
+     * true for others cases
+     * @author Christophe GARCIA <christopheg@taotesting.com>
+     * @return boolean
+     */
+    public static function isApiCall() {
```
bugalot: @chrisphpgarcia, Is a request like media (e.g. image, video, sound, ...) or other miscellaneous things (.zip file, ...) also considered as API Call (::isApiCall returns true)? I'm not sure this will suit to any kind of scenarios?

models/classes/mvc/Bootstrap.php
```diff
@@ -214,111 +214,17 @@ public function dispatch()
...
-    		    $previous = $previous->getPrevious();
-    		}
-    		
-    		common_Logger::e($msg);
-    		
-    		$trace = '';
-
-			if (DEBUG_MODE) {
-				$message = $e->getMessage();
-				$trace = $e->getTraceAsString();
```
bugalot: @chrisphpgarcia, Does that mean we loose the PHP backtrace in DEBUG_MODE? It was usefull when dealing with customer errors in case of Field Test or criteria acceptance phase to know directly what is going on.

models/classes/mvc/error/NonAccepteble.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
bugalot: This class should be called "_NonAcceptable_" instead of "_NonAccepteble_" 😀

bugalot:
In addition, in the unit tests, we have an additional error in the unit tests that do not occur on the develop branch:

Undefined index: REQUEST_METHOD in /home/jerome/Development/TAO/taoplatform300/vendor/clearfw/clearfw/core/Request.class.php on line 225

bugalot:
@chrisphpgarcia , following our discussion, checked that the trace is displayed on the 500 error template in debug mode. OK. When disabling debug mode, it disappears. Perfect.

models/classes/mvc/error/ExceptionInterpretor.php
```diff
@@ -0,0 +1,121 @@
...
+            case 'ActionEnforcingException':
+            case 'tao_models_classes_FileNotFoundException':   
+                $this->returnHttpCode    = 404;
+                $this->responseClassName = 'MainResponse';
+            break;
+            default :
+                $this->responseClassName = 'MainResponse';
+                $this->returnHttpCode    = 500;
+            break;
+
```
bugalot: @chrisphpgarcia ,

It seems we do not cover the case 'common_exception_BadRequest'. Indeed, it should return a 400 code in case of such an exception thrown. You can see it in action in taoQtiTest/actions/RestQtiTests.php

bugalot:
@chrisphpgarcia,

This one is unfortunately critical. When using the bootstrap refactoring, the session timeout behaviour on the client side fails. To reproduce, go in the Items extension, and clear your cookes. Click an item, and you will see the application crashing on the client side. This does not occur on develop.

PR #893, opened 2016-07-12
check origin header

hutnikau:
Security team found this issue:

_Application is vulnerable to CSRF that permits ability to add a new user. An attacker can lure an admin user to an HTML page on a different site that is created to automatically submit a request to create a new user on Workkeys.  This requires social engineering a victim user that has an active session and has admin privileges on Workkeys. The proof of concept below demonstrates how this vulnerability can be used to create an administrator level user._

_CSRF HTML Proof of Concept_

``` html
<form action=""https://act-stress.taocloud.org/tao/Users/add"" method=""post"" name=""main"">
<input type=""hidden"" name=""user_form_sent"" value=""1"">
<input type=""hidden"" name=""tao.forms.instance"" value=""1"">
<input type=""hidden"" name=""http_2_www_0_w3_0_org_1_2000_1_01_1_rdf-schema_3_label"" value=""User+5"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_userFirstName"" value=""Test"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_userLastName"" value=""CSRF"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_userMail"" value=""testcsrf@mailinator.com"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_login"" value=""testcsrf"">
<input type=""hidden"" name=""password1"" value=""P@ssw0rd"">
<input type=""hidden"" name=""password2"" value=""P@ssw0rd"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_userDefLg"" value=""http_2_www_0_tao_0_lu_1_Ontologies_1_TAO_0_rdf_3_Langen-US"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_userUILg"" value=""http_2_www_0_tao_0_lu_1_Ontologies_1_TAO_0_rdf_3_Langen-US"">
<input type=""hidden"" name=""http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_userRoles_0"" value=""http_2_www_0_tao_0_lu_1_Ontologies_1_TAO_0_rdf_3_SysAdminRole"">
<input type=""hidden"" name="""" value="""">
<input type=""submit"" id=""btn"">
</form>
<script>
  document.forms[0].submit();
</script>
```

The better approach would be implement checking of csrf token for TAO forms, but this is easier solution. I guess we can implement checking of token later.

### DIFF
--- a/actions/form/class.Users.php
+++ b/actions/form/class.Users.php
@@ -94,7 +94,8 @@ class tao_actions_form_Users extends tao_actions_form_Instance
     	}
     	
     	$options['topClazz'] = CLASS_GENERIS_USER;
-    	
+    	$options['checkOrigin'] = true;
+
     	parent::__construct($clazz, $this->user, $options);
     }
 
--- a/helpers/class.Http.php
+++ b/helpers/class.Http.php
@@ -383,4 +383,20 @@ class tao_helpers_Http
             header('HTTP/1.1 416 Requested Range Not Satisfiable');
         }
     }
+
+    /**
+     * CSRF Protection via Origin Header. Verify request value matches the tao's origin.
+     * @throws common_exception_BadRequest
+     */
+    public static function checkOrigin()
+    {
+        $request = new Request();
+        if ($request->getMethod() === Request::HTTP_POST && $request->hasHeader('origin')) {
+            $rootInfo = parse_url(ROOT_URL);
+            $originInfo = parse_url($request->getHeader('origin'));
+            if (!isset($originInfo['host']) || $originInfo['host'] !== $rootInfo['host']) {
+                throw new \common_exception_BadRequest("CSRF protection in POST request: detected invalid Origin header: " . $_SERVER["HTTP_ORIGIN"]);
+            }
+        }
+    }
 }
--- a/helpers/form/xhtml/class.Form.php
+++ b/helpers/form/xhtml/class.Form.php
@@ -152,21 +152,23 @@ class tao_helpers_form_xhtml_Form
      */
     protected function validate()
     {
-        $returnValue = (bool) false;
+        $this->valid = true;
 
-        
-		
-		$this->valid = true;
-		
-		foreach($this->elements as $element){
-			if(!$element->validate()){
-				$this->valid = false;
-			}
-		}
-		
-        
+        foreach($this->elements as $element){
+            if(!$element->validate()){
+                $this->valid = false;
+            }
+        }
 
-        return (bool) $returnValue;
+        if (isset($this->options['checkOrigin']) && $this->options['checkOrigin'] === true) {
+            try {
+                \tao_helpers_Http::checkOrigin();
+            } catch (\common_exception_BadRequest $e) {
+                $this->valid = false;
+            }
+        }
+
+        return $this->valid;
     }
 
 }
--- a/manifest.php
+++ b/manifest.php
@@ -27,7 +27,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '5.2.0',
+    'version' => '5.2.1',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => array(
         'generis' => '>=2.22.1'
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -504,6 +504,8 @@ class Updater extends \common_ext_ExtensionUpdater {
         }
 
         $this->skip('2.22.0', '5.2.0');
+        
+        $this->skip('5.2.0', '5.2.1');
     }
     
     private function migrateFsAccess() {


krampstudio:
FYI we already have a XSRF token management that waits someone to generalize it and apply him everywhere... you can have a look at https://github.com/oat-sa/extension-tao-test/tree/master/models/classes/runner
However, using a token requires a strict workflow in HTTP requests (order needs to be guarantee). 

cc @jbout

krampstudio:
Hum, your approach won't work. The only way to prevent XSRF is to ensure the received response is linked to the previous request, by identifying the previous the requests (using a token).
In your approach, you can fake the `origin` header the same way you can fake any other request part.

Even if we don't implement the token in every form, we should do it for this one : 
 1. move the token classes to tao (or generis ? - ask joel).
 2. Generate a token to add to the form.
 3. Check the response contains the generated token.

jbout:
@krampstudio Is it possible to forge the headers using java-script?

concerning (1) : Tao is handling everything relating to web requests, so tao-core would be the appropriate extension.

krampstudio:
Yes, you can [create headers in javascript](https://developer.mozilla.org/en-US/docs/Web/API/Headers).

hutnikau:
You can set headers but you can't set `Origin` header. As far as I know the value of `Origin` cannot be changed form javascript. Otherwise this header does not make sense.

Try to run this in console:
```javascript
$.ajax({
    type: 'POST',
    url: '/',
    beforeSend: function (request) {
        request.setRequestHeader("Origin", 'fake');
    },
}).done(function(data) { 
    console.log(data);
});
```
You will see `Refused to set unsafe header "Origin"` error

hutnikau:
I saw all the classes in taoTest extension related to XSRF token. It seems to be very useful thing and it definitely should be moved to tao extension, but as I said it assumes a lot of work whereas my solution is simple and fast.

krampstudio:
Ok, 
First, the forbidden headers are a limitation that's not implemented the same way across browser. Your example is a jQuery polyfill of the expected behavior of setRequestHeader... 

And I think there's a misunderstanding of the problem. 
We don't care if the request can be forged using JavaScript. You can forge a request with any other tool. For example, you can replay requests in the context of a session using the developer tools.

Then the origin doesn't guarantee anything in that context. It only guarantee the request comes from same domain. An XSRF attack usually consist with a valid user on the same system that have lower privileges and will forge a request to run an action he shouldn't. 

I think the given "proof of concept" leads you to the wrong direction.

You must use a token.

hutnikau:
Example in "proof of concept" is the simplest and most usable way to do CRSF attack. Of course request can be forged using any different tool, but in this case malefactor must have cookies (session id) to perform such request. Example above shows how you can execute any request just by sending link on page which contains form to logged in user. 

Origin must be checked and this checking is the first thing which should be done to improve security.

Look at this article:
https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#General_Recommendations_For_Automated_CSRF_Defense

Checking of token is next necessary step, but checking of Origin header is also necessary action.

krampstudio:
Again, checking the Origin header is something to be implemented by the browsers, and protects you only if you run the query from another web page if you have a recent browser. 
Recent browser will do the check for you everywhere Same Origin Policy applies, and old browser will let you forge this header, for example the Origin header is in the forbidden list only since Firefox 43...

You can look at this discussion http://security.stackexchange.com/questions/91165/why-is-the-synchronizer-token-pattern-preferred-over-the-origin-header-check-to

hutnikau:
Origin header supports by all modern browsers:
http://caniuse.com/#search=CORS

Not only recent browsers use this header. For example IE use it since 8 version.

This fix avoid to use such vulnerability as shown in example, in my opinion checking of this header is absolutely necessary for POST requests. And all recommendations advice to check it.

krampstudio:
No you don't understand. 
This form is only one way to create a request, you won't prevent other ways. It's not a valid protection if all the mediums aren't protected. 
It's even worth because the customer will think the problem is solved when it is not. 
What prevent you to add a token ?

You know in the reality most case of XSRF are executed from emails (or even from the same site when they've been installed using an XSS), do you know if it's possible to change the Origin header in the embed browsers of Outlook Express or whatever ? I don't.  
Obviously checking origin from another site is something that make sense, but if the goal is to protect again a case of XRSF, then this is not what we want.

hutnikau:
JavaSrcipt code cannot be executed in email clients. This is security restriction.
The only way how checking of the Origin header can be bypassed is embedding html/javascript on the page of Tao. But this is another issue. All html/JS code should be sanitized to prevent embedding it on page by somebody.

I can't start work on tokens now because it involves a lot of work. I guess such work should be scheduled by Thomas for one of the following sprints and appropriate task should be created in Jira.

krampstudio:
Closing because the token API has been implemented by https://github.com/oat-sa/tao-core/pull/1295

hutnikau:
I still think we need to enable same-origin policy restriction. All the wikis and security tutorials suggest that as a first step to protect web application against CRSF attack.

PR #894, opened 2016-07-13
Tao 2715/roles users logging

boajer:
depends on https://github.com/oat-sa/generis/pull/205


models/classes/class.UserService.php
```diff
@@ -522,4 +526,16 @@ public function getRootClass()
...
+
+    /**
+     * @param core_kernel_classes_Class $clazz
+     * @param string $label
+     * @return core_kernel_classes_Resource
+     */
+    public function createInstance(core_kernel_classes_Class $clazz, $label = '')
+    {
+        $user = parent::createInstance($clazz, $label); // TODO: Change the autogenerated stub
+        $this->getEventManager()->trigger(new UserCreatedEvent($user));
```
Alroniks: This line is causing the error `Fatal error: Call to undefined method tao_models_classes_UserService::getEventManager() in /home/vagrant/Code/tao/act/tao/models/classes/class.UserService.php on line 538`

PR #911, opened 2016-07-18
Refactor/directory fly interface

hutnikau:


models/classes/service/class.StorageDirectory.php
```diff
@@ -218,4 +245,25 @@ public function getIterator()
...
+        return rtrim($this->getRelativePath(), '\\/') . '/' . ltrim($path, '\\/');
+    }
+
+    /**
+     * @param string $path
+     * @return string
+     */
+    protected function fixSlashes($path)
+    {
+        if (!$this->getFileSystem()->getAdapter() instanceof Local) {
```
siwane: undefined

hutnikau: [Fixed](https://github.com/oat-sa/tao-core/pull/911/commits/8c181cdd339bcb9178439e2ce8e84b1dbb610c98)

siwane: Does this method has to do the reverse?

> **>>>>!<<<<** $this->getFileSystem()->getAdapter() instanceof Local

models/classes/service/class.StorageDirectory.php
```diff
@@ -111,32 +112,58 @@ public function getPublicAccessUrl() {
...
     }
 
     /**
+     * Store a file in the directory from string
+     *
+     * @param string $path
+     * @param string $content
+     * @return boolean
+     */
+    public function write($path, $content, $mimeType = null)
```
siwane: Config parameters for write method does not take under consideration the mimetype for local, dropbox, ftp.. but for AWS yes.
@jbout what do u think about that?

@see https://github.com/thephpleague/flysystem/blob/master/src/Adapter/Local.php#L136
@see https://github.com/thephpleague/flysystem-aws-s3-v3/blob/master/src/AwsS3Adapter.php#L561

jbout: Not all file-systems support the mime-type storage as a separate field. The local file-system relies on a function Util::guessMimeType() to guess the mime-type from the content.

This is not blocking, but something to keep in mind when selecting our storage.

models/classes/service/class.StorageDirectory.php
hutnikau: No, reverse is not necessary here.

PR #912, opened 2016-07-19
Added additional check in case, when http accept not defined in SERVE…

Alroniks:
…R variable


Alroniks:
@chrisphpgarcia Please, take a look at this fix when you will have time.

PR #914, opened 2016-07-19
Use named filesystems for services

jbout:


jbout:
Now depends on https://github.com/oat-sa/generis/pull/207

PR #917, opened 2016-07-20
TAO-2811 Now you can deselect all values from a treeview.

bugalot:
Without this patch, it is impossible to remove all the selected elements of a Treeview widget. You can check that by installing taoClientRestrict. Without this patch, it's impossible to unslect all the browsers/os previously checked.


boajer:
As it has been assigned to me initially I took a look and seems it works, at least in context of OS/Browser trees

PR #926, opened 2016-07-22
Refreshing tree after edit instane

hutnikau:
See comments

### DIFF
--- a/views/js/layout/tree.js
+++ b/views/js/layout/tree.js
@@ -222,6 +222,11 @@ define([
                     var treeState       = $elt.data('tree-state') || {};
                     var selectNode      = treeState.selectNode || options.selectNode;
                     var nodeSelection   = function nodeSelection(){
+                        //afrer refreashing tree previously node will be already selected.
+                        if (tree.selected !== undefined) {
+                            return tree.selected;
+                        }
+
                         //the node to select is given
                         if(selectNode){
                              $selectNode = $('#' + selectNode, $elt);
--- a/views/js/layout/tree.js
+++ b/views/js/layout/tree.js
@@ -222,10 +222,6 @@ define([
                     var treeState       = $elt.data('tree-state') || {};
                     var selectNode      = treeState.selectNode || options.selectNode;
                     var nodeSelection   = function nodeSelection(){
-                        //afrer refreashing tree previously node will be already selected.
-                        if (tree.selected !== undefined) {
-                            return tree.selected;
-                        }
 
                         //the node to select is given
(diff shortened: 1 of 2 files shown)


views/templates/footer.tpl
```diff
@@ -15,7 +15,7 @@ use oat\tao\helpers\Template;
 <?php endif; ?>
 <?php if(get_data('reload')): ?>
     require(['jquery'], function ($) {
-        $('.tree').trigger('refresh.taotree', [{
+        $('.tree:visible').trigger('refresh.taotree', [{
```
hutnikau: Without this fix all the trees will be updated and it lead to bugs. Steps to reproduce:
1. Go to "Settings" -> "Web Browsers List" page
2. Then select  "Settings" -> "OS Lists" page (both trees exists in DOM, only OS tree is visible)
3. Select any node in tree and press "Save" button on the form

See three `getOntotogyData` requests and five (!) `editInstance` requests
See Web Browser form instead OS form.

views/js/layout/tree.js
```diff
@@ -222,6 +222,11 @@ define([
                     var treeState       = $elt.data('tree-state') || {};
                     var selectNode      = treeState.selectNode || options.selectNode;
                     var nodeSelection   = function nodeSelection(){
+                        //afrer refreashing tree previously node will be already selected.
+                        if (tree.selected !== undefined) {
+                            return tree.selected;
+                        }
+
```
hutnikau: After saving instance tree reloads and selects node automatically. Without this fix node selected twice (once automatically  and one manually).

Steps to reproduce:
1. Go to any page with tree (for example "Tests").
2. Select any node and press "Save" on the form.
See two `editInstance` requests in row.

krampstudio:
That's a good catch! 
But I'd like to avoid using `:visible` as far as much as possible. The main reason is this selector cause repaint and redraw of stacking contexts. 
In replacement you can retrieve the displayed panel to contextualize the selection : 

```js
    require(['jquery', 'layout/section'], function ($, section) {
        var $trees;
        var currentSection = section.current().selected;
        if(currentSection && currentSection.panel && currentSection.panel.length){
            $trees = $('.tree', currentSection.panel);
        } else {
            $trees = $('.tree:visible');
        }
        $trees.trigger('refresh.taotree', [{
            selectNode : <?=json_encode(get_data('selectNode'))?>
        }]);
    });
```

hutnikau:
@krampstudio thanks for your comment. The branch has been updated. Please review it again.

PR #931, opened 2016-07-27
Fixed several unit tests

jbout:


bugalot:
It's better, but some tests are still failing !

PR #932, opened 2016-07-27
Only numerics can be passed to time-limit field

boajer:


krampstudio:
Could you please also check the case in the unit test ?

boajer:
@krampstudio I've added new unit test to reflect changes, could you check it again please.

PR #946, opened 2016-08-04
Remove deprecated conflicting getClass()

jbout:


chrisphpgarcia:
no more getClass method founds.

PR #948, opened 2016-08-05
Feature/tao 2853 unique store

krampstudio:
The goal of this feature is to be able to identify a store. So, if we move to another store (ie. another browser or to private mode) we have a way to detect the change.


krampstudio:
:warning:  Wait, I've forgotten a part...

PR #949, opened 2016-08-05
base64 select2 icon in sass to fix missing icon inclusion

ssipasseuth:


dietertaotesting:
This breaks the down arrow on regular select 2. As discussed 

- remove all refs to the original image
- add a comment  referring to f9dd14d0185142223f1649a98aa85f7c12dfa924 which still has the picture)
- write a ::before style that adds the code of _icon-result-nok_

ssipasseuth:
I remove the retina media query, and all seems to work fine.
could you double check please ? 
thanks !

PR #951, opened 2016-08-08
Added check if controller exists before call start methods

Alroniks:
https://github.com/oat-sa/extension-tao-mpart/pull/295


ssipasseuth:
This PR is no longer required.

PR #957, opened 2016-08-11
Feature/tao 2899 pdf viewer

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2899

First part of the work. It only provide the fallback to the native browser implementation.

The second part, using the PDF.js library is still under work. It will be submitted soon.


ssipasseuth:
oops, pdfViewer/test.html is failling in phantomJs
```
grunt connect qunit:single --test=/tao/views/js/test/ui/documentViewer/pdfViewer/test.html
Running "connect:test" (connect) task
Started connect web server on http://127.0.0.1:8082

Running "qunit:single" (qunit) task
Testing http://127.0.0.1:8082/tao/views/js/test/ui/documentViewer/pdfViewer/test.html .
>> PhantomJS timed out, possibly due to:
>> - QUnit is not loaded correctly.
>> - A missing QUnit start() call.
>> - Or, a misconfiguration of this task.
Warning: 1/1 assertions failed (0ms) Use --force to continue.

Aborted due to warnings.
```

jsconan:
Yes, this is due to the iframe. I suppose PhantomJS does not support the PDF format... In that case, the fallback won't work, and we'll need to discard this test with headless browsers

jsconan:
I updated the component, with use of PDF.js if installed

ssipasseuth:
Would need to find a better way to pass file url into the pdf previewer (especially base64 encoded file data, in e.g. uploadInteraction)

PR #965, opened 2016-08-19
Fix/tao 2929 calc roundoff

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-2929

Add a third-party library to fix round-off errors on floating-point computations.

Patch the calculator to use this library.

A noteworthy article about floating-point and related issues: http://blog.chewxy.com/2014/02/24/what-every-javascript-developer-should-know-about-floating-point-numbers/


jsconan:
Also contains the fix for https://oat-sa.atlassian.net/browse/TAO-2918

PR #973, opened 2016-08-24
expand the property form if there is an error in it

antoinerobin:


views/js/form/post-render-props.js
```diff
@@ -218,6 +218,25 @@ define([
             _upgradeButtons($container, 'radio');
             _upgradeButtons($container, 'checkbox');
             _toggleModeBtn('disabled');
+            _showErrors($properties);
+        }
+
+        function _showErrors($properties){
```
ssipasseuth: This works but a more performant way would be searching for error before ... unless the parent selector cannot be defined easily
```
$error = $container.find('.error');
if($error.length){
    $error.parents('selector')
}

PR #976, opened 2016-08-30
Feature/event context

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3008

Add event name and namespace in the context object provided to the `before` handlers.
```javascript
emitter.before('foo', function(e) {
    // name of event
    console.log(e.name);

    // namespace of event
    console.log(e.namespace);
});
```

Add global namespace to listen to every namespaced events:

```javascript
// listen to "foo" only
emitter.on('foo');

// listen to "foo" and "foo.bar"
emitter.on('foo.bar');

// listen to all "foo" events
emitter.on('foo.*');

// notify all "foo" listeners
emitter.trigger('foo');

// notify only "foo.bar" and "foo.*"
emitter.trigger('foo.bar');
```

Can now use promises in async event:

```javascript
emitter.before('foo', function() {
    return new Promise(resolve, reject) {
        if (ok) {
            resolve();
        } else {
            reject();
        }
    });
});
```

Add method to erase all registered handlers:

```javascript
emitter.removeAllListeners();
```

### DIFF
--- a/views/js/core/eventifier.js
+++ b/views/js/core/eventifier.js
@@ -289,23 +289,23 @@ define([
             return eventHandlers[ns][name][type];
         };
 
-       /**
-        * The API itself is just a placeholder, all methods will be delegated to a target.
-        */
+        /**
+         * The API itself is just a placeholder, all methods will be delegated to a target.
+         */
         var eventApi = {
 
-           /**
-            * Attach an handler to an event.
-            * Calling `on` with the same eventName multiple times add callbacks: they
-            * will all be executed.
-            *
-            * @example target.on('foo', function(bar){ console.log('Cool ' + bar) } );
-            *
-            * @this the target
-            * @param {String} eventNames - the name of the event, or multiple events separated by a space
-            * @param {Function} handler - the callback to run once the event is triggered
-            * @returns {Object} the target object
-            */
+            /**
+             * Attach an handler to an event.
+             * Calling `on` with the same eventName multiple times add callbacks: they
+             * will all be executed.
+             *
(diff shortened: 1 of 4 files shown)


krampstudio:
:metal: This is a great improvment!

views/js/core/eventifier.js
```diff
@@ -56,32 +72,40 @@
  *      });
  * });
  *
- * TODO replace before done syntax by promises
+ * TODO replace before done syntax by promises (work in progress: promise support added, need now to update every extension with new syntax, then finish the work)
```
krampstudio: Could you elaborate ? 

My plan was to be able to return a Promise inside a before callback : 

```js
emiter.before('foo', function(){

      return Promise.resolve();
});
```
is equivalent to 

```js
emiter.before('foo', function(e){
     e.done();
});
```

views/js/test/core/eventifier/test.js
```diff
@@ -95,16 +109,19 @@ define(['core/eventifier'], function(eventifier){
...
-        QUnit.expect(2);
-
         var emitter = eventifier();
 
+        QUnit.expect(3);
+
         emitter.on('foo', function(){
             assert.ok(true, 'the foo handler is called');
         });
+        emitter.on('foo.*', function(){
```
krampstudio: Could you also add a test case where you listen to the default ns explicitely : 

```js
emitter.on('foo.@', cb);
```

The same for trigger, even if there's no reason to fail.

jsconan:
I made the requested ajustments

PR #990, opened 2016-09-20
Fix/exporter error

siwane:


krampstudio:
Is there any issue linked ?
Could you  please provides me steps to reproduce and what's expected now.

siwane:
When an error occurs at the export (for exemple, no portable element ffound to export):
- before : no message
- expected: a view displays an error

Tell me if u want more details

krampstudio:
Getting back on this old issue : 
tao3 2 0-sprint40-firefoxdeveloperedition_168

Is it the message that should be displayed ?

siwane:
It's the output of tao_helpers_report_Rendering::render($report) and the type of report is error. An error at the export is very special case, I think error message is enough.

views/templates/form/export.tpl
```diff
@@ -1,68 +1,65 @@
...
+                            }else{
+                                params[param.name] = param.value;
+                            }
+                        });
+                        params.instances = instances;
+
+                        $.fileDownload(helpers._url("<?=get_data('export_action')?>", "<?=get_data('export_module')?>", "<?=get_data('export_extension')?>", params), {
+                            failCallback: function (html) {
+                                $('#export-container').html(html);
+                                $('#export-container > .feedback-nesting-0 ').removeClass('feedback-error');
```
krampstudio: Why do you remove the _error_ style ? I think you should keep it.

PR #998, opened 2016-09-26
Fix/tao 3147 pdf chrome issue

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3147

Fix the render issue of PDF under Chrome/Safari and IE.

Fix page navigation issue.

You will need PDF.js installed.
Report to this page to install it: http://forge.taotesting.com/projects/tao/wiki/Install_PDFjs_viewer

Steps to reproduce:
- Create an item and add some interaction to it.
- Attach a PDF file to this item by editing the prompt and inserting a media file (this file needs to be big enough to be streamed).
- A document viewer is inserted, but you do not see any content, and the console display a "Bad offset" error

### DIFF
--- a/views/js/ui/documentViewer/providers/pdfViewer.js
+++ b/views/js/ui/documentViewer/providers/pdfViewer.js
@@ -396,8 +396,8 @@ define([
                                 self.pdf.refresh();
                             });
 
-                            self.controls.navigation.on('click', function (e) {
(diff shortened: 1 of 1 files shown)


jsconan:
The fix has been embedded in the refactoring PR: https://github.com/oat-sa/tao-core/pull/1003

PR #1000, opened 2016-09-27
Feature/multiple datatable filters

hutnikau:


krampstudio:
The approach is correct but instead using your own `strategyFactory` you should  use `core/providerRegistry`.

hutnikau:
I changed `filterStrategy.js` in order to use `providerRegistry`.

hutnikau:
@krampstudio I updated the branch according to your comments. Please review it again.

PR #1003, opened 2016-09-28
Refactoring of PDF viewer (split into submodules)

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3153
Main task: https://oat-sa.atlassian.net/browse/TAO-3131

Refactoring of the PDF viewer to isolate each functionality, especially the fallback and the PDF.js wrapper.

The goal is to facilitate the extension of this viewer.
Also the unit test coverage will be better.

This PR also embed the fix regarding the rendering issue in Chrome/Safari and IE: https://github.com/oat-sa/tao-core/pull/998

You will need PDF.js installed.
Report to this page to install it: http://forge.taotesting.com/projects/tao/wiki/Install_PDFjs_viewer

This is the first of three tasks. A integration branch will be used to gather the code. So the PR merge time should not matter.


views/js/test/ui/documentViewer/pdfViewer/fallback/test.js
```diff
@@ -0,0 +1,103 @@
...
+
+        QUnit.expect(13);
+
+        assert.equal($container.children().length, 0, 'The container does not contain any children');
+
+        instance = fallbackFactory($container);
+        promise = instance.load(pdfUrl);
+
+        assert.equal(typeof promise, 'object', 'The load() function returns an object');
+        assert.equal(typeof promise.then, 'function', 'The object returned by the load() function is a promise');
```
krampstudio: I'm used to do check for Promise using `instanceof` : 
`assert.ok(promise intanceof Promise, 'The load() function returns a Promise');`

This is just informational, you don't need to change your code.

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/viewer.js
```diff
@@ -0,0 +1,226 @@
...
+        var controls = {};
+        var pdf = null;
+        var enabled = true;
+
+        /**
+         * Will update the displayed page number, and toggle the input enabling
+         */
+        function updatePageNumber() {
+            var page = pdf.getPage();
+            if (page !== Number(controls.$pageNum.val())) {
```
krampstudio: Isn't `parseInt` more appropriate ? Casting using `Number` is more capricious. For example : 
```js
var val = "12a";
parseInt(val, 10) //-> 12
Number(val) //->NaN
```
Use `Number` only if you want a float or a int.

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/wrapper.js
```diff
@@ -0,0 +1,330 @@
...
+        return {
+            /**
+             * Loads a PDF document using PDF.js
+             * @param {String} url
+             * @returns {Promise}
+             */
+            load: function load(url) {
+                unload();
+                return PDFJS.getDocument(processUri(url)).then(function (pdfDoc_) {
+                    pdfDoc = pdfDoc_;
```
krampstudio: Could you rename `pdfDoc_` to something meaningful like `newPdfDoc`

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/viewer.js
jsconan: You are right, just a bad habit. Changes made.

PR #1007, opened 2016-09-29
Fix/php7.1 compatibility

boajer:
The reason is to improve compatibility with PHP 7.1


helpers/class.Environment.php
```diff
@@ -84,15 +84,14 @@ public static function getOperatingSystem()
...
+     * @param  string $phpSyntax The PHP syntax to describe the variable.
      * @return int The size in bytes.
      */
     private static function toBytes($phpSyntax)
     {
-        $returnValue = (int) 0;
-
         $val = trim($phpSyntax);
         	$last = strtolower($val[strlen($val)-1]);
+        $val = (int)trim($val, $last);
```
bugalot: what if $first === $last ? :-)

boajer: Do you mean if value in php_ini is incorrect and looks like upload_max_filesize = 'M' ?

jbout: why not just use (int)substr($val, 0, -1) ?

bugalot:
Hi @boajer,

In order to understand a little bit better the cases where $last = 'g' or 'm' or 'k', could you please just show me one or two invocation samples of the toBytes() method ? I'll had unit tests then and try with multiple versions of PHP.

Thanks,
Jérôme

bugalot:
Finally got it... :)

PR #1017, opened 2016-09-30
Feature/temp filsystem unit test

siwane:


boajer:
Maybe it would be better to use something like https://github.com/thephpleague/flysystem-memory instead of real filesystem?

siwane:
Memory adapter will be used if installed:

> composer require league/flysystem-memory

PR #1018, opened 2016-10-03
Added ability to remove NonEmpty check from password field in login form

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-3139

### DIFF
--- a/actions/form/class.Login.php
+++ b/actions/form/class.Login.php
@@ -19,79 +19,56 @@
  * 
  */
 
-use oat\tao\helpers\Layout;
-
 /**
  * This container initialize the login form.
- *
  * @access public
  * @author Bertrand Chevrier, <bertrand.chevrier@tudor.lu>
  * @package tao
- 
  */
-class tao_actions_form_Login
-    extends tao_helpers_form_FormContainer
+class tao_actions_form_Login extends tao_helpers_form_FormContainer
 {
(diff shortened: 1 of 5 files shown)


actions/form/class.Login.php
```diff
@@ -19,79 +19,56 @@
...
+
+        $loginElt = tao_helpers_form_FormFactory::getElement('login', 'Textbox');
+        $loginElt->setDescription(__("Login"));
+        $loginElt->setAttributes(['autofocus' => 'autofocus']);
+        $loginElt->addValidator(tao_helpers_form_FormFactory::getValidator('NotEmpty'));
+        $this->form->addElement($loginElt);
+
+        $passElt = tao_helpers_form_FormFactory::getElement('password', 'Hiddenbox');
+        $passElt->setDescription(__("Password"));
+        if (!empty($this->data['emptyPassword'])) {
```
antoinerobin: this field is never set, you didn't pass the params to the constructor

Alroniks:
@antoinerobin Fixed issue and version

Alroniks:
@antoinerobin After discussion with a customer this feature not needed more for them so I can close it but maybe it may need for us in the future. What do you think?

antoinerobin:
I think you can close it, I don't see a use case for an empty password

Alroniks:
No need more

PR #1024, opened 2016-10-10
Feature/tao 3154 pdf viewer search

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3154

Implement a search engine using the PDF.js API.

PDF.js just provides a rendering API, so most of the UI features have to be added by our own.
By default, the PDF is rendered in a canvas, and the text is not accessible by the viewer. We can request the API to get access to the text, and then process a search in it. However we cannot update the canvas to make the search matches highlighted. To display the search results we need to add a layer above the canvas that will contain the text, so we are able to highlight the results.

Hopefully, the rendering of the text layer is handled by PDF.js. But we still need to make the link between the text layer and the canvas. This is the purpose of this PR: implement the search engine and provides a text layer to display the results.

To do so, some changes have been made. The wrapper has been split into several modules:
- a pages managers has been added to manages the views (for now only one for all pages, but we can manage to have one view per page in order to allow scrolling across pages)
- the view itself is handled by a pageView, which is responsible of its rendering
- a text manager centralizes the access to the PDF text content, and also handles the rendering of the text layer

Then the search engine itself, which relies on the text manager to access to the text and update the layer.

This PR only covers the search engine feature, the UI will be addressed by another task.


views/js/ui/documentViewer/providers/pdfViewer/pdfjs/wrapper.js
```diff
@@ -289,41 +194,82 @@ define([
...
+                    destroyed: true
+                };
             }
         };
+
+        config = config || {};
+        PDFJS = config.PDFJS;
+
+        if ('object' !== typeof PDFJS) {
+            throw new TypeError('You must provide the entry point to the PDS.js library! [config.PDFJS is missing]');
```
krampstudio: there's a typo here: PDS.js

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/textManager.js
```diff
@@ -0,0 +1,250 @@
...
+                return pdfDoc;
+            },
+
+            /**
+             * Gets the content of the PDF
+             * @returns {Promise}
+             */
+            getContents: function getContents() {
+                if (pdfDoc) {
+                    return getPageContents();
```
krampstudio: Should be `this.getPageContents()` or you'll loose your context

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/textManager.js
```diff
@@ -0,0 +1,250 @@
...
+             * @param {Number} pageNum
+             * @param {Object} viewport
+             * @param {Number} [timeout]
+             * @returns {Promise}
+             */
+            renderPage: function renderPage(pageNum, viewport, timeout) {
+                return this.getContents().then(function (content) {
+                    var pageIndex = Math.min(Math.max(0, pageNum - 1), content.length - 1);
+                    var pageContent = content[pageIndex];
+                    var textLayerFrag = document.createDocumentFragment();
```
krampstudio: Could you please just do a quick test on IE11. I had bad experiences with `DocumentFragments`  on IE in the past.

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/viewer.js
```diff
@@ -121,6 +124,11 @@ define([
         }
 
         config = config || {};
+        PDFJS = config.PDFJS;
+
+        if ('object' !== typeof PDFJS) {
+            throw new TypeError('You must provide the entry point to the PDS.js library! [config.PDFJS is missing]');
```
krampstudio: We are in the same case (for the typo also)

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/wrapper.js
```diff
@@ -91,163 +67,92 @@ define([
...
-            } else {
-                return Promise.resolve(num);
-            }
-        }
+        var wrapper = {
+            /**
+             * The wrapped API (i.e.: the PDF.js library)
+             * @type {Object}
+             */
+            get wrapped() {
```
krampstudio: Did you checked the `getter`syntax is well supported ?

krampstudio:
Code and test review only.

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/textManager.js
jsconan: Well... This is not a method but a closure. Should I transform it to an object's method ?

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/textManager.js
jsconan: Yes, I have made a check under IE11, and it works. In facts, this code is inspired from the PDF.js official viewer. (https://github.com/mozilla/pdf.js/blob/master/web/text_layer_builder.js#L100)

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/wrapper.js
jsconan: Yes, this feature, in its simple form, is supported since IE9. In facts, PDF.js does a strong use of this feature, so if it is not supported, PDF.js won't work at all.

https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Fonctions/get

views/js/ui/documentViewer/providers/pdfViewer/pdfjs/textManager.js
krampstudio: I've mixed it with `getPageContent`

krampstudio:
I've reviewed the PR and merge it against develop. But this PR is against master... I'll close it

PR #1029, opened 2016-10-14
Feature/tao 3155 pdf viewer search ui

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3155

Yet another large pull request. Sorry!

Implements a find bar that relies on the search engine to search terms within the viewed PDF document.

The Unit Tests do not need PDF.js installed. However if you want to test manually the interface you will have to install it.

Some important changes have been made in the architecture of the viewer:
- an area broker has been added to simplify the access to the UI
- an events hub is used to centralize the communication between main components.


views/js/ui/documentViewer/providers/pdfViewer/pdfjs/findBar.js
```diff
@@ -0,0 +1,467 @@
...
+            if (query) {
+                if (searchEngine) {
+                    /**
+                     * Notifies a search is processing
+                     * @event searching
+                     * @param {String} query
+                     */
+                    events.trigger('searching', query);
+
+                    searchEngine.search(query, pageNum).then(function (page) {
```
jsconan: In fact the catch is not needed since the search will just do nothing in case of error. However I added a catch that trigger an `error` event.

PR #1034, opened 2016-10-18
fix French translation mistakenly used for Italian UI

chazotte:


chazotte:
I cancel this request then close this PR, as the backport should also fix this issue without reinstalling (needs this in Updater.php script). I will submit a stronger PR.

PR #328, opened 2015-05-19
Tao 737 component autocomplete

jsconan:

PR #1044, opened 2016-10-26
Feature/tao 3251 shortcut namespace

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3251

Refactor the `shortcut` module in order to:
- allow the use of namespaces
- provide the ability to register several listener for a same shortcut
- allow to target an other element than the window

A factory has been added, and the existing component has been updated to rely on the factory to target and listen to the window.

Note: the code coverage is not 100% because the component provides cross-browser code paths, and not all code paths can be reached by the tested browser. We need to figure out how test those code paths in a single browser.


views/js/test/util/shortcut/registry/test.js
```diff
@@ -0,0 +1,689 @@
...
+            assert.ok(true, 'The shortcut has been caught');
+            assert.equal(typeof event, 'object', 'The event object is provided');
+            assert.equal(keystroke, 'meta+c', 'The keystroke is provided');
+            shortcuts.remove('Meta+C');
+            QUnit.start();
+        });
+
+        assert.equal(res, shortcuts, 'The helper returns itself');
+
+        $target.simulate('keydown', {
```
jsconan: Yes if it was not in a unit test, this is what I would do.
However unit test is the only place where inlining and cut/paste are authorized and even encouraged as this prevent to introduce bugs with third parties.

https://github.com/oat-sa/tao-core/pull/328#r30785347

PR #1054, opened 2016-11-04
Feature/configurable default url

chrisphpgarcia:


antoinerobin:
what if we pass here : 
* https://github.com/oat-sa/tao-core/blob/master/actions/class.Main.php#L78
* https://github.com/oat-sa/tao-core/blob/master/views/templates/passwordRecovery/password-recovery-info.tpl#L18

PR #1061, opened 2016-11-11
Feature/datatable model action type

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-3305
https://oat-sa.atlassian.net/browse/TAO-3308
https://oat-sa.atlassian.net/browse/TAO-3309

### DIFF
--- a/views/js/test/ui/datatable/test.js
+++ b/views/js/test/ui/datatable/test.js
@@ -214,6 +214,82 @@ define([
         }, dataset);
     });
 
+    QUnit.asyncTest('Model loading with the "action" type property using predefined data', function(assert){
+        QUnit.expect(11);
+
+        var $elt = $('#container-1');
+        assert.ok($elt.length === 1, 'Test the fixture is available');
+
+        QUnit.stop(4);
+
+        $elt.on('create.datatable', function(){
+            assert.ok($elt.find('.datatable').length === 1, 'the layout has been inserted');
+            assert.ok($elt.find('.datatable thead th').length === 7, 'the table contains 7 heads elements');
+            assert.equal($elt.find('.datatable thead th:eq(6) div').text(), 'Pause', 'the pause label is created');
+            QUnit.start();
+        });
+        $elt.on('query.datatable', function(event, ajaxConfig) {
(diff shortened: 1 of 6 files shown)


views/js/test/ui/datatable/test.js
```diff
@@ -214,6 +214,217 @@ define([
...
+            },{
+                id: 'pauseCl',
+                label: 'Pause',
+                type: 'actions',
+                actions: [{id: 'pause'}]
+            },{
+                id: 'administration',
+                label: 'Administration',
+                type: 'actions',
+                actions: [{id: 'pause'}, {id: 'stop'}, {id: 'run'}]
```
ssipasseuth: I would have only one question: why did you choose to assign actions first in the datatable root-level configuration then reassign them by their id in the column model ?
I can only see one use case for doing it this way, which is to enable multiple columns to share the same action but maybe I missed something.

views/js/ui/datatable.js
```diff
@@ -90,6 +92,7 @@ define([
          * @param {Function} options.listeners.xxx - the callback function for event xxx, parameters depends to event trigger call.
          * @param {Boolean} options.selectable - enables the selection of rows using checkboxes.
          * @param {Boolean} options.rowSelection - enables the selection of rows by clicking on them.
+         * @param {Boolean} options.showActions - show added actions in new column 'Actions' (default 'true').
```
ssipasseuth: Sorry, but I am afraid that adding an extra option here will confuse future people using the datatable.
Indeed, this showActions option is too tightly linked to the list of actions. 
My suggestion was rather to have the new column type=action stand-alone (so we dont have to define a global actions list and link it by id in order to activate an action into a column in the middle). 
The legacy option "actions" can stay as an alias for retrocompability but new feature should not rely on it any more. 
This means that the callbacks need to be defined within the column directly.
Does this makes sense ? What do you think about it ?

zagovorichev: I think that all these changes make datatable component more difficult for understanding. If I don't use actions, then I'll create same initialization for newActions like for actions and tools

PR #1073, opened 2016-11-23
Fix/dynamic component resize

krampstudio:
When trying to resize a component close to the container border the resize was prevented (move a component to the left border of the container increase the width, then you weren't always able to reduce the width).

I've also changed the "reset size" icon to something more meaningful.


views/scss/inc/_dynamic-component.scss
```diff
@@ -25,7 +25,7 @@
             padding-left : 5px;
             &:before{
                 @include tao-icon-setup;
-                @include icon-reset;
+                @include icon-resize;
```
ssipasseuth: The icon-resize looks like a maximize button so might be confusing (but that is better than existing one and I know adding a new icon is a pain so I guess that is alright)

ssipasseuth: It is not part of this PR, but I noticed that the two icons close and resize overlap. An issue with the padding left it seems. What about fixing it by the way ?

ssipasseuth: on second thought, that is fine, I have to work on that component anyway =)

krampstudio: Ok, but in a separate PR then

PR #1074, opened 2016-11-24
Feature/logout url configurable

chrisphpgarcia:


config/default/urlroute.conf.php
```diff
@@ -13,6 +13,13 @@
                     'ext'        => 'tao',
                     'controller' => 'Main',
                     'action'     => 'login',
+                ],
+                'logout' => 
+                [
+                    'ext'        => 'tao',
+                    'controller' => 'Main',
+                    'action'     => 'logout',
+                    'redirect'   => _url('login', 'Main', 'tao'),
```
antoinerobin: why is it ```_url('login', 'Main', 'tao'),```and not ```_url('entry', 'Main', 'tao'),```?

PR #1075, opened 2016-11-24
Fix/tao 3392 datatable sortable indicator

krampstudio:
Layout changes to display a sortable icon when a column is sortable.
Change also the way the sort/sortable icons are displayed in different states and prevent them to push the content of the columns.


Alroniks:
Hi @krampstudio. Could you resolve conflicts and update branch, please?

PR #1079, opened 2016-11-28
Feature/tao 3396 calculator font resizing

ssipasseuth:
Tweaked the calculator css and template to ensure proportional font resizing when the component resizes.


krampstudio:
That's great. And I know this comment comes too late but did you try the viewport units ? There's a good explanation here : https://css-tricks.com/viewport-sized-typography/

ssipasseuth:
I didn't try it but I indeed checked that vw ad vh units when I was searching for a solution.
Since vw and vh are relative to the viewport, it won't help here as I want to resize the font relative to the component's size whereas the viewport units only applies to window, not to parent div.

views/js/ui/calculator.js
```diff
@@ -79,6 +91,17 @@ define([
                 //reset the calculator input
                 this.calc.press('C');
             })
+            .on('resize', function(){
+                var element = this.getElement();
```
krampstudio: This should be `$element` and the check  below on `$element.length`

views/js/ui/calculator.js
```diff
@@ -27,9 +27,21 @@ define([
...
     };
-    
+
+    /**
+     * The constant ratio to be applied to font size scaling during component resizing.
+     * It has been calculated to match a reference font-size of 10px when the width of the component is 240px.
+     * @type {number}
+     * @private
+     */
+    var _fontSizeRatio = 10/240;
```
krampstudio: Shouldn't this be part of the defaultConfig, so if we change the default width, we can also change the default ratio ?

views/js/ui/calculator.js
ssipasseuth: Ok fixed !

views/js/ui/calculator.js
ssipasseuth: No actually because this is the font to width ratio and match the component css. The ratio should be fixed because the css is fixed: https://github.com/oat-sa/tao-core/pull/1079/files#diff-0b536ad53889ffe6f36cb2d8d0f78aefR49

Presently, changing the default width of the calculator to say 400px would just has the same effect as using any intial width of 240px and resize it manually to 400px. 
Initial width of 240px -> fontsize = 240 * 10 /240 
Initial width of 400px -> fontsize = 400 * 10 /240, so proportional bigger

Actually I could have selected another base font-size in the css, in which case, the children element's font-size would have their font-size value in em modified.

PR #1080, opened 2016-11-28
Feature/component size

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3377

Add max size constraints on the movableComponent.

Set the hidden state of the component at render if the HTML markup is already hidden.

### DIFF
--- a/views/js/ui/component.js
+++ b/views/js/ui/component.js
@@ -151,6 +151,7 @@ define([
                 }
 
                 this.setState('rendered', true);
+                this.setState('hidden', this.$component.hasClass('hidden'));
 
                 delegatedResize.call(this);
 
--- a/views/js/ui/component.js
+++ b/views/js/ui/component.js
@@ -151,7 +151,6 @@ define([
                 }
 
                 this.setState('rendered', true);
-                this.setState('hidden', this.$component.hasClass('hidden'));
 
                 delegatedResize.call(this);
 
--- a/views/js/ui/movableComponent.js
+++ b/views/js/ui/movableComponent.js
@@ -128,19 +128,15 @@ define([
          * @returns {Object}
          */
         getPosition: function getPosition() {
-            var $element, position;
+            var $element, position, transform;
 
             if (this.is('rendered')) {
                 $element = this.getElement();
-
-                if ($element.hasClass('transform-translate')) {
-                    position = $element.position();
-                } else {
-                    position = {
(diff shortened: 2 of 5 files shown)


views/js/ui/component.js
```diff
@@ -151,6 +151,7 @@ define([
                 }
 
                 this.setState('rendered', true);
+                this.setState('hidden', this.$component.hasClass('hidden'));
```
krampstudio: This should be the other way. If a component starts hidden, it should execute `this.hide()`.

jsconan: Yes, but:
- can introduce flickering, as the the component will first be shown, and then hidden
- `this.hide()` triggers an event, that may not be wanted for the render step (otherwise an explicit call can be made on top if the event is needed)
- the goal here is to reflect the init state as set by the template, which allows to simplify the design

krampstudio: I have to disagree. A component is a visual representation of a state at a time. We should'nt encourage a 2-way state machine for our component. 
If the component is hidden, the event must be sent. 
Then the flickering is only a side effect. For example you can avoid it by setting the state before the component is attached or use a smooth rendering (opacity + transition).

views/js/ui/movableComponent.js
```diff
@@ -115,10 +121,55 @@ define([
...
         }
     };
 
     /**
+     * Gets the transform#translate coordinates
+     * @param {jQuery} $el
+     * @returns {Object}
+     */
+    function getTranslatedCoords($el) {
+        var transform = $el.css('transform').split(/[()]/)[1];
```
krampstudio: We have an util for that : `transformer.getTransformation($el)`  that also takes prefixed transform into account. It's worth to use it.

views/js/ui/movableComponent.js
```diff
@@ -115,10 +121,55 @@ define([
...
+         * @returns {Object}
+         */
+        getPosition: function getPosition() {
+            var $element, position, transform;
+
+            if (this.is('rendered')) {
+                $element = this.getElement();
+                transform = getTranslatedCoords($element);
+                position = {
+                    top: parseFloat($element.css('top')) + transform.top,
```
krampstudio: Isn't it the same as  `$element.css('top') + this.config.x` ?

jsconan: Yes you are right. This should be better.

PR #1087, opened 2016-12-01
Feature/mutation observer

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3377

Add a polyfill for the MutationObserver API


krampstudio:
According to http://caniuse.com/#feat=mutationobserver the polyfill targets only IE10. It's sad to load 25kb of code everywhere only for this browser.

jsconan:
Ok, so I will only rely on the native object. However I will keep this apart in case of it is needed for PhantomJS. I close the PR and will update the magnifier accordingly.

PR #1091, opened 2016-12-02
input of select get the same width than the parent

antoinerobin:


krampstudio:
This is too dangerous, this could potentially change the layout of every select2 in the GUI. I'll suggest you to override the rule into [the creator file](https://github.com/oat-sa/extension-tao-testqti/blob/master/views/scss/creator.scss)

antoinerobin:
will be transfered to the creator.scss of taoQtiTest

PR #1099, opened 2016-12-08
Fix/dialog remove

jsconan:
Fix issue happening when a modal dialog is closed by the ESC key: the DOM element is not removed.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -31,7 +31,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '7.37.0',
+    'version' => '7.37.1',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => array(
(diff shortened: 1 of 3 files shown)


ssipasseuth:
the code looks fine but the unit test fails !

```
grunt connect:test qunit:single --test=/tao/views/js/test/ui/dialog/test.html
Running "connect:test" (connect) task
Started connect web server on http://127.0.0.1:8082

Running "qunit:single" (qunit) task
Testing http://127.0.0.1:8082/tao/views/js/test/ui/dialog/test.html ...........
>> PhantomJS timed out, possibly due to:
>> - QUnit is not loaded correctly.
>> - A missing QUnit start() call.
>> - Or, a misconfiguration of this task.
Warning: 1/1 assertions failed (0ms) Use --force to continue.

Aborted due to warnings.

jsconan:
Yes, you are right!
I am wondering why I didn't get the same error when I checked it. Maybe browser cache issue.
Anyway, it is fixed now.

PR #1102, opened 2016-12-08
Fix/tao 3256 dynamic comp zoom

krampstudio:
Added a `position` util that works on bounding rectangle (takes transforms into account).

On zoom (resize + change of pixelRatio), I'm checking if the component is still inside the container, if not, I reset it.


ssipasseuth:
ok it works !

PR #1104, opened 2016-12-09
Tao 3461/flysystem upload

boajer:


helpers/uploadReferencerTrait.php
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        throw new \common_Exception('Unsupported file reference');
+    }
+
+    /**
+     * @param $serial
+     * @return string
+     */
+    protected function getLocalCopy($serial)
```
siwane: I understand you need an absolute path to allow the legacy method, work with path.
Discussed with @jbout, could you split these two ways into two separated methods.
It will be better, to identify legacy, and it will be more understandable for developers.

helpers/uploadReferencerTrait.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
siwane: As this trait is the skeleton of file upload, could u add detailled PHP doc please?

scripts/update/Updater.php
```diff
@@ -638,6 +638,12 @@ public function update($initialVersion) {
         }
 
         $this->skip('7.35.0', '7.36.2');
+
+        if ($this->isVersion('7.36.2')) {
+            (new AddTmpFs())->__invoke([]);
```
siwane: Could u add $this->getServiceManager()->propagate?

boajer:
@siwane Could you take a look again, please? I will resolve version conflicts after your approval.

models/classes/upload/UploadService.php
```diff
@@ -0,0 +1,194 @@
...
+    static protected $SESSION_ATTRIBUTE_FILES = 'tracked_uploads';
+    static protected $SESSION_ATTRIBUTE_LOCAL = 'tracked_uploads_local_copies';
+
+
+    /**
+     * @param array $postedFile
+     * @param string $folder
+     * @return array
+     */
+    public function uploadFile(array $postedFile, $folder)
```
siwane: Could you check variables please?
like here https://github.com/oat-sa/tao-core/pull/1104/files#diff-dff2c278c5a3dde51de93ab2f85e6f8cR49

models/classes/upload/UploadService.php
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Either create local copy or use original location for tile
+     * Returns absolute path to the file, to be compatible with legacy methods
+     * @param string|File $serial
+     * @return string
+     * @throws \common_Exception
+     */
+    public function getUploadedFile($serial)
```
siwane: Could you create similar method to be able to work with File? Just return the File
Put getUploadedFile() on @deprecated. 
Return a path will not be the final solution.

boajer:
@siwane Could you take a look again? ( and to the 4 more related PRs also )

boajer:
@siwane  I'll update versions and version constraints after you approval

PR #1107, opened 2016-12-09
Fix/tao 3493 justify

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3493

Always ensure than the left side of the calculator result is visible.

Known issue: the caret is always placed on the left. I tried several solutions, but fallen on other issues, so rolled back to this solution.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -31,7 +31,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '7.37.3',
+    'version' => '7.37.4',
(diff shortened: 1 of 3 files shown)


ssipasseuth:
This meet the AC but I am afraid that having the cursor on the left is too strange.... could we simply hide it somehow ? in a four function calculator, the cursor cannot be moved so it makes sense to remove/hide it after all...

Some example of online calculator: 
http://www.online-calculator.com/fr/   https://demos.learnosity.com/assessment/questions/featuretypes.php#calculator

ssipasseuth:
impressive =)

PR #1112, opened 2016-12-14
Refactor ui/feedbacks

krampstudio:
For the need of https://oat-sa.atlassian.net/browse/TAO-3453 I have migrated the module `ui/feedback` to a proper component, mostly in order to override the config at the level of the factory method.

The main methods are backward compatible but now each feedback is a component. This should also facilitate the implementation of stacked messages.


views/js/ui/feedback.js
```diff
@@ -18,328 +18,276 @@
...
+                    this.render($container);
+                }
+                return this;
+            },
+        }, defaultOptions);
+
+        return feedback
+            .setTemplate(tpl)
+            .on('init', function(){
+                this.config.id = 'feedback-' + (currents.length + 1);
```
jsconan: Caution, the id may not be unique, as the list of feedbacks is rebuilt after each [remove](https://github.com/oat-sa/tao-core/pull/1112/files#diff-4418ad0cf4fde256721eae5af94e1ddbR287).
Isn't it an issue ? If so you should probably rely on another counter to ensure the uniqueness of the id.

krampstudio: Yes, I was also wondering but since this is how it was done on the previous version I prefer to keep it backward compatible. We can consider changing it if we are going to support more than one feedback at a time.

jsconan: Ok, no problem

PR #1113, opened 2016-12-14
fixed the returned value in case the property value is a Resource

ssipasseuth:


ssipasseuth:
ok version updated in both extension. See related PR: https://github.com/oat-sa/extension-tao-mpart/pull/407

PR #1121, opened 2016-12-20
Fix/lang rdf

bugalot:


llecaque:
Updating current installation have a lower priority but it is a nice to have

PR #1133, opened 2016-12-30
Allow to guess extension id  by controller name

boajer:


jbout:
ControllerHelper::getControllers($ext->getId()) is an expensive operation that parses the files and should only be used sparingly

PR #1137, opened 2017-01-04
Context support for required action

boajer:


models/classes/requiredAction/RequiredActionAbstract.php
```diff
@@ -66,24 +66,26 @@ public function __construct($name, array $rules = [])
 
     /**
      * Whether the action must be executed.
+     * @param null $context
      * @return boolean
      */
-    public function mustBeExecuted()
+    public function mustBeExecuted($context = null)
```
antoinerobin: it should also be a change to the interface
You should also specify the type of the $context, at least in the documentation

boajer: There is no reason to change the signature in the interface I think, as this parameter should be optional.  If it will be specified in interface it became required  ( with default value) for all implementations, that is not really intended to be

PR #1144, opened 2017-01-12
Extra PO Service

bugalot:
Ability for extensions (specifically customer extensions) to register extra PO files (either compiled in PHP or regular POs) in order to override TAO Core translations without having impact on the existing core translations.


ssipasseuth:
Cool you made it ! TT-31 !

PR #1151, opened 2017-01-19
Feature/tao 3642 mediaplayer replay timeout option

no-chris:
Companion PR for https://github.com/oat-sa/extension-tao-itemqti-pci/pull/51

### DIFF
--- a/views/js/ui/mediaplayer.js
+++ b/views/js/ui/mediaplayer.js
@@ -96,6 +96,13 @@ define([
     var volumePositionThreshold = 150;
 
     /**
+     * The polling interval used to update the progress bar while playing a YouTube video.
+     * Note : the YouTube API does not provide events to update this progress bar...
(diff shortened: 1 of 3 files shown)


views/js/ui/mediaplayer.js
```diff
@@ -103,31 +96,39 @@ define([
     var volumePositionThreshold = 150;
 
     /**
+     * The polling interval used to update the progress bar while playing a YouTube video.
+     * Note : the YouTube API does not provide events to update this progress bar...
+     * @type {Number}
+     */
+    var youtubePolling = 100;
```
jsconan: This breaks unit tests.
The reason why this property was exposed is to allow to fine tune the youtube polling.
Thus it could be better to add it to the config entries of the player, even if it is not a common config set.
At least you should update the unit test.

PR #1165, opened 2017-01-25
Feature/statifier

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3714

Add a core component that delegate API to manage states.


views/js/core/statifier.js
```diff
@@ -0,0 +1,133 @@
...
+             * @returns {Boolean}
+             */
+            getState: function getState(name) {
+                return !!states[name];
+            },
+
+            /**
+             * Sets a state
+             * @param {String} name
+             * @param {Boolean} value
```
krampstudio: value should be optional and true by default

PR #1168, opened 2017-01-26
Feature/tao 3730 add renderer to area broker

no-chris:
see https://github.com/oat-sa/extension-tao-act/pull/466 for instructions


no-chris:
The current implementation of the areabroker is missing a destroy step for the custom renderers. Please do not merge yet.

PR #1170, opened 2017-01-26
Feature/tao 3651 log logger

krampstudio:
bunyan inspired logger, supports multiple providers but only the console provider is implemented. 
I've added a debug entry on the eventifier's trigger as the first use the logger.

Logger level can be changed either from the module configuration or by executing the following method : `window.setTaoLogLevel(level)` to enable logging on any browser


views/js/core/eventifier.js
```diff
@@ -260,19 +266,23 @@ define([
...
      * @param {Object} [target = {}] - the target object, a new plain object is created when omited.
-     * @param {logger} [logger] - a logger to trace events
      * @returns {Object} the target for conveniance
      */
-    function eventifier(target, logger){
+    function eventifier(target){
 
-        var targetName;
+        //try to get something that looks like a name, an id or generate one only for logging purposes
+        var targetName = target.name || target.id || target.serial || uuid(6);
```
jsconan: I see an issue there: as `target` is an optional parameter it can be null or undefined.

krampstudio:
Updated

jsconan:
The unit tests are mostly fixed, but I still have a failed test when using the CLI.

And the installer class is missing.

PR #1191, opened 2017-02-17
FlyFS compatibility improved

boajer:
https://oat-sa.atlassian.net/browse/TT-67


models/classes/service/class.FileStorage.php
```diff
@@ -109,22 +109,38 @@ public function deleteDirectoryById($id)
...
+                    $file = $directory->getFile($iterator->getSubPathName());
+                    $fileExists = $file->exists();
+
+                    if ($fileExists && 0 !== strcmp($file->read(), file_get_contents($item))) {
+                        throw new common_Exception('Different file content');
+                    }
+
+                    if (!$fileExists) {
+                        $fh = fopen($item, 'rb');
+                        $file->put($fh);
```
siwane: Could u use stream to handle large file?

boajer: But I stream already used here ?

siwane: You are right, sorry read too fast ;)

PR #1201, opened 2017-02-23
Feature/tao 3860 plugins z index

no-chris:
related pr : 
https://github.com/oat-sa/extension-tao-testqti/pull/758
https://github.com/oat-sa/extension-tao-act/pull/497

### DIFF
--- a/views/js/test/ui/stacker/test.js
+++ b/views/js/test/ui/stacker/test.js
@@ -147,5 +147,4 @@ define([
         assert.equal($div3.css('z-index'), 1025, 'div13has the correct z-index');
     });
 
-
 });
--- a/views/js/test/ui/stacker/test.js
+++ b/views/js/test/ui/stacker/test.js
@@ -41,10 +41,17 @@ define([
             $container = $(fixtureContainer),
             $div1 = $container.find('#div1');
 
-        QUnit.expect(1);
+        QUnit.expect(3);
+
+        stacker.resetScope();
+
+        assert.equal($div1.css('z-index'), 'auto', 'no z-index is set');
 
         stacker.bringToFront($div1);
-        assert.equal($div1.css('z-index'), 1001, 'correct z-index has been set');
+        assert.equal($div1.css('z-index'), 1010, 'z-index has been set');
+
+        stacker.reset($div1);
+        assert.equal($div1.css('z-index'), 'auto', 'z-index has been removed');
     });
 
     QUnit.test('does not increase z-index if already max', function (assert) {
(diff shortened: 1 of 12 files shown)


views/js/ui/component/stackable.js
```diff
@@ -0,0 +1,54 @@
...
+ * If you want to use a stacking scope, the component should be initialised with a stackingScope property
+ *
+ * @example
+ * var component = componentFactory({}, { stackingScope: 'myScope' });
+ * makeStackable(component);
+ *
+ * @author Christophe Noël <christophe@taotesting.com>
+ */
+define([
+    'lodash',
```
krampstudio: it seems you don't need the lodash dependency

views/js/test/ui/stacker/test.js
```diff
@@ -0,0 +1,275 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
+ */
+/**
+ * @author Christophe Noël <christophe@taotesting.com>
+ */
+define([
+    'jquery',
+    'lodash',
```
krampstudio: It seems lodash is not used.

views/js/ui/stacker.js
```diff
@@ -0,0 +1,124 @@
...
+    return function stackerFactory(scope) {
+        scope = scope || defaultScope;
+        initScope(scope);
+
+        return {
+            /**
+             * Set the z-index, on the given element, to the next available value
+             * @param {jQuery} $element
+             */
+            bringToFront: function bringToFront($element) {
```
krampstudio: You should check at least if `$element` is defined (everywhere, since it is a parameter), to prevent `TypeError: Cannot read property 'css' of undefined`, etc. (same on every method)

views/js/ui/dynamicComponent.js
```diff
@@ -98,6 +102,16 @@ define([
         },
 
         /**
+         * compute x/y coords of the component according to the start position and the dragged offset
+         */
+        setCoords : function setCoords() {
+            var $element = this.getElement();
+
+            this.position.x = parseFloat($element.attr('data-x')) + this.config.left;
```
krampstudio: From where does the `$element.attr('data-x')` comes from ? Where is this value set ?

views/js/ui/stacker.js
```diff
@@ -0,0 +1,124 @@
...
+ * @author Christophe Noël <christophe@taotesting.com>
+ */
+define([], function() {
+    'use strict';
+
+    var ns = '.stacker',
+        indexes = {},
+        increment = 10,
+        zIndexStart = 1000,
+        defaultScope = 'global';
```
krampstudio: It could be useful to be able to change those values for a stacking context : having the start and increment step per scope. (Don't do it if this is too heavy, this is just an idea).

no-chris: I also thought about it but my main issue here is that, due to the component architecture, we cannot control who chooses those parameters, nor be certain that the correct values will be set at a certain moment. Maybe we could wait until a use case present itself?

views/js/ui/dynamicComponent.js
no-chris: Indeed I am not very happy with this either. They are set by the interactUtils helper. The resize/move behavior of the dynamic component has been improved with this PR, but remains buggy. Proper fix could be to factorize it in the same way than the "makeStackable" module, we could have a "makeMovable" and "makeResizable" helpers, that could be shared by components requiring this behavior. The need for this leaky parameter would then go away. For now I can add a comment explaining the issue.

PR #1203, opened 2017-02-24
Feature/tao 3496 data proxy

jsconan:

Add data handling components:
- The `core/dataProvider/proxy` module provides a common API to serve several implementations, with respect to a CRUD architecture.
- The `core/dataProvider/proxy/ajax` module implements an AJAX proxy that provides a request handler to get/set data from/to a server

### DIFF
--- a/views/js/core/dataProvider/proxy.js
+++ b/views/js/core/dataProvider/proxy.js
@@ -30,8 +30,6 @@ define([
 
     var _defaults = {};
 
-    var _slice = [].slice;
-
     /**
      * Defines a CRUD proxy bound to a particular adapter. Each adapter will have to provide the following API:
      *
@@ -259,7 +257,7 @@ define([
          * @throws Error
          */
         function delegate(fnName) {
-            var request = {command: fnName, params: _slice.call(arguments, 1)};
+            var request = {command: fnName, params: Array.prototype.slice.call(arguments, 1)};
             if (!initialized && fnName !== 'init') {
                 return Promise.reject(new Error('Proxy is not properly initialized or has been destroyed!'));
             }
--- a/views/js/core/dataProvider/proxy.js
+++ b/views/js/core/dataProvider/proxy.js
@@ -263,12 +263,17 @@ define([
             if (!initialized && fnName !== 'init') {
                 return Promise.reject(new Error('Proxy is not properly initialized or has been destroyed!'));
             }
-            return delegateProxy.apply(null, arguments).then(function (data) {
-                if (middlewares) {
-                    return middlewares.apply(request, data);
-                }
-                return data;
-            });
+            return delegateProxy.apply(null, arguments)
+                .then(function (data) {
+                    if (middlewares) {
+                        return middlewares.apply(request, data);
+                    }
+                    return data;
+                })
+                .catch(function (err) {
+                    proxy.trigger('error', err);
+                    return Promise.reject(err);
+                });
         }
 
 
--- a/views/js/core/dataProvider/proxy.js
+++ b/views/js/core/dataProvider/proxy.js
@@ -1,0 +1,279 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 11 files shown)


views/js/core/dataProvider/proxy.js
```diff
@@ -0,0 +1,284 @@
...
+    'core/eventifier',
+    'core/promise',
+    'core/providerRegistry',
+    'core/tokenHandler'
+], function (_, delegator, eventifier, Promise, providerRegistry, tokenHandlerFactory) {
+    'use strict';
+
+    var _defaults = {};
+
+    var _slice = [].slice;
```
krampstudio: Please do not keep a ref at the module scope (not garbage collected) for just using it once in the delegate function

views/js/core/dataProvider/proxy.js
```diff
@@ -0,0 +1,284 @@
...
+
+        /**
+         * Delegates the call to the proxy implementation and apply the middleware.
+         *
+         * @param {String} fnName - The name of the delegated method to call
+         * @returns {Promise} - The delegated method must return a promise
+         * @private
+         * @throws Error
+         */
+        function delegate(fnName) {
```
krampstudio: Shouldn't this be handled by the delegator's wrapper config ?

views/js/core/dataProvider/proxy/ajax.js
```diff
@@ -0,0 +1,193 @@
...
+         * - 'read'   (GET)
+         * - 'write'  (POST)
+         * - 'remove' (GET)
+         *
+         * Other actions are applied with POST method by default. You can override the method in each action descriptor.
+         *
+         * @param {Boolean} [config.noCache] - Prevent the request to be cached by the client (default: true)
+         * @param {Boolean} [config.noToken] - Prevent the request to be use the security token when available (default: false)
+         */
+        init: function ajaxInit(config) {
```
krampstudio: please use the same name : `init : function init` because some browsers optimize them. (Same for other methods)

views/js/core/dataProvider/request.js
```diff
@@ -39,16 +39,20 @@ define([
      * Create a new error based on the given response
      * @param {Object} response - the server body response as plain object
      * @param {String} fallbackMessage - the error message in case the response isn't correct
+     * @param {XMLHttpRequest} xhr - the response header
      * @returns {Error} the new error
      */
-    var createError = function createError(response, fallbackMessage){
+    var createError = function createError(response, fallbackMessage, xhr){
```
krampstudio: Instead of giving the `xhr` ref, give directly the error code as parameter

views/js/core/dataProvider/proxy/ajax.js
```diff
@@ -0,0 +1,193 @@
...
+            };
+
+            _.defaults(config, _defaults);
+        },
+
+        /**
+         * Cleans up the instance when destroying
+         */
+        destroy: function htmlDataDestroy() {
+            this.processRequest = null;
```
krampstudio: is it to free some memory ?

views/js/core/dataProvider/proxy.js
jsconan: I do not see the problem here: instead of wasting resources (and execution time) by creating a useless array each time I need to apply the `slice()` method on another context, I put it in cache. Maybe should I target `Array.prototype.slice` instead of using a thrown anonymous array?

views/js/core/dataProvider/proxy/ajax.js
jsconan: As the `processRequest` method is added on init, it should be logically removed on destroy.
And this is also a helper to better garbage the object.

views/js/core/dataProvider/request.js
jsconan: This was my first idea. Then I was wondering if we need to access more than the error code. So I added the full ref to avoid to have plenty of parameters later.
If you think this is useless, I will replace the ref by only the code.

views/js/core/dataProvider/proxy.js
jsconan: The processing of promise's response could be handled by the wrapper.
But the check cannot, as the wrapper is invoked at the end and the need to be done before.

views/js/core/dataProvider/request.js
krampstudio: Yes, since you only use the code, set the code. This is consitent with the fallbackMessage

views/js/core/dataProvider/proxy.js
krampstudio: you don't need to cache those kind of calls, inside a function, the creation of an array will be inlined by the optimizer and then the ref garbage collected directly. Keeping a ref in a closure forever will cost more.

views/js/core/dataProvider/proxy.js
jsconan: In fact, I cannot move this code into the wrapper for the following reasons:
- the arguments provided to the wrapper are not the same, and I need to access to original arguments
- the check need to be done before the delegate mechanism operate (`!initialized && fnName !== 'init'`)

PR #1204, opened 2017-02-24
Feature/tao 3496 middleware

jsconan:

Add data handling components:
- The `core/middleware` module provides a mechanism that injects a list of filters between a request and its response.

### DIFF
--- a/views/js/core/middleware.js
+++ b/views/js/core/middleware.js
@@ -23,7 +23,7 @@ define([
     'async',
     'core/eventifier',
     'core/promise'
-], function (_, async, eventifier, Promise) {
+], function (_, asyncUtil, eventifier, Promise) {
     'use strict';
 
     /**
@@ -88,7 +88,7 @@ define([
 
                 // apply each middleware in series, then resolve or reject the promise
                 return new Promise(function (resolve, reject) {
-                    async.series(list, function (err) {
+                    asyncUtil.series(list, function (err) {
                         // handle implicit error from response descriptor
                         if (!err && response.success === false) {
                             err = response;
--- a/views/js/core/middleware.js
+++ b/views/js/core/middleware.js
@@ -1,0 +1,141 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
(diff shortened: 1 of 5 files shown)


views/js/core/middleware.js
```diff
@@ -0,0 +1,141 @@
...
+ */
+/**
+ * @author Jean-Sébastien Conan <jean-sebastien@taotesting.com>
+ */
+define([
+    'lodash',
+    'async',
+    'core/eventifier',
+    'core/promise'
+], function (_, async, eventifier, Promise) {
```
krampstudio: Please change the name of `async` to something else to prevent semantic collides with the upcoming `async/await` spec.

views/js/test/core/middleware/test.js
```diff
@@ -0,0 +1,204 @@
...
+        .cases(middlewareApi)
+        .test('instance API ', function (data, assert) {
+            var instance = middlewaresHandlerFactory('default');
+            QUnit.expect(1);
+            assert.equal(typeof instance[data.title], 'function', 'The middlewaresHandlerFactory instance exposes a "' + data.title + '" function');
+        });
+
+
+    QUnit.asyncTest('middlewares.apply() #success', function (assert) {
+        var middlewares = middlewaresHandlerFactory('default');
```
krampstudio: What the `default` parameter means ?

views/js/test/core/middleware/test.js
```diff
@@ -0,0 +1,204 @@
...
+            .apply(request, response, context)
+            .then(function () {
+                assert.ok(false, 'The promise should be rejected');
+                QUnit.start();
+            })
+            .catch(function (err) {
+                assert.deepEqual(err, response, 'The error has been provided');
+                QUnit.start();
+            });
+    });
```
krampstudio: Could you please add a test case where `next` is not called. This is going to be one of the most common error.

jsconan: I added a test case for a missing call to `next()`. However, if no call is made, the middleware cannot be resolved. I don't see how to counter that, except by adding a timeout.

views/js/test/core/middleware/test.js
krampstudio: I meant the [factory](https://github.com/oat-sa/tao-core/pull/1204/files/2a4db58d7d1609242885503d7d464ad371231350#diff-85eda88c9ac1c3d190fc9ff7c65f7a26R34) has no parameter.

views/js/test/core/middleware/test.js
krampstudio: May be we should instead use the same behavior than express & co. : if you don't run `next` your callback is the last one : 

```js
 middlewares
    .use('context',  (req, res, next) => { console.log('called'); next(); })
    .use('context',  () => console.log('last called because no next'))
    .use('context',  () => console.error('should not be called'))
    .apply(req, res)
    .then( () => console.log('OK'))
    .catch( err => console.error(err))
```

views/js/test/core/middleware/test.js
jsconan: Yes, the unit test was wrong because I started with a provider based architecture, and then changed my mind, but the unit test was not fixed at the same time.

views/js/test/core/middleware/test.js
jsconan: Yes, but the connect plugin relies on other behavior, as the chain is synchronous. 
In this implementation, I use `async.series()`, so I cannot detect whether there is no call tho next. Unless I change the implementation to be synchronous.

PR #1205, opened 2017-02-24
Feature/tao 3496 container

jsconan:

Add a component that manage a container in order to:
- handle the CSS scope
- access to the content
- handle context data in the DOM attributes

This component is intended to be used in the context of a single page application, as the container should not change but the scope and the context.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -33,7 +33,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '7.69.7',
+    'version' => '7.70.0',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => array(
         'generis' => '>=3.17.0',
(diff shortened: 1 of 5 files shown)


jsconan:
It is not related to `ui/component`. This container is intended to be used in the context of a single page app, where a same container is used with different contexts, the scope being set by a CSS class.
For instance: https://github.com/oat-sa/extension-lti-proctoring/blob/17c1619c332eaa177ced8d347ab7ee6f02aabbcd/views/js/controller/delegate.js

krampstudio:
So you are using this structure to "forward" the current controller to another controller after an ajax request ? and the `container` is just a way to retrieve data attributes ?
It means your request returns some HTML, you add it to the DOM, then inspect it to see if you have to execute another controller ?
If your use case is to execute another controller than the one bound to the request URL, you
could  replace it by the `X-Tao-Forward` mechanism ?

jsconan:
Ok for the `X-Tao-forward`, to redirect the controller, but there is other use cases, especially in https://github.com/oat-sa/extension-tao-proctoring/pull/295

PR #1210, opened 2017-02-24
Feature/tao 3496 data broker

jsconan:

Requires: https://github.com/oat-sa/tao-core/pull/1204

Add data handling components:
- The `core/dataProvider/dataBroker` module manages a list of data providers.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -33,7 +33,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '7.69.7',
+    'version' => '7.70.0',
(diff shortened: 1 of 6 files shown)


krampstudio:
@ssipasseuth please check how useful this is. I'm wondering why we would like to keep references of dataProviders when they have to be handled by the controllers themselves (we should not keep references during all the app lifecycle but only for a given controller).

jsconan:
The purpose is not to keep providers during all the app lifecycle, but to a have a mechanism to share providers inside the controller itself. You can see an usage here: https://github.com/oat-sa/extension-tao-proctoring/pull/295/files#diff-d4f049018ec1c3723f17bfa3c047745aR166

jsconan:
As discussed, this feature is not needed anymore.

PR #1217, opened 2017-03-06
Feature/middleware promises

jsconan:
Remove the use of async and use promises instead of callbacks.
The behavior is now:
- each middleware is called in order
- the next middleware is called just after the current one has returned
- a middleware can have asynchronous behavior, in that case it should return a promise to defer the call to the next
- a middleware can break the chain either by returning `false` or a promise that resolve `false`
- any error in the middlewares will break the chain

The advantage is there is no more infinite wait if the `next()` callback is never invoked, like in the previous implementation. Obviously, an unresolved promise can still block the chain.

### DIFF
--- a/views/js/core/middleware.js
+++ b/views/js/core/middleware.js
@@ -20,14 +20,15 @@
  */
 define([
     'lodash',
-    'async',
     'core/eventifier',
     'core/promise'
-], function (_, asyncUtil, eventifier, Promise) {
+], function (_, eventifier, Promise) {
     'use strict';
 
     /**
-     * Defines a middlewares handler
+     * Defines a middlewares chain handler.
+     * It will manage a chain of middlewares that can interact with request responses.
+     * Each middleware must return a promise.
      *
      * @returns {middlewareHandler} - The middlewares handler instance
      */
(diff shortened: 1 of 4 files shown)


views/js/core/middleware.js
```diff
@@ -79,43 +80,54 @@ define([
...
                 // apply each middleware in series, then resolve or reject the promise
                 return new Promise(function (resolve, reject) {
-                    asyncUtil.series(list, function (err) {
-                        // handle implicit error from response descriptor
-                        if (!err && response.success === false) {
-                            err = response;
+                    function next() {
+                        var middleware = stack[pointer ++];
+                        if (middleware) {
+                            Promise.resolve(middleware.call(context, request, response))
```
krampstudio: Why do you call  `Promise.resolve(middleware.call(context, request, response))` instead of `middleware.call(context, request, response).then(..)` ?

views/js/core/middleware.js
```diff
@@ -79,43 +80,54 @@ define([
...
+                                .then(function(res) {
+                                    if (res !== false) {
+                                        next();
```
krampstudio: In that situation the Promise from line 87 is never resolved ?

jsconan: It is a call chain. `next()` will either call the next middleware, or stop the chain.

views/js/core/middleware.js
jsconan: The middleware may not return a promise. So this code ensure there is always a promise.

PR #1223, opened 2017-03-08
Feature/tao 3295 isolate components behavior

no-chris:
This PR breaks down some component behavior in multiple modules : 
- placeable
- draggable
- resizable

The goal of this PR is multiple : 
- groundwork for https://oat-sa.atlassian.net/browse/TAO-3995 (integrate mathquill in QTI authoring / fix popups in QTI authoring) and https://oat-sa.atlassian.net/browse/TAO-3945 (line reader tool)
- fix bugs in resizing behavior of existing components (dynamicComponent still have to be refactored to benefit from this)
- encourage code reuse for those behaviors

It has a companion PR
https://github.com/oat-sa/extension-tao-testqti/pull/767

How to test : 
- all modules have a visual test in their corresponding unit test
- plugins using the movable component (area masking / magnifier) should work as before, with the resizing bug fixed (component still moving when resizing hit the min/max boundaries)


views/js/ui/component.js
```diff
@@ -48,12 +48,15 @@ define([
 
         if ($element) {
             if (_.isNumber(width)) {
-                $element.width(width);
+                $element.css({ width: width });
```
krampstudio: If it's a number (unit less), you should use `.width(width)`.

views/js/ui/component/draggable.js
```diff
@@ -0,0 +1,99 @@
...
+                    element     = $element[0],
+                    $container  = this.getContainer(),
+                    container   = $container[0],
+                    rootNode    = document.querySelector('html');
+
+                interact(element)
+                    .draggable({
+                        autoScroll: true,
+                        restrict: {
+                            restriction: container,
```
krampstudio: Are we always restricting the drag within the component container ? 
The component's container is where it is appended. And you may want to restrict to another container.

views/js/ui/component/resizable.js
```diff
@@ -0,0 +1,199 @@
...
+
+            if (this.is('rendered') && !this.is('disabled')) {
+                /**
+                 * @event Component#beforeresize the component is about to be resized (or not)
+                 * @param {Number} width - the new expected width
+                 * @param {Number} height - the new expected height
+                 * @param {Boolean} resizeFromLeft - if resize happens from the left
+                 * @param {Boolean} resizeFromTop - if resize happens from the top
+                 */
+                this.trigger('beforeresize', newWidth, newHeight, resizeFromLeft, resizeFromTop);
```
krampstudio: I don't get the goal of this event

views/js/ui/component/resizable.js
```diff
@@ -0,0 +1,199 @@
...
+                    $element    = this.getElement(),
+                    element     = $element[0],
+                    $container  = this.getContainer(),
+                    container   = $container[0];
+
+                interact(element)
+                    .resizable({
+                        autoScroll: true,
+                        restrict: {
+                            restriction: container
```
krampstudio: Same here

views/js/ui/component.js
no-chris: I'd rather add the units, if not jQuery produce unexpected results when the DOM element has the `box-sizing: border-box` property (= setting the .width(x) will result on an element having x + border size total width). see http://api.jquery.com/width/

views/js/ui/component.js
```diff
@@ -54,6 +54,9 @@ define([
                 $element.height(height);
             }
         }
+
+        $element.data('width', width);
+        $element.data('height', height);
```
no-chris: That's right, they are not meant to be used outside and so will be better as plain JS variables

views/js/ui/component/draggable.js
no-chris: Yes, we might also want to customize other interact behavior, which can be done by gradually exposing more options in the config object as we need them. But I can already expose this option no pb

views/js/ui/component/resizable.js
no-chris: I happen to need it in the line reader tool for a very specific use case. Basically I need to set the maxWidth/Height constrain depending on whether the resizing occurs on a specific border. Interact event "resizestart" is not suitable for this as it doesn't provide the information I need on the event object (which edges are used for resizing)

views/js/ui/component/placeable.js
```diff
@@ -0,0 +1,231 @@
...
+        _translate: function _translate(xOffsetAbsolute, yOffsetAbsolute) {
+            var $element = this.getElement(),
+                newX,
+                newY;
+
+            transformer.translateXY($element, xOffsetAbsolute, yOffsetAbsolute);
+
+            // retrieving current translate values is a costly process (see ui/transformer and/or lib/unmatrix)
+            // thus, we store them as custom attributes for later use, and especially when a relative transform will be needed (eg, .moveBy())
+            $element.data('translateX', xOffsetAbsolute);
```
no-chris: YEs

views/js/ui/component.js
no-chris: Also, I will use `this._width`and `this._height`as I'd rather not change the values given by the consumer in `this.config`. We might also want to restore the original size at some point.

no-chris:
Requested changes have been made

ssipasseuth:
ok tested, and working

PR #1224, opened 2017-03-08
Fix/tao 3496 error response

jsconan:

When using the `dataProvider/request` module, an error is triggered if the response contains `success=false`. However this response is not provided with the error, making difficult to react precisely to the error. This PR fix that.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -34,7 +34,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '7.81.0',
+    'version' => '7.81.1',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => array(
         'generis' => '>=3.17.0',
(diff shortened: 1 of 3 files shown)


krampstudio:
If I understand well, this is to support endpoints that doesn't format their response correctly ? Because if the response contains `success : false`, it must also contain `errorCode` and `errorMsg`. This is the contract : https://github.com/oat-sa/tao-core/pull/1224/files#diff-55663eacffddd3217253f3da541a3dabL26

jsconan:
This is for endpoints that handle errors by themselves based on more complex data. Just having a code and string message is not enough in that case.

krampstudio:
My point is the `response`  is already part of the `err` if it is formatted correctly : https://github.com/oat-sa/tao-core/pull/1224/files#diff-55663eacffddd3217253f3da541a3dabL49. 
If the response is not formatted as it should this is an issue.

jsconan:
Ok, so I will fix the controllers in order to provide an error code.

PR #1225, opened 2017-03-09
Fix/tao 3950 navigator space key

krampstudio:
When hitting the space bar within a navigable element like a text field, we don't activate the navigable in order to let people write content. 
This can be tested within any test that have both the navigator configured on the action bar and the comment plugin activated : 

 - Open the comment 
- Write anything with a space
- The element is closed (blured)

### DIFF
--- a/views/js/ui/keyNavigation/navigator.js
+++ b/views/js/ui/keyNavigation/navigator.js
@@ -173,9 +173,9 @@ define([
             }
         }
 
-         _.each(navigables, function(navigable){
+        _.each(navigables, function(navigable){
                 //check if it is a valid navigable element
-                navigable.init();
+            navigable.init();
         });
 
         if(config.group){
@@ -439,8 +439,9 @@ define([
             navigable.getElement()
                 //requires a keyup event to make unselecting radio button work with space bar
                 .on('keyup'+_ns, function keyupSpace(e){
+                    var $target = $(e.target);
                     var keyCode = e.keyCode ? e.keyCode : e.charCode;
-                    if(keyCode === 32){//space bar
+                    if(keyCode === 32 && !$target.is(':input')){//space bar
                         e.preventDefault();
                         keyNavigator.activate(e.target);
                     }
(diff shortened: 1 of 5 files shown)


ssipasseuth:
the issue is that this fix will prevent the activation of an input with space for instance, if you have a choice interaction, you should be allowed to activate/deactivate with bot enter and space

views/js/ui/keyNavigation/navigator.js
```diff
@@ -439,8 +439,9 @@ define([
             navigable.getElement()
                 //requires a keyup event to make unselecting radio button work with space bar
                 .on('keyup'+_ns, function keyupSpace(e){
+                    var $target = $(e.target);
                     var keyCode = e.keyCode ? e.keyCode : e.charCode;
-                    if(keyCode === 32){//space bar
+                    if(keyCode === 32 && !$target.is(':input')){//space bar
```
ssipasseuth: this should rather be if(keyCode === 32 && !$target.is('input[type="text"],textarea')) to allow input checkbox and radio to be activable

krampstudio: But if I have a radio within a navigable we will have the same problem. So do we have to activate the only if the target is the navigable element and not a child ? 
Because I've tested with choice interactions and it works as expected since the event bubbles up to the navigable element.

krampstudio:
The fix handles now arrows and activate the radios of they're the navigable

krampstudio:
The fix handles now arrows and activate the radios of they're the navigable

ssipasseuth:
The radio button activation with space and up/left issue in the allow comment text area are fixed now but it appears that we still have an issue with right/down key.

To fix the arrow key issue completely, we can return the target information to the trigger here https://github.com/oat-sa/tao-core/blob/fix/TAO-3950_navigator-space-key/views/js/ui/keyNavigation/navigator.js#L435
then use that information to prevent navigation the toolbar for text input here https://github.com/oat-sa/extension-tao-testqti/blob/master/views/js/runner/plugins/content/accessibility/keyNavigation.js#L54-L57

I understand it is slightly out of scope of the related JIRA issue, so could be potentially be part of another PR.

krampstudio:
No, I'll try to also fix them.

krampstudio:
And the `enter` key produces also the same effect.

krampstudio:
I've fixed the issues, but directly on the shortcut event listener.

PR #1233, opened 2017-03-15
make key navigation work inside a large test suite

krampstudio:
PhantomJS did not support the mix of the `navigator` variable name and a missing `QUnit.start` resulting in a broken test suite while running all tests : https://jenkins.taocloud.org/job/build-tao-js/767/console

### DIFF
--- a/views/js/test/ui/keyNavigation/test.js
+++ b/views/js/test/ui/keyNavigation/test.js
@@ -44,32 +44,32 @@ define([
     QUnit.module('API');
 
     QUnit.test('factory', function(assert) {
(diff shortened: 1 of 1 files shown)


ssipasseuth:
so _navigator_ is a sort of reserved word ? 
have you found a source that gave you this information or you discovered it randomly ?-> might be work communicating to other devs

krampstudio:
It looks like it was a side effect only of PhantomJS. `navigator` isn't a keyword but is always available globally (since it is bound to `window.navigator`) but the spec allows you to use it as a variable name in lexical scope.

PR #1234, opened 2017-03-15
Feature/tao 3949 json uri

antoinerobin:


actions/class.Export.php
```diff
@@ -83,14 +83,27 @@ public function index()
...
-                for ($i = 0; $i < $instanceCount; $i++) {
-                    $exportData['instances'][$i] = tao_helpers_Uri::decode($exportData['instances'][$i]);
+                $instances = json_decode(urldecode($exportData['instances']));
+                unset($exportData['instances']);
+
+                //allow to export complete classes
+                if(isset($exportData['type']) && $exportData['type'] === 'class'){
+
+                    $children = array();
+                    foreach ($instances as $i => $instance){
```
jsconan: Is the `$i` variable really needed? It does not seem to be used.
The same for for the other foreach loop below.

views/templates/form/export.tpl
```diff
@@ -48,9 +48,13 @@
                                 params[param.name] = param.value;
                             }
                         });
-                        params.instances = instances;
 
-                        $.fileDownload(helpers._url("<?=get_data('export_action')?>", "<?=get_data('export_module')?>", "<?=get_data('export_extension')?>", params), {
+                        params.instances = encodeURIComponent(JSON.stringify(instances));
```
jsconan: Just a question, nothing to change, I am just curious.
Is the call to `encodeURIComponent` really needed? The AJAX requester should handle that on its hands.

PR #1235, opened 2017-03-16
Feature/tao 3868 frontend forms

eacraft13:


krampstudio:
Here are my remarks. I think @ssipasseuth can also help us improving the API :

krampstudio:
Here are my remarks. I think @ssipasseuth can also help us improving the API. But I think the best is to start by : 
1. changing the `ui/form` to become a `ui/component`
2. `ui/field/fields` must not embed all kind of fields, we should have a separate component per field. Also each field must be a `ui/component` with shared API and events
3. The form will be configured based on the description of a class
4. You should start directly with unit tests

For example, let's start with the text field component : `ui/form/field/text` (or whatever how you call it).
Each field can be used as is, but they all share the same API : 

```js
//textField is a ui/component
var textField = textFieldFactory({
  name : 'Name',
  placeholder : 'Enter your name',
  value : 'John'
})
//textField is a component, so it has the common lifecycle (init->render->destroy)
textField.on('render', function(){
     textField.getValue();
    textField.setValue('Jack');
});
//let's say "change" is an event all fields trigger
textField.on('change', function(value){
       
});
textField.on('error', function(err){
       
});
textField.render(container);
```

So if all fields have the same API (`getValue`, `setValue`, etc.) and the same events (`change`, 'error`, etc.) 

Now let's instantiate a form :

eacraft13:
requires changes to backend api - https://oat-sa.atlassian.net/browse/TAO-3989

views/js/ui/form/field/field.js
```diff
@@ -0,0 +1,151 @@
...
+    'jquery',
+    'ui/component',
+    'tpl!ui/form/field/tpl/checkBox',
+    'tpl!ui/form/field/tpl/comboBox',
+    'tpl!ui/form/field/tpl/hiddenBox',
+    'tpl!ui/form/field/tpl/textBox'
+], function (
+    $,
+    component,
+    checkBoxTpl,
```
krampstudio: We expect to have one component per field :  
 - one for checkboxes
 - one for comboboxe 
 - etc.

Since they share the same API, they can all use the same base API.
Some fields can be complex (we have even trees).

views/js/ui/form/field/field.js
```diff
@@ -0,0 +1,151 @@
...
+
+
+    /**
+     * Property for ui/form/field templates mapping
+     */
+    var _templates = {
+        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#CheckBox' : checkBoxTpl,
+        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#ComboBox' : comboBoxTpl,
+        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HiddenBox' : hiddenBoxTpl,
+        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#TextBox' : textBoxTpl
```
krampstudio: The mapping should be handled outside

views/js/ui/form/field/field.js
```diff
@@ -0,0 +1,151 @@
...
+
+        .setTemplate(_templates[config.type])
+
+        .on('render', function () {
+            var self = this,
+                timer,
+                $input = this.getElement().find('input');
+
+            if ($input.length) {
+                $input.on('keyup', function () {
```
krampstudio: It's the reason why we need one component per field : each field has different DOM element, with different events, etc.

views/js/ui/form/field/tpl/checkBox.tpl
```diff
@@ -0,0 +1,25 @@
+<div class="ui-form-field">
+    <label
+        class="label"
+        for="{{input.name}}">
+        {{label}}
+        {{#if required}}
+        <abbr title="This field is required">*</abbr>
```
krampstudio: Every textual string needs to be wrapped in the translation function `{{__ 'This field is required'}}`

views/js/ui/form/form.js
```diff
@@ -0,0 +1,138 @@
...
+
+            _.each(this.fields, function (existingField) {
+                if (!existingField.is('rendered')) {
+                    existingField.render(_fieldContainer);
+                }
+            });
+
+            $form.on('submit', function (e) {
+                e.preventDefault();
+                self.trigger('submit', this);
```
krampstudio: send the data instead of the component instance : an object with `{ id : value}` should be more appropriate

views/js/ui/form/generis/user.js
```diff
@@ -0,0 +1,250 @@
...
+
+
+    /**
+     * Factory for ui/form/generis/user
+     * @param {Object} config
+     * @param {String} [config.action] - URI that processes form (default is `'/tao/users/add'`)
+     * @param {String} config.container - Container to render form
+     * @param {String} [config.method] - The HTTP method the browser uses to submit the form (default is `'get'`)
+     */
+    var userGenerisFormFactory = function userGenerisFormFactory(config, user) {
```
krampstudio: We're expecting 2 modules instead : 
 - `ui/form/generis` that turns any Generis/RDF metadata into a `ui/form`with the appropriate fields (this module handle the mapping bewteen widgets and fields (`http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HiddenBox` to `ui/form/field/password`)
 - optionally a  `ui/form/generis/user` that take care of the user form specificities

krampstudio:
Regarding the retrieval of the data we expect to have a dedicated data provider : 

//a brand new user
```js
generisFormDataProvider
  .getClass("http://www.tao.lu/Ontologies/generis.rdf#User")
  .then(function(userClassData){
         //userClassData contains the list of properties and the values for the combobox, etc.
   })
  .catch(function(err){ 
    //...
   });
```


get the user data of the user `http://foo.org/tao.rdf#123456` (in order to edit him)
```js
generisFormDataProvider
  .getResource("http://www.tao.lu/Ontologies/generis.rdf#User", 'http://foo.org/tao.rdf#123456')
  .then(function(userClassData){
         //userClassData contains the list of properties and the values for the combobox, etc.
         // and the values of the selected user
   })
  .catch(function(err){ 
    //...
   });
```

save user data of the user (but we can separate add and edit, up to you)
```js
generisFormDataProvider
  .saveInstance("http://www.tao.lu/Ontologies/generis.rdf#User", 'http://foo.org/tao.rdf#123456', values) // values are the values from the form 
  .then(function(response){
         //response may contain the URI of the new user
   })
  .catch(function(err){ 
    //...
   });
```

PR #1236, opened 2017-03-16
Added seconds to date interval duration formatting

Alroniks:
https://oat-sa.atlassian.net/browse/TAO-3774


Alroniks:
@jbout Please, have a look at too.

Alroniks:
It doesn't make sense anymore.

PR #1239, opened 2017-03-17
Fix/breadcrumbs

jsconan:
Add a helper to resolve routes into extension, controller and actions.

Use the new helper in the breadcrumbs.

TODO: use the helper in the `oat\tao\model\routing\Resolver` class.


jsconan:
The Resolver in TAO will be refactored. hence, the helper introduced here is useless.

PR #1244, opened 2017-03-22
Feature/tao 4027 cache buster

krampstudio:
Add the cache buster to most of the assets  : 
 - modify the `AssetService` to use the cache buster
 - modify the `ThemeRegistry`to use the cache buster

@antoinerobin is asked for a code review since this is a backend development.
@no-chris is the PR assignee an will verify the feature


no-chris:
This is not part of the AC of the story but if the purpose when we discussed this is sprint planning, the purpose was to make sure that no one would ever have again to refresh the browser cache to see an update (including on "intermediate" releases). On proposed solution was to compute the cache buster key on each taoUpdate instead of relying only on the sprint build number.

Don't know if this can be an easy win for this story, if not no pb

models/classes/ThemeRegistry.php
```diff
@@ -238,26 +239,25 @@ public function unregisterTheme($id)
...
             if(strpos($theme['path'] , ThemeRegistry::WEBSOURCE) === 0) {
                 $websource = WebsourceManager::singleton()->getWebsource($this->get(ThemeRegistry::WEBSOURCE));
                 $webUrl = $websource->getAccessUrl(substr($theme['path'],strlen(ThemeRegistry::WEBSOURCE)));
                 $theme['path'] = $webUrl;
             }
             else {
-                // normalizing makes sure that whatever\\comes/in gets/out/properly
-                $theme['path'] = ROOT_URL . FsUtils::normalizePath($theme['path']) ;
-
+                $assetService = ServiceManager::getServiceManager()->get(AssetService::SERVICE_ID);
```
antoinerobin: could you put the getServiceManager in a function so eache time you need it you do : 
```php
$this->getServiceManager->get(MY_SERVICE_ID)
```
You can pu it here or in the parent class

models/classes/asset/AssetService.php
```diff
@@ -27,25 +27,78 @@
...
+    }
+
+
+    /**
+     * Get a the cache buster value
+     * @return string the busteri value
+     */
+    protected function getCacheBuster()
+    {
+        return $this->hasOption('buster') ? $this->getOption('buster') : TAO_VERSION;
```
antoinerobin: please you the constant that you defined above

models/classes/asset/AssetService.php
```diff
@@ -27,25 +27,78 @@
...
+    public function getAsset($asset, $extensionId = null)
     {
-        return $this->getJsBaseWww($extensionId) . FsUtils::normalizePath($asset);
+        if( ! is_null($extensionId)){
+            $url = $this->getJsBaseWww($extensionId) . FsUtils::normalizePath($asset);
+        } else {
+            $url = $this->getAssetBaseUrl() . FsUtils::normalizePath($asset);
+        }
+
+        if(!$this->hasOption('noBuster')) {
```
antoinerobin: can you add a constant for this option and the 'base' ?

krampstudio:
@no-chris I've not been part of those discussions... and this is not either part of the story. I'll see with the PM

krampstudio:
I've made all the requested changes, even the Update ID.
So by default the URL to each asset contains the `TAO_VERSION` value, and each call to the updater generates a new id.

PR #1251, opened 2017-03-24
Feature/tao 3949 async export

antoinerobin:
I'm not sure about the version number.
When you export items / tests with taskqueue extension aren't loaded so constants aren't accessible.
I put them in class constants to avoid this issue, but if you get this version of tao with older version of taoQtiItem / taoItem ... it will break.


antoinerobin:
@ssipasseuth We just need your review on the javascript part, since you did the taskqueue table

actions/class.TaskQueueData.php
```diff
@@ -0,0 +1,148 @@
...
+        /**
+         * @var $taskQueue oat\Taskqueue\Persistence\RdsQueue
+         */
+        $taskQueue = $this->getServiceManager()->get(Queue::SERVICE_ID);
+
+        if(is_a($taskQueue , 'oat\Taskqueue\Persistence\RdsQueue', true)) {
+            /** @var oat\Taskqueue\Action\TaskQueueSearch $dataPayLoad */
+            $dataPayLoad =  $taskQueue->getPayload($user->getIdentifier());
+
+            echo json_encode($dataPayLoad);
```
jbout: please use return "$this->returnJson($dataPayLoad)"

actions/class.TaskQueueData.php
```diff
@@ -0,0 +1,148 @@
...
+    public function getTasks()
+    {
+        $user = common_session_SessionManager::getSession()->getUser();
+
+        /**
+         * @var $taskQueue oat\Taskqueue\Persistence\RdsQueue
+         */
+        $taskQueue = $this->getServiceManager()->get(Queue::SERVICE_ID);
+
+        if(is_a($taskQueue , 'oat\Taskqueue\Persistence\RdsQueue', true)) {
```
jbout: Should not check for a specific class name, but a feature defined by an interface

actions/class.TaskQueueData.php
```diff
@@ -0,0 +1,148 @@
...
+                $file = $directory->getFile($filename);
+                $this->output($file);
+                return;
+            }
+        }
+        return $this->returnJson([
+            'success' => false,
+        ]);
+    }
+    protected function output(File $file)
```
jbout: please use tao_helpers_Http::returnStream() instead, to not create temporary files

models/classes/TaskQueueActionTrait.php
```diff
@@ -22,6 +22,8 @@
 
 use oat\oatbox\task\Queue;
 use oat\oatbox\task\Task;
+use oat\oatbox\user\User;
```
jbout: User unused

antoinerobin:
requires : https://github.com/oat-sa/generis/pull/326

PR #1254, opened 2017-03-28
Feature/tao 3961 abstract payload

antoinerobin:


models/classes/datatable/AbstractPayload.php
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Class AbstractPayload
+ *
+ * It will return data formatted to feed the client datatable tao/views/js/ui/datatable.js
+ *
+ * @package oat\oatbox\task
+ * @author Antoine Robin, <antoine@taotesting.com>
+ */
+abstract class AbstractPayload implements \JsonSerializable
```
hutnikau: Why don't you use current `AbstractDatatablePayload` implementation?

hutnikau:
We already have abstract class to build response for datatable component. Please consider to use current implementation (AbstractDatatablePayload) and implement function for sorting, filtration and pagination.

PR #1272, opened 2017-04-10
fix active page calculation

hutnikau:


views/js/ui/pagination.js
```diff
@@ -149,6 +149,9 @@ define(['jquery', 'lodash', 'i18n', 'ui/component', 'ui/pagination/paginationStr
                 totalPages = validTotalPages(config.totalPages);
 
                 activePage = config.activePage || 1;
+                if (activePage > totalPages) {
+                    activePage = totalPages;
+                }
                 activePage = validActivePage(activePage, totalPages);
```
zagovorichev: That validation should be implemented in the `validActivePage` method

zagovorichev: But why do you need it?
This is error, when you give incorrect parameters:
```
if (totalPages === false || activePage === false) {
    this.trigger('error', __('Undefined amount of the pages for pagination'));
}
```

PR #1274, opened 2017-04-11
Fix/tao 4177 fix scratchpad icons

no-chris:
https://oat-sa.atlassian.net/browse/TAO-4177

The helper Template::js returns a cache busted url even when a folder is passed as a parameter. Thus, when litterallyCanvas tries to load its assets, it requests something like http://xxxx?buster=XXX/icon.png

This PR prevents the cache buster from being added when a folder is requested as an asset

### DIFF
--- a/models/classes/asset/AssetService.php
+++ b/models/classes/asset/AssetService.php
@@ -62,8 +62,10 @@ class AssetService extends ConfigurableService
             $url = $this->getAssetBaseUrl() . FsUtils::normalizePath($asset);
         }
 
+        $isFolder = (substr_compare($url, '/', strlen($url) - 1) === 0);
+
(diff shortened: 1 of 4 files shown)


models/classes/asset/AssetService.php
```diff
@@ -62,8 +62,10 @@ public function getAsset($asset, $extensionId = null)
             $url = $this->getAssetBaseUrl() . FsUtils::normalizePath($asset);
         }
 
+        $isFolder = (substr_compare($url, '/', strlen($url) - 1) === 0);
```
krampstudio: Could you just add a note in the method comment saying it supports also folders now

no-chris: Requested change has been made

PR #1279, opened 2017-04-20
new security level for install

antoinerobin:


install/index.php
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
+ *
+ *
+ */
+require_once '../../vendor/autoload.php';
+
+if(tao_install_utils_System::isTAOInstalled()){
```
hutnikau: Even if TAO installed you still can reach installation screen by this url:
http://yourdomain/tao/install/index.html
I guess we have to get rid of index.html and move all it's content to index.php file (and of course do this checking there).

PR #1283, opened 2017-04-21
in some situation, shownExtension can be empty

krampstudio:


actions/class.ClientConfig.php
```diff
@@ -118,8 +118,11 @@ private function getExtension($extensionId)
     private function getShownExtension()
     {
         if($this->hasRequestParameter('shownExtension')){
-            $extension = $this->getExtension($this->getRequestParameter('shownExtension'));
-            return $extension->getName();
+            $shownExtension = $this->getRequestParameter('shownExtension');
+            if(strlen(trim($shownExtension)) > 0){
+                $extension = $this->getExtension($this->getRequestParameter('shownExtension'));
```
jsconan: Why do you call twice the `getRequestParameter()` ? you already have the value in the `$shownExtension` variable.
Nothing blocking, I am just wondering.

PR #1287, opened 2017-04-21
:not rules where wrongly applied

dietertaotesting:
Install any TAO version
- resize the window to a size where the menus would normally overlay
- mobile menu is applied

To fix the original NCCER case the TAO dependency needs to be increased to whatever the outcome of this PR will be


no-chris:
Works fine for the right menu, but when the window gets smaller, hovering the mouse in the empty black brings back the main icons (see screenshot)

screen shot 2017-04-24 at 15 54 39

dietertaotesting:
We recommand a minimum screen width of 1024px - although this is already unreaonable in authoring mode. Smaller screens have never been tested in authoring.
Won't fix

PR #1290, opened 2017-04-24
removed referring to a variable BASE_WWW

zagovorichev:


install/class.ExtensionInstaller.php
```diff
@@ -133,12 +134,11 @@ public function applyAccessRules()
...
-            ClientLibRegistry::getRegistry()->register($this->extension->getId(),$extManifestConsts['BASE_WWW']. 'js');
-            ClientLibRegistry::getRegistry()->register($this->extension->getId().'Css',$extManifestConsts['BASE_WWW']. 'css');
-            
-        }
+    {
+        $jsPath = trim(Template::js('', $this->extension->getId()), '/');
+        ClientLibRegistry::getRegistry()->register($this->extension->getId(), $jsPath);
+
+        $cssPath = trim(Template::css('', $this->extension->getId()), '/');
+        ClientLibRegistry::getRegistry()->register($this->extension->getId().'Css', $cssPath);
```
antoinerobin: you should not have this 'css' here

scripts/update/Updater.php
```diff
@@ -137,13 +138,8 @@ public function update($initialVersion) {
...
             foreach ($extensionManager->getInstalledExtensions() as $extension) {
-                $extManifestConsts = $extension->getConstants();
-                if (isset($extManifestConsts['BASE_WWW'])) {
-
-                    ClientLibRegistry::getRegistry()->register($extension->getId(), $extManifestConsts['BASE_WWW'] . 'js');
-                    ClientLibRegistry::getRegistry()->register($extension->getId() . 'Css', $extManifestConsts['BASE_WWW'] . 'css');
-
-                }
+                ClientLibRegistry::getRegistry()->register($extension->getId(), Template::js('', $extension->getId()));
+                ClientLibRegistry::getRegistry()->register($extension->getId().'Css', Template::css('', $extension->getId()));
```
antoinerobin: here neither

antoinerobin: Why don't you trim your js and css here ?

zagovorichev: because of unit tests, they failed when I did it here

install/class.ExtensionInstaller.php
zagovorichev: Why do you think so? Could you explain me, please? I think that used by frontend, but if you see another new way, that not for this task
I mean: for me important to delete BASE_WWW usage, each another changes should be in other PR

antoinerobin: the css is already part of the $cssPath so you don't have to add it. Plus the case wasn't right

scripts/update/Updater.php
antoinerobin: don't get why, it is just an update script it shouldn't impact unit test

install/class.ExtensionInstaller.php
antoinerobin: I read it too fast, your right

scripts/update/Updater.php
zagovorichev: because someone wrote test before me, and I don't want to broke it

PR #1295, opened 2017-04-27
Feature/tao 4215 xsrf token

krampstudio:
The token management API has been moved from `taoTests` to `tao` and refactored  in order to support the following features : 
 - It's now a proper Service
 - it works with a pool instead of a unique token
 - the pool storage is represented using a `TokenStore` (the PHP session storage has been implemented)
 - invalidation of the pool based on a maxsize or a TTL

In a first time the protection has been implemented into the resources forms, so the user form is covered.


models/classes/security/xsrf/TokenService.php
```diff
@@ -0,0 +1,218 @@
...
+            usort($reduced, function($a, $b){
+                if($a['ts'] == $b['ts']){
+                    return 0;
+                }
+                return $a['ts'] < $b['ts'] ? -1 : 1;
+            });
+
+            while(count($reduced) >= $this->getPoolSize()){
+                array_shift($reduced);
+            }
```
jsconan: You could also use `$reduced = array_slice($reduced, -$this->getPoolSize());`

models/classes/security/xsrf/TokenService.php
```diff
@@ -0,0 +1,218 @@
...
+        return $reduced;
+    }
+
+    /**
+     * Get the configured pool size
+     * @return int the pool size, 10 by default
+     */
+    protected function getPoolSize()
+    {
+        $poolSize = 10;
```
jsconan: Maybe should you use a constant here instead of a magic value.

test/model/security/xsrf/TokenServiceTest.php
```diff
@@ -0,0 +1,209 @@
...
+            'store' =>  $store
+        ]);
+
+        $this->assertEquals(0, count($store->getTokens()), 'The store is empty');
+
+        $token = $service->createToken();
+        $this->assertEquals(40, strlen($token), 'The token has the expected length');
+        $this->assertRegExp('/^[0-9a-f]{40}$/', $token, 'The token is correctly formatted');
+
+        $this->assertEquals(1, count($store->getTokens()), 'The store contains now one token');
```
jsconan: Maybe should you also check if the right token has been stored instead of just counting it.

test/model/security/xsrf/TokenServiceTest.php
```diff
@@ -0,0 +1,209 @@
...
+        $this->assertEquals(3, count($store->getTokens()), 'The store contains now three tokens');
+
+        $service->createToken();
+        $this->assertEquals(4, count($store->getTokens()), 'The store contains now four tokens');
+
+        $service->createToken();
+        $this->assertEquals(4, count($store->getTokens()), 'The store remains at four tokens, the max pool size');
+
+        $service->createToken();
+        $this->assertEquals(4, count($store->getTokens()), 'The store remains at four tokens, the max pool size');
```
jsconan: Maybe should you also check whether the removed token are the old ones

test/model/security/xsrf/TokenServiceTest.php
```diff
@@ -0,0 +1,209 @@
...
+
+        $token1 = $service->createToken();
+
+        $this->assertEquals(1, count($store->getTokens()), 'The store contains now one token');
+        $this->assertTrue($service->checkToken($token1), 'This first token is valid');
+
+        sleep(1);
+
+        $token2 = $service->createToken();
+        $this->assertEquals(2, count($store->getTokens()), 'The store contains only one token, the 1st has been invalidated');
```
jsconan: Wrong test label, the 1st token is still valid at this point

test/model/security/xsrf/TokenServiceTest.php
```diff
@@ -0,0 +1,209 @@
...
+        $this->assertTrue($service->checkToken($token1), 'This first token is valid');
+        $this->assertTrue($service->checkToken($token2), 'This second token is also valid');
+
+        sleep(1);
+
+        $this->assertFalse($service->checkToken($token1), 'This first token is not valid anymore');
+        $this->assertFalse($service->checkToken($token2), 'This second token is not valid anymore');
```
jsconan: oO I would  say the second is still valid as only 1 second has elapsed since its creation and the max time is 2 seconds. Am I wrong?

actions/form/class.Instance.php
```diff
@@ -63,8 +66,13 @@ protected function initForm()
...
 		$instanceElt->setValue('1');
-		$this->form->addElement($instanceElt);
-		
+                $this->form->addElement($instanceElt);
+
+                //add a token to protect against xsrf
+                $tokenElt = tao_helpers_form_FormFactory::getElement('token', 'Token');
+                $tokenElt->addValidator(new XsrfTokenValidator());
+                $this->form->addElement($tokenElt);
+
```
jsconan: Nothing blocking, but it seems there is a bad indent here. The same a few lines below.

krampstudio:
Updated

PR #1296, opened 2017-04-27
Datatable no more lockes layout after backend errors

boajer:


views/js/ui/datatable.js
```diff
@@ -191,8 +193,17 @@ define([
...
-            $.ajax(ajaxConfig).done(function(response) {
+            $.ajax(ajaxConfig).done(function (response) {
                 self._render($elt, response);
+            }).fail(function (response) {
+                var errorDetails = JSON.parse(response.responseText);
+
+                if (response.status === 403) {
+                    logoutEvent();
+                } else {
+                    feedback().error(response.status + ': ' + errorDetails.message);
```
jsconan: You should also log the error using `core/logger` ([see an example here](https://github.com/oat-sa/tao-core/blob/master/views/js/controller/app.js#L159)).
Maybe should you also triggers an `error` event? This way the container/parent should be notified is something goes wrong.

boajer: Done

PR #1302, opened 2017-05-04
TAO-4326 Updater ACLProxy change

chilly86:


manifest.php
```diff
@@ -35,7 +35,7 @@
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '9.3.3',
+    'version' => '9.3.4',
```
gyszucs: Do we have to change the version in this case as well? Because the only change here is in Updater and it's for an old version change.

PR #1311, opened 2017-05-10
Feature/tao 4318 plugins service

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4318

Generalize the frontend plugins service from the taoTests extension.


jsconan:
I merged develop as I needed the new plugin loader version for a companion PR I am working on

models/classes/plugins/PluginService.php
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Copyright (c) 2016-2017 (original work) Open Assessment Technologies SA;
+ */
+
+namespace oat\tao\model\plugins;
+
+use oat\oatbox\service\ConfigurableService;
+
+/**
+ * Manage module plugins. Should be overridden to provide the right PluginRegistry instance and a SERVICE_ID constant.
```
krampstudio: If it has to be overridden, it could be declared `abstract`, no ?

PR #1312, opened 2017-05-11
Fix/mime detection

ssipasseuth:
The mimetype detection issue by the filesystem getMimetype() is not limited to css files.
We have another issue with the PCI audio recording where svg files are interpreted as text/plain instead of image/svg+xml. To prevent further issue, I replaced it with the TAO's file helper.


ssipasseuth:
@hutnikau as suggested, I revert the change to the old way and added a case for svg file in a similar way as the existing css. This should be compatible yith flysystem now.

PR #1313, opened 2017-05-11
Feature/base64 unicode

jsconan:
Add a base64 encoder/decoder that take care of Unicode strings.

Needed to fix a decoding issue when UTF-8 strings are used.


jsconan:
Indeed. And this helper is using those native methods.
However, there is an issue with unicode strings, especially since the native decoder does not take care of unicode strings.

You can find more explanations on the MDN page: https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding.

PR #1314, opened 2017-05-11
Feature/tao 4279 export exclude empty

ssipasseuth:
Improved the readonly checkboxes form element to optionally show a tooltip.
To be tested with https://github.com/oat-sa/extension-tao-itemqti/pull/929


eacraft13:
Something in here breaks 'APIP Content Package' export.

ssipasseuth:
Ok so Apip and RDf export should be fixed now. Please check again.

helpers/form/elements/class.Checkbox.php
```diff
@@ -72,20 +72,24 @@ protected function getReadOnly()
...
 
     /**
+     * Set an option as readonly
      *
-     * @param $option
+     * @param string $optionId - the optionId to be marked as disabled
+     * @param string $reason - the reason the option is disabled
      */
-    public function addReadOnly( $option )
+    public function addReadOnly($optionId, $reason)
```
eacraft13: I have no problem with this, but it doesn't appear to be used anywhere. Why update it? Instead of removing it?

ssipasseuth: Since it is an existing API, I just change it.
But maybe we could remove it completely.

ssipasseuth:
Updated with latest change

helpers/form/elements/xhtml/class.Checkbox.php
```diff
@@ -66,32 +66,46 @@ public function render()
         }
         $i = 0;
         $checked = 0;
-        $returnValue .= '<div class="form_radlst form_checklst">';
+        $returnValue .= '<div class="form_radlst form_checklst plain">';
```
eacraft13: This appears to be the convention for these form elements to dynamically create html, so perhaps it's better as is. But maybe it would be better to do an include(checkboxTpl.php) instead.

ssipasseuth: yes I see what you meant. I also hesitated to do it in fact but I prefer to stick to the current convention as this form generation library is a legacy code that is going to be replaced by your client side form generator soon, so I see little use to actually improving that a lot

PR #1315, opened 2017-05-11
Fix/tao 3958/file upload security issue

chilly86:


helpers/form/validators/class.FileMimeType.php
```diff
@@ -62,29 +62,30 @@ public function evaluate($values)
...
             }
 
-			if (file_exists($values['uploaded_file'])) {
-				$mimetype = tao_helpers_File::getMimeType($values['uploaded_file']);
-			}
-
-			if (!empty($mimetype) ) {
-                common_Logger::d($mimetype);
+            if (!empty($mimeType) ) {
+                common_Logger::d($mimeType);
```
siwane: Could u remove this debug please

PR #1316, opened 2017-05-11
Feature/tao 4278 add indeterminate checkbox style

no-chris:
related to https://github.com/oat-sa/extension-tao-testqti/pull/828


views/scss/inc/_forms.scss
```diff
@@ -308,6 +308,9 @@ label {
         &:checked ~ .icon-checkbox:before {
             @extend .icon-checkbox-checked:before;
         }
+        &:indeterminate ~ .icon-checkbox:before {
+            @extend .icon-checkbox-indeterminate:before;
+        }
```
krampstudio: could you add it also for radio ? Even if the support is quite recent (only chrome & ff) others will support the property very soon (and at worth the radio will look unchecked)

PR #1337, opened 2017-05-19
Allow dbal installs

jbout:


jbout:
check the PR number ^^

PR #1353, opened 2017-05-31
Turn TAO_VERSION in a more verbose form

dietertaotesting:
Did not do the truncate class, have doubts whether this should be a function

### DIFF
--- a/helpers/Layout.php
+++ b/helpers/Layout.php
@@ -335,6 +335,21 @@ class Layout
     }
 
     /**
+     * Turn TAO_VERSION in a more verbose form
+     *
+     * @return string
+     */
+    public static function getVerboseVersionName() {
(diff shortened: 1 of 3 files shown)


helpers/Layout.php
```diff
@@ -335,6 +335,21 @@ public static function isUnstable() {
     }
 
     /**
+     * Turn TAO_VERSION in a more verbose form
+     *
+     * @return string
+     */
+    public static function getVerboseVersionName() {
+        preg_match('~(?<revision>([\d\.]+))(-?(?<specifics>(.*)?))~', TAO_VERSION, $components);
```
no-chris: Is that possible to have a unit test assessing the behavior of the regexp to ease future maintenance?

PR #1355, opened 2017-05-31
Feature/tao 4415 add operated by info

no-chris:


no-chris:
The "operated by" string is set at install by the system operator (see https://oat-sa.atlassian.net/browse/TAO-4414), those are the default values.

Regarding the obfuscated email address, you'll see that it is encoded if you look at the source HTML (and not at the code inspector in the debugger tools). I'm using the provided TAO helper, whether this is sufficient or not is of course questionable.

Test values have been removed from template

ssipasseuth:
Ok for the obfuscated email on the source. That looks alright to me.

PR #1375, opened 2017-06-13
Refactor/tao 4481/absolute url urlservice

chrisphpgarcia:


bugalot:
Perfectly retro-compatible, so I can merge dependent PRs quietly... thanks :) !

PS: Please not that I had to commit a little fix to avoid a PHP Error in the logs when logging out ;)

PR #1379, opened 2017-06-15
Hotfix/tao 4537/fixing interface install

chilly86:


bugalot:
Excellent thx it works fine now. Please merge this in develop and let's re-release :)

PR #1382, opened 2017-06-15
Fix/fix parser construct

siwane:


bugalot:
Seems good to me! @antoinerobin to confirm!

PR #1396, opened 2017-06-28
Tao 4371 add craete class method to rest controller

hutnikau:


actions/class.RestController.php
```diff
@@ -172,6 +178,50 @@ protected function getClassFromRequest(\core_kernel_classes_Class $rootClass)
...
+        $comment = $this->hasRequestParameter(static::CLASS_COMMENT_PARAM)
+            ? $this->getRequestParameter(static::CLASS_COMMENT_PARAM)
+            : '';
+
+        $class = null;
+
+        /** @var \core_kernel_classes_Class $subClass */
+        foreach ($rootClass->getSubClasses() as $subClass) {
+            if ($subClass->getLabel() === $label) {
+                throw new \common_exception_InconsistentData(__('Class already exists. Class uri: '.$subClass->getUri()));
```
antoinerobin: Shouldn't keep the behavior : get the class that have this label instead of throwing an exception ?

PR #1402, opened 2017-07-05
Implementing JSON serialization of Service Call

tverhoof:
[https://oat-sa.atlassian.net/browse/TAO-4686](https://oat-sa.atlassian.net/browse/TAO-4686)


jbout:
replaced

PR #1405, opened 2017-07-11
Feature/devops 226

tverhoof:


tverhoof:
Indeed! It's fixed now ;)

tverhoof:
I'll ask them. As far as I know, I think the new index.php will only be used on deployed versions on AWS with the healthcheck.

bugalot:
Ok excellent! The thing we developers really need to know is where to store that to make them available to them (important if we have to provide fixes later on)

tverhoof:
I see. But currently, we don't have tested the healthcheck, so we're not 100% sure if we implemented it the correct way, maybe it can still be changed, and we need this PR to be merged to make the tests ;)

bugalot:
I see, can you send me via another communication channel the index.php file you developed so that I can have a look ?

install/utils/class.System.php
```diff
@@ -61,6 +61,46 @@ public static function isTAOInstalled(){
...
+        $generisConf = dirname(__FILE__).'/../../../config/generis.conf.php';
+
+        if (!file_exists($generisConf)) {
+            return false;
+        }
+
+        include_once($generisConf);
+        $installationConf = dirname(__FILE__).'/../../../config/generis/installation.conf.php';
+
+        if (!file_exists($installationConf)) {
```
bugalot: maybe we could use is_readable() instead of file_exists()? Indeed, is_readable checks

if the file exists AND it is readable. Better two checks than a single one isn't it ? :)

tverhoof: OK, but then we may need to change it in System::isTAOInstalled too because file_exists() is used there

install/utils/class.System.php
```diff
@@ -61,6 +61,46 @@ public static function isTAOInstalled(){
...
+
+        include_once($generisConf);
+        $installationConf = dirname(__FILE__).'/../../../config/generis/installation.conf.php';
+
+        if (!is_readable($installationConf)) {
+            return false;
+        }
+
+        $conf = include_once($installationConf);
+        $extIterator = is_array($conf) ? $conf : $conf->getConfig();
```
bugalot: In which context could we have another value than a `oat\oatbox\config\ConfigurationService` object in $conf? Is that really worth? Maybe I'm missing something regarding backward compatibility? :hankey:

tverhoof: I've tested this with the TAO version to download from taotesting.com and it crashed with $conf->getConfig(), I saw that the installation.conf.php returned an array with the extension's version directly as keys.

PR #1408, opened 2017-07-13
Feature/cache item resources

hutnikau:
ACT team found an issue with item caching.
Cache itself works perfect, but resources of items (such as images, css etc) are loaded twice, once during loading of next item data and then during loading item itself:
image
I added `Expires` header to prevent loading already loaded resources:
image


krampstudio:
Linked to TAO-4777

hutnikau:
@krampstudio thanks for review.
Just updated code according to your comments and resolved conflicts.

PR #1420, opened 2017-07-24
Hotfix/tao 4780 cannot move into empty class

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4780

Backport a patch that fix the drag'n drop issue affecting empty classes: cannot move an instance into an empty class.


chazotte:
Hotfix applied but doesn't seem to solve the issue on 3.2-rc.

jsconan:
Ops, yes you are right! I forgot to rebuild the js bundles!

chazotte:
Works!

PR #1423, opened 2017-07-26
Icons Connect/Disconnect

dietertaotesting:
Related to https://github.com/oat-sa/extension-tao-testqti/pull/916
and https://github.com/oat-sa/extension-tao-devtools/pull/97


manifest.php
```diff
@@ -36,7 +36,7 @@
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '12.2.2',
+    'version' => '12.2.3',
```
krampstudio: Adding a new icons increases the feature number. It should be `12.3.0`.

PR #1425, opened 2017-07-31
Feature/tao 3989 resource generic api

siwane:


actions/form/class.RestUserForm.php
```diff
@@ -0,0 +1,281 @@
...
+use \oat\generis\model\user\PasswordConstraintsService;
+use \oat\oatbox\validator\ValidatorInterface;
+use \Zend\ServiceManager\ServiceLocatorAwareTrait;
+use \Zend\ServiceManager\ServiceLocatorAwareInterface;
+
+/**
+ * Class tao_actions_form_RestForm
+ *
+ * Form object to handle rdf resource
+ * It manages data to create and edit a resource into a given class
```
antoinerobin: wrong documentation

actions/form/class.RestUserForm.php
```diff
@@ -0,0 +1,280 @@
...
+        foreach ($properties as $property) {
+            if ($this->isEdition() && $property['uri'] == 'http://www.tao.lu/Ontologies/generis.rdf#login') {
+                $property['widget'] = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#Readonly';
+            }
+        }
+
+        if ($this->isCreation()) {
+            $properties[] = [
+                'uri' => 'password1',
+                'label' => 'Password',
```
krampstudio: Could you please use the translation function around labels

actions/form/class.RestForm.php
```diff
@@ -0,0 +1,436 @@
...
+            if (isset($property['validators'])) {
+                foreach ($property['validators'] as $validatorName) {
+                    $validatorClass = 'tao_helpers_form_validators_' . $validatorName;
+                    if (!class_exists($validatorClass)) {
+                        throw new common_Exception('Validator is not correctly set (unknown)');
+                    }
+                    /** @var ValidatorInterface $validator */
+                    $validator = new $validatorClass();
+                    if (!$validator->evaluate($value)) {
+                        throw new tao_helpers_form_Exception($property['label'] . ' : ' . $validator->getMessage());
```
krampstudio: Can we have more elaborated exceptions when the validation fails ?
But instead of throwing an exception at the 1st invalid field, I'd prefer to have them all and send back a structured response, for example : 

```
422 Unprocessable Entity
Content-Type: application/json

{
   "success" : false,
   "erroCode" : 422,
   "errorMsg" : "Some fields are invalids",
   "data"  : {
         "http://foo.com#LoginProperty" : "A login is required to login",
         "http://foo.com#PasswordProperty" : "A password must contains at least 45 characters",
          "http://foo.com#FirstNameProperty" : "The firstname cannot contains emojis",
   }
}
```

If it's too much work, we can delay this part but this would be awesome. 
At least we should differentiate runtime exception to form validation (they both use  `tao_helpers_form_Exception`).

actions/class.RestResourceController.php
```diff
@@ -0,0 +1,172 @@
...
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
+ *
+ */
+
+/**
+ * Class tao_actions_RestResourceController
+ *
+ * The rest controller to manage resource APIs
+ */
+class tao_actions_RestResourceController extends tao_actions_RestController
```
krampstudio: Those requests are intended to be within a logged in session in TAO. You should disable the HTTP Basic Auth mechanism otherwise any ajax request made will end into https://github.com/oat-sa/tao-core/blob/master/models/classes/routing/TaoFrontController.php#L67

siwane:
@krampstudio @antoinerobin Code updated

krampstudio:
@antoinerobin I don't think. We can handle the PRs separated and merging this PR into develop (testing the Rest API can be sufficient).

actions/class.RestResourceController.php
```diff
@@ -0,0 +1,262 @@
...
+
+use \oat\generis\model\OntologyAwareTrait;
+use \oat\tao\helpers\RestExceptionHandler;
+
+/**
+ * Class tao_actions_RestResourceController
+ *
+ * The rest controller to manage resource APIs
+ */
+class tao_actions_RestResourceController extends tao_actions_CommonModule
```
krampstudio: Could you please rename this controller to "RestResource" otherwise the URL will look like `/tao/RestResourceController/create` and for consistency we never name controller `*Controller`.

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+    {
+        $data = ['data' => []];
+        /** @var common_report_Report $error */
+        foreach ($report->getErrors() as $error) {
+            $data['data'][$error->getData()] = $error->getMessage();
+        }
+
+        if ($withMessage) {
+            $data['success'] = false;
+            $data['errorCode'] = 422;
```
eacraft13: Why not 400? It's more widely understood (422 is much more obscure) and easily reflects that modifications to the request need to be made.

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+            $data['errorMsg'] = 'Some fields are invalids';
```
eacraft13: i18n and 'invalid' with the 's'

*CHANGE

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+            $data['version'] = TAO_VERSION;
+        }
+
+        header('HTTP/1.1 422 Unprocessable Entity');
```
eacraft13: Why is this header needed?

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+        Context::getInstance()->getResponse()->setContentHeader('application/json');
+        echo json_encode($data);
```
eacraft13: Can this be `returnJson()`? I think it's best to let the framework manage returns.

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+        exit(0);
```
eacraft13: I don't think using `exit()` is a good practice. I can see a scenario where you may want to perform cleanup after calling this method. Perhaps this is harmless though...

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+     */
+    protected function returnFailure(Exception $exception, $withMessage=true)
+    {
+        $handler = new RestExceptionHandler();
+        $handler->sendHeader($exception);
+
+        $data = array();
+        if ($withMessage) {
+            $data['success'] = false;
+            $data['errorCode'] = $exception->getCode();
```
eacraft13: Exception codes don't correspond to http status codes. If this is a catch all return, I would suggest 500.

*CHANGE

actions/class.RestResource.php
siwane: This is requirement from @krampstudio , I will put which code you want

actions/class.RestResource.php
```diff
@@ -0,0 +1,264 @@
...
+
+use \oat\generis\model\OntologyAwareTrait;
+use \oat\tao\helpers\RestExceptionHandler;
+
+/**
+ * Class tao_actions_RestResourceController
+ *
+ * The rest controller to manage resource APIs
+ */
+class tao_actions_RestResource extends tao_actions_CommonModule
```
siwane: I'm not very imaginative when I choose my classname. Normally it should follows the rest standart. I can still change the class name, however `tao_actions_Generis` is not very intuitive to understand that this controller/endpoint/api is managing a rdf resource.

actions/class.RestResource.php
siwane: normally header is manager through `returnJson`method with $code as argument. Unfortunately, 422 is not managed by this clearFW (used by this method). It's why I have to do it manually.

actions/class.RestResource.php
siwane: Same as above

actions/form/class.RestUserForm.php
```diff
@@ -0,0 +1,275 @@
...
+                'uri' => 'password1',
+                'label' => __('New password'),
+                'widget' => 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HiddenBox',
+            ];
+
+            $properties[] = [
+                'uri' => 'password2',
+                'label' => __('Repeat new password'),
+                'widget' => 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HiddenBox',
+            ];
```
siwane: @krampstudio What do u think about that? It's posssible

actions/form/class.RestForm.php
```diff
@@ -0,0 +1,445 @@
...
+            // Validators
+            $validators = $this->getPropertyValidators($property);
+            if (!empty($validators)) {
+                $propertyData['validators'] = $validators;
+            }
+
+            // Range values
+            /** @var core_kernel_classes_Class $range */
+            $range = $property->getRange();
+            if(!is_null($range) && $range->getUri() != 'http://www.w3.org/2000/01/rdf-schema#Literal') {
```
siwane: All properties does not have a range rdf property. I put it into response only if I have values on it

actions/class.RestResource.php
siwane: I did like that to have the same behaviour of RestController method.

actions/class.RestResource.php
siwane: I don't agree on this point, if a notFoundException is thrown 404 code has to be rendered, same for 403

actions/form/class.RestForm.php
```diff
@@ -0,0 +1,445 @@
...
+use \oat\tao\helpers\form\ValidationRuleRegistry;
+use \oat\oatbox\validator\ValidatorInterface;
+
+/**
+ * Class tao_actions_form_RestForm
+ *
+ * Form object to handle rdf resource
+ * It manages data to create and edit a resource into a given class
+ */
+class tao_actions_form_RestForm
```
siwane: I did it on form to be consistent with other. Should be on model, agreed

actions/class.RestUser.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param $instance
+     * @return tao_actions_form_RestUserForm
+     */
+    protected function getForm($instance)
+    {
+        $form = new \tao_actions_form_RestUserForm($instance);
+        $form->setServiceLocator($this->getServiceManager());
+        return $form;
+    }
+}
```
siwane: agreed

actions/class.RestResource.php
krampstudio: This [post](https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data#answer-20215807) sums up the difference. At the end, I don't really care. 400 is more generic when 422 means the requests is correct but not it's content.

actions/form/class.RestUserForm.php
krampstudio: yes, let's handle this by the frontend. At the end the backend should only ingest (valid) data. Password confirmation is a UX mecanism to ensure the guy is typing what he thinks he's typing. It can be handle by the frontend

PR #1430, opened 2017-08-02
Fix/copyright 2017

eacraft13:
Updates all copyrights to 2017

Also adds new line characters at the end of files (as is a best practice).


eacraft13:
@krampstudio, this change was implemented via
`git grep -l 'Copyright (c) 2010 (original work) Open Assessment Technologies SA' . | xargs sed -i '' -e 's/Copyright (c) 2010 (original work) Open Assessment Technologies SA/Copyright (c) 2017 (original work) Open Assessment Technologies SA/g'`.
The only changes should be the copyright date and newlines at the end of files.

krampstudio:
That's a good idea, but it should be based on last modification date instead. For example, the file `models/classes/media/TaoMediaException.php` has the correct copyright (2015). This file has been created in 2015, modified in 2015. A `git log` tell us:

```
commit ea6dfb6e3f6716529431348c8381b4a4f7ddd1f0
Author: Joel Bout <joel@taotesting.com>
Date:   Fri Apr 17 11:03:33 2015 +0200

    Fixed license and phpdoc

commit 39a3a04f3d5581f070d0fc31e23ff14fb7b19d6e
Author: Joel Bout <joel@taotesting.com>
Date:   Wed Apr 15 16:33:07 2015 +0200

    Added web source
```

eacraft13:
@krampstudio I looked into copyright dates a little more and I'm not sure my logic was correct. I thought it should be a single year for when the project was last active (i.e. 2017). But apparently, the original creation date is more important. And of less importance is the date last modified. So, 2014, for example, shouldn't be updated to 2017, but instead to 2014 - 2017... I think. 

Anyways, because origination date is most important, I don't believe this is a good pr after all. (At least from my understanding.) Probably best to just close this...

krampstudio:
The idea is good, it should just be improved. In TAO we use the copyright date (creation) (creation-lastModification) to track the intellectual property. This is very important since the project has been develop by a different orgs : public research center, then a German department of education (they have founded a part of the soft), then OAT. 
So if you can adapt the command to take into account the first/last modification year, this would help fixing the headers not updated.

krampstudio:
Closing the PR for now.

PR #1435, opened 2017-08-08
Tao 4893 end session dialog

hutnikau:


views/js/ui/dialog/confirm.js
```diff
@@ -26,11 +26,12 @@ define(['lodash', 'ui/dialog'], function (_, dialog) {
      * @param {String} message - The displayed message
      * @param {Function} accept - An action called when the message is accepted
      * @param {Function} refuse - An action called when the message is refused
+     * @param {Object} options - Dialog options
      * @returns {dialog} - Returns the dialog instance
      */
-    return function dialogConfirm(message, accept, refuse) {
+    return function dialogConfirm(message, accept, refuse, options) {
```
krampstudio: It's good to be able to override the buttons, however I'd have limited the override to the labels. If you change the buttons ids or close property, the confirm won't work anymore. And if you want other buttons layouts, you should use `ui/dialog`.

PR #1436, opened 2017-08-10
Fix/tao 4802 use styled checkboxes

krampstudio:
- Use the checkbox style from the the [style guide](https://hub.taocloud.org/styleguide/checkboxes-and-radio-buttons)
 - With no value, the widget did return an empty jQuery object, ie. `{ length: 0, context: {...}, selector: '.option input:checked', ...}` instead of an empty array.


views/js/ui/generis/widget/checkBox/checkBox.js
```diff
@@ -58,11 +58,13 @@ define([
...
                 if (this.is('rendered')) {
-                    ret = this.getElement()
-                    .find('.option > input:checked')
-                    .map(function () {
-                        return $(this).val();
-                    });
+                    ret = [];
+                    this.getElement()
+                        .find('.option input:checked')
+                        .map(function () {
```
eacraft13: `map` implies a return. If you prefer this style, `each` would be better.

views/js/ui/generis/widget/checkBox/checkBox.js
```diff
@@ -116,4 +118,4 @@ define([
     }
 
     return factory;
-});
```
eacraft13: Good to know I'm causing 'no newline at end of file'. It was never my intent.

views/js/test/ui/generis/widget/checkBox/test.js
```diff
@@ -180,10 +180,85 @@ define([
...
+
+            assert.equal($('.option:nth-child(2) label', $element).text().trim(), 'v2', '2nd option label is correct');
+            assert.equal($('.option:nth-child(2) input', $element).val(), 'http://foo#v2', '2nd option value is correct');
+            QUnit.start();
+        })
+        .render($container);
+    });
+
+
+    QUnit.asyncTest('change value', function (assert) {
```
eacraft13: Thanks for adding this.

views/js/test/ui/generis/widget/checkBox/test.js
```diff
@@ -180,10 +180,85 @@ define([
     /**
      * Events
      */
-    QUnit.module('Events');
+    QUnit.module('Behavior');
 
-    QUnit.test('change & blur', function (assert) {
-        assert.ok(true, 'on(\'change blur\')');
+    QUnit.asyncTest('DOM', function (assert) {
```
eacraft13: This is very much my opinion, but I believe that tests that rely on or check DOM structure and interaction are better left out. 

Reason being that content and styles tend to be changed frequently. And these tests tend to be updated just to make them pass (if they're updated at all). Further, they don't really guarantee a good style and ux. So, regardless of their existence, manual testing is still necessary. Which means that both this test needs to be maintained, and manual testing needs to be performed.

This is something where I think less tests is better.

views/js/ui/generis/widget/checkBox/checkBox.js
eacraft13: After reading the commit message, I see the point of this change. Nonetheless, `each` would be best.

krampstudio:
You shouldn't have merged it with changes required, at least for `each` / `map`,

views/js/test/ui/generis/widget/checkBox/test.js
krampstudio: I understand your point of view, but disagree with it. We don't test style but the rendering computation : 
 - does the template add the values at the correct position (it's more relevant with more complex templates)
 - some nodes have events and behaviour bound to them, we should assert they've not been changed or removed.

eacraft13:
No changes were required. The comments made were opinions, but nothing I would impose.

PR #1437, opened 2017-08-15
Feature/tao 3868 create the add edit user form

eacraft13:
Feature - TAO
Jira: https://oat-sa.atlassian.net/browse/TAO-3868

Requires
- ~~https://github.com/oat-sa/tao-core/pull/1444~~
- ~~https://github.com/oat-sa/tao-core/pull/1493~~

Acceptance Criteria:
- No resource is created unless all fields are valid
- Server-sided validation errors get displayed correctly to the end user


actions/class.RestResource.php
```diff
@@ -99,7 +99,7 @@ public function edit()
             }
         }
 
-        if ($this->isRequestPut()) {
+        if ($this->isRequestPost()) {
```
eacraft13: Should be `PUT` (actually, `PATCH` because the entire resource is not replaced), but only `POST` works.

actions/class.Users.php
```diff
@@ -300,53 +269,25 @@ public function checkLogin()
     }
 
     /**
+     * form to add a user
+     * @return void
+     */
+    public function add()
+    {
+        // ???
+        // $this->setData('loginUri', tao_helpers_Uri::encode(PROPERTY_USER_LOGIN));
```
eacraft13: What was `loginUri` used for?

views/js/ui/generis/form/form.js
```diff
@@ -124,9 +135,12 @@ define([
              * @returns {Object[]}
              */
             serializeArray: function serializeArray() {
-                return _.map(this.widgets, function (widget) {
+                return _(this.widgets)
+                .map(function (widget) {
                     return widget.serialize();
-                });
+                })
+                .flatten()
```
eacraft13: Required for checkBox widget... if multiple key values are required for the roles property. If not, remove this change.

views/js/ui/generis/form/readme.md
```diff
@@ -58,13 +58,20 @@ form.errors;
 form.widgets;
 
 @returns this
-form.addWidget(object widgetData);
-form.removeWidget(string name);
+form.addWidget(object widgetOptions);
+form.removeWidget(string widgetUri);
 form.validate();
+form.toggleLoading(boolean isLoading);
```
eacraft13: Add other new methods.

views/js/ui/generis/validator/validator.scss
```diff
@@ -11,4 +11,8 @@
         border-color: $error;
         color: $error;
     }
+
+    .validation-error {
+        overflow: hidden;
+    }
 }
```
eacraft13: All files should have new lines at end of files.

views/js/ui/generis/widget/checkBox/checkBox.js
```diff
@@ -76,21 +76,40 @@ define([
...
                 }
 
                 return this.config.values;
+            },
+
+            /**
+             * Serializes checkBox widget into a name/value object array for form submission
+             * @returns {Object[]}
+             */
+            serialize: function serialize() {
```
eacraft13: This is required for multiple form data for multiple roles. Remove if this is not needed.

views/js/ui/generis/widget/readme.md
```diff
@@ -54,5 +54,5 @@ widget.set(string value);
 widget.serialize();
```
eacraft13: Add new methods

eacraft13 (commented):
Todo:
- don't require password for RestUser@edit
- allow multiple values for user roles on RestUser

Possibly add frontend validations.

eacraft13:
@siwane, I'll need to make some changes to the backend api to complete this. Specifically, RestUser@edit will need to allow blank passwords (as they are already set). And the api must handle multiple user roles (i.e. a list of values for a single property).

siwane:
@eacraft13 The two points that you requires are normally included in API. 
For multiple properties (userRoles):
``"http://www.tao.lu/Ontologies/generis.rdf#userRoles"=> [
           'http://www.tao.lu/Ontologies/TAOProctor.rdf#ProctorRole',
           'http://www.tao.lu/Ontologies/TAO.rdf#SysAdminRole'
],``
About password, in case of edition, it's not required to have a password field:
https://github.com/oat-sa/tao-core/blob/develop/actions/form/class.RestUserForm.php#L90

Could you double check if these points are not implemented. Otherwise, I can help you.

actions/class.RestResource.php
eacraft13: @siwane, does our framework have issues with PUT? Honestly, unless this were strict REST, I have no issue with this being a POST.

actions/class.RestResource.php
```diff
@@ -128,6 +128,10 @@ public function getRequestParameters()
                 $argumentSplited = explode('=', $argument);
                 $key = urldecode($argumentSplited[0]);
                 $value = urldecode($argumentSplited[1]);
+                // for multiple values
+                if (strpos($value, ',')) {
+                    $value = explode(',', $value);
+                }
```
eacraft13: @siwane, the nature of www-form-urlencoded is to send an array as a comma separated list (e.g. one,two,three). Because of this, I think it best to handle multiple values by parsing on the backend (as opposed to manipulating the data on the frontend). So, is this how you would like to handle multiple values in the backend?

actions/form/class.RestForm.php
```diff
@@ -408,21 +408,21 @@ protected function getFieldValue(core_kernel_classes_Property $property, $range
     }
 
     /**
-     * Check if current form is for edition by checking if resource is not null
+     * Check if current form exists
      *
      * @return bool
      */
-    protected function isEdition()
+    protected function doesExist()
```
eacraft13: @siwane, The word 'edition' in english does not make sense for the context. Following 'isCreation', transforming 'edit' is a bit difficult. The name 'isEditing' or 'isEditifiable' closer... but not great. I switched them to more Rails like syntax.

models/classes/class.UserService.php
```diff
@@ -311,7 +311,7 @@ public function logout()
      */
     public function getAllUsers($options = [], $filters = [PROPERTY_USER_LOGIN => '*'])
     {
-        $userClass = new core_kernel_classes_Class(CLASS_TAO_USER);
+        $userClass = new core_kernel_classes_Class(CLASS_GENERIS_USER);
```
eacraft13: @siwane, is this an acceptable change?

eacraft13 (commented):
@siwane, can you review these changes? Just the ones I tagged you in a comment from this review.

eacraft13:
@krampstudio, there is one more consideration to make for this pr - the position/ordering of the widgets. Accommodations could be made on the frontend to account for the 'position' property of the backend data. The simpler solution is to let the backend sort the properties/widgets passed, which will be rendered on the frontend according to that ordering. I'm not sure ordering on the frontend is worth the effort (or maybe I'm just lazy). But if you think it is needed I can add it.

eacraft13:
Todo:
- The 'label' and 'roles' fields should be positioned/sorted on backend
- Empty 'login' field returns 'Login is already in use.' validation error from backend (should have a 'notEmpty' validator).
- Some backend validation error messages have format, for example, 'Password: Password is empty'. Should remove the redundancy of 'Password'.
- The 'login' field should be checked for uniqueness on edit.
- Using 'CLASS_GENERIS_USER' returns a user with random string login (e.g. RAwwgNOj). This user looks like it shouldn't be touched. Perhaps this shouldn't be returned/displayed when listing users.

actions/class.RestResource.php
krampstudio: I just don't understand why the parameters are handled that way. In my opinion, we have to choose either to use use a `Content-Type:  application/x-www-form-urlencoded` or `Content-Type:  application/json` (or support both).
In the first case you can access the parameters through `$this->getRawParameters()` or `json_decode` on the `php://input` stream. But parsing parameters manually is risky.

actions/class.Users.php
```diff
@@ -300,53 +269,23 @@ public function checkLogin()
...
-                    $this->getEventManager()->trigger(new UserUpdatedEvent($user, $values));
-                    $this->setData('message', __('User saved'));
-                }
-            }
-        }
-
-        $this->setData('formTitle', __('Edit a user'));
-        $this->setData('myForm', $myForm->render());
-        $this->setView('user/form.tpl');
+        $this->setData('uri', tao_helpers_URI::decode($this->getRequestParameter('uri')));
```
krampstudio: Do we still need to decode the URI ? We are trying to get rid of this everywhere it's possible

actions/form/class.RestForm.php
```diff
@@ -211,7 +211,7 @@ public function validate()
                         $validator = new $validatorClass();
                         if (!$validator->evaluate($value)) {
                             throw new common_exception_ValidationFailed(
-                                tao_helpers_Uri::encode($property['uri']), $property['label'] . ' : ' . $validator->getMessage()
+                                tao_helpers_Uri::encode($property['uri']), $validator->getMessage()
```
krampstudio: Please don't encode URIs

actions/form/class.RestUserForm.php
```diff
@@ -36,6 +36,32 @@ class tao_actions_form_RestUserForm extends tao_actions_form_RestForm implements
     /** @var bool If password is change, set it to true to save new one */
     protected $changePassword = false;
 
+    public function initFormProperties(array $properties) {
+        parent::initFormProperties($properties);
+
+        // reorder label property
+        foreach ($this->formProperties as $index => $property) {
+            if ($property['uri'] === 'http://www.w3.org/2000/01/rdf-schema#label') {
```
krampstudio: Please use constants for such kind of URIs

models/classes/class.UserService.php
krampstudio: The difference between the both users' class need to be verified, if one doesn't add _system_ or _technical_ users.

views/js/controller/users/add.js
```diff
@@ -1,27 +1,94 @@
...
+
+                    formData.push({ name: 'classUri', value: classUri });
+
+                    this.toggleLoading();
+                    this.validate();
+
+                    if (!this.errors.length) {
+                        request(route, formData, 'post')
+                        .then(function () {
+                            setTimeout(function () {
```
krampstudio: Why do you need this timeout ?

krampstudio:
@eacraft13 I'm in favor of doing display stuff on the frontend side, the backend has just to serve you the data. Having the label always at the top, then ordering the properties by the order field should be the responsibility of `ui/generis/form`. 
The main reason, is that we could use this service in other situations.

views/js/controller/users/add.js
eacraft13: It was simply for an effect. I will remove it.

models/classes/class.UserService.php
siwane: This is dangerous because all components using this service trust to this method as "get TAO Users".  Please do not do that

krampstudio:
What the current state of this ? 
Is the change of the user class the last blocking change ?

siwane:
Hello, I have other task with higher priority. I go back to this PR asap. I would like to remove the dirty parsing. I want to speak with eric about that.

eacraft13:
@siwane, is this pr now ready for review? Or do your changes from https://github.com/oat-sa/tao-core/pull/1493 need to be merged beforehand?

siwane:
The PR https://github.com/oat-sa/tao-core/pull/1493 is an improvement of parameters parsing, it depends on code from this PR. I separate it in another PR to have more visibility on parsing changes.

eacraft13:
@krampstudio this is now ready for review.

PR #1438, opened 2017-08-16
Feature/tao 4905 persistent audio volume

tikhanovichA:


views/js/ui/mediaplayer.js
```diff
@@ -1874,6 +1884,40 @@ define([
...
 
         /**
+         * Update volume in DBIndex store
+         * @param volume
+         * @private
+         */
+        _storeVolume: function _storeVolume(volume) {
+            return store('mediaVolume')
+                .then(function(volumeStore) {
+                    return volumeStore.clear().then(function() {
```
krampstudio: You don't need to `clear` the store if you do a `setItem` just after

views/js/ui/mediaplayer.js
```diff
@@ -1874,6 +1884,40 @@ define([
...
+         * Get volume from DBIndex store
+         * @private
+         */
+        _getStoredVolume: function _getStoredVolume() {
+            var self = this;
+            return store('mediaVolume')
+                .then(function(volumeStore) {
+                    return volumeStore.getItem('volume')
+                        .then(function(volume) {
+                            if (typeof(volume) !== 'undefined' && jQuery.type !== 'object') {
```
krampstudio: `jQuery.type` ?

manifest.php
```diff
@@ -36,7 +36,7 @@
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '12.8.2',
+    'version' => '12.9.2',
```
krampstudio: increasing the feature number, always resets the hotfix numver to 0 in semver. You should increase 
`12.8.2` to `12.9.0`

views/js/ui/mediaplayer.js
```diff
@@ -1874,6 +1884,40 @@ define([
         },
 
         /**
+         * Update volume in DBIndex store
+         * @param volume
```
krampstudio: If we expect `volume` to be a number, the documentation should be : 

```
/**
 * Store the volume 
 * @param {Number} volume
 */
```

views/js/ui/mediaplayer.js
krampstudio: I'd instead check the type  and apply the rule : 

```js
if(_.isNumber(volume)){
    self.volume = Math.max(_volumeMin, Math.min(_volumeMax, parseFloat(volume)));
}
```

tikhanovichA:
Fixed issues and updated version.

views/js/ui/mediaplayer.js
```diff
@@ -1698,10 +1702,16 @@ define([
                 self.seek(value, true);
             });
 
+            $(document).on('updateVolume', function(event, value) {
+                self.setVolume(value);
+            });
+
             this.$volume.on('change' + _ns, function(event, value) {
                 self.unmute();
+                $(document).trigger('updateVolume', value);
```
eacraft13: Namespace the event: `'updateVolume' + _ns`

views/js/ui/mediaplayer.js
```diff
@@ -1698,10 +1702,16 @@ define([
                 self.seek(value, true);
             });
 
+            $(document).on('updateVolume', function(event, value) {
```
eacraft13: Unbind this event on 'destroy'. There is an `_unbindEvents` method already.

views/js/ui/mediaplayer.js
```diff
@@ -1596,6 +1597,9 @@ define([
             this.autoStart = this.config.autoStart;
             this.autoStartAt = this.config.autoStartAt;
             this.startMuted = this.config.startMuted;
+
+            this._getStoredVolume();
```
eacraft13: Could this be moved to `init()`? `_reset()` is called on `destroy`, which doesn't need to retrieve the stored volume.

views/js/ui/mediaplayer.js
```diff
@@ -1874,6 +1884,35 @@ define([
...
+                .then(function(volumeStore){
+                    volumeStore.setItem('volume', volume);
+                });
+        },
+
+        /**
+         * Get volume from DBIndex store
+         * @private
+         */
+        _getStoredVolume: function _getStoredVolume() {
```
eacraft13: This isn't a requirement, but I'd like to see the following change...

The name `_getStoredVolume` is misleading. Perhaps `_updateVolumeFromStore` is more appropriate.

views/js/ui/mediaplayer.js
```diff
@@ -1874,6 +1884,35 @@ define([
...
+            var self = this;
+            return store('mediaVolume')
+                .then(function(volumeStore) {
+                    return volumeStore.getItem('volume')
+                        .then(function(volume) {
```
eacraft13: I believe these can be chained better to not have so much indentation. If `getItem()` returns a promise, you can simply return that and still use a `then`. 

Better explained through example...

```
store('mediaVolume')
.then(function (volumeStore) {
    return volumeStore.getItem('volume');
})
.then(function (volume) {
    ...
});
```

... I believe this should work.

tikhanovichA:
Fixed

PR #1439, opened 2017-08-16
Fix/tao 4802 fix selector

krampstudio:
Some minor fixes and improvements of the resource selector (right now in the test autthoring) :  
 - Fix the bug with the "load more button" that appear on sub-requests or on search requests
 - Every search request now switch to the list view, since the backend doesn't have yet any way to search within the hierarchy
 - Modify the placeholder text of the search field once a filter query is done. This is a quick fix and we should think to redo this part (from a UX point of view).
 - filters unwanted & unimplemented widget

Added some new methods to the `ui/generis/form` module : 
 - getWidget
 - getValues (to the form values as an object)
 - reset
 - minor UI improvments


views/js/ui/generis/form/form.js
```diff
@@ -173,13 +208,25 @@ define([
 
             $form.on('submit', function (e) {
                 e.preventDefault();
+
                 /**
                  * @event form#submit
                  * @param {Object[]} formData
                  */
-                self.trigger('submit', self.serializeArray());
+                self.trigger('submit change', self.serializeArray());
```
eacraft13: I don't feel that 'change' belongs here. Submit shouldn't imply changes to the form.

views/js/ui/resource/filters.js
```diff
@@ -63,43 +75,111 @@ define([
...
+
                     this.form = generisFormFactory({
-                        properties : data.properties,
+                        properties : properties,
                         values     : data.ranges
                     }, {
                         submitText : this.config.applyLabel,
                         title      : this.config.title
-                    }).on('submit', function(values){
+                    }).on('change', function(){
```
eacraft13: You could listen for `submit` and `reset` instead of `change`. Then, `change` could be removed from ui/generis/form. I believe `change` is more of a ui/generis/widget event than a form event.

views/js/ui/generis/form/form.tpl
```diff
@@ -3,9 +3,14 @@
     <form action="{{formAction}}" method="{{formMethod}}">
         <fieldset></fieldset>
         <div class="toolbar">
-            <button type="submit">
+            {{#if reset}}
+            <button type="reset" class="btn-neutrali small">
```
krampstudio: This is one of the button form https://hub.taocloud.org/styleguide/buttons (btw we don't use bootstrap ;)

PR #1444, opened 2017-08-20
Feature/tao 3868 generis resource backend api

eacraft13:
Feature - TAO
Jira: https://oat-sa.atlassian.net/browse/TAO-3868

Backend changes required for frontend user form


actions/class.RestResource.php
```diff
@@ -128,6 +128,10 @@ public function getRequestParameters()
                 $argumentSplited = explode('=', $argument);
                 $key = urldecode($argumentSplited[0]);
                 $value = urldecode($argumentSplited[1]);
+                // for multiple values
+                if (strpos($value, ',')) {
+                    $value = explode(',', $value);
+                }
```
eacraft13: via @krampstudio 

I just don't understand why the parameters are handled that way. In my opinion, we have to choose either to use use a `Content-Type:  application/x-www-form-urlencoded` or `Content-Type:  application/json` (or support both).
In the first case you can access the parameters through `$this->getRawParameters()` and `json_decode` on the `php://input` stream for the other case. 

But parsing parameters manually is risky.

actions/form/class.RestForm.php
```diff
@@ -211,7 +211,7 @@ public function validate()
                         $validator = new $validatorClass();
                         if (!$validator->evaluate($value)) {
                             throw new common_exception_ValidationFailed(
-                                tao_helpers_Uri::encode($property['uri']), $property['label'] . ' : ' . $validator->getMessage()
+                                tao_helpers_Uri::encode($property['uri']), $validator->getMessage()
```
eacraft13: via @krampstudio 

Please don't encode URIs

actions/form/class.RestUserForm.php
```diff
@@ -36,6 +36,32 @@ class tao_actions_form_RestUserForm extends tao_actions_form_RestForm implements
     /** @var bool If password is change, set it to true to save new one */
     protected $changePassword = false;
 
+    public function initFormProperties(array $properties) {
+        parent::initFormProperties($properties);
+
+        // reorder label property
+        foreach ($this->formProperties as $index => $property) {
+            if ($property['uri'] === 'http://www.w3.org/2000/01/rdf-schema#label') {
```
eacraft13: via @krampstudio 

Please use constants for such kind of URIs

actions/form/class.RestUserForm.php
```diff
@@ -36,6 +36,32 @@ class tao_actions_form_RestUserForm extends tao_actions_form_RestForm implements
...
+            if ($property['uri'] === PROPERTY_USER_ROLES) {
+                $rolesIndex = $index;
+            }
+        }
+        if ( isset($rolesIndex) ) {
+            $out = array_splice($this->formProperties, $rolesIndex, 1);
+            array_splice($this->formProperties, -1, 0, $out);
+        }
+    }
+
```
eacraft13: @siwane, I believe this method override is going to be removed. Sorting the form will be done solely on the frontend. As for the 'label' and 'roles' properties not having a position, I don't think that will be an issue.

actions/class.RestResource.php
siwane: I totally agree about this point, this is dirty... But first key of array are escaped, and uris are not consistent. I will take time will @eacraft13 to fix this point

eacraft13:
@siwane, would you be able to switch `CLASS_GENERIS_USER` to `CLASS_TAO_USER`?

actions/form/class.RestUserForm.php
```diff
@@ -211,4 +223,4 @@ protected function getUserClass()
     {
         return $this->getClass(CLASS_GENERIS_USER);
```
eacraft13: The original form uses `CLASS_TAO_USER`. Could this be changed to match? Using `CLASS_GENERIS_USER` shows unwanted users when listing users on tao (specifically, the user created with a random string label).

actions/class.RestUser.php
```diff
@@ -82,6 +82,10 @@ protected function getClassParameter()
         $class = parent::getClassParameter();
         $rootUserClass = $this->getClass(CLASS_GENERIS_USER);
```
eacraft13: Change to `CLASS_TAO_USER`.

PR #1446, opened 2017-08-22
Feature/tao 4930 feedback chart bundle

eacraft13:
Required for bundling a requirejs module as a standalone library


krampstudio:
I'll merge this one. almond can be useful and since it's in node_modules it has no impact on the source tree

PR #1450, opened 2017-08-27
Feature/tao 4482 redesign test selector for delivery creation 2

eacraft13:
Feature - TAO
Jira: https://oat-sa.atlassian.net/browse/TAO-4482

Ancillary:
- https://github.com/oat-sa/extension-tao-delivery-rdf/pull/103

Acceptance Criteria:
- I can quickly find my test on the list
- long lists of tests do not pose a problem


views/js/ui/generis/widget/comboSearchBox/comboSearchBox.js
```diff
@@ -0,0 +1,221 @@
...
+            var $document, $el, $input, $dropdown, $dropdownSearch, $dropdownMenuItem;
+
+            $el = this.getElement();
+
+            $document = $(document);
+            $dropdown = $el.find('> .right > .dropdown');
+            $dropdownSearch = $el.find('> .right > .dropdown > .search > input');
+            $dropdownMenuItem = $el.find('> .right > .dropdown > .menu > .item');
+            $input = $el.find('> .right > .input');
+
```
no-chris: Why not use a classe or a data attributes directly on the element to select it ? Those kind of selector have a tight coupling with the html markup.
Or at the very least re-use the previous selected element:
`$dropdown = $el.find('> .right > .dropdown');
$dropdownSearch = $dropdown.find('> .search > input');
$dropdownMenuItem = $el.find( > .menu > .item');`

views/js/ui/generis/widget/widget.scss
```diff
@@ -58,4 +67,114 @@
         text-align: right;
         width: 99%;
     }
+
+    &.combo-search-box {
+        position: relative;
+
+        > .right {
+            > .input {
```
no-chris: Again I don't see why enforcing the markup coupling here with the descendant selector *everywhere* instead of accessing elements directly with a specific class name. If something needs to be wrapped one day in a div or a span then both the CSS and the JS selectors will break. The structure of the HTML should not be repeated in the CSS.
If there is a specific reason for this, please explain.

views/js/ui/generis/widget/comboSearchBox/comboSearchBox.js
eacraft13: I've updated this but not exactly how you had in mind. My intention is to keep as much specificity as justifiable. Where I agree with you is that `.right`, a class that implies positioning, does not belong in this jquery lookup. 

However, reviewing the component template structure, the `.left` and `.right` elements is also meant to imply the difference between the label and widget/input/non-label (so I added classes `.label` and `.widget` respectively). 

Now, from the above code, a dev can read that a `$dropdownMenuItem` is an `.item` that is part of a `.menu` that is part of a `.dropdown` that is part of the `.widget`.

views/js/ui/generis/widget/widget.scss
eacraft13: Again, I would argue that this specificity assists a dev to envision the dom. That is, it adds a classical structuring, which could be helpful when adding a div as well.

eacraft13:
@no-chris, in case they're missed, I replied to the now outdated comments you made.

no-chris:
The use of the new resources selector instead of this component has been agreed on

PR #1455, opened 2017-08-30
Fix/tao 4964/no js error message

tikhanovichA:


views/js/layout/requirement-check.js
```diff
@@ -31,7 +31,9 @@
 (function () {
     'use strict';
 
-    var jsFeedback = document.getElementById('js-check');
+    var noscript = document.getElementsByTagName("noscript")[0];
+    var jsFeedback = noscript.textContent||noscript.innerHTML;
```
chilly86: Handle the situation when noscript is not presented!

tikhanovichA:
@chilly86 fixed

PR #1475, opened 2017-09-21
Refactoring/tao 5102/remove globals from core extension

ionutpad:


ionutpad:
This can be merged after this PR https://github.com/oat-sa/extension-tao-item/pull/283 will be merged.

helpers/form/elements/Model.php
```diff
@@ -56,7 +57,7 @@ public function getOptions() {
         foreach (parent::getOptions() as $optUri => $optLabel) {
             $model = new core_kernel_classes_Resource(tao_helpers_Uri::decode($optUri));
             $status = $model->getOnePropertyValue($statusProperty);
-            if (!is_null($status) && $status->getUri() != TAO_ITEM_MODEL_STATUS_DEPRECATED) {
+            if (!is_null($status) && $status->getUri() != ItemModelStatus::INSTANCE_DEPRECATED) {
```
antoinerobin: You should remove this part of the condition so we don't have a implicit dependency to taoItems. (plus this deprecated item model shouldn't be used)

ionutpad: Done

PR #1478, opened 2017-09-22
Fix/tao 5032/handle empty theme name

chilly86:


models/classes/theme/ThemeService.php
```diff
@@ -166,12 +166,12 @@ protected function getThemeIdFromThemeDetailsProviders()
         foreach ($providers as $provider) {
             if ($provider instanceof ThemeDetailsProviderInterface) {
                 $themeId = $provider->getThemeId();
-                if (!empty($themeId)) {
+                if (!empty($themeId) && $themeId !== ' ') {
```
krampstudio: Can we have a `themeId` with a single space ?

chilly86: It's the default value of the combobox which is a surprise for me, because an empty string would be better. But I guess the reason was to being able to "reset" the field. So I need to handle this situation too.

scripts/update/Updater.php
```diff
@@ -915,7 +915,7 @@ public function update($initialVersion) {
         }
 
 
-        $this->skip('12.2.2', '12.21.4');
+        $this->skip('12.2.2', '12.21.54');
```
krampstudio: It should be `'12.21.5`

PR #1479, opened 2017-09-26
Fix/tao 5178/ignoring commas between quotes

tikhanovichA:


helpers/data/class.CsvFile.php
```diff
@@ -220,8 +220,8 @@ public function load($path)
 			foreach ($rows as  $row){
 				if (trim($row) != ''){
 					$data[$lineNumber] = array();
-					
-					$rowFields = array_map('rtrim', explode($DELIM, $row));
+
+                    $rowFields = array_map('rtrim', str_getcsv($row, $DELIM, '"'));
```
Alroniks: Need to get field encloser from options against hardcoding.

PR #1489, opened 2017-10-03
Fix/prevent no login

krampstudio:
A few times I've observed impossibilities to login (and some customers have reported us this issue too). I can only reproduce it very rarely but this time I was able to catch it : 

screenshot

It seems the DOM is not always ready when creating a `feedback` (from `ui/feedback`). It usually will take more time to load the js modules than the DOM, but it seems we have situations where this occur.

Since the bug itself is very difficult to reproduce, Please test there's no regressions.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -36,7 +36,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '13.0.2',
+    'version' => '13.0.3',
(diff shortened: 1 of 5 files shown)


manifest.php
```diff
@@ -36,7 +36,7 @@
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '13.0.1',
+    'version' => '13.0.3',
```
jsconan: Why is there such a jump? Is it for backport purpose?

PR #1490, opened 2017-10-04
Fix/clean request body read

siwane:
PR to improve the request body reading for user rest api.

Require https://github.com/oat-sa/tao-core/pull/1437


eacraft13:
@siwane, I'm going to reject this pr in favor of https://github.com/oat-sa/tao-core/pull/1493 so that it is easier to understand the proposed changes.

PR #1493, opened 2017-10-05
Clean request body input

eacraft13:


eacraft13:
These changes are currently failing due to 'missing classUri'. However, I'm not even able to make it to the @create action (https://github.com/oat-sa/tao-core/pull/1493/files#diff-fc69783ec85d0bb667b666ecb4b0e066R54), so I'm not sure why this is happening or how to diagnose.

views/js/controller/users/add.js
```diff
@@ -79,7 +79,12 @@ define([
                     this.validate();
 
                     if (!this.errors.length) {
-                        request(route, formData, 'post')
+                        request(
+                            route,
+                            this.toJson(),
```
eacraft13: I moved serialization to `ui/generis/form`. Previously, the form was serialized to name/value pairs (e.g. `{ name: '#label', value: 'user label' }`). This was to accommodate `form-url-encoded`. As I understand, you want the json to be serialized as `{ '#label': 'user label' }`. If not, this can be reverted.

eacraft13: Regardless of how the data is sent, I still receive a 500 error from server.

siwane: You have improved the serialization, and the backend was not in sync with it. Now, 500 error is solved

siwane: Another problem is no **classUri** or **uri** parameter is sent to the server during POST request. Could you check that?

PR #1499, opened 2017-10-10
Fix/ui modal qunit test

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-5161

### DIFF
--- a/views/js/test/ui/modal/test.js
+++ b/views/js/test/ui/modal/test.js
@@ -162,15 +162,14 @@ define([
 
     // Modal position does not depend on the main page scroll (BODY OR HTML)
     QUnit.asyncTest('mainScroll', function(assert) {
-        QUnit.expect(2);
+        QUnit.expect(1);
 
         var $container = $('#modal-1');
         var $modal = $(modalTpl());
         $container.append($modal);
 
-        $modal.on('opened.modal', function() {
-            assert.ok(true, "The modal is visible");
-            assert.equal($modal.css('top'), '40px', 'Default scroll value used for the position');
+        $modal.on('animate.modal', function(op, to) {
+            assert.equal(to.top, '40px', 'Default scroll value used for the position');
             QUnit.start();
         });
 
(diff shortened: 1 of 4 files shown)


views/js/test/ui/modal/test.js
```diff
@@ -189,9 +188,8 @@ define([
         $container.append($scrolledContainer);
         $scrolledContainer.scrollTop(500);
 
-        $modal.on('opened.modal', function() {
-            assert.ok(true, "The modal is visible");
-            assert.equal($modal.css('top'), '540px', 'Modal window opened in the scrolled position');
+        $modal.on('animate.modal', function(op, to) {
+            assert.equal(to.top, '540px', 'Modal window opened in the scrolled position');
```
ssipasseuth: I am afraid that testing now the to.top value is not as good as testing the  $modal.css('top') you have before. Indeed $modal.css('top') tests really what the users sees whereas testing the value to.top does not offer the warranty that the modal is properly positioned. Would you agree ?

views/js/ui/modal.js
```diff
@@ -240,6 +242,7 @@ define(['jquery', 'core/pluginifier', 'core/dataattrhandler'], function ($, Plug
 
                 $element.show();
 
+                $element.trigger('animate.' + pluginName, to);
                 if (options.animate && $element.is(':visible')) {
```
ssipasseuth: so I guess that the actually issue is due to the fact that the callback onOpen() in `$element.animate(to, options.animate, onOpen);` is called. The real fix, isn't to understand why ? maybe something in the doc of $.animate would gives some hint ?

views/js/test/ui/modal/test.js
zagovorichev: I think that the cause of this is PhantomJs, I don't know why top always auto

zagovorichev: I found the reason of this (we need position absolute for the element to move it)

zagovorichev: I'm going to prepare new PR, this one will be closed

zagovorichev: https://github.com/oat-sa/tao-core/pull/1501

ssipasseuth: Ok thanks !

PR #1505, opened 2017-10-12
Feature/switch component

krampstudio:
Added a switch component (1st implementation, some improvements will be required)


views/js/ui/switch/switch.js
```diff
@@ -0,0 +1,199 @@
...
+            })
+            .on('render', function(){
+                var self = this;
+                var $component = this.getElement();
+                onElt = $('.' + states.on, $component);
+                offElt = $('.' + states.off, $component);
+
+                //switch on click
+                $component.on('click', 'span', function(e){
+                    e.preventDefault();
```
ssipasseuth: There is something wrong with the even binding and state management: indeed if I use this component, the events change, on or off are not triggered. If I click on it, visually it toggles however it does not seem that the internal state has been updated. 
If I inspect the dom, the active class is always set to the same element after the toggle. It works visually only because the style relies on the state of the checkbox but not on the component state. So what about choosing only one source of truth to store the state of the widget, either the checkbox or the component state ?
Maybe removing the checkbox and relying solely on the component state to toggle would be enough and make the component more clear ? you could simply use the the state of the component to show and hide the spans and there would be no need to add remove the active class to the spans:
```
.switch {
    label {
        //inactive by default
        .on {
            @include switch-default;
        }
        .off {
            @include switch-active-off;
        }
    }
    &.on {
        //activate style with component's class
        label {
            .on {
                @include switch-active-on;
            }
            .off {
                @include switch-default;
            }
        }
    }
}

Comment (krampstudio) [reply views/js/ui/switch/switch.js]: I'll fix it but I'll keep the checkbox, this will ensure accessibility.

**Comment by krampstudio** [comment]
I've updated the implementation, the binding is now done only on the checkbox which allows you to use your keyboard to control it.

**Comment by ssipasseuth** [comment]
Cool thanks for the update. The toggle works ok and I can listen to the events now.
But I have noticed now 2 problems: 
1. the unit test 31 "Behavior: toggle by clicking" doesn't start
2. the initialization of the state via initial config `{on:{active:true}}` does not seem to work either (I always have the switch turned off with this option on). I used it this way:
```
switchFactory($switch, {
                            on : {
                                label : __('active'),
                                active : !$li.hasClass('disabled')
                            },
                            off : {
                                label : __('disabled'),
                                active : $li.hasClass('disabled')
                            }
                        })
```
where console.log($li.hasClass('disabled')) = false

PR #1506, opened 2017-10-12
Feature/tao 4814 requirejs bundles

eacraft13:
Feature - TAO
Jira: https://oat-sa.atlassian.net/browse/TAO-4814

Note: each extension should have a pr with a dependency on this pr


krampstudio:
Could you please explain your approach. 

How are you going to handle the initial problem which is to not include nested dependencies from other extensions ?
Why are you moving the configuration to the server ?

eacraft13:
The main difference to this approach is in the use of requirejs bundles feature (http://requirejs.org/docs/api.html#config-bundles). And to make this work, every 'bundle' will need to declare which packages it will contain in the requirejs config (thus, every extension will need to be touched).

For how this relates to removing other extensions code from a bundle...

Using taoQtiTest as an example. Currently, taoQtiTest bundles many taoQtiItem modules (let's estimate around 40). If I simply remove those from the taoQtiTest bundle, then that bundle will only be aware of each individual location for each module, and will still need 40 or so requests to get each taoQtiItem module (that is, the taoQtiTest bundle doesn't know that the taoQtiItem modules it needs is all in another optimized bundle). To summarize, by only removing taoQtiItem from taoQtiTest bundle in our current setup, it would still require about 41 http requests.

Using the bundle features essentially allows optimized files (bundles) to know about other optimized files. So, taoQtiTest bundle can get all the needed taoQtiItem modules with one http request.

PR #1512, opened 2017-10-17
Fix/tao 5385 tree classname

dietertaotesting:


views/js/lib/jsTree/themes/scss/inc/_extension-icons.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../../.././../../scss/inc/fonts/tao-icon-classes';
```
krampstudio: This folder is in the include path, you can use : 
```
@import "inc/fonts/tao-icon-classes";
```

views/js/lib/jsTree/themes/scss/inc/_extension-icons.scss
```diff
@@ -0,0 +1,25 @@
...
+    'Test':          'icon-test',
+    'Trees':         'icon-tree',
+);
+
+@each $extensionId, $extensionIcon in $extensionIcons {
+    $extensionClass: to-lower-case(str-replace($extensionId, ' ', $replace: '_'));
+    #{'.taotree.taotree-' + $extensionClass} .node-instance ins:before {
+        @extend #{'.' + $extensionIcon + ':before'};
+    }
+}
```
krampstudio: We can do something simpler. Why using a string transformation when we can list the target classes directly :  
```
$extensionIcons: (
    'battery':       'icon-filebox',
    'restult_server' : 'icon-result-server'
   //etc.
)
```

This will prevent us from using string manipulation functions

views/scss/inc/_functions.scss
```diff
@@ -281,3 +281,20 @@ Usage:
 @mixin disableSelect() {
     @include vendor-prefix(user-select, none, property);
 }
+
+
+/// Replace `$search` with `$replace` in `$string`
+/// @author Hugo Giraudel
```
krampstudio: If you still want to add it, you need to add the source (ie. css-tricks.com)

PR #1516, opened 2017-10-18
Fix/tao 5367/js tree horizontal scroll

zagovorichev:


krampstudio:
I'd like to have @dietertaotesting review too

views/js/lib/jsTree/themes/scss/style.scss
```diff
@@ -14,6 +14,7 @@
         overflow: visible;
     }
     ul {
+        display: table;
```
krampstudio: Why do you use a `table` display ?

zagovorichev: to make this space

dietertaotesting: You need to fix the underlying problem instead. When you look at the blue overlays of the tree items you see that they are too short if a text is rather long. This is an old bug and also the reason you cannot add this space at the right without much trouble.

Override `.tree li a { display: block }` with the following snippet and remove the table stuff again.
```.tree li a {
    display: inline-block;
    margin-right: 10px;
    padding-right: 5px;
}
```

zagovorichev: yes, it works, but without external space after the row:

dietertaotesting: check if the margin is taken in account, it might be overwritten somewhere

dietertaotesting: Turns out that above code solves the problem but does not add a bit of extra space on the right hand side in Chrome (margin no effect). I suggest to use the `display` and `padding` part only. You might be able to add that space with `:after`, but if not it's also acceptable to just increase the padding a bit, maybe 10 or 12 px.

dietertaotesting: Should mention that adding padding to any of the parent elements would also not extra space in Chrome, hence the `:after` suggestion

zagovorichev:
New implementation without tables: https://github.com/oat-sa/tao-core/pull/1523

PR #1520, opened 2017-10-20
Feature/config theme

dietertaotesting:
# [WIP] do not merge!
All PRs are now complete:
https://github.com/oat-sa/extension-tao-premium/pull/5
https://github.com/oat-sa/package-multitenancy/pull/2 (merged)

### DIFF
--- a/helpers/Layout.php
+++ b/helpers/Layout.php
@@ -25,6 +25,7 @@ use Jig\Utils\StringUtils;
 use oat\tao\model\menu\Icon;
 use oat\tao\model\OperatedByService;
 use oat\tao\model\theme\ConfigurablePlatformTheme;
+use oat\tao\model\theme\ConfigurableTheme;
 use oat\tao\model\theme\Theme;
 use oat\tao\model\theme\ThemeService;
 use oat\oatbox\service\ServiceManager;
@@ -210,7 +211,10 @@ class Layout
     public static function getLogoUrl()
     {
         $theme = self::getCurrentTheme();
-        if ($theme instanceof ConfigurablePlatformTheme) {
+        if (
+            $theme instanceof ConfigurableTheme ||
+            $theme instanceof ConfigurablePlatformTheme
+        ) {
             $logoFile = $theme->getLogoUrl();
             if (! empty($logoFile)) {
                 return $logoFile;
@@ -266,7 +270,10 @@ class Layout
     public static function getLinkUrl()
     {
         $theme = self::getCurrentTheme();
-        if ($theme instanceof ConfigurablePlatformTheme) {
+        if (
+            $theme instanceof ConfigurableTheme ||
+            $theme instanceof ConfigurablePlatformTheme
+        ) {
             $link = $theme->getLink();
             if (! empty($link)) {
                 return $link;
@@ -302,7 +309,10 @@ class Layout
     public static function getMessage()
     {
         $theme = self::getCurrentTheme();
-        if ($theme instanceof ConfigurablePlatformTheme) {
+        if (
+            $theme instanceof ConfigurableTheme ||
+            $theme instanceof ConfigurablePlatformTheme
+        ) {
             $message = $theme->getMessage();
             if (! empty($message)) {
                 return $message;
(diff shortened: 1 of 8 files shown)


models/classes/theme/ConfigurablePlatformTheme.php
```diff
@@ -26,118 +26,185 @@
...
+     * @param $arguments
+     * @return mixed
+     * @throws \common_exception_NotFound
      */
-    public function getThemeData()
+    public function __call($optionKey, $arguments)
     {
-        if ($this->hasOption(static::THEME_DATA) && is_array($this->getOption(static::THEME_DATA))) {
-            return $this->getOption(static::THEME_DATA);
+        $optionKey = strtolower($optionKey[3]) . substr($optionKey, 4);
```
siwane: could u check if the method start with `get`?

models/classes/theme/ConfigurablePlatformTheme.php
```diff
@@ -26,118 +26,185 @@
...
+        if ($this->hasOption($optionKey)) {
+            return $this->getOption($optionKey);
         }
+        throw new \common_exception_NotFound('Unknown option "' . $optionKey . '"');
```
siwane: No need to throw an exception, return false should enough, isn't it?

helpers/Layout.php
```diff
@@ -210,7 +210,7 @@ public static function getContentTemplate() {
     public static function getLogoUrl()
     {
         $theme = self::getCurrentTheme();
-        if ($theme instanceof ConfigurableTheme) {
+        if ($theme instanceof ConfigurablePlatformTheme) {
```
siwane: To avoid future refactor, could you put an interface instead of `ConfigurablePlatformTheme` to identify this theme class

dietertaotesting: That would lead to an incompatibility. `DefaultTheme extends Configurable implements Theme` has no method `getLogoUrl()` and has no knowledge of the Logo URL either, so it can't be implemented.

models/classes/theme/ThemeService.php
```diff
@@ -223,4 +232,37 @@ protected function getThemeDetailsProviders()
...
     }
+
+
+    /**
+     * Remove all themes that share a certain class name
+     *
+     * @param $className
+     * @return mixed
+     */
+    public function removeThemesByClassName($className) {
```
chilly86: Why does this return with all of the themes? (From a remove method I'd expect bool return value)

models/classes/theme/ThemeService.php
```diff
@@ -223,4 +232,37 @@ protected function getThemeDetailsProviders()
...
+        $themes = $this->getOption(self::OPTION_AVAILABLE);
+        foreach($themes as $registryId => $theme) {
+            $current = is_array($theme) ? $theme['class'] : get_class($theme);
```
chilly86: themes will be always arrays but I'm not sure they have a class offset

models/classes/theme/ThemeService.php
```diff
@@ -223,4 +232,37 @@ protected function getThemeDetailsProviders()
...
+    }
+
+
+    /**
+     * Remove a theme with a certain id
+     *
+     * @param $id
+     * @return mixed
+     */
+    public function removeThemeById($id) {
```
chilly86: Why does this return with all of the themes? (From a remove method I'd expect bool return value)

models/classes/theme/ConfigurablePlatformTheme.php
```diff
@@ -222,59 +301,184 @@ public function getLabel()
...
-     * 
-     * @return string
-     * 
-     * @TODO: this method can be reusable move to a helper class if you need to use it!
+     * Retrieve all existing custom translatable strings
+     *
+     * @return array
      */
-    public static function convertTextToId($text)
+    public function getAllTexts()
```
chilly86: and you don't need this because of the previously mentioned reason

helpers/Layout.php
```diff
@@ -302,7 +302,7 @@ public static function getLinkUrl()
     public static function getMessage()
     {
         $theme = self::getCurrentTheme();
-        if ($theme instanceof ConfigurableTheme) {
+        if ($theme instanceof ConfigurablePlatformTheme) {
```
chilly86: or ConfigurableTheme or create an interface which implemented by both of them (contains getLink, getMessage, getLogoUrl methods)

dietertaotesting:
Used latest version of this code in TAO 4270. There will be a new PR eventually

PR #1523, opened 2017-10-24
Fix/tao 5367/horizontal scroll for js tree

zagovorichev:


views/js/lib/jsTree/themes/scss/style.scss
```diff
@@ -66,12 +66,18 @@
...
-            margin: 0;
-            display: block;
+            display: inline-block;
+            padding-right: 5px;
             @include font-size(13);
             @include simple-border(transparent);
             overflow: visible;
             text-overflow: ellipsis;
+            &:after {
+                content: "\00a0\00a0\00a0\00a0";
```
krampstudio: Why 4 breaking spaces ? and not only one. For the size ?

krampstudio: I'm just curious

PR #1528, opened 2017-10-26
Feature/tao 5340 single multiple resource selector

krampstudio:
Those changes are required in the resource selector in order to implement the media selector : 

 - Enhanced single/multiple selection mode for the resource selector
 - Trigger change only once on single mode
 - change the default config (to single instead of multiple)
 - expose the selectionModes values through the factory

Companion PR https://github.com/oat-sa/extension-tao-testqti/pull/1120


views/js/test/ui/resource/selector/test.js
```diff
@@ -228,16 +269,16 @@ define([
                 var selection = this.getSelection();
 
                 assert.equal($control.length, 1, 'The selection control exists');
-                assert.equal($control.prop('indeterminate'), false, 'The selection control says no values');
+                assert.equal($control.prop('indeterminate'), false, 'hThe selection control says no values');
```
jsconan: Most useful change ever :) :)

PR #1530, opened 2017-10-30
Feature/tao 5340 select class

krampstudio:
This PR contains changes required for the task it's linked to. New APIs have no implementations so you can test them only through the tests and check for regressions.

 - `ui/class/selector` let's you get the class node 
 - `ui/resource/selector` has an option `selectClass`. When `true` this option makes the class nodes selectable. In this situation the _arrow_ is used to open/close the class.


views/js/test/ui/resource/selector/test.js
```diff
@@ -567,6 +611,9 @@ define([
                 if(params.format === 'list'){
                     this.update(listData, params);
                 }
+            })
+            .on('change', function(selection){
+                window.console.log(selection);
```
jsconan: Is it wanted? or just a forgotten debug line...

krampstudio: an oversight. removed

PR #1532, opened 2017-10-31
Feature/tao 5340 add remove node

krampstudio:
This PR adds new features to the resource selector : the ability to add and remove nodes manually.
It contains only API changes, please validate using the unit test and check for regressions.


jsconan:
I am unable to reproduce the errors in the unit tests. Maybe a performance issue, as I was doing other things in the same time (video record for a demo)...

PR #1536, opened 2017-11-02
Feature/tao 5238 allow origin on placeable tranforms

no-chris:


ssipasseuth:
Oops, have you already pushed the change to taoQtiItem to adapt the breaking change for table alignment ?

no-chris:
I actually did the change in my work branch for the tooltip authoring but yes, it may be better in a seperate PR, especially if we get mid sprint releases

no-chris:
https://github.com/oat-sa/extension-tao-itemqti/pull/1066

PR #1545, opened 2017-11-14
Tao 5448 frontend log

hutnikau:
The reason of changing log levels is to have them compatible with RFC 5424 and the same as used on the server side (PSR-3).


hutnikau:
RFC 5424 is software world standard but not only js world. In my opinion it's better to use one common standard in the TAO then have two different standards for one purpose. We are going to use Fluentd and it has syslog input plugin and supports it. 
Do we need to use not common standard only because it corresponds to javascript's `console` module API? Maybe it's better to create map on client side and have one common standard levels (from server, client or whatever else) in TAO log?
@jbout I would like to know also your opinion.

krampstudio:
The selected levels aren't based on the console API but on _bunyan_ styles (a compatible subset of log4j styles). It's a pure pragmatic choice.

hutnikau:
bunyan is JSON logging module for node.js services (as per description). It's even not used in TAO. Why we need to follow it's logging levels?
Logging levels of this library are only opinion of author (as said in readme). My opinion is that better to use standard (common for entire application) than someone's opinion.

hutnikau:
Bertrand, you said that bunyan levels are used large number of JavaScript libraries. Are we going to use Tao logger for them (maybe in some of libraries there is a possibility to inject custom logger provider) or this logger is supposed to be used only in Tao code?

krampstudio:
No as said, it's a de facto standard. We've introduced it recently and it's very convenient because other libraries use the same levels too, everybody knows them and integration is also easiest. 
I seriously don't see the point of changing them, having to update all extensions where the logger is used (there's a shift in the levels meaning if we use syslog like levels).

hutnikau:
Now client logger is used only in a few places in Tao. If we will decide to change levels now is good time to do that. 
There are popular js libraries which use RFC 5424 (for example winstonjs/winston)

My point to use common levels is that it would be good to have the full sequence of events in the Tao log from client and server side. It will make investigation of support issue much easier and will give full picture of what is happening. Using standard log levels on both sides is good improvement for this reason.

krampstudio:
This is a discussion that should occur during planning since it would require to change all extensions to have the logs to comply with the syslog style, to change the levels threshold (from 7 to 0 in the RFC). and maybe attach this to the task that requires the backend migration to PSR-3

PR #1552, opened 2017-11-17
Tao 5547/specify language for tts in items

IlyaYarkavetsOnePoint:
This additions are within TAO-5547 task. New handlebars helper function is added. It allows to check equation of two variables.
Syntax:
```
{{#ifequals var1 var2}}
    do_something
{{/if}}
```

### DIFF
--- a/views/js/lib/handlebars.js
+++ b/views/js/lib/handlebars.js
@@ -312,6 +312,16 @@ var __module2__ = (function(__dependency1__, __dependency2__) {
       var level = options.data && options.data.level != null ? parseInt(options.data.level, 10) : 1;
       instance.log(level, context);
     });
+
+    instance.registerHelper('ifequals', function(lvalue, rvalue, options) {
+      if (arguments.length < 3)
+          throw new Error("Handlebars Helper equal needs 2 parameters");
+      if( lvalue!=rvalue ) {
+          return options.inverse(this);
+      } else {
+          return options.fn(this);
+      }
+    });
   }
 
   var logger = {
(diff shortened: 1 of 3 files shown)


siwane:
Could u assign the PR to a frontend guy please

views/js/lib/handlebars.js
```diff
@@ -312,6 +312,16 @@ var __module2__ = (function(__dependency1__, __dependency2__) {
       var level = options.data && options.data.level != null ? parseInt(options.data.level, 10) : 1;
       instance.log(level, context);
     });
+
+    instance.registerHelper('ifequals', function(lvalue, rvalue, options) {
+      if (arguments.length < 3)
+          throw new Error("Handlebars Helper equal needs 2 parameters");
+      if( lvalue!=rvalue ) {
```
jsconan: Please use strict operators: `===` or `!==`. Simple comparison operators should not be used, unless you rely on their sick behavior for performance purpose.

views/js/lib/handlebars.js
```diff
@@ -312,6 +312,16 @@ var __module2__ = (function(__dependency1__, __dependency2__) {
       var level = options.data && options.data.level != null ? parseInt(options.data.level, 10) : 1;
       instance.log(level, context);
     });
+
+    instance.registerHelper('ifequals', function(lvalue, rvalue, options) {
```
jsconan: Maybe should you use a simpler name, like `equals`.

IlyaYarkavetsOnePoint:
> For Handlebars, there is already a module to add new helpers, please use it
Thanks for the info! Didn't know that the one already exists.

Closing this pr as another helper with same function already exists.

PR #1556, opened 2017-11-20
Fix/tao 5366/rdf duplications

zagovorichev:
FYI:
quick way to delete duplicates it's a direct query:
```
DELETE FROM statements s where s.id IN (SELECT s1.id FROM statements s1
LEFT OUTER JOIN (
   SELECT MIN(s2.id) as id, s2.modelid, s2.subject, s2.predicate, s2.object, s2.l_language  
   FROM statements s2
   GROUP BY s2.modelid, s2.subject, s2.predicate, s2.object, s2.l_language
) as KeepRows ON
   s1.id = KeepRows.id
WHERE
   KeepRows.id IS NULL)
```

### DIFF
--- a/scripts/tools/rdf/RdfDuplicates.php
+++ b/scripts/tools/rdf/RdfDuplicates.php
@@ -27,6 +27,22 @@ use oat\oatbox\extension\AbstractAction;
 
 
 /**
+ * Counts all duplicates of the rdf properties
+ *   [properties]
+ *      -- fix - delete all duplicate rows from the statements table
+ *          (duplicate row - this is row which has same fields: modelid, subject, predicate, object, l_language)
+ *
+ * To quickly delete all duplicates can be used SQL query:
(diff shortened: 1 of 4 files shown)


scripts/tools/rdf/RdfDuplicates.php
```diff
@@ -0,0 +1,105 @@
...
+namespace oat\tao\scripts\tools\rdf;
+
+use common_report_Report as Report;
+use oat\generis\model\OntologyAwareTrait;
+use oat\oatbox\extension\AbstractAction;
+
+
+/**
+ * Class FindDuplicates
+ * @package oat\taoNccer\scripts\tools
```
krampstudio: Could you just add a small warning saying data will be changed, we will publish it on a customer env

zagovorichev:
@krampstudio Update

PR #1557, opened 2017-11-21
add long format with microseconds

hutnikau:


helpers/class.Date.php
```diff
@@ -40,6 +40,8 @@ class tao_helpers_Date
 
     const FORMAT_ISO8601 = 3;
 
+    const FORMAT_LONG_MICROSECONDS = 4;
```
zagovorichev: You have to define this new format in the default formatters: `tao/helpers/dateFormatter`

PR #1561, opened 2017-11-24
Feature/allow json ld not resource

antoinerobin:


ionutpad:
Moved to https://github.com/oat-sa/tao-core/pull/1774

PR #1565, opened 2017-11-29
Fix/tao 5671 bulk action shortcuts

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-5671

The `bulkActionPopup` was not taking care of the keyboard. The modal dialog implementation is basically supporting the Escape key, but that is not enough.

So, this fix is a refactoring of the component with support of keyboard shortcut (optional, but activated by default).

The `bulkActionPopup` component now offers an API to validate/cancel the dialog, and is listening to keyboard (Escape to cancel, Enter to validate)


jsconan:
Good catch! I will fix that

PR #1566, opened 2017-11-30
Feature/tao 5663 better itemstore

krampstudio:
Create an "in memory" backend for `core/store`. The goal is to use the same API for the offline/pre-caching itemstore, so if we need to use another way to store items the API matches.
In addition, this kind of store can match other use cases. 

Please validate the PR by running the unit tests.


views/js/core/store/memory.js
```diff
@@ -0,0 +1,145 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2016 (original work) Open Assessment Technologies SA ;
```
jsconan: Wrong copyright year?

views/js/core/store/memory.js
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+/**
+ * LocalStorage backend of the client store
```
jsconan: Wrong description?

krampstudio:
I've updated the comments. Could you please review it again?

PR #1581, opened 2017-12-14
Feature/tao 5701 provider loader

jsconan:

Add a core component that manages the loading of providers. It will behalf slightly like the plugins loader, and will be used aside the providers registry.

There is no integration for now, so please check by using the unit test:
`grunt connect:dev qunit:single --test=/tao/views/js/test/core/providerLoader/test.html`


no-chris:
maybe a backend developer could review this also ?

krampstudio:
The overall looks good, however I'm asking if we won't benefit also to have a shared `core/moduleLoader` so the `core/pluginLoader`, `core/providerLoader` (and eventually `taoQtiTest/runner/proxy/loader`) can benefit. I'm asking because except a few methods they share more or less the same behavior. Maybe you already thought about it and decided to go that way for a good reason.

jsconan:
Yes you are right! I did not make the change now because I was unsure on the full use case, and wanted to first validate the architecture (my first try was simpler, for instance I didn't see a need for categories in providers at a first glance, but then I added it back). However I plan to factorize the two components now the use case seems stabilized.

jsconan:
I factorized `pluginLoader` and `providerLoader` with the shared `moduleLoader`.

PR #1582, opened 2017-12-14
Feature/tao 5701 provider registry

jsconan:

Add an abstraction for a providers registry. It will behalf slightly like the plugins registry, and will be used aside the providers loader.

There is no integration for now, so please check by using the unit test:
`vendor/phpunit/phpunit/phpunit tao/test/providers`


no-chris:
maybe a backend developer could review this also ?

jsconan:
Sure! I added @antoinerobin as reviewer on every related PR that introduce backend stuff.
The PR are not so urgent, so we can wait for his return next week. I will use integration branches meanwhile.

jsconan:
I factorized the plugins and the providers registries into one common module registry

PR #1583, opened 2017-12-18
Feature/tao 5637/fixed extra role rights

zagovorichev:


jbout:
Quick-fix for urgent security use case, but requires full solution

PR #1586, opened 2017-12-19
Feature/ajax mock

jsconan:

Add a tool for unit tests: an AJAX mock.
This pattern was used in several unit tests, but instead of copying it once again I decided to generalize it.

You can test the tool using the provided unit test:
`grunt connect:dev qunit:single --test=/tao/views/js/test/mocks/ajax/test.html`

To use the tool in a unit test:
```javascript
define([..., 'test/mocks/ajax', ...], function(..., ajaxMock, ...) {
    QUnit.module('myTest');

    // backup/restore AJAX manager between each test
    QUnit.testStart(function() {
        ajaxMock.push();
    });
    QUnit.testDone(function() {
        ajaxMock.pop();
    });

    QUnit.asyncTest('test with ajax', function(assert) {
        ...
        function calledBeforeResolve(...) {
            assert...
        }
        ...
        ajaxMock.mock(booleanSuccessOrFail, expectedResponse, calledBeforeResolve);
        ...
        $.ajax(...)
            .done(function(data) {
                assert...
            })
            .fail(function(err) {
                assert...
            });
        ...
    });
});
```


krampstudio:
What about https://github.com/jakerella/jquery-mockjax, I think we can use it as QUnit plugin, no ?

jsconan:
Yes, mockjax is a way better, but it is also a lot more complex.

krampstudio:
The idea behind the use of a library is to not bias in the testing. So we consider the library as part of the testing framework (like the js runtime or qunit). I'm not against merging it, just saying we should consider the case to not maintain a mock.

Personally, I prefer to not  use mock  objects for ajax requests when possible, but to use stubs (you don't want to assert a call is done, but to assert a response is given, how your module behaves for a given response).  Most of the time a plain json file is sufficient.

PR #1588, opened 2017-12-19
Feature/tao 5698 access resource by uri

no-chris:
This is the first step to implement this:
https://oat-sa.atlassian.net/browse/TAO-5698

This part is only about having resources accessible from a url parameter. I've also implemented a `generisRouter`so we can navigate into the tree with the browser Back/Forward buttons, along with navigating between sections which was already implemented.

I've tried to test it extensively but this obviously requires a lot of testing, including cross browser. I haven't found regressions although I found some cases where navigation does not work as expected (= when opening item or test authoring is involved), but those issues were already there. Also, see https://github.com/oat-sa/extension-tao-itemqti/pull/1094 for a potential fix (?) of another issue.

For the direct access by URL to work with the uri parameter, the section to which the resource is displayed must be specified. Otherwise, default section is used.


views/js/layout/generisRouter.js
```diff
@@ -0,0 +1,193 @@
...
+    _,
+    eventifier,
+    urlUtil
+) {
+    'use strict';
+
+    /**
+     * The router instance
+     */
+    var generisRouter;
```
krampstudio: If you keep a unique reference to `generisRouter` you don't need a factory, but instead you should return the router with and `init` and a `destroy` method. 
I know you don't use it for now, but it will keep the script idempotent. For example we can do : 

```js
define([], function(){
  return eventifier({
       init(){
          $(window).on('popstate.generisRouter', () => this..restoreState(true));
       },
       //...
       destroy(){
          $(window).off('popstate.generisRouter');
       },
   });
});
```

views/js/layout/generisRouter.js
```diff
@@ -0,0 +1,193 @@
...
+                    // we are restoring in section initialisation: we only need to deal with the section,
+                    // as uri will be read and set during tree initialisation
+                    } else {
+                        this.trigger('section' + state.restoreWith, state.sectionId);
+                    }
+                    topState = state;
+                }
+            },
+
+            hasRestorableState: function hasRestorableState() {
```
krampstudio: Missing doc

no-chris:
@antoinerobin as discussed please review the php modification. Thx!

no-chris:
Requested changes have been made. Could not reproduce the error but I added an extra check on a parameter that could be the cause.

no-chris:
About URI you cannot use them unencoded as they contain a '#'. Best effort is to use a regular URL encoding scheme.

krampstudio:
Oh yes, obviously. The parameters use by default `encodeURIComponent` in the helpers so that should be good. This is just about the "in house encoding" which is abandoned.

views/js/controller/main.js
```diff
@@ -98,7 +100,8 @@ function (module, _, $, __, context, router, helpers, uiForm, section, actions,
                                 section      : context.section,
                                 classUri     : rootNode ? rootNode : undefined
                             },
-                            actions : treeActions
+                            actions : treeActions,
+                            loadNode : section.defaultUri
```
krampstudio: Could you use `uri.encode` here instead of [layout/section](https://github.com/oat-sa/tao-core/pull/1588/files#diff-9d8b21648d369967d99db77edd1ad55fR68) ? 
It will help in the transition to the resource selector...

no-chris: Done !

PR #1598, opened 2017-12-29
Fix/tao 5756/task queue position

zagovorichev:


ssipasseuth:
looks better now, thanks !

PR #1600, opened 2018-01-02
Breaking/tao 5239 brs integration

krampstudio:
This PR aims to integrate the "Resource Selector" in place of "jstree", using a smooth transition.

### Main changes 

1. **tree providers** :  the notion of tree provider has been introduced, with 2 providers `jstree` and `resource-selector`. The default provider is `jstree`. But we can change the default provider by config. We can also select/force the provider by setting the property `type` of the tree node in `structures.xml`. Each provider instantiate the given tree. 

2. **ResourceLookup** : in order to get the data required by the resource selector a new `ResourceService` has been introduced, with different implementation of `ResourceLookup` per supported data layout (`list` or `tree` layout).  All resources and classes are retrieved using a RESTfull way (`RestClass` and `RestResource` controllers). 

3. **updated resource selector** : the component itself has been modified in order to match the resources needs, mostly possiblities to update the tree or the list when something change, to syncrhonize the class list on change.

4. **actionManager** : backoffice actions are handled by the module `layout/actions` based on a register/exec model. In order to dispatch an action over all tree providers, the module is now a `core/eventifier`. 

5. **deleteAll** : a new action available from the `RdfController` is `deleteAll` that let's you remove all the resources (classes included) from a list of URIs.

###  Additional notes 
 - Actions `search` and `filter` from the `structures.xml` have been removed since they didn't do anything and their bindings were not working...

### Testing

The best way to test is to switch the tree providers, either manually on the `structures.xml` or globally, using the script `php index.php 'oat\tao\scripts\install\SetResourceSelector'` or by changing the value in `config/tao/client_lib_config_registry.conf.php` : `layout/tree/loader : { treeProvider }`.
You should test with the 2 providers since `jstree` has also be migrated to provider.

### Known issues
There are some issues that should be fixed, but could be done in a later stage : 

 - The scrollbar is fixed to prevent flickering but this introduce a small shift when using floating scrollbars.
 - The permissions are not handled inside the component.
 - We could improve the class selection style and the hovering too
 - The advanced filtering (by property) works only on a limited set of properties, for example properties with radio buttons are not implemented.

### Companion PR

 - [x] https://github.com/oat-sa/extension-tao-backoffice/pull/36
 - [x] https://github.com/oat-sa/extension-tao-outcome/pull/104
 - [x] https://github.com/oat-sa/extension-tao-mediamanager/pull/91
 - [x] https://github.com/oat-sa/extension-tao-test/pull/206
 - [x] https://github.com/oat-sa/extension-tao-testtaker/pull/74
 - [x] https://github.com/oat-sa/extension-tao-devtools/pull/119
 - [x] https://github.com/oat-sa/extension-tao-item/pull/294
 - [x] https://github.com/oat-sa/extension-tao-testcenter/pull/49
 - [x] https://github.com/oat-sa/extension-tao-group/pull/64
 - [x] https://github.com/oat-sa/extension-tao-delivery-rdf/pull/158
 - [x] https://github.com/oat-sa/extension-tao-community/pull/120
 - [x] https://github.com/oat-sa/extension-tao-dac-simple/pull/57 
 - [x] https://github.com/oat-sa/extension-tao-outcomeui/pull/145
 - [x] https://github.com/oat-sa/extension-tao-monitoring/pull/26


krampstudio:
@jsconan do not hesitate to assign a backend reviewer to review the backend side if needed

jsconan:
Problems I encountered:

- after having updated the platform and applied the mode 'resource-selector', I am logged out from every page where the selector is used (error 403, Access to tao::tao_actions_RestClass::getAll denied)

- in legacy mode (jstree), both in update or fresh install, the actions `delete` and `deleteAll` are displayed (with the same name), for all pages except Booklet, and depending on which mode is applied one of these actions don't work properly

- with the resource selector mode, on page TestTaker, I got this error in the browser's console: `Error: Script error for: ui/generis/widget/textBox/textBox
http://requirejs.org/docs/errors.html#scripterror require.js:166:17
	makeError http://invalsi.tao/tao/views/js/lib/require.js:166:17
	onScriptError http://invalsi.tao/tao/views/js/lib/require.js:1689:36
`

- on Results and Monitoring pages, error 500 (cannot construct the resource because the uri cannot be empty, debug:  (common_exception_Error))

scripts/update/Updater.php
```diff
@@ -1033,6 +1036,14 @@ public function update($initialVersion) {
             $this->setVersion('14.21.0');
         }
         $this->skip('14.21.0', '14.22.0');
+
+        if($this->isVersion('14.22.0')){
+            $this->getServiceManager()->register(ResourceService::SERVICE_ID, new ResourceService());
+            $this->getServiceManager()->register(ListResourceLookup::SERVICE_ID, new ListResourceLookup());
+            $this->getServiceManager()->register(TreeResourceLookup::SERVICE_ID, new TreeResourceLookup());
```
antoinerobin: you should add the ACL here

PR #1607, opened 2018-01-09
Tao 5780/wk1009 unable to add delivery in the test center

IlyaYarkavetsOnePoint:
https://oat-sa.atlassian.net/browse/TAO-5780
needed for  oat-sa/extension-tao-testcenter#50


IlyaYarkavetsOnePoint:
Yep, it seems to be unneeded at first, but... Without this check wrong top margin will be calculated (initial top margin equal to 40px + scrollOffset of the modal parents).

PR #1610, opened 2018-01-10
core/eventifier.spread

krampstudio:
Adds a `forward|` method on `core/eventifier` in order to forward events, this will be useful in components (like the mediaSelector) to forward events between components. 
The PR needs to be validated through the unit test

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -40,7 +40,7 @@ return array(
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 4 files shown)


krampstudio:
Indeed, I should think of another name but forward was suitable :/ 
I'll also add a warning in the `eventifier` itself in case one of the target object has already one of the eventifier method.

jsconan:
Maybe `chain` as replacement name?

krampstudio:
I've changed to `spread` and added a warning when an overwrite is detected.
Please review it again.

PR #1617, opened 2018-01-18
Move getTimeKeys to tao_helpers_Date

hutnikau:


helpers/class.Date.php
```diff
@@ -203,4 +203,70 @@ static function getTimeStampWithMicroseconds(DateTime $dt)
...
+     *     )),
+     *       ...
+     *   )
+     *
+     * @param \DateInterval $interval
+     * @param \DateTime|null $date
+     * @param null $amount
+     * @return \DateTime[]
+     */
+    public static function getTimeKeys(\DateInterval $interval, \DateTime $date = null, $amount = null)
```
Alroniks: Better to use DateTimeInterface instead `\DateTime $date = null` to have ability to use different implementations of DateTime (DateTimeImmutable, implementation from dtms).

PR #1621, opened 2018-01-22
Feature/tao 5148 store get all

krampstudio:
_No rush on this, it's an old story I had the time to finish. Please take the time to perform additional tests_

This PR contains some new features of the `core/store` and the backends module to support the unified store for the test runner and some extras.
 - we keep the date of the last time we open a store, in the store index.
 - new backend _static_ method : `getAll` to list the store. 
 - the backend _static_ method `getAll` and `removeAll` have a `validate` method to filter the stores. This callback has now a new parameter `storeDetails` that contains for example the name of the store and the last time the store was opened.*
  - new backend _static_ method : `cleanUpSpace` that let's you remove _old_ stores to free some space. Default invalidation rules  are based on 2 levels (2 weeks and 2 months).
 - if the [API](https://developer.mozilla.org/en-US/docs/Web/API/StorageManager/estimate) is available we check the disk quotas and perform an upfront clean up 
 - storage backends expose also the `getItems` method


no-chris:
Functionality looks good also. Well done!

krampstudio:
I've fixed the message.
I've also added one change to rename that I'd like to do from a long time : rename `indexdb.js` to `indexeddb.js` (by luck the file was only required by `core/store`)

PR #1622, opened 2018-01-23
Feature/tao 5666/sprint 69/search

tikhanovichA:


manifest.php
```diff
@@ -105,7 +106,8 @@
             SetContainerService::class,
             RegisterResourceWatcherService::class,
             RegisterResourceEvents::class,
-            RegisterActionService::class
+            RegisterActionService::class,
+            RegisterIndexService::class
```
jbout: This service should not be required as the Service already gets set using the default mechanism

PR #1624, opened 2018-01-24
Feature/tao 5816/set sevice state upgrade

chilly86:
depends on https://github.com/oat-sa/generis/pull/453


scripts/install/SetServiceState.php
```diff
@@ -0,0 +1,69 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ * 
+ * Copyright (c) 2014 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
antoinerobin: please use the correct date in the header

PR #1636, opened 2018-02-02
Feature/tao 5898 allow to toggle password visibility on login

no-chris:
https://oat-sa.atlassian.net/browse/TAO-5898

Added a new _ViewableHiddenbox_ Generis Widget. Template and javascript are inlined which is ugly but is the currently used pattern until we move generis form on the frontend.
- the eye is not actually in the input box to avoid conflicts with browser behavior (some add custom icons on password fields, for example) or password managers such as Last pass, that also add icons to the login forms input fields. This makes styling a bit complex.
- to avoid that the password is saved if the test taker submit a plain-text password, the field is reverted back to a password type as soon as a click is made outside of it. Also, the plain text field gets the `autocomplete=off` attribute.
- This has been tested in Chrome, Firefox, Safari Desktop, Safari mobile (iPad), Edge and IE 11.

Related PRs:
- https://github.com/oat-sa/generis/pull/457
- https://github.com/oat-sa/extension-tao-depp/pull/89


helpers/form/elements/xhtml/class.ViewableHiddenbox.php
```diff
@@ -0,0 +1,115 @@
...
+use oat\tao\helpers\form\elements\xhtml\XhtmlRenderingTrait;
+
+/**
+ * Short description of class tao_helpers_form_elements_xhtml_ViewableHiddenbox
+ *
+ * @access public
+ * @author Bertrand Chevrier, <bertrand.chevrier@tudor.lu>
+ * @package tao
+ */
+class tao_helpers_form_elements_xhtml_ViewableHiddenbox extends tao_helpers_form_elements_ViewableHiddenbox
```
antoinerobin: this should be `tao_helpers_form_elements_xhtml_Viewablehiddenbox` because of the way calssname are retrieved `"tao_helpers_form_elements_xhtml_".ucfirst(strtolower($name))`

PR #1637, opened 2018-02-06
Fix/tao 5925 uploader deleter

ssipasseuth:
- Ensure that the deleter is called after each file addition to make the `ui/uploader` also works outside of the backoffice
- Disable the undo function which is causing user confusion


views/js/ui/uploader.js
```diff
@@ -352,6 +353,9 @@ define([
                             .removeClass('placeholder');
                     }
 
+                    //Initialize the deleter on the new file entry
+                    deleter($elt);
```
krampstudio: It appears the deleter to be triggered twice since it has already been setup. By chance the element is already removed so it's deleted only once and the event listener is attached to the element, but that's a possible source of problem.
We have multiple solutions : 
 - don't use the module `ui/deleter` in the uploader anymore, but a plain `$elt.on('click', () =>  $elt.remove())`. So doesn't depend on an external state
 - let it as it is and expect the deleter to be set up, outside. It was the concept of thoose modules : once initialized the markup produces itself the behavior (the Bootstrap way), so your component expects the ui/deleter to be always set up. This means to initialize it in the other GUIs.
 - replace the `ui/deleter` by a contextualized only component
 - make the `ui/uploader` a propoer component

The main issue comes from 2 different conceptions : auto sufficient markup vs self-contained components, and that we're doing a transition from the 1st to the 2nd.

My personal advice would be  to implement the 1st solution, especially if you don't use the _undo_ feature anymore.

ssipasseuth:
makes sense, I have removed the dependency to `ui/uploader` and simply listen to the click

PR #1647, opened 2018-02-13
Improve log, drop trace

jbout:


siwane:
Could u remove unused import.
Could u replace logger to all ActionEnforcer class
it should implement `\oat\oatbox\log\TaoLoggerAwareInterface` to be able to propagate is own logger

PR #1659, opened 2018-02-20
Fix/frontend log messages

hutnikau:


views/js/core/logger/api.js
```diff
@@ -181,7 +181,12 @@ define([
                 };
 
                 if(checkMinLevel(levels.error, level) || message instanceof Error){
-                    err = message instanceof Error ? message : new Error(message);
+                    if (message instanceof Error) {
+                        err = message;
+                    } else {
+                        message = _.isObject(message) ? JSON.stringify(message) : message;
```
krampstudio: That's great, but since calling `log.error({ obj : obj })` is not expected, the correct form is `log.error('message')` or `log.error(new Error())` I'm wondering if we shouldn't also trigger a warning message, like : 

```js  
 if (message instanceof Error) {
                        err = message;
                    } else {
                        message = _.isObject(message) ? JSON.stringify(message) : message;
                        this.log('warning', 'Unexpected call to log.error with an object');
                     }
```

The goal is to help developers fix their code

hutnikau: I'm afraid that call `log` inside of logger itself is not a proper way to warn developers. 
At least almost all the unit test will be broken. Also it may lead to some unexpected results.
What do you think about `console.warn()`?

PR #1666, opened 2018-03-02
Added RestPublisher role

zagovorichev:


zagovorichev:
@jbout pls, check realisation (as an example of the usage you could review this https://github.com/oat-sa/extension-tao-item/pull/305/files)

PR #1667, opened 2018-03-02
Fix/tao 3904/sprint71/fixed tag spaces

tikhanovichA:
The rawurlencode works with RFC 3986. And if I non mistaken urlencode uses RFC 1738. Because of this we have some problems with decode some symbols. For example urlencode encodes whitespace to +, and front part can't decode it.


test/helpers/UriHelperTest.php
```diff
@@ -0,0 +1,39 @@
...
+class UriHelperTest extends TaoPhpUnitTestRunner
+{
+
+    public function testUrlEncode()
+    {
+        $url = \tao_helpers_Uri::url('index', 'Main', 'tao', ['param' => 'test 1']);
+        $this->assertContains('test%201', $url);
+        $this->assertNotContains('+', $url);
+    }
+
```
krampstudio: I'm sorry but you should cover more cases by the test : 

```

public function uriDataProvider()
{
  return [
      ['index', 'Main', 'tao', ['param' => 'test 1'], '/tao/Main/index?param=test%201'],
      ['index', 'Main', 'tao', ['param' => 'test+1'], '/tao/Main/index?param=test+1'],
      ['index', 'Main', 'tao', ['param' => 'multiple space case '], '/tao/Main/index?param=multiple%20space%20case'],
      ['index', 'Main', 'tao', ['p1' => 't 1', 'p2' => 't 2'], '/tao/Main/index?p1=t%201&p2=t%202'],
//etc.
  ];
} 

public function testUrlEncode($action, $module, $extension, $params, $expected) 
{
    $url = \tao_helpers_Uri::url('index', 'Main', 'tao', ['param' => 'test 1']);
    $this->assertContains($exptected, $url);
    
} 
```
And at least test also an URI in parameter which is a common use case

PR #1675, opened 2018-03-08
Feature/tao 5720 oauth2

siwane:
!the purpose of this PR is to use a service to load rest session (to be configured by taoOAuth)
Doc to review: https://github.com/oat-sa/extension-tao-sync/pull/6


manifest.php
```diff
@@ -41,8 +41,8 @@
     'label' => 'Tao base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '17.8.1',
-    'author' => 'Open Assessment Technologies, CRP Henri Tudor',
+    'version' => '17.9.0',
+    'author' => 'Open Ass11essment Technologies, CRP Henri Tudor',
```
ionutpad: Typo

models/classes/session/restSessionFactory/RestSessionFactory.php
```diff
@@ -0,0 +1,77 @@
...
+     *
+     * @param $request
+     * @param $resolver
+     * @return bool
+     * @throws LoginFailedException
+     */
+    public function createSessionFromRequest($request, $resolver)
+    {
+        /** @var SessionBuilder $builder */
+        foreach ($this->getSessionBuilders() as $builder) {
```
ionutpad: Based on this means that the order we add builder to config matters, in case of `'oat\\taoOauth\\model\\bootstrap\\Oauth2SessionBuilder'` should be first

PR #1677, opened 2018-03-13
Added search by properties to the GenerisSearch

zagovorichev:


krampstudio:
@jbout could you please have a quick look

jbout:
1. Agreed
2+3: $queryString was originally intended to be search engine dependent and thus a complex user generated string. Building this string programmatically could become problematic, as it would require to define a clear syntax as interface.
If we already have a structured search query then it might be better to extend the interface of the search engine, so that each implementation can treat it 
4. Agreed

zagovorichev:
@krampstudio @jbout Added new method, check it, please. Also was changed ElasticSearch https://github.com/oat-sa/lib-tao-elasticsearch/pull/5

models/classes/resources/ListResourceLookup.php
```diff
@@ -50,8 +50,11 @@ public function getResources(\core_kernel_classes_Class $rootClass, array $selec
     {
         /** @var Search $searchService */
         $searchService = $this->getServiceLocator()->get(Search::SERVICE_ID);
+
         /** @var ResultSet $result */
-        $result = $searchService->query(current($propertyFilters), $rootClass, $offset, $limit);
+        $result = $searchService
+            ->addFiltersByProperties($propertyFilters)
+            ->query(current($propertyFilters), $rootClass, $offset, $limit);
```
krampstudio: Sorry but I still don't get the API. I think I get your intent to not break the existing method `query` but it still odd at the usage.

models/classes/search/strategy/GenerisSearch.php
```diff
@@ -127,4 +132,14 @@ public function supportCustomIndex()
...
         return false;
     }
+
+    /**
+     * (non-PHPdoc)
+     * @see \oat\tao\model\search\Search::addFiltersByProperties()
+     */
+    public function addFiltersByProperties($properties = [])
+    {
+        $this->properties = $properties;
```
krampstudio: So it means you set the properties once and all next search using the service instance will also use the same properties.

zagovorichev: yep, I've missed this, I could fix it if the idea is generally correct

zagovorichev:
@jbout @krampstudio New implementation is ready, check it, please

models/classes/resources/ListResourceLookup.php
```diff
@@ -78,6 +93,21 @@ public function getResources(\core_kernel_classes_Class $rootClass, array $selec
         ];
     }
 
+    private function getResourceData($resource)
```
krampstudio: Could you please comment this method

models/classes/resources/ListResourceLookup.php
```diff
@@ -48,26 +49,40 @@ class ListResourceLookup extends ConfigurableService implements ResourceLookup
      */
     public function getResources(\core_kernel_classes_Class $rootClass, array $selectedUris = [], array $propertyFilters = [], $offset = 0, $limit = 30)
     {
-        /** @var Search $searchService */
-        $searchService = $this->getServiceLocator()->get(Search::SERVICE_ID);
-        /** @var ResultSet $result */
-        $result = $searchService->query(current($propertyFilters), $rootClass, $offset, $limit);
-        $count = $result->getTotalCount();
+        if (count($propertyFilters) == 1 && isset($propertyFilters[OntologyRdfs::RDFS_LABEL])) {
```
krampstudio: Please add a one line comment explaining that we use the indexed search when only a label is set and the generis search instance when properties are given

zagovorichev:
@krampstudio Update

krampstudio:
This looks ok for me.  I'm waiting @jbout  approval to test and merge.

zagovorichev:
@jbout Update

PR #1685, opened 2018-03-21
Tao 5940 reduce nr of indexed resources

hutnikau:
depends on https://github.com/oat-sa/generis/pull/485


models/classes/search/index/IndexResourceIterator.php
```diff
@@ -0,0 +1,74 @@
...
+use oat\search\helper\SupportedOperatorHelper;
+use core_kernel_classes_Resource;
+use core_kernel_classes_Class;
+
+/**
+ * Class IndexResourceIterator
+ * @package oat\tao\model\search\index
+ * @author Aleh Hutnikau, <hutnikau@1pt.com>
+ */
+class IndexResourceIterator extends \core_kernel_classes_ResourceIterator implements ServiceLocatorAwareInterface
```
jbout: This resource iterator is not specific to indexing and could be used in any situation where we need to iterate over a large subset of data, so I would give it a more generic name

models/classes/search/index/IndexService.php
```diff
@@ -42,21 +42,35 @@ class IndexService extends ConfigurableService
...
+            $iterator = $this->getResourceIterator($this->getLastIndexTime(), $time);
+        } else {
+            $iterator = $this->getResourceIterator();
+        }
         $indexIterator = new IndexIterator($iterator);
         $indexIterator->setServiceLocator($this->getServiceLocator());
         $searchService = $this->getServiceLocator()->get(Search::SERVICE_ID);
-        return $searchService->index($indexIterator);
+        if ($sinceLast) {
+            $this->updateLastIndexTime($time);
```
jbout: The time should only be updated after a successful run, to prevent some resources from never being updated in case of an error

models/classes/search/index/IndexService.php
```diff
@@ -42,21 +42,35 @@ class IndexService extends ConfigurableService
     const SERVICE_ID = 'tao/IndexService';
     const INDEX_MAP_PROPERTY_DEFAULT = 'default';
     const INDEX_MAP_PROPERTY_FUZZY = 'fuzzy';
+    const OPTION_LASTRUN_STORE = 'lastrun_store';
+    const OPTION_INDEX_SINCE_LAST_RUN = 'index_since_last_run';
```
jbout: Would it not be easier to extend the IndexService with a new service (IndexSinceLastRunService is a bit long, but precise) that includes the additional functionality of allowing partial updates instead of adding a switch to turn it on and off in the same implementation?

hutnikau:
@jbout I've made changes according to your comments.

PR #1686, opened 2018-03-21
Added custom rule for processing mime types for javascript files load…

Alroniks:
…ed dynamically


models/classes/websource/BaseWebsource.php
```diff
@@ -124,6 +124,11 @@ public function getMimetype($filePath)
         if (isset($pathParts['extension'])) {
             //manage bugs in finfo
             switch ($pathParts['extension']) {
+                case 'js':
+                    if ($mimeType === 'text/plain' || $mimeType === 'text/x-asm') {
+                        return $mimeType = 'text/javascript';
```
IlyaYarkavetsOnePoint: Maybe it is better to remove variable setting as it is not needed?

Alroniks:
@uncleempty Updated

PR #1693, opened 2018-03-27
parse response to get error message

hutnikau:


views/js/test/util/httpErrorParser/test.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8">
+        <title>Test - URL Util</title>
```
krampstudio: The title is not correct

views/js/util/httpErrorParser.js
```diff
@@ -0,0 +1,66 @@
...
+            var error;
+            try {
+                json = JSON.parse(xhr.responseText);
+                msg = json.message ? json.message : errorThrown;
+            } catch(e) {
+                msg = errorThrown;
+            }
+            error = new Error(msg);
+            error.response = xhr;
+            error.errorThrown  = errorThrown;
```
krampstudio: could you please also add the HTTP status code. This could be useful in some situations.

PR #1694, opened 2018-04-02
Fix/tao core unit tests

hutnikau:


models/classes/search/strategy/GenerisSearch.php
```diff
@@ -106,7 +106,11 @@ private function getTotalCount( $queryString, $rootClass = null )
     public function index($document = [])
     {
         // nothing to do
-        return true;
+        $i = 0;
+        foreach ($document as $resuource) {
```
zagovorichev: something weird here, doesn't it?

PR #1698, opened 2018-04-04
Allow to disable datatable columns

hutnikau:
In order to be able to disable/enable user data language we need to be able to disable/enable `Data Language` column on users table.

### DIFF
--- a/views/js/test/ui/datatable/test.js
+++ b/views/js/test/ui/datatable/test.js
@@ -1185,7 +1185,7 @@ define([
         });
     });
 
-    QUnit.asyncTest('Disabled columns', function(assert){
+    QUnit.asyncTest('Hidden columns', function(assert){
         var $container = $('#container-1');
 
         QUnit.expect(5);
@@ -1209,12 +1209,12 @@ define([
                     id: 'login',
                     label: 'Login',
                     sortable: true,
-                    enabled: true
+                    visible: true
                 }, {
                     id: 'name',
                     label: 'Name',
                     sortable: true,
-                    enabled: false
+                    visible: false
                 }, {
                     id: 'email',
                     label: 'Email',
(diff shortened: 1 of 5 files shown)


views/js/test/ui/datatable/test.js
```diff
@@ -1184,4 +1184,41 @@ define([
...
+                'model' : [{
+                    id: 'login',
+                    label: 'Login',
+                    sortable: true,
+                    enabled: true
+                }, {
+                    id: 'name',
+                    label: 'Name',
+                    sortable: true,
+                    enabled: false
```
krampstudio: When you want to disable en entry from the model, shouldn't you just remove the entry from the model ? 
You _should_ be able to use the following `model` with the same effect : 
```js
         datatable({
                url : 'js/test/ui/datatable/data.json',
                'model' : [{
                    id: 'login',
                    label: 'Login',
                    sortable: true
                }, {
                    id: 'email',
                    label: 'Email',
                    sortable: false
                }]
            });
```

No ?

hutnikau: Use a property is easier and more readable. You do not need to add conditions like:
```
model = [{
    id: 'login',
    label: 'Login',
    sortable: true
}, {
    id: 'email',
    label: 'Email',
    sortable: false
}];

if (someColumnEnabled) {
    model.push({
        id: 'somecol',
        label: 'Some Column',
        sortable: false
    });
}
```

And even worse if we need to disable/enable several columns.

Also I have added ability to use callback function to check if column enabled or not. It will allow to update columns dynamically. For example we may use row of checkboxes to choose which columns should be visible for wide tables.

krampstudio: Ok, why not.
Could you please just adjust the vocabulary : 
`disabled` / `enabled`make me expect the column will be there but grayed out. 
It could be `visible : false` or `display : false`

hutnikau: Just pushed the commit with option renamed to `visible`

PR #1700, opened 2018-04-05
Tao 6077 disable user data lang

hutnikau:
depends on:
https://github.com/oat-sa/generis/pull/494
https://github.com/oat-sa/tao-core/pull/1698


hutnikau:
opened https://github.com/oat-sa/tao-core/pull/1701 instead

PR #1709, opened 2018-04-17
Fix/avoid oauth controller

siwane:


ionutpad:
Please add some unit tests for `RestSessionFactory`

PR #1738, opened 2018-05-16
Fix/tao 6176/add more specific import report

ionutpad:


models/classes/import/service/ImportMapperInterface.php
```diff
@@ -61,4 +61,9 @@ public function getProperties();
      * @return mixed
      */
     public function getOption($option);
+
+    /**
+     * @return \common_report_Report
```
SergiiTao: it also could be `null`

PR #1740, opened 2018-05-17
Feature/tao tao 6273 ckeditor tab untab plugins

ssipasseuth:
the icons has been added in https://github.com/oat-sa/extension-tao-devtools/pull/130
Integrate the new ckeditor plugins for tab and un-tab features that comes from https://github.com/oat-sa/ckeditor-dev/pull/11

To test this PR, just test an extended text interaction with the `html` format.

**Important info** - there is one bug remaining:
When the cursor is at the beginning of an empty element (empty body, p or div), the tab function does not work. The fix will come in another PR.


ssipasseuth:
You are right, even if the bug is here, we should at least ensure that the error is handled gracefully.
I made the fix here https://github.com/oat-sa/ckeditor-dev/pull/12 and integrated in this branch in my last commits. Please check again.

PR #1750, opened 2018-06-04
Feature/tao 5701 config util

jsconan:

Add a small helper that ensure a config object is properly set and at least filled with default values.


views/js/test/util/config/test.html
```diff
@@ -0,0 +1,34 @@
...
+        <script  type="text/javascript">
+
+             //don't start the test now
+            QUnit.config.autostart = false;
+
+            //load the config
+            require(['/tao/ClientConfig/config'], function(){
+
+                //load the test
+                require(['test/util//config/test'], function(){
```
krampstudio: There's a `//` in the module name

views/js/util/config.js
```diff
@@ -0,0 +1,63 @@
...
+    return {
+        /**
+         * Builds a config object, ensure default values are set.
+         * @param {Object} config
+         * @param {Object} [defaults]
+         * @returns {Object}
+         * @throws Error if a required entry is missing
+         */
+        build: function build(config, defaults) {
+            config = _.assign({}, config);
```
krampstudio: `_.defaults` already creates a clone, you might not need to call  `_.assign` to prevent iterating the `config` object if not necessary. `return _.defaults(config || {}, defaults)` should be sufficient.

PR #1754, opened 2018-06-11
Feature/tao 5868 relocate task queue

ssipasseuth:
Copied Sync Queue related components from `taoTaskQueue` to tao-core.

Testing would be ideal with a fresh install and updating an instance where `taoTaskQueue` is already installed. Good scenario can be to save a task with the old code first, update the code base and run the saved task with the new code.

For both test, you will also have to checkout the other extensions where the task queue related modifications are made.


jbout:
I'm not sure a meaning full review of a pull request this big (10k lines, 119 files) is even possible.
But it appears to me it contains some components which are backoffice specific, can you move those to https://github.com/oat-sa/extension-tao-backoffice ?

PR #1758, opened 2018-06-15
Feature/tao 6401/import export events refactored

gyszucs:


actions/class.Import.php
```diff
@@ -19,98 +19,97 @@
...
+    /**
+     * Returns the currently selecte importhandler
+     * or the importhandler to use by default
+     *
+     * @return tao_models_classes_import_ImportHandler
+     */
+    protected function getCurrentImporter()
+    {
+        if ($this->hasRequestParameter('importHandler')) {
+            //$importHandlerClass = $this->getRequestParameter('importHandler');
```
SergiiTao: Could you please remove the comment?

actions/class.Import.php
```diff
@@ -19,98 +19,97 @@
...
+            }
+        }
 
-	/**
-	 * Helper to get the selected class, needs to be passed as hidden field in the form
-	 */
-	protected function getCurrentClass() {
-		return new core_kernel_classes_Class(tao_helpers_Uri::decode($this->getRequestParameter('classUri')));
-	}
+        return current($this->getAvailableImportHandlers());
```
SergiiTao: imho [0] is more explicit.

actions/class.Import.php
```diff
@@ -19,98 +19,97 @@
...
+    }
 
-	protected function getValidators(){
-		return array();
-	}
+    /**
+     * Gets the available import handlers for this module
+     * Should be overwritten by extensions that want to provide
+     * additional Importhandlers
+     */
```
SergiiTao: @return tao_models_classes_import_ImportHandler[]

actions/class.Import.php
```diff
@@ -19,98 +19,97 @@
...
+    protected function getAvailableImportHandlers()
+    {
+        return [
```
SergiiTao: on every call we will have new instances. do we need to cache this return?

actions/class.Import.php
jbout: I would advise against, in case it's not a numeric array. Even current() has a risk, if someone has iterated over the array before, so I often use https://secure.php.net/manual/en/function.reset.php

PR #1763, opened 2018-06-19
Feature/tao 6323/media editor

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-6323

`mediaDimensionComponent` - shows resizer
http://tao.loc/tao/views/js/test/ui/mediaEditor/plugins/mediaDimension/component/test.html
`mediaDimension/helper` - calculate media sizes
http://tao.loc/tao/views/js/test/ui/mediaEditor/plugins/mediaDimension/helper/test.html
`mediaEditorComponent` - initialize mediaDimensionComponent with media
http://tao.loc/tao/views/js/test/ui/mediaEditor/mediaEditorComponent/test.html


krampstudio:
On the overall structure, that's ok. 
I'll give you a few tricks before the actual review, mostly there's a lot of stuffs already managed by the component model.


## mediaEditorComponent
 - Each `ui/component` is a GUI component. I advise you to always use a template and to follow the lifecycle.
 
 ```
 <div class="media-editor"> <div>
 ```
 
 - The configuration override is already builtin
 - It's an habit, but we are used to preinitlalize and render the component within the factory
 - ensure the media is loaded  outside the editor 
 
```
    var defaultConfig = {
        $media: null,
        tools: {
            mediaDimension: {
                active: false
            }
        }
    };
    
    return function mediaEditorFactory($container, media, config) {

        /**
         * Current component
         */
        var mediaEditorComponent = component({
            // all public methods goes here for example
            getMedia : function getMedia(){
                return media:
            }
         
        }, defaultConfig)
            .setTemplate(mediaEditorTpl)
            .on('init', function () {
                //on init ensure the config is correct
                //before the template is applied rendering
                
                 //if you need to load the media first, then render once loaded
                this.render($container)
             })
             .on('render', function(){ 
                var $element = this.getElement();
                 var config = this.getConfig();
                 
                  if(config.tool.dimension){
                       this.dimensionComponent = mediaDimensionComponentFactory($element, media)
                          .on('change', function(){
                            // check if change shouldn't be spread too so we can listen for change of the mediaEditor
                         });
                         
                         //you can spread some events from the sub component to the parent component
                         this.dimensionComponent.spread('error', this);
                   }
    
            });
         
         //always defer to be able to listen for the `init` event
         _.defer(function(){
              mediaEditorComponent.init(config);
         }):
        return mediaEditorComponent;
    };

```

## mediaHelper

 - all async code needs to be handled by a Promise
 - please check but I'm sure there is already a tool to load a media 
 - I'm wondering if the editor won't give you already the informations about the media (not only an image : width, height, type, source and node)
 - even with the load it looks like a POJO/a data object, 
 
 ## mediaDimensionComponent
 
 Same remarks than the mediaEditor : 
   - use the builtin template, config and lifecycle
   
   ```
    var defaultConfig = {
          responsive: true,
        showSync: true,
        showReset: true,
        sizeProps: {},
        denyCustomRatio: false,
        syncDimensions: true,
        width: 0,
        height: 0,
        minWidth: 0,
        maxWidth: 0,
        precision: 5

    };
    
    return function mediaDimensionFactory($container, media, config) {

        /**
         * Current component
         */
        var mediaDimensionComponent = component({
            // all public methods goes here 
            toggleMode : function toggleMode(){
                  if(this.is('rendered')){
                      if(this.is('responsive')){
                             //set relative mode
                             
                             this.setState('responsive', false);
                      } else {
                            this.setState('responsive', true);
                      }
                      
                      //always trigger through the component not the DOM when
                      //it's about custom events
                      this.trigger('modechange', this.is('responsive'));
                  }
                  return this;
            }
        }, defaultConfig)
            .setTemplate(mediaDimensionpl)
            .on('init', function () {
                //on init ensure the config is correct
                //before the template is applied rendering
                
                var _config = this.getConfig();
                
                 //you can also change the component state (this will change the css class accordingly)

                     this.setState('synced', _config.syncDimensions);
                     //it will add the CSS class "synced" on the main element, you can then check the component
                     //state by this.is('synced') 
                
                     this.setState('synced', _config.responsive);
                
                 //if you need to load the media first, then render once loaded
                this.render($container)
             })
             .on('render', function(){ 
                var self = this;
                var $element = this.getElement();
                
                  //get inner DOM elements, scoped
                  this.controls = {
                       $responsiveToggleField = $('.media-mode-switch', $element ),
                       //...
                  };
                  
                  //and initialize the bindings
                  this.controls.responsiveToggleField.on('click', function () {
                         self.toggleMode();
                  })
                

                  
     
            });
         
         //always defer to be able to listen for the `init` event
         _.defer(function(){
              mediaDimensionComponent.init(config);
         }):
        return mediaDimensionComponent;
    };

```

File: views/js/ui/mediaEditor/plugins/mediaDimension/mediaDimensionComponent.js [hunk 1]
```diff
@@ -0,0 +1,492 @@
...
+         */
+        var $blocks, $slider, $fields;
+
+        /**
+         * Current component
+         */
+        var mediaDimensionComponent = component({
+            update: function update() {
+                // slide sliders
+                $slider.val(_config.sizeProps['%'].current.width);
```
krampstudio: You should access the configuration through `this.getConfig`

views/js/ui/mediaEditor/plugins/mediaDimension/mediaDimensionComponent.js
```diff
@@ -0,0 +1,492 @@
...
+        precision: 5
+    };
+
+    /**
+     * Creates mediaDimension component
+     * @param config
+     * @fires "changed" - on State changed
+     * return {component|*}
+     */
+    return function mediaDimensionFactory($container, media, config) {
```
krampstudio: As a good practice, I suggest you : 
- `defaultConfig` on the top
- `initialConfig` in the factory
- `var config = this.getConfig()` within the methods

views/js/ui/mediaEditor/plugins/mediaDimension/mediaDimensionComponent.js
```diff
@@ -0,0 +1,492 @@
...
+                _blocks[unit].prop('unit', unit);
+                _blocks[unit].find('input').data('unit', unit).after($('<span>', {
+                    'class': 'unit-indicator',
+                    text: unit
+                }));
+            });
+
+            $responsiveToggleField.on('click', function () {
+                _checkMode();
+                $elt.trigger('responsiveswitch', [$responsiveToggleField.is(':checked')]);
```
krampstudio: What the goal of this event ?

views/js/ui/mediaEditor/plugins/mediaDimension/mediaDimensionComponent.js
```diff
@@ -0,0 +1,492 @@
...
+                this.render($container);
+            })
+            .on('render', function () {
+                var $tpl, $mediaSizer;
+
+                $tpl = $(tpl({
+                    responsive: (typeof _config.responsive !== 'undefined') ? _config.responsive : true
+                }));
+
+                $tpl.appendTo(this.getContainer());
```
krampstudio: Wihy don't you use the component's template instead ? 

```
mediaDimensionComponent
   .setTemplate(tpl)
   .on('init', function() { ...
```

views/js/ui/mediaEditor/plugins/mediaDimension/mediaDimensionComponent.js
```diff
@@ -0,0 +1,492 @@
...
+            var $btn = $elt.find('.media-sizer-reset');
+
+            if(!_config.showReset) {
+                $elt.find('.media-sizer').addClass('media-sizer-reset-off');
+            }
+
+            // this stays intact even if hidden in case it will be
+            // displayed from somewhere else
+            $btn.on('click', function() {
+                _config.sizeProps = _.cloneDeep(_config.originalSizeProps);
```
krampstudio: It's better to make a proper API method for this kind of use case : 
```js
var mediaDimensionComponent = component({

    reset : function reset(){
         var config = this.getConfig();
         if(this.is('rendered')){

              //update the config
              config.sizeProps = _.cloneDeep(config.originalSizeProps);
              this.setConfig(config);

              //apply changes
              calculateCurrentSizes();
 
              //notify the world
              this.trigger('reset');
         }
         return this;
    },
});
//...
$btn.on('click', function(e) {
    e.preventDefault();
    mediaDimensionComponent.reset();
});
```

IMO you can apply this pattern widely on the component, the goal is to separate input from side effects.

views/js/ui/mediaEditor/mediaEditorComponent.js
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * target - jQuery element with media $()
+     * container - container to which an target is attached
+     *
+     * @type mediaEditorConfig
+     * @private
+     */
+    var _defaults = {
+        mediaDimension: {
+            $container: null,
```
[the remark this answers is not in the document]
zagovorichev: It does not exist yet, so I had no idea. But now I think that I have to do it

views/scss/inc/_mediasizer.scss
```diff
@@ -1,138 +1,136 @@
...
-                right: 69px;
-            }
-        }
-
-        &.media-sizer-responsivetoggle-off {
-            .media-sizer-responsive-toggle {
-                display:none;
-            }
-        }
+.media-sizer {
```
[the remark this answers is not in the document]
zagovorichev: All colors will be hardcoded, is it OK for you?

views/js/ui/mediaEditor/plugins/mediaDimension/mediaDimensionComponent.js
zagovorichev: I can't find `.setConfig` in the `ui/component`

views/js/ui/mediaEditor/mediaEditorComponent.js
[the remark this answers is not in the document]
zagovorichev: but what if it needs to be changed by the content? I don't think that we can do it in the item editor

views/js/ui/mediaEditor/plugins/mediaDimension/helper.js
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Round a decimal value to n digits
+     *
+     * @param {number} value
+     * @param {number} precision
+     * @returns {number}
+     * @private
+     */
+    var _round = function _round(value, precision) {
```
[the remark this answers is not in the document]
zagovorichev: it's really weird method gamp.round, I need to use Math.round in some places to make it worked.

zagovorichev:
@krampstudio @ssipasseuth Update

zagovorichev:
1. The goal of the `getContainerWidth` is to calculate the initial values. (On the initialization or on the `reset` action)
If I change container width then I press reset - media sizes will be changed.

views/js/test/ui/mediaEditor/mediaEditorComponent/test.html
```diff
@@ -45,6 +45,10 @@
             height: initial;
             max-width: initial;
         }
+
+        .media-sizer .unit-indicator {
+            margin: -20px;
+        }
```
ssipasseuth: why is it fixed in the unit test and not in the component scss ?

zagovorichev:
@ssipasseuth Update

PR #1769, opened 2018-07-10
Fix/tao 6416/item theme not applied when set through delivery edit

uncleempty:


jbout (changes requested):
Breaks all multiple dependencies

PR #1775, opened 2018-07-12
Fix/tao 6416/access namespace in themes config script

uncleempty:


views/js/ui/themes.js
```diff
@@ -86,6 +86,30 @@ define([
...
+
+        /**
+         * Set active namespace for themes
+         *
+         * Explicitly sets activeNamespace property into config.
+         *
+         * @param {String} ns - activeNamespace value to be set into config
+         */
+        setActiveNamespace: function setActiveNamespace(ns){
+            var config = module.config();
```
krampstudio: The config (from `module.config()`) is not intended to be written, instead you'll have to work on a config clone : 
 ```js
var themeConfig = _.deepClone(module.config());

return {
   /// ...
    setActiveNamespace: function setActiveNamespace(namespace){
        themeConfig.activeNamespace = namespace;
    }
};
``` 

and of course use the clone everywhere

File: manifest.php [hunk 2]
```diff
@@ -45,7 +45,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '19.13.1',
+    'version' => '19.13.2',
```
krampstudio: This a feature version not a fix version since you've added methods to a public API

views/js/ui/themes.js
```diff
@@ -28,12 +28,31 @@ define([
...
+        /**
+         * Gets module config clone. Checks if there are differences between actual module config and a clone. If needed -
+         * clones it again.
+         *
+         * @returns {Object}
+         */
+        getConfig : function getConfig() {
+            var moduleConfig = module.config();
+
+            if (!_.isEqual(themesConfig, moduleConfig)) {
```
krampstudio: Why do you need to clone it again ?  `module.config` doesn't change but if you set  the `activeTheme` this will be erased.
Why don't you do a something like : 

```js
var themeConfig;
return {
    getConfig : function getConfig(){
       var initialConfig;
       if(!themeConfig){
           initialConfig = module.config();
           themeConfig = _.cloneDeep(initialConfig);
      }
      return themeConfig;
    }
};
```

krampstudio: The issue in the unit test comes from way the config is "mocked".

PR #1777, opened 2018-07-16
Fix/tao 6416/expose api

uncleempty:


views/js/ui/themes.js
```diff
@@ -52,6 +53,10 @@ define([
             return themesConfig;
         },
 
+        getDefaultNamespacePrefix : function getDefaultNamespacePrefix() {
```
krampstudio: Missing jsdoc

views/js/ui/themes.js
```diff
@@ -126,6 +131,21 @@ define([
...
             this.getConfig().activeNamespace = ns;
+        },
+
+        /**
+         * Gets the current theme data from config
+         *
+         * @param what
+         * @returns {*}
+         */
+        getCurrentThemeData : function getCurrentThemeData(what){
```
krampstudio: Wrong jsdoc, it should be : 
```js
  * @param {String} [what] - themes are classified, what is the theme for ?
  * @returns {Object} the current theme data
```

krampstudio (changes requested):
Could you please cover the new methods through the  unit test.

PR #1800, opened 2018-08-22
Tao 6676 tasl queue checkup

hutnikau:


actions/class.TaskQueue.php
```diff
@@ -82,6 +85,36 @@ public function get()
         }
     }
 
+    /**
+     * @throws \common_exception_NotImplemented
+     */
+    public function getAll()
```
gyszucs: This action will give back the tasks only in the new task queue. Is that okay?

PR #1802, opened 2018-08-27
Removed unnecessary includes of raw_start, fixed TestCase parents, de…

alexeyvy:
According to https://oat-sa.atlassian.net/browse/TAO-6469


SashaTAO:
There are a lot of broken tests and errors in tests, but they all will be fixed in https://oat-sa.atlassian.net/browse/TAO-6912 which is based on this branch.
Approved.

PR #1811, opened 2018-09-12
Feature/tao 6786/prevent installation when tao environment not in debug mode

SashaTAO:


install/api.php
```diff
@@ -330,6 +335,11 @@
     header("Content-Type:text; charset=UTF-8");
     echo $e->getMessage();
 }
+catch (tao_install_api_NotAllowedAPICallException $e) {
+    header('HTTP/1.0 405 Method Not Allowed');
```
no-chris: Is `405` really the error code that we should get here?

```
The HyperText Transfer Protocol (HTTP) 405 Method Not Allowed response status code indicates that the request method is known by the server but is not supported by the target resource.

The server MUST generate an Allow header field in a 405 response containing a list of the target resource's currently supported methods.
```
https://tools.ietf.org/html/rfc7231#section-6.5.5

Here:
- the method is known by the server AND supported by the resource, only the context (production mode) makes it unavailable
- the Allow header with the supported method is missing (and does not really make sense in our case)

install/api.php
```diff
@@ -281,6 +281,12 @@
     	}
     	else{
 	    	switch ($input['type']){
+                case 'Install':
+                    if (tao_install_utils_System::isTAOInstalled() && !tao_install_utils_System::isTAOInDebugMode()) {
+                        $errorMessage = 'System is already installed in production mode and can not be reinstalled.';
+                        throw new tao_install_api_NotAllowedAPICallException($errorMessage);
+                    }
```
no-chris: missing `break`

install/api.php
```diff
@@ -330,6 +335,11 @@
     header("Content-Type:text; charset=UTF-8");
     echo $e->getMessage();
 }
+catch (tao_install_api_NotAllowedAPICallException $e) {
```
no-chris: I wonder if this is now an appropriate exception and error messages.
If the goal is to increase security, then I would not give the reason for the rejection as this give more information to an attacker on what is blocking him. A plain 403 forbidden with a generic error message should be enough (or even a 404 as the HTTP spec suggest in that case)

SashaTAO: The idea was to provide more specific error message since it's an API call. But for security reasons in this case it makes sense to return general error message. I'll fix it.

install/api/class.NotAllowedAPICallException.php
```diff
@@ -0,0 +1,27 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ * 
+ * Copyright (c) 2002-2008 (original work) Public Research Centre Henri Tudor & University of Luxembourg (under the project TAO & TAO2);
```
siwane: Could u update the legal mention please? Only need to specify 2018, because of file legal mention not project

install/api/class.NotAllowedAPICallException.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
siwane: It is possible to add a namespace here?

install/init.php
```diff
@@ -29,6 +29,10 @@
 
 require_once ($root . 'vendor/autoload.php');
 
+if (tao_install_utils_System::isTAOInstalled()) {
```
siwane: Why it's needed? If it is installed, do we need to load configs before new install?

install/services/class.Service.php
```diff
@@ -126,7 +126,7 @@ protected static function getRequestMethod(){
      * in their implementation.
      * @return tao_install_services_Data The resulting Data.
      */
-    protected abstract function execute();
+    public abstract function execute();
```
siwane: why?

install/api.php
```diff
@@ -229,6 +229,20 @@
...
+/**
+ * @param string $rawInput
+ * @param string $serviceName
+ * @return tao_install_services_Service
+ * @throws ReflectionException
+ */
+function getService($rawInput, $serviceName)
+{
+    $data = new tao_install_services_Data($rawInput);
+    $class = new ReflectionClass('tao_install_services_' . $serviceName . 'Service');
```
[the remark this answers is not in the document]
SashaTAO: Fixed

install/init.php
SashaTAO: We need to load generis.conf.php to check if it's installed in debug mode.

install/services/class.Service.php
SashaTAO: This method is public in all child classes and it's used outside of the class in api.php file to execute instantiated service.

composer.json
```diff
@@ -59,7 +59,8 @@
 			"oat\\tao\\model\\" : "models/classes/",
 			"oat\\tao\\helpers\\" : "helpers",
 			"oat\\tao\\test\\" : "test",
-			"oat\\tao\\scripts\\" : "scripts"
+			"oat\\tao\\scripts\\" : "scripts",
+			"oat\\tao\\install\\" : "install"
```
siwane: Does it mean that I need to add a line foreach folder I want to namespacify?

SashaTAO: It looks like yes, because of this line "oat\\tao\\model\\" : "models/classes/"

jbout: Alternatively, the long term goal is to move the classes in /models/classes/... to /model/ and then we can unify the autoloaded

siwane:
@no-chris , approved

no-chris:
I now get the following error message:

```
br/>
<b>Fatal
    error</b>:  Uncaught Error: Class 'oat\tao\install\api\NotAllowedAPICallException' not found in /www/tao/tao/install/api.php:301
Stack trace:
#0 {main}
thrown in <b>/www/tao/tao/install/api.php</b> on line <b>301</b><br/>
```

SashaTAO:
You have this error because we converted new exception class to PSR-4 standard with namespaces and it's not registered in your composer yet. Could you please try to run following commands:
composer clear-cache
composer dump-autoload

no-chris:
I get the correct error message but a 500 error code, is this normal?

```
POST http://tao.local/tao/install/api.php

HTTP/1.0 500 Internal Server Error
Date: Mon, 17 Sep 2018 17:33:54 GMT
Server: Apache/2.4.10 (Debian)
X-Powered-By: PHP/7.0.30
Set-Cookie: PHPSESSID=f15e458c23a59a4fd72005684b57a04e; path=/
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate
Pragma: no-cache
Access-Control-Allow-Origin: *
Content-Length: 48
Connection: close
Content-Type: text; charset=UTF-8

Fatal error: The requested service is forbidden.

Response code: 500 (Internal Server Error); Time: 93ms; Content length: 48 bytes
```

SashaTAO:
Sorry, I forgot to import namespace in **catch** block. Now it works as expected.

no-chris:
Working good 👍 

Merging now

PR #1813, opened 2018-09-14
Protecting endpoints from ZIP bombs

alexeyvy:
According to https://oat-sa.atlassian.net/browse/TAO-6770


helpers/class.File.php
```diff
@@ -513,6 +513,36 @@ public static function renameInZip(ZipArchive $zipArchive, $oldname, $newname)
...
+    public static function checkWhetherArchiveIsBomb(\ZipArchive $archive)
+    {
+        if (!$archive->filename) {
+            throw new common_Exception('ZIP archive should be opened before checking for a ZIP bomb');
+        }
+
+        $contentSize = 0;
+        for ($fileIndex = 0; $fileIndex < $archive->numFiles; $fileIndex++) {
+            $stats = $archive->statIndex($fileIndex);
+            $contentSize += $stats['size'];
```
gyszucs: Does this work for a nested zip as well?

helpers/class.File.php
```diff
@@ -513,6 +513,36 @@ public static function renameInZip(ZipArchive $zipArchive, $oldname, $newname)
...
+        }
+
+        $archiveFileSize = filesize($archive->filename);
+
+        $minCompressionRatioToBeBomb = 200;
```
gyszucs: What is this ratio? Why is it exactly 200?

alexeyvy: Ratio is the `the archive size / archive content size`, why 200 - that's too much for a regular archive and good enough to be an attack attempt

helpers/class.File.php
alexeyvy: It doesn't, and we don't need it to work for a nested zip, because we don't open archives recursively

helpers/class.File.php
gyszucs: Is this based on some industrial standard or just some personal experience? Anyway, this value seems a config one for me, it would be nice if we could configure this from outside somehow so in the future we can change it easily if we want. Hard-coded values should be avoided as much as possible.

alexeyvy: it's personal experience; maybe you could suggest some existing config which is appropriate for containing this value?

gyszucs: Check for example https://github.com/oat-sa/tao-core/blob/master/actions/class.Main.php#L136. That is how you can reach a config value living in an extension. You will have to create a default config file like this: https://github.com/oat-sa/tao-core/blob/master/config/default/login.conf.php

gyszucs: Sorry, I just noticed this helper method a static one so probably you won't be able to get the extension manager. So only put this ratio value as an argument of the method with default value so every client has the change to set different value.

alexeyvy:
I've added the tests

helpers/class.File.php
alexeyvy: ok, done

PR #1819, opened 2018-09-26
Update Dutch language

peetya:


manifest.php
```diff
@@ -45,7 +45,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '20.0.0',
+    'version' => '20.0.1',
```
siwane: As a devops, do I want to update translations at fix updates?

PR #1821, opened 2018-10-02
Fix/tao 6420 focus colors change

rkarolis:
Dependancies:

https://github.com/oat-sa/extension-tao-testqti/pull/1368
https://github.com/oat-sa/extension-tao-testqti-previewer/pull/12
https://github.com/oat-sa/extension-tao-itemqti/pull/1189


views/scss/inc/_forms.scss
```diff
@@ -3,10 +3,8 @@ $autofillFocusBorderColor: $colorWheel-01;
 $autofillBgColor: rgb(250, 255, 189); // = Chrome default setting
 
 @mixin focus-style {
-    border: 1px solid $info !important;
-    &:-webkit-autofill {
-        border-color: $autofillFocusBorderColor !important;
-    }
+    box-shadow: 0 0 9px 3px $info !important;
+    transition: box-shadow 200ms ease-out;
```
no-chris: What was the purpose of the `&:-webkit-autofill` rule and why was it removed ?

views/scss/inc/_colors.scss
```diff
@@ -15,6 +15,7 @@ $lightBlueGrey: rgb(228, 236, 239);
 $brownRedGrey: rgb(154, 137, 123);
 $darkBrown: rgb(111, 99, 89);
 $websiteBorder: rgb(141, 148, 158);
+$itemHover: rgb(232,239,245);
```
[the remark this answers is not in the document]
rkarolis: Luc, suggests using #e8eff5 for focus on items and #f9f8f7 for item background.

no-chris:
I would like our CSS expert @dietertaotesting 's opinion on the need to have a new color variable here.

views/scss/inc/_forms.scss
dietertaotesting: I understand this has been re-added (see line 9ff)

views/scss/inc/_colors.scss
```diff
@@ -12,6 +12,7 @@ $grey: rgb(173, 161, 148);
 $darkBlueGrey: rgb(164, 187, 197);
 $mediumBlueGrey: rgb(193, 212, 220);
 $lightBlueGrey: rgb(228, 236, 239);
+$lightGrey: rgb(232,239,245);
```
dietertaotesting: I am not against this, but there is not a huge difference between the two shades and they both build not much of a contrast to white which is usualy in the background. $lightBlueGrey vs. $lightGrey. I haven't tested the colors in a real life example but I doubt that they will work on a crappy screen - I might be mistaken about that though. If the color comes from the UX team ask them if they would be OK with `$lightBlueGrey` instead.

views/scss/inc/_forms.scss
```diff
@@ -3,9 +3,10 @@ $autofillFocusBorderColor: $colorWheel-01;
 $autofillBgColor: rgb(250, 255, 189); // = Chrome default setting
 
 @mixin focus-style {
-    border: 1px solid $info !important;
+    box-shadow: 0 0 9px 3px $info !important;
```
dietertaotesting: The mixin must be in /tao/views/scss/inc/_functions.scss. Make sure you use the grunt compiler, not you local SASS

rkarolis:
Color requested by UX

dietertaotesting:
If the color has been requested by UX I have no objection

views/scss/inc/_colors.scss
```diff
@@ -72,3 +73,10 @@ $colorWheel-09: #BA122B;
 $colorWheel-10: #C34713;
 $colorWheel-11: #C36F13;
 $colorWheel-12: #C39413;
+
+// Harmonize focus style between browsers
+$autofillFocusBorderColor: $colorWheel-01;
+$autofillBgColor: rgb(250, 255, 189); // = Chrome default setting
+
+$selectableHover: $lightGrey;
+$activeHover: $lightBlueGrey;
```
no-chris: whiten() is definitely a better option here than having 2 colors, especially since lightGrey is not grey (looking at the color code all channel values are distinct).

dietertaotesting: FYI UX has insisted on that specific color, I have approved of this solution

PR #1824, opened 2018-10-04
Feature/tao 6956 add move all to rdf controller

siwane:


krampstudio:
I don't get the structure of this Pull request.
Why @jsconan is reviewer of change he has made ? 
If this is the result of an integration branch; separated PR should have been done against the integration branch.

jsconan:
@krampstudio my bad, I directly commited my changes in the @siwane's PR.
Here is the PR for frontend only: https://github.com/oat-sa/tao-core/pull/1833

actions/class.RdfController.php
```diff
@@ -586,6 +586,71 @@ public function moveInstance()
...
+
+            $this->validateWritePrivileges($ids);
+            $this->validateMoveRequest();
+
+            $ids = $this->getRequestParameter('ids');
+            $response = $this->moveAllInstances($ids);
+            $this->returnJson($response);
+        } catch (\InvalidArgumentException $e) {
+            $response = [
+                'success' => false,
```
SergiiTao: code duplication. 
could you create a method to generate this array?

actions/class.RdfController.php
```diff
@@ -586,6 +586,71 @@ public function moveInstance()
...
+
+            $ids = $this->getRequestParameter('ids');
+            if (empty($ids)) {
+                throw new InvalidArgumentException('No instances specified.');
+            }
+
+            $this->validateWritePrivileges($ids);
+            $this->validateMoveRequest();
+
+            $ids = $this->getRequestParameter('ids');
```
SergiiTao: code duplication

actions/class.RdfController.php
```diff
@@ -824,6 +889,141 @@ public function validateCsrf()
...
+     * @throws common_exception_Error|InvalidArgumentException
+     */
+    protected function validateWritePrivileges(array $uris)
+    {
+        $privileges = [];
+        foreach ($uris as $uri) {
+            $privileges[$uri] = 'WRITE';
+        }
+        if (!DataAccessControl::hasPrivileges(common_session_SessionManager::getSession()->getUser(), $privileges)) {
+            throw new InvalidArgumentException('Resources to move have restricted access.');
```
SergiiTao: wrong permissions is not the same as invalid argument :)

actions/class.RdfController.php
```diff
@@ -824,6 +889,141 @@ public function validateCsrf()
...
+     *     "http://resource2",
+     *     "http://class1",
+     *     "http://class2"
+     *   ]
+     * }
+     * @requiresRight destinationClassUri WRITE
+     *
+     * @throws common_exception_Error
+     */
+    protected function moveAllInstances(array $ids)
```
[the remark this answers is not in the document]
SergiiTao: agree ;)

siwane:
@krampstudio , I had the same error, it was coming from ACL, the array $ids is evaluated directly in pregmatch, try to clear ur cache

krampstudio:
@siwane I still have the error
@jsconan I've foynd another issue : when you change your selection, the destination component disappear (it looks it is intentional but it's very annoying to have to select the the move action again)

models/classes/accessControl/data/DataAccessControl.php
```diff
@@ -42,14 +42,17 @@ public function hasAccess(User $user, $controller, $action, $parameters) {
         try {
             foreach (ControllerHelper::getRequiredRights($controller, $action) as $paramName => $privileges) {
                 if (isset($parameters[$paramName])) {
-                    if (preg_match('/^[a-z]*_2_/', $parameters[$paramName]) != 0) {
-                        common_Logger::w('url encoded parameter detected for '.$paramName);
-                        $cleanName = \tao_helpers_Uri::decode($parameters[$paramName]);
+                    if (is_array($parameters[$paramName])) {
```
krampstudio: Which Test covers this use case ?

actions/class.RdfController.php
```diff
@@ -824,6 +876,189 @@ public function validateCsrf()
...
+
+    /**
+     * Get movable instances from the list of instances
+     *
+     * @param array $classes
+     * @param array $instances
+     *
+     * @return array
+     */
+    private function getMovableInstances(array $classes = [], array $instances = [])
```
antoinerobin: Here we have no restriction on the root class, so I can move my root class into a subclass => it breaks everything

actions/class.RdfController.php
```diff
@@ -824,6 +876,189 @@ public function validateCsrf()
...
+    {
+        $movableInstances = [];
+
+        // Check if a class belong to class to move
+        /** @var core_kernel_classes_Resource|core_kernel_classes_Class $instance */
+        foreach ($instances as $instance) {
+            $isValid = true;
+            foreach ($classes as $class) {
+                if ($instance instanceof core_kernel_classes_Class) {
+                    if ($class->getUri() != $instance->getUri() && $instance->isSubClassOf($class)) {
```
antoinerobin: some comments here would be welcome to understand why we don't have to move subclasses

actions/class.RdfController.php
```diff
@@ -824,6 +876,189 @@ public function validateCsrf()
...
+                        $statuses[$instance->getUri()] = [
+                            'success' => false,
+                            'message' => sprintf('Instance "%s" cannot be moved to class to move "%s"', $instance->getUri(), $class->getUri()),
+                        ];
+                        $isValid = false;
+                        break;
+                    }
+                } else {
+                    if ($instance->isInstanceOf($class)) {
```
antoinerobin: same here to understand that these instances will automatically be moved by moving the class above

actions/class.RdfController.php
antoinerobin: maybe a more explicit name like getInstancesToMove or something like that would be better and more explicit

actions/class.RdfController.php
```diff
@@ -824,6 +876,189 @@ public function validateCsrf()
...
+        // Check if a class belong to class to move
+        /** @var core_kernel_classes_Resource|core_kernel_classes_Class $instance */
+        foreach ($instances as $instance) {
+            $isValid = true;
+            foreach ($classes as $class) {
+                if ($instance instanceof core_kernel_classes_Class) {
+                    if ($class->getUri() != $instance->getUri() && $instance->isSubClassOf($class)) {
+                        $statuses[$instance->getUri()] = [
+                            'success' => false,
+                            'message' => sprintf('Instance "%s" cannot be moved to class to move "%s"', $instance->getUri(), $class->getUri()),
```
antoinerobin: we could maybe have clearer messages here and in the instance status also + these statuses are not used, do we really need them ?

actions/class.RdfController.php
```diff
@@ -824,6 +876,189 @@ public function validateCsrf()
...
+    private function getInstancesList(array $ids)
+    {
+        $statuses = $instances = $classes = [];
+
+        foreach ($ids as $key => $instance) {
+            $instance = new core_kernel_classes_Resource($instance);
+            if ($instance->isClass()) {
+                $instance = new core_kernel_classes_Class($instance);
+                $classes[] = $instance;
+            } else {
```
antoinerobin: you could put a elseif directly here, it might help code climate

models/classes/accessControl/data/DataAccessControl.php
krampstudio: I'm agree with @antoinerobin the changes on the ACL needs to be tested since they impact all controllers.

PR #1825, opened 2018-10-04
Fix/tao 6924 brs permissions new node

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-6924

Fix permissions issues on the BRS:
- after a node add, the permissions were not refreshed, so no actions were available on it
- the computation of access denied was not properly done

To test, you will need to activate the `resourceSelector`. In the config file `config/tao/client_lib_config_registry.conf.php`, simply add:
```
        'layout/tree/loader' => array(
            'treeProvider' => 'resource-selector'
        ),
```


views/js/layout/tree/provider/resourceSelector.js
```diff
@@ -72,133 +72,134 @@ define([
...
+                                        self.selectDefaultNode(defaultNode);
+                                    });
+                                    actionManager.on('removeNode', function(actionContext, node){
+                                        self.removeNode(node);
+                                        self.selectDefaultNode(defaultNode);
+                                    });
+                                    actionManager.on('subClass instanciate duplicateNode', function(actionContext, node){
+                                        self.changeSelectionMode('single');
+                                        self.addNode(node, node.classUri);
+                                        self.query({ classUri : node.classUri });
```
jsconan: Sorry for the apparently huge change. This is because I fixed the indent in the file to get rid of my IDE complaining about that.

This is the line I actually changed to fix the lack of permissions on newly added node.

ssipasseuth:
oops, still a problem:

jsconan:
Good catch! I  was not sure it is part of the same story, so ended up by putting that aside. However, I managed to fix it.

ssipasseuth:
It works !

PR #1827, opened 2018-10-04
Tao 6860/clean out all the file statements linked to non existent files

boajer:
https://oat-sa.atlassian.net/browse/TAO-6860

Cleans additionally to specified in story
1) Files that no more stored as per https://github.com/oat-sa/extension-tao-item/pull/327


boajer:
closed as moved to generis https://github.com/oat-sa/generis/pull/538

PR #1830, opened 2018-10-09
Fix / TAO-6994 Cover cases like .5 float input; 2 tests for that

oatymart:
The bug arose because we are using our own locale-specific `parseFloat` function in `util/locale.js`. 

There was an edge case which was missed: splitting a string which begins with the decimal separator -> the part before the separator will end up empty. The first part will now become 0 in this case.

### DIFF
--- a/views/js/util/locale.js
+++ b/views/js/util/locale.js
@@ -77,15 +77,18 @@ define(['module', 'moment'], function (module, moment) {
          * @returns {Number}
          */
         parseFloat: function (numStr) {
+            var thousandsSeparator = this.getThousandsSeparator(),
+                decimalSeparator = this.getDecimalSeparator();
+
             // discard all thousand separators:
-            if (this.getThousandsSeparator().length) {
-                numStr = numStr.replace(new RegExp('\\' + this.getThousandsSeparator(), 'g'), '');
+            if (thousandsSeparator.length) {
+                numStr = numStr.replace(new RegExp('\\' + thousandsSeparator, 'g'), '');
             }
         
             // standardise the decimal separator as '.':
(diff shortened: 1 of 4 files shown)


oatymart:
This rewrite of `locale.parseFloat` handles negative float inputs as well as omission of the leading zero.

Please note that the default locale for TAO is set in `config/tao/client_lib_config_registry.conf.php` as
```
    'config' => array(
        'util/locale' => array(
            'decimalSeparator' => '.',
            'thousandsSeparator' => '',
            'dateTimeFormat' => 'DD/MM/YYYY HH:mm:ss'
        ),
```
In other words we are currently set up to use a decimal point and no thousands separator.

If a user or client chooses to modify this configuration, the new code will handle whichever separators they specify here.

views/js/util/locale.js
```diff
@@ -73,22 +73,23 @@ define(['module', 'moment'], function (module, moment) {
...
             if (this.getThousandsSeparator().length) {
-                ones = parts[0].replace(new RegExp('\\' + this.getThousandsSeparator(), 'g'), '');
+                numStr = numStr.replace(new RegExp('\\' + this.getThousandsSeparator(), 'g'), '');
             }
-
-            return parseFloat(ones) + parseFloat('0.' + parts[1]);
+        
+            // standardise the decimal separator as '.':
+            if (this.getDecimalSeparator() != '.') {
+                numStr = numStr.replace(new RegExp('\\' + '.', 'g'), '_')
```
dietertaotesting: type comparison required `this.getDecimalSeparator() !== '.'`

oatymart: Changed.

PR #1831, opened 2018-10-09
Act 239 improve frontend log

hutnikau:
By call `JSON.stringify(message.err, ["message", "arguments", "type", "name", "stack"]); ` instead of `JSON.stringify(message.err);` we may get more information about the error (message, name and stack trace):
image


hutnikau:
Not needed anymore

PR #1833, opened 2018-10-11
Feature/tao 6924 move to

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-6924

Implement the `moveTo` action

### DIFF
--- a/views/js/layout/actions/common.js
+++ b/views/js/layout/actions/common.js
@@ -42,6 +42,15 @@ define([
     };
 
     /**
+     * Cleans up the main panel and creates a container
+     * @returns {jQuery}
+     */
+    var emptyPanel = function cleanupPanel() {
+        section.current().updateContentBlock('<div class="main-container flex-container-form-main"></div>');
+        return $(section.selected.panel).find('.main-container');
+    };
+
+    /**
      * Register common actions.
      *
      * TODO this common actions may be re-structured, split in different files or moved in a more obvious location.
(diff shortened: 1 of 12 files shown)


views/js/layout/actions/common.js
```diff
@@ -523,6 +533,130 @@ define([
                 .on('error', reject);
             });
         });
+
+        /**
+         * Register the moveAll action: select a destination class to move resources
```
krampstudio: The action is called `moveTo`

views/js/layout/actions/common.js
```diff
@@ -523,6 +533,130 @@ define([
...
+         *
+         * @this the action (once register it is bound to an action object)
+         *
+         * @param {Object|Object[]} actionContext - multiple action contexts
+         * @returns {Promise<String>} with the new resource URI
```
krampstudio: It resolve with the destination class URI

krampstudio:
There's also 2 points, but to confirm with PM : 
 -  in the previous move implementation (one resource) you were warned about a potential lost of metadata only when the properties of the destination did not match the properties from the copies (there was a method server side to compute the property diff)
 - the warning message may need to be reviewed. I find it understandable because I know the underlying concepts, but for an end user, this message can be confusing. I don't know if we can find a more concise version.

jsconan:
Indeed @krampstudio, the warning message was not so good. It is now changed to something clearer and more concise.

For the backend diff/confirm feature, it would be a little tricky to deal with an huge selection of resources to move. After having discussed that with PM, we agreed to keep the generic message approach.

PR #1835, opened 2018-10-11
Fix/sass compile on linux x64

uncleempty:
For: https://oat-sa.atlassian.net/browse/TAO-7052.
On Linux 64bits, install TAO.
Install last versions of node.js and npm.
Got to `tao/views/build` and run `npm install`

node-sass is installed you can try to run `npx grunt taosass` to compile SASS to CSS.

### DIFF
--- a/views/build/Gruntfile.js
+++ b/views/build/Gruntfile.js
@@ -59,6 +59,17 @@ module.exports = function(grunt) {
      // Load local tasks.
     grunt.loadTasks('tasks');
 
+    //instantiate sass module
+    const sass = require('node-sass');
+    //set initial config for sass to make 3.x.x branches of grunt-sass work, see https://github.com/nodejs/nan/issues/504#issuecomment-385296082, https://github.com/sourcey/spectacle/issues/156#issuecomment-401731543
+    grunt.initConfig({
+        sass: {
+                options: {
+                           implementation: sass
+                }
+        }
+    });
+
     /*
      * Load separated configs into each extension
      */
--- a/views/build/package.json
+++ b/views/build/package.json
@@ -31,7 +31,7 @@
     "grunt-notify": "^0.4.3",
     "grunt-qunit-junit": "^0.3.0",
     "grunt-replace": "^0.11.0",
-    "grunt-sass": "^2.0.0",
+    "grunt-sass": "^3.0.0",
     "load-grunt-tasks": "^3.4.0",
     "lodash": "^4.13.1",
     "time-grunt": "^1.0.0",
--- a/views/build/package.json
+++ b/views/build/package.json
@@ -34,6 +34,7 @@
     "grunt-sass": "^2.0.0",
     "load-grunt-tasks": "^3.4.0",
     "lodash": "^4.13.1",
-    "time-grunt": "^1.0.0"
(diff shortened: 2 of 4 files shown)


views/build/Gruntfile.js
```diff
@@ -59,6 +59,17 @@ module.exports = function(grunt) {
      // Load local tasks.
     grunt.loadTasks('tasks');
 
+    //instantiate sass module
+    const sass = require('node-sass');
+    //set initial config for sass to make 3.x.x branches of grunt-sass work, see https://github.com/nodejs/nan/issues/504#issuecomment-385296082, https://github.com/sourcey/spectacle/issues/156#issuecomment-401731543
+    grunt.initConfig({
+        sass: {
+                options: {
```
krampstudio: please define the options in the main sass config file : `tao/views/build/grunt/sass.js`  : https://github.com/oat-sa/tao-core/blob/develop/views/build/grunt/sass.js#L9

views/build/package.json
```diff
@@ -31,9 +31,10 @@
     "grunt-notify": "^0.4.3",
     "grunt-qunit-junit": "^0.3.0",
     "grunt-replace": "^0.11.0",
-    "grunt-sass": "^2.0.0",
+    "grunt-sass": "^3.0.0",
     "load-grunt-tasks": "^3.4.0",
     "lodash": "^4.13.1",
-    "time-grunt": "^1.0.0"
+    "time-grunt": "^1.0.0",
+    "node-sass": "^4.7.2"
```
krampstudio: I think the targeted version of node-sass is `4.9.3`

views/build/package.json
```diff
@@ -31,9 +31,10 @@
     "grunt-notify": "^0.4.3",
     "grunt-qunit-junit": "^0.3.0",
     "grunt-replace": "^0.11.0",
-    "grunt-sass": "^2.0.0",
+    "grunt-sass": "^3.0.0",
```
krampstudio: The minimum node version for this version is `8.0.0` 
I was about to update the require version in https://github.com/oat-sa/tao-core/blob/breaking/TAO-4814_atomic-bundles/views/build/package.json#L14
so you can do it here as well

views/build/package.json
uncleempty: Yes, it installs from this constraint, but as I saw in various comments on github it can be like that. But would change to `4.9.3`

views/build/package.json
uncleempty: Ok, would apply min version

krampstudio:
Also please, do not forget to add the ticket number in the branch name, ie. `fix/TAO-7052/sass-compile-on-linux-x64` or `fix/TAO-7052_sass-compile-on-linux-x64` to ease the build of release notes

uncleempty:
@krampstudio, already seen that forgotten the ticket number, just wanted you to test/review before closing this PR and moving to another branch, correctly named.

views/build/package.json
krampstudio: yes, the `^` means the most recent feature version ie `4.x.x`, so  you got version `4.9.3` in you lock file. It's just to be safe and ensure we start from the last.

PR #1842, opened 2018-10-16
Breaking/tao 4814 atomic bundles

krampstudio:
Migrate TAO to the atomic bundles process

Requires 
 - [x] https://github.com/oat-sa/grunt-tao-bundle/pull/1

 - use the new bundler library
 - use a unique configuration for the AMD build and test `views/build/config/requirejs.build.json` 
 - update the router to load bundles (and test it by the way)
 - update the AMD loader tag to always load the vendor module
 - update the tao build config

Using the PR you should be able to run build the bundles for the tao extension : 
```
cd views/build
npm i
npx grunt taobundle
```


jsconan:
all good! 👍

PR #1843, opened 2018-10-17
Feature/tao 7019 maths parser

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-7019

Implement a maths expression evaluator based on:
- a configurable parser: [expr-eval](https://github.com/silentmatt/expr-eval)
- a robust maths computation library: [decimal.js](https://github.com/MikeMcl/decimal.js)

The `expr-eval` parser allows to override the built-in operators and functions, so they are almost all mapped to those from `decimal.js`. The reason why is because `expr-eval` is vulnerable to the round-off error, while `decimal.js` is not (by design).
Also, `decimal.js` supports good precision and huge numbers.

The new util is a factory that packages the parser and returns a simple evaluator function.
It accepts an expression, and even allows it to be parametric, as the second argument is the list of mapped variables.

Some examples:
```javascript
var evaluate = mathsEvaluatorFactory();

// simple arithmetic
var result = evaluate("3*4+30"); // will return '42';

// advanced expression
var result = evaluate("(10! - 5!) * 4 * (18 / 4) + sqrt(56^4)"); // will return '65319376';

// parametric expression
var result = evaluate("2*a*x+b", {a:5, x:3, b:15}); // will return '45';
```

Other PR will follow to present the basic UI on top of this evaluator. For now, the unit test provides a minimalistic UI.

About the unit tests, while they check essential aspects, they do not cover any operators and functions yet. They will be augmented later.


jsconan:
cleaned useless and inappropriate `codeclimate` comments

PR #1844, opened 2018-10-17
Fix interface compatibility issue

gyszucs:
During `tao/scripts/taoSetup.php` using PHP 7.2 `Declaration of oat\tao\model\api\ApiClientConnector::request($method, $uri, array $options = Array) must be compatible with GuzzleHttp\ClientInterface::request($method, $uri = NULL, array $options = Array)` error occurred.


gyszucs:
Outdated

PR #1845, opened 2018-10-19
Added empty message for datatable when no data found

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-6429


Alroniks:
@ssipasseuth Please, have a look from UX perspective.

Alroniks:
I've added option emptyText to have the ability to configure datatable object when it is needed, but for old usages, all will stay as is, that's safe.

PR #1848, opened 2018-10-23
Added exception class for prevented path traversal attempts

alexeyvy:


models/classes/import/InvalidSourcePathException.php
```diff
@@ -0,0 +1,63 @@
...
+     * @var string
+     */
+    private $sourcePath;
+
+    /**
+     * @param string $basePath
+     * @param string $sourcePath
+     * @param Exception|null $previous
+     */
+    public function __construct($basePath, $sourcePath, Exception $previous = null)
```
SergiiTao: $sourcePath supposed to be a relative path or full path?

models/classes/import/InvalidSourcePathException.php
```diff
@@ -0,0 +1,63 @@
...
+     * Get the human-readable message for the end-user. It is supposed
+     * to be translated and does not contain any confidential information
+     * about the system and its sensitive data.
+     *
+     * @return string A human-readable message.
+     */
+    public function getUserMessage()
+    {
+        return __(
+            'Invalid path of a source "%s". Path must point to the existed file inside the package.',
```
SergiiTao: if `sourcePath` is full path we should not show it

alexeyvy: It's a relative path to a file. Full path is $basePath

PR #1853, opened 2018-10-26
tao.min is considered always loaded since the router belongs to it

krampstudio:
One of the unit test was failing after a late change made in the [router](https://github.com/oat-sa/tao-core/blob/develop/views/js/core/router.js#L115) : we consider the `tao.min.js` bundle always loaded in the routing. 
The test to check is `/tao/views/js/test/core/router/test.html`


krampstudio:
Let me fix the typo

PR #1854, opened 2018-10-26
Investigate and fix xss vulnerabilities in some libs shown on PEN testing reports

handleman:
All libs are updated to safest versions, except of jquery.qtip.js Seems that it is became abandoned now, Its site is broken and there is no newer compatible version in their repository. Fixing issue manually inside its files needs a lot of investigation, PEN testing environment setup and etc...

all other libs are updated to their xss-solved-versions:
jquery-ui-1.8.23 - > jquery-ui-1.9.2.
ckeditor.js- 4.4.7 TAO-2 - > ckeditor.js- 4.4.8 TAO-1
jquery-1.8.0 - > jquery-1.9.1

task with description TAO-6774


handleman:
I've reverted jquery back to 1.8 because of many dependencies across files, which definitely needs to be made separately and with accuracy.

handleman:
@krampstudio  all done

PR #1859, opened 2018-10-31
Feature/tao 6188 migrate qunit puppeter

krampstudio:
Replace PhantomJS by puppeteer.

 - I had to rewrite the bridge ti support our version of QUnit.
 - The timeout seems to not work so I had to simulate it in the bridge also
 - Some tests were using the user agent to detect headless. Some have been updated (Chrome headless doesn't contain third parts like some media playing codecs or pdf viewer).
 - Some test (pageStatus) are failing in Chrome, so I let them fail here too

All grunt commands work the same way.


views/build/config/chrome-bridge.js
```diff
@@ -0,0 +1,94 @@
...
+     * A test case get started
+     */
+    QUnit.testStart( ({ name }) =>  {
+
+        //start a timeout and
+        //keep it in the map under the test name
+        runningTestsTimeouts.set(name,
+            setTimeout( () => {
+                emit('fail.timeout', name);
+                runningTestsTimeouts.delete(name);
```
jsconan: Is this the only place where the timeout is cleared? How is it supposed to be after end of test?

krampstudio: Indeed, I've removed the `clearTimeout` during the refactoring...

krampstudio:
@jsconan the build platform is on the way. Please merge when you want.

PR #1870, opened 2018-11-12
fix ipad error for launching test taker

rkarolis:
The TAO-7140 is still not replicated. But by taking steps to replicate I found bug, that could not launch test taker. It's in `unmatrix.js` parser which asks for string or object, when we pass null from `transformer.js` it gets stuck, so we send empty string and it works.


rkarolis:
Used by @ricproenca-tao

PR #1871, opened 2018-11-13
Feature/tao 7050 dropdown component

oatymart:
Initial PR for code review, still preparing the work in taoInvalsi where it will be used.

To do:
~~Fix unit tests~~
~~Finalise styles~~
~~Updater & manifest~~


views/js/ui/dropdown.js
```diff
@@ -0,0 +1,240 @@
...
+     * @param {String} [config.activatedBy] - hover or click
+     * @param {String} [config.vDirection] - not implemented
+     * @param {String} [config.hAlign] - not implemented
+     * @returns {dropdown}
+     */
+    function dropdownFactory(config) {
+        return component(dropdownSpecs, defaults)
+        .setTemplate(dropdownTpl)
+        // dropdown-specific init:
+        .on('init', function() {
```
jsconan: this is where you can init the `open` state:

```
this.setState('open', this.config.isOpen);
```

If you want the state to be reflected as CSS class, you must put that in the `on('render'` instead.

views/js/ui/dropdown.js
```diff
@@ -0,0 +1,240 @@
...
+            }
+            $component
+            .on('focus', self.open)
+            .on('blur', self.close);
+
+            $toggler
+            .on('click', self.toggle)
+            .on('focus', self.open);
+        })
+        .init(config);
```
jsconan: Also, pay attention to the indent. I know this is part of personal preferences, but this should at least respect the linter.

views/js/ui/dropdown/tpl/dropdown.tpl
```diff
@@ -0,0 +1,29 @@
+<div class="dropdown-container">
+    <div class="dropdown {{#if cls}} {{cls}}{{/if}} {{#if isOpen}} open{{/if}}" {{#if id}}id="{{id}}"{{/if}} data-control="{{id}}" role="navigation">
```
jsconan: No need to test about inexistence of `cls`. Simply invoke the value. If it exists it will placed, else it will be ignored.
If you rely on the internal state management, the `open` class will be directly set.

views/js/ui/dropdown.js
```diff
@@ -0,0 +1,240 @@
...
+    'use strict';
+
+    /**
+     * Some default config
+     * @type {Object}
+     */
+    var defaults = {
+        isOpen: false,
+        headerItem: {text: 'Menu'},    // li component or just html
+        innerItems: [], // array of li components or html items
```
jsconan: Not sure the data should be part of the config.
Instead you should provide a way to set them at init, and also a bulk setter/getter.

views/js/ui/dropdown/tpl/dropdown.tpl
```diff
@@ -0,0 +1,29 @@
...
+            {{{headerItem.html}}}
+            {{else}}
+            {{#if headerItem.icon}}<span class="icon icon-{{headerItem.icon}}"></span>{{/if}}
+            <span class="text">{{headerItem.text}}</span>
+            {{/if}}
+            <span class="toggler clickable" tabindex="0"></span>
+        </div>
+        <ul class="dropdown-submenu plain" aria-label="submenu">
+            {{#each innerItems}}
+            {{! each list item can contain raw html, or structured data }}
```
jsconan: A good idea would be to have a separate template for a single item. This way it will ease the insert of new items. The main template should contain only the placeholders

views/scss/inc/_dropdown.scss
```diff
@@ -0,0 +1,61 @@
...
+            display: block;
+        }
+        a, .a {
+            cursor: pointer;
+            height: auto;
+            padding: 1.2rem 0.6rem;
+            text-align: left;
+        }
+        a:hover, a:focus, .a:hover, .a:focus {
+            background: #f3f7fa !important; // was: whiten($brand, 0.9) !important;
```
jsconan: Colors should come from the main scheme. Also they should be provided through variables

views/scss/inc/_dropdown.scss
```diff
@@ -0,0 +1,61 @@
...
+        .dropdown-header {
+            padding: 1.2rem 0 0.8rem;
+        }
+        .sep-before:before {
+            content: "";
+            display: block;
+            height: 1px;
+            width: 90%;
+            margin: 0.6rem auto;
+            background: #999;
```
jsconan: same about colors

oatymart:
I've updated the code style and changed some of the architecture based on your feedback @jsconan .
No longer calling render() manually, but using jQuery in the API methods to modify the DOM.
The tests now pass.
I haven't implemented the 2-level template at this point.

oatymart:
I added the 2nd template for the list item.

addItem() is now called with an object that must be prepared, but the template is still flexible enough to allow custom HTML within the mandatory <li>.

views/js/ui/dropdown.js
```diff
@@ -0,0 +1,222 @@
...
+            var self = this;
+            var $component = this.getElement();
+            // wire up main behaviour:
+            if (self.config.activatedBy === 'hover') {
+                $component
+                .on('mouseenter', self.open)
+                .on('mouseleave', self.close);
+            }
+            else if (self.config.activatedBy === 'click') {
+                $component.on('click', self.toggle);
```
jsconan: Caution, you are listening for click on the whole component. That means it will also be triggered when clicking on an item. That might result in events mess. You should rather target the exact control that should be clickable, I guess `this.controls.$toggler` or eventually `this.controls.$headerItem`.

oatymart: Made the event listener more specific.

oatymart:
I changed the way the data is stored in the component, and added a visual test.

views/js/ui/dropdown.js
```diff
@@ -0,0 +1,228 @@
...
+        return component(dropdownSpecs, defaults)
+        .setTemplate(dropdownTpl)
+        // dropdown-specific init:
+        .on('init', function() {
+            this.setState('open', this.config.isOpen);
+
+            this.data = {
+                headerItem: this.config.headerItem || '',
+                innerItems: this.config.innerItems || []
+            };
```
jsconan: Well, this was not what I have in mind. The data should be separated from config. And be managed as a private variable. Exposing it as a property allows to alterate it from the outside. The only way to deal with it should be through API: `setItems`, `addItem`, `setHeader`, `removeItem`, `clearItems`...
Roughly, something like that (obviously you should add some checks):
```javascript
dropdownFactory(config, data) {
    var api = {
        ...
        clearItems: function clearItems() {
            data.items = [];
            if (this.is('rendered')) {
                this.controls.$list.empty();
            }
            return this;
        },
        setItems: function setItems(items) {
            data.items = items;
            if (this.is('rendered')) {
                this.controls.$list.empty();
                _.forEach(items, function(item) {
                    this.controls.$list.append(itemTpl(item));
                });                
            }
            return this;
        },
        addItem: function addItem(item) {
            data.items.push(item);
            if (this.is('rendered')) {
                this.controls.$list.append(itemTpl(item));
            }
            return this;
        },
        removeItem: function removeItem(idx) {
            data.items.splice(idx, 1);
            this.setItems(data.items);
            return this;
        },
        ...
    };
    var dropdown = component(api, default)
        .on('render', function() {
            ....
            if (!_.empty(data)) {
                this.setItems(data.items);
                this.setHeader(data.header);
            }
            ...
        });
        .on(...);
    if (!_.empty(data)) {
        data = [];
    }
    _.defer(function() {
        dropdown.init(config);
    });
    return dropdown;
}
```

views/js/ui/dropdown.js
```diff
@@ -0,0 +1,259 @@
...
+                 */
+                self.trigger('item-click', id);
+                /**
+                 * @event item-click-<id>
+                 */
+                self.trigger('item-click-' + id);
+            });
+        })
+        .on('item-click', function() {
+            self.close();
```
jsconan: This is not working. The reason why is the `self` variable does not refer to `this` but to `window`... This is a drawback of the naming convention.

Since you are in the context of the component here, you should use `this.close();` instead.

views/js/ui/dropdown.js
```diff
@@ -0,0 +1,259 @@
...
+            else if (this.config.activatedBy === 'click') {
+                this.controls.$headerItem.on('click', self.toggle);
+            }
+            $component
+            .on('focus', self.open)
+            .on('blur', self.close);
+
+            // list item events
+            this.controls.$listContainer.on('click', 'li', function() {
+                var id = $(this).attr('id');
```
jsconan: caution, depending on the event context, `this` could refer to a sub element. You should rather use `$(this).closest('li').attr('id')`  or `$(e.target).attr('id')`;

PR #1880, opened 2018-11-21
breaking / tao 7150 tooltip

handleman:
**task is**
The library jquery.qtip is not maintained anymore and contains some vulnerabilities.
Either fork and fix or replace the library.
So it was decided to replace jquery.qtip with popper.js https://popper.js.org/

JIRA Card: https://oat-sa.atlassian.net/browse/TAO-7150

requirements: backward compatibility: remain all old lib API the same and cover with tests.

Related PRs:
https://github.com/oat-sa/extension-tao-itemqti/pull/1204
https://github.com/oat-sa/popper.js/pull/1
https://github.com/oat-sa/extension-tao-dac-simple/pull/78
https://github.com/oat-sa/extension-tao-testqti/pull/1408
https://github.com/oat-sa/extension-tao-itemqti-pic/pull/68
https://github.com/oat-sa/extension-tao-depp/pull/128
https://github.com/oat-sa/popper.js/pull/2


views/scss/inc/_tooltip.scss
```diff
@@ -1,194 +1,91 @@
...
-  background-color: #FFEF93;
-}
 
-.qtip-default .qtip-icon{
-  border-color: #CCC;
-  background: #F1F1F1;
-  color: #777;
+// qtip plugin legacy code
+// http://qtip2.com/
+.qtip-rounded,
```
krampstudio: is it necessary to keep the qtip CSS ?

views/js/test/ui/tooltip/test.js
```diff
@@ -21,53 +21,203 @@
...
+                assert.ok(!tooltipApi._isOpen, 'api visibility parameter is changed to "false"');
+                assert.equal($tooltip.css('visibility'), 'hidden', 'The tooltip become hidden when mouse cursor leave the reference item (button)');
+            }, 50);
+            QUnit.stop();
+        },100);
+    });
+
+    // former tooltip plugin compatibility API (jQuery.qtip plugin)
+    QUnit.asyncTest('Backward compatibility: Jquery.qtip("show") and Jquery.qtip("hide")', function (assert) {
+        $el.qtip('show');
```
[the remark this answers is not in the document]
handleman: It require stop/start when setTimeout use. 
Cant find proof on documentation because we use old version of Qunit. 
But in practice: when I delete .stop()/.start() on asyncTest() ( in this particular tests) it hangs forever

handleman:
@krampstudio  All your points are fixed except some where I answered here above. Please check. 
And visual playground added.

views/js/test/ui/tooltip/test.js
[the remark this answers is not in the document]
krampstudio: `QUnit.stop` should only be used if you wait multiple async paths, for example 2 unrelated `setTimeout`, but here since they are waiting each other, you can easily rely, on something like  : 
```
    QUnit.asyncTest('JQuery wrapper: physical user interactions (mouseEnter, MouseLeave)', function (assert) {
        var $el = $('<div id="tip-toggle"/>')
            .appendTo('#'+containerName)
            .qtip(defaultOpts);
        var tooltipApi = $el.qtip('api');

        QUnit.expect(4);

        $el[0].dispatchEvent(mouseenter);
        setTimeout(function() {
            var $tooltip = $('.tooltip');
            assert.ok(tooltipApi._isOpen, 'api visibility parameter is changed to "true"');
            assert.equal($tooltip.css('visibility'), 'visible', 'The tooltip become visible on when mouse cursor is in area of reference element (button)');

            $el[0].dispatchEvent( mouseleave );
            setTimeout(function () {

                assert.ok(!tooltipApi._isOpen, 'api visibility parameter is changed to "false"');
                assert.equal($tooltip.css('visibility'), 'hidden', 'The tooltip become hidden when mouse cursor leave the reference item (button)');

                QUnit.start();
            }, 50);
        },100);
    });
```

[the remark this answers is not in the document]
krampstudio: I think this helps in readability

views/js/ui/tooltip.js
```diff
@@ -16,45 +16,135 @@
...
     var themes = ['dark', 'default', 'info', 'warning', 'error', 'success', 'danger'],
         themesMap = {
-            'default' : 'qtip-rounded qtip-plain',
-            'dark' : 'qtip-rounded qtip-dark',
-            'error' : 'qtip-rounded qtip-red',
-            'success' :'qtip-rounded qtip-green',
-            'info' : 'qtip-rounded qtip-blue',
-            'warning' : 'qtip-rounded qtip-orange',
-            'danger' : 'qtip-rounded qtip-danger'
+            'default' : '<div class="tooltip qtip-rounded qtip-plain" role="tooltip"><div class="tooltip-arrow"></div><div class="tooltip-inner qtip-content"></div></div>',
```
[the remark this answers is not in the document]
krampstudio: You can move this into a tpl file, something like 
```
<div class="tooltip qtip-rounded qtip-{{theme}}" role="tooltip">
    <div class="tooltip-arrow"></div><div class="tooltip-inner qtip-content"></div>
</div>
```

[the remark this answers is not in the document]
krampstudio: By convention, we try to move every bit of HTML into templates.

views/js/ui/tooltip.js
```diff
@@ -16,45 +16,135 @@
...
-    var qtipConstructor = $.fn.qtip;
+    /**
+     * redefinition of jquery.qtip plugin http://qtip2.com/
+     * the goal is to substitude outdated lib code with new solution (https://popper.js.org/)
+     * so it is just a jquery rapper for Popper.js library.
+     * The Popper Instance will be stored inside .data('$popper') of wrapped element
+     * leaving its original interfaces that are widely used through project.
+     * https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/tooltip-documentation.md
+     */
+    var qtip = function (command, message, messageData) {
```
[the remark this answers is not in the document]
krampstudio: I think this is part of the library replacement. Hopefully it is used directly in a very few locations. 
Merging it in this state will create a technical debt, on if we built more on the tooltip API it will become more difficult to fix.

handleman:
@krampstudio  I've made all that were discussed:
Removed all qtip related references,  both from .js files and css styles.

So now it is ui/tooltip AMD module that use popper.js functionality in its native way, instead of being wrapped by jquery and imitating jquery.qtip methods.

And all your comments to this PR were worked out too.

Could you do another round of review?

manifest.php
```diff
@@ -45,7 +45,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '22.5.0',
+    'version' => '22.6.0',
```
krampstudio: Oh and this is goinf to be a breaking change, no ?

handleman:
@krampstudio 

Pending changes:

1) Made a split in tooltip API so now it has `tooltip.lookup()` and `tooltip.instance()` methods as you requested.
  1.1) Here I moved `.data('$tooltip')` attribute out of `tooltip.instance()`  and put it at exact places through the code where there is need in such signature. So now it is not a part of `ui/tooltip` component, but part of the exact places where it was made in such way, for example:
 `taoQtiItem/views/js/qtiCommonRenderer/renderers/Tooltip.js`
 `taoQtiItem/views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js` and in some other place ( see related PR's)

  1.2) Elimination completely `.data('$tooltip')` from all the code from all the extensions it is used seems useless for me, because tooltips are invoked asynchronously in some places and fixes could led to deep complicated refactoring of such things as our inner eventifier system and "Component" factory call-backs. For example, invalid form fields on qtiItem, invokes tooltips after render or after form validation occurs. It seems too complicated for this particular task, on which more than month is spent already.

2) Some other extensions became affected Please look at "Related PR's" section in description.

3)  Version of tao-core extension was changed accordingly your request.

4) All other conversations, that I marked as 'resolved' Should be fixed too.

views/js/ui/tooltip.js
```diff
@@ -13,67 +13,97 @@
...
+                        _.merge(opt, { title: $content[0] }, defaultOptions);
+                    }
+                    setTooltip(this, new Tooltip(this, opt));
+                });
+            } else {
+                throw new TypeError("Tooltip should be connected to DOM Element");
+            }
+
+        },
+        instance: function (el, settings){
```
krampstudio: - the method name `instance` is too generic in my opinion. What about `create` ? 
 - by convention we use `options` instead of `settings`
 - I'd also extract the message from the `settings/options` 
 - add alias for each themes for a user friendly API 

We would have : 
```js

   /**
    *  Create a tootlp
    * @param {jQueryElement|HtmlElement}  element -  the element to attach the tooltip to
    * @param {String}  message - the tooltip message
    * @param {Object} [options]
    * @param {String} [options.theme = default] - the tooltip theme
    * @param {String} [options.trigger] - ...
    * @returns {Tooltip} the tooltip instance
    */
create : function create(element, message,  options){

},
error : function error(element, message, options){
  return this.create(element, message, _.default({ theme : 'error'}, options)): 
},
success : function success(element, message, options){
  return this.create(element, message, _.default({ theme : 'success'}, options)): 
},

```

As a consumer of the API this is something I expect : 
```js
var dangerFeedback = tooltip.danger( myElement, 'You have one minute left before the test session expires');
```

views/js/ui/tooltip.js
```diff
@@ -13,67 +13,97 @@
...
+            if(!el && !(el instanceof Element || el instanceof HTMLDocument || el.jquery)){
```
krampstudio: Since this is used twice could you please factorize it

views/js/ui/tooltip.js
```diff
@@ -13,67 +13,97 @@
...
+                throw new Error("Tooltip should be connected to DOM Element");
+            }
+            var themeName = _.contains(themes, settings.theme) ? settings.theme : 'default';
```
krampstudio: Hum, ESLint should have reported a warning about hoisting, we enforce [var-on-top](https://eslint.org/docs/rules/vars-on-top#require-variable-declarations-to-be-at-the-top-of-their-scope-vars-on-top) on ES5 code.

views/js/ui/tristateCheckboxGroup.js
```diff
@@ -173,21 +173,27 @@ define([
...
                             $icon = $input.siblings('.icon')
                                 .addClass('cross')
-                                .qtip({
-                                    theme : 'warning',
-                                    content : {
-                                        text : self.config.maxMessage
-                                    }
-                                }).qtip('show');
+                                .each(function( ) {
+                                    $(this).data('$tooltip', Tooltip(this, {
```
krampstudio: `ui/tooltip` doesn't return a `Tooltip` prototype! 
I also don't think it's necessary to link this instance to the DOM, does it ?
Shouldn't it be something like : 
```js
self.tooltip = tooltip.warn( $icon, self.config.maxMessage )
```
and later 
```js
$icon.parent('label').on('mouseleave', function (){
    if(self.tooltip){
        self.tooltip.dispose();
       self.tooltip = null;
   }
});
```

handleman:
@krampstudio  have done all that you requested. Would you make another round of review on latest fixes?

views/js/test/ui/tooltip/test.js
```diff
@@ -22,52 +22,280 @@ define([
...
 
-        $el.trigger('mouseover');
+            setTimeout(function () {
+                var $hidden = $('.tooltip-blue');
+                var styles = window.getComputedStyle($hidden[0]);
+                QUnit.start();
+                assert.equal(styles.visibility, 'hidden', 'tooltip became hidden on mouse out');
+            },);
+            QUnit.stop();
+        },);
```
krampstudio: I would either have the parameter at `0`, `1` or do not add a trailing comma in function call

views/js/test/ui/tooltip/test.js
```diff
@@ -22,52 +22,280 @@ define([
...
+            $single[0].dispatchEvent(mouseleave);
 
-        $el.trigger('mouseover');
+            setTimeout(function () {
+                var $hidden = $('.tooltip-blue');
+                var styles = window.getComputedStyle($hidden[0]);
+                QUnit.start();
+                assert.equal(styles.visibility, 'hidden', 'tooltip became hidden on mouse out');
+            },);
+            QUnit.stop();
```
krampstudio: Instead of the `stop/start` pattern we are used to : 
1. define the function as async using `QUnit.asyncTest('tooltip.lookup()', function (assert) {`
2. always add the number of expected assert using `Qunit.assert` 
3. That way you can call `QUnit.start()` only when the test is `done` 

Your test will look like : 
```js
QUnit.asynTest('tooltip.lookup()', function (assert) {
       var $container = $('#' + containerName);
       var $single =  $( '[data-tooltip]', $container).first();

       QUnit.expect(2);

        tooltip.lookup($cointainer);
        $single[0].dispatchEvent(mouseenter);

        setTimeout(function () {
            var $tooltip = $('.tooltip-blue', $container);
            assert.equal($tooltip.length, 1, 'tooltip binds to DOM');
            $single[0].dispatchEvent(mouseleave);
            setTimeout(function () {
                var $hidden = $('.tooltip-blue', $container);
                var styles = window.getComputedStyle($hidden[0]);
                assert.equal(styles.visibility, 'hidden', 'tooltip became hidden on mouse out');
                QUnit.start();
            }, 0);
       }, 0);
});
```

views/js/test/ui/tooltip/test.js
```diff
@@ -22,52 +22,280 @@ define([
...
+        // $single.trigger('mouseenter');
+        setTimeout(function () {
+            var $tooltip = $('.tooltip-blue');
+            QUnit.start();
+            assert.equal($tooltip.length, 1, 'tooltip binds to DOM');
+            $single[0].dispatchEvent(mouseleave);
 
-        $el.trigger('mouseover');
+            setTimeout(function () {
+                var $hidden = $('.tooltip-blue');
```
krampstudio: Every selection in a test needs to be scoped `$('.tooltip-blue', $container)`

views/js/test/ui/tooltip/test.js
krampstudio: Obviously this applies to every async test case

views/js/loader/tao.min.js
```diff
@@ -46,46 +46,116 @@ link.href=url,document.getElementsByTagName("head")[0].appendChild(link)},_url:f
 ;var helper,options,buffer="",escapeExpression=this.escapeExpression,helperMissing=helpers.helperMissing;return buffer+='<div class="undobox feedback-info">\n    <span class="icon-info"></span>\n        ',buffer+=escapeExpression((helper=helpers.undoMessage)?helper.call(depth0,{hash:{},data:data}):"function"==typeof(helper=depth0&&depth0.undoMessage)?helper.call(depth0,{hash:{},data:data}):helper)+'\n        <a class="undo" href="#">'+escapeExpression((options={hash:{},data:data},(helper=helpers.__||depth0&&depth0.__)?helper.call(depth0,"Undo",options):helperMissing.call(depth0,"__","Undo",options)))+'</a>\n    <span title="'+escapeExpression((options={
```
krampstudio: Could you please remove compiled files from the PR, we create the bundles only during the release process

views/js/ui/formValidator/highlighters/tooltip.js
```diff
@@ -73,12 +61,17 @@ define([
...
             },
-
+            /**
+             * remove tooltip with error message from given field
+             * @param $field
+             */
             destroy : function destroy($field) {
-                if ($field.data('hasqtip') !== undefined) {
-                    $field.qtip('destroy', true);
+                if ($field.data('hasqtip')) {
```
krampstudio: this will be wrong

handleman:
problem with tooltip margin solved.

Also please note the https://github.com/oat-sa/extension-tao-depp/pull/128

image-2

views/scss/inc/_tooltip.scss
```diff
@@ -19,7 +19,9 @@ $dark : #404040;
 }
 
 .popper,
-.tooltip {
+.tooltip,
+.qti-item .tooltip
```
krampstudio: based on the dependency model the tao extension is not aware of items nor qti, please specify style linked to the qti item rendering at https://github.com/oat-sa/extension-tao-itemqti/tree/master/views/scss/qti

handleman:
@krampstudio  Tooltip padding when inside qti Item fixed

handleman:
The latest problem that was found when using tooltip with MathJax interaction

image-3

Unfortunately, while working wit this problem, found issue affected forked version of popper.js,
So you need to check  https://github.com/oat-sa/popper.js/pull/2 too.

PR #1881, opened 2018-11-21
Force numeric values in json output

Alroniks:


krampstudio:
What's the context of this PR. 
I think can have some side effects, no ?

Alroniks:
@krampstudio Probably yes, side effect possible. Context is that mysql driver does not support proper passing types of values and always returns data as string values. It means that if we have 0 as an integer value in resulted JSON we will get `value: "0"` instead of `value: 0`. Postgres based envs (and almost all prods) already works with real types but on SP we have some small annoying issues.

krampstudio:
I'm wondering if we need to solve database issues at the level of the HTTP layer. My main concern is on the client side we often check the type of the value, for example : `if (typeof response.code === 'string')` but got a number instead. Thanks to type coercion this should work most of the time but this can create unexpected bugs.

Alroniks:
But postgres already returns int as int. So if such bugs have existed it should be fixed already I guess. So this PR make json the similar for all cases. Fix the database issue is not possible (easily) because it's even not a PHP bug (btw it exists on bugs.php.net) but of the driver that used for connecting mysql and php.

krampstudio:
You're right but the concern isn't really for the part of the response that comes directly from the database (I think there's a lot of stored number moving all the way to the HTTP response) but more on manually forged values in all controllers, everything that looks like a number will be casted to a number, so if someone did something like `'code' => '12'` it will be parsed to `"code" : 12'` or what if we enter a dynamic value that looks like a number in a string, `"label" => "12"` will be served as `"label" : 12`

PR #1886, opened 2018-11-23
Feature/tao 6968 gdpr message

oatymart:
https://oat-sa.atlassian.net/browse/TAO-6968

Allows a GDPR message to be shown on the login page, only if the config option has been turned on in a customer extension.


actions/class.Main.php
```diff
@@ -279,6 +279,8 @@ public function login()
             $this->setData('msg', $this->getRequestParameter('msg'));
         }
 
+        $this->setData('show_gdpr', $config['show_gdpr'] || false);
```
jsconan: Caution, this won't work in PHP. You should use instead:
```php
$this->setData('show_gdpr', !empty($config['show_gdpr']));
```

PR #1894, opened 2018-11-29
Feature/tao 7188/create sync dashboard

SashaTAO:


models/classes/taskQueue/TaskLogInterface.php
```diff
@@ -134,10 +135,11 @@ public function updateParent($parentTaskId);
     public function search(TaskLogFilter $filter);
 
     /**
-     * @param TaskLogFilter $filter
+     * @param TaskLogFilter             $filter
+     * @param DatatableRequestInterface $request
      * @return DataTablePayload
      */
-    public function getDataTablePayload(TaskLogFilter $filter);
+    public function getDataTablePayload(TaskLogFilter $filter, DatatableRequestInterface $request);
```
ionutpad: Because this change of interface you will have to bump a major version.

PR #1901, opened 2018-12-03
digest support for frontend added

boajer:
https://oat-sa.atlassian.net/browse/TAO-7186


krampstudio:
If I could review this PR ( I can't since I've written the original code) I'd suggest to also take the unit tests from https://github.com/oat-sa/tao-mobile-app/tree/master/src/js/test/core/digest and place the polyfill under `views/js/lib/polyfill` but  can we have it reviewed by someone else ?

siwane:
I didn't know it was a backport (even if it's written in branch name). So I'm not sure we need other review. Thanks @krampstudio

PR #1902, opened 2018-12-04
Fix/tao 7238/fix viewable hiddenboxes

uncleempty:


helpers/form/elements/xhtml/class.Viewablehiddenbox.php
```diff
@@ -99,13 +100,36 @@ public function render()
         iconHide.classList.add('icon-eye-slash');
         hide();
         
+        //allowing to focus span on tab key
+        input.addEventListener('keydown', function(e) {
+            if (e.which == 9) {
```
jsconan: There are several issues here:
- the lazzy equality should be avoided, unless a very good reason (for instance `a == null` to check against both `undefined` and `null`, but this is a edge case). The strict equality should be always used: `e.which === 9`
- the `key` property should be preferred in `keydown`/`keyup` events, since `which` is deprecated (see [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent))

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
```diff
@@ -99,13 +100,36 @@ public function render()
...
+                toggle.focus();
+                e.preventDefault();
+                return false;
+            } else {
+                return false;
```
jsconan: The `return false` has the same meaning as `preventDefault`. (see [Event/returnValue](https://developer.mozilla.org/en-US/docs/Web/API/Event/returnValue)).

It is also better to avoid to have several exit points, for the same reason `goto` is deprecated in many languages and is not implemented anymore.

The question here is: what do you expect with this construction? If it is "always prevent default action", then simply call `e.preventDefault();` at the beginning of the listener. But be careful, you might prevent the input to behave well.

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
```diff
@@ -99,13 +100,36 @@ public function render()
...
             } else {
                 hide();
             }
         });
+        
+        //allowing hide/show from keyboard, by space(#32) key press
+        toggle.addEventListener('keyup', function(e) {
+            if (e.which !== 32) {
+                return false;
+            }
```
jsconan: Please use `key` instead of `which` as this property is deprecated. Also, avoid early function exit when possible, that breaks the algorithmic good practices. Instead you should do:
```javascript
if (e.key === ' ') {
    if (input.type === 'password') {
        show();
    } else {
        hide();
    }
} else {
    // but I'm still wondering why it should be default prevented...
    e.preventDefault();
}
```

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: Will change to strict equality.
About e.key - will change + a question: I know that e.which && e.keyCode are deprecated, but maybe you know - is there any way of getting key code (which is not deprecated) instead of getting its value? Or since they are deprecated the only way is to compare with string representation of character not its code?

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: Will remove `return false` - just didn't know that it is also the same meaning in clear js (as it is in jQuery event model).

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: Same as  the first one.
About why it should be prevented - in the case we are not preventing default - `toggle` will immediately loose focus as it will be transferred to the login button

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: Also `keydown` listener would be removed. The problem with toggle focus loose - came from the keydown listener. When it is removed (or changed to `keyup` event) - nothing strage happens

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: This one should be prevented to disallow toggle work on other keys, because it will even autotoggle the password reveal on navigating from input to span with tab.

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: As said in previous - this listener won't exist anymore

helpers/form/elements/xhtml/class.Viewablehiddenbox.php
uncleempty: Great, thank you!

PR #1905, opened 2018-12-06
TAO-7049 - Resolve issue where user timezone is not updated

martijn-tao:
Turns out the timezone was updated and stored just fine, but we were passing a Litteral as a value. String casting this value resolves the issue.


actions/class.UserSettings.php
```diff
@@ -1,157 +1,166 @@
...
+        if (!empty($props[GenerisRdf::PROPERTY_USER_UILG])) {
+            $langSettings['ui_lang'] = current($props[GenerisRdf::PROPERTY_USER_UILG])->getUri();
+        }
+        if (!empty($props[GenerisRdf::PROPERTY_USER_DEFLG])) {
+            $langSettings['data_lang'] = current($props[GenerisRdf::PROPERTY_USER_DEFLG])->getUri();
+        }
+
+        $langSettings['timezone'] = TIME_ZONE;
+        if (!empty($props[GenerisRdf::PROPERTY_USER_TIMEZONE])) {
+            $langSettings['timezone'] = (string) current($props[GenerisRdf::PROPERTY_USER_TIMEZONE]);
```
martijn-tao: The `(string)` casting here is basically the fix for the issue. The rest of the changes is simply code cleanup.

PR #1908, opened 2018-12-11
Fix/tao 6125/prevent statenment insertion

SergiiTao:
To prevent changes in the form fields new GignatureGenerator service was implemented. It generates a signature for readonly field(-s) in order to check if the field was changed manually or not. 
The idea is: backend calculates a signature for some form/data and send it to frontend. This signature must be sent to backend with some action request. While processing a request backend calculates a  signature again and compares it with the signature provided in the request. If a protected field was changed the signatures will be different and backend ends up with an error.

How to test:
All actions like move, copy, delete, edit, create, etc. need to be tested with instances and subclasses. With BRT and without.


models/classes/security/SecurityException.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->messageToLog = $message;
+
+        parent::__construct('', $code, $previous);
+
+        common_Logger::singleton()->handleException($this);
+    }
+
+    public function __toString()
+    {
+        return get_class($this) . " '{$this->messageToLog}' in {$this->file}({$this->line})\n
```
Skyld: Please use sprintf and fix conventional issues

models/classes/security/SecurityException.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    private $messageToLog;
+
+    public function __construct($message = null, $code = 0, \Exception $previous = null)
+    {
+        $this->messageToLog = $message;
+
+        parent::__construct('', $code, $previous);
+
+        common_Logger::singleton()->handleException($this);
```
Skyld: Are you sure we have to handle the exception here? I don't know if it covers all the cases but Monolog's ErrorHandler registers a custom error handler function, which will log any exceptions, please check. Also, it's not a good practice to log in the constructor, you cannot guarantee that the exception is actually throwed.

models/classes/security/SignatureGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+namespace oat\tao\model\security;
+
+use oat\oatbox\service\ConfigurableService;
+use oat\tao\model\metadata\exception\InconsistencyConfigException;
+
+class SignatureGenerator extends ConfigurableService
+{
+    const SERVICE_ID = 'tao/SignatureGenerator';
+
+    const SALT = 'salt';
```
Skyld: OPTION_SALT, if it's an option.

models/classes/security/SignatureGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return string
+     *
+     * @throws InconsistencyConfigException
+     */
+    public function generate(...$data)
+    {
+        $salt = $this->getOption(self::SALT);
+
+        if (empty($salt)) {
+            throw new InconsistencyConfigException('Salt is not defined');
```
Skyld: You should use the constant value here to avoid inconsistency later.

models/classes/security/SignatureValidator.php
```diff
@@ -0,0 +1,55 @@
...
+        $signatureGenerator = ServiceManager::getServiceManager()->get(SignatureGenerator::class);
+
+        foreach ($list as $item) {
+           if ($signatureGenerator->generate($item[$idFieldName]) !== $item[$signatureFieldName]) {
+               throw new SecurityException('Invalid signature');
+           }
+        }
+    }
+
+    public function checkSignature($signature, ...$data)
```
Skyld: The `data` parameter is very confusing here, you can have no clue about what is expected here. Please create a Signature object or something to make is easy to understand what is expected here.

models/classes/security/SignatureValidator.php
```diff
@@ -0,0 +1,55 @@
...
+ * Copyright (c) 2018 (update and modification) Open Assessment Technologies SA
+ */
+
+namespace oat\tao\model\security;
+
+use oat\oatbox\service\ServiceManager;
+
+class SignatureValidator
+{
+    public function checkSignatures(array $list, $signatureFieldName = 'signature', $idFieldName = 'id')
```
Skyld: This method should invoke the `checkSignature` method instead of code duplication. Please refactor

models/classes/security/SignatureGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+    const SALT = 'salt';
+
+    /**
+     * @param string[] $data
+     *
+     * @return string
+     *
+     * @throws InconsistencyConfigException
+     */
+    public function generate(...$data)
```
Skyld: This parameter is very confusing, it has zero meaning. Please use appropriate parameter names.

scripts/install/RegisterSignatureService.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace oat\tao\scripts\install;
+
+use oat\tao\model\security\SignatureGenerator;
+use oat\taoLti\scripts\install\InstallServices;
+
+class RegisterSignatureService extends InstallServices
+{
+    public function __invoke($params)
```
Skyld: Missing typehint

scripts/install/RegisterSignatureService.php
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) 2018 (original work) Open Assessment Technologies SA
+ *
+ */
+
+namespace oat\tao\scripts\install;
+
+use oat\tao\model\security\SignatureGenerator;
+use oat\taoLti\scripts\install\InstallServices;
+
+class RegisterSignatureService extends InstallServices
```
Skyld: It registers the signature generator, not the signature. Please rename.

scripts/install/RegisterSignatureService.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function __invoke($params)
+    {
+        $taskSerializer = new SignatureGenerator([
```
Skyld: Task serializer or signature generator? Please adjust variable name

scripts/install/RegisterSignatureService.php
```diff
@@ -0,0 +1,38 @@
...
+            SignatureGenerator::SALT => uniqid(mt_rand(), true)
```
Skyld: Please use the signature register service here instead of re-declaring the salt. I'm also not sure about the salt usage, having it as a constructor parameter means that you have fixed salt during an entire request...

scripts/update/Updater.php
```diff
@@ -875,5 +876,15 @@ public function update($initialVersion) {
         }
 
         $this->skip('21.5.0', '22.4.2');
+
+        if ($this->isVersion('22.4.2')) {
+            $taskSerializer = new SignatureGenerator([
```
Skyld: Same comment here, please see above

Skyld:
Please fix merge conflicts

scripts/install/RegisterSignatureService.php
SergiiTao: `Please use the signature register service here instead of re-declaring the salt` sorry. don't undrestand what it means.

`fixed salt during an entire request` i'd say, fixed salt during entire life.

models/classes/accessControl/data/DataAccessControl.php
```diff
@@ -19,21 +19,40 @@
...
 use oat\tao\model\controllerMap\ActionNotFoundException;
-use common_exception_MissingParameter;
+use oat\tao\model\lock\LockManager;
 
 /**
  * Interface for data based access control
  */
 class DataAccessControl implements AccessControl
 {
+    private function getUris(array $parameters, array $filterNames)
```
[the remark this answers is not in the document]
SergiiTao: for there is nothing to split. do you have ideas?

helpers/form/validators/ResourceSignatureValidator.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        return 'Signature is not valid';
+    }
+
+    /**
+     * @param string $message
+     *
+     * @return $this
+     */
+    public function setMessage($message)
```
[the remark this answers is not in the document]
Skyld: It's not about that. This class is a perfect example of violation of interface segregation principle, but since the interface forces you be able to set custom message, please do it. If you 'hack' it around, it just further decrease the trustfulness of the code. Please implement the features demanded by the interface.

helpers/form/validators/ResourceSignatureValidator.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        return $this;
+    }
+
+    /**
+     * @param array $options
+     *
+     * @return $this
+     */
+    public function setOptions(array $options)
```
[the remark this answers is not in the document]
Skyld: See my comment above. You don't have to refactor the interface, but respect it to avoid more confusion that it's necessary. I don't get your point with the loop thing, could you please explain?

models/classes/GenerisTreeFactory.php
```diff
@@ -224,6 +225,10 @@ private function buildClassNode(core_kernel_classes_Class $class, core_kernel_cl
     {
     	$label = $class->getLabel();
         $label = empty($label) ? __('no label') : $label;
+
+        /** @var SignatureGenerator $signatureGenerator */
+        $signatureGenerator = ServiceManager::getServiceManager()->get(SignatureGenerator::class);
```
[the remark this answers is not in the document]
Skyld: For the same reason why you have a `getSearchService()` method instead of calling `$this->getModel()->getSearchInterface()`. It's more clean.

models/classes/accessControl/data/DataAccessControl.php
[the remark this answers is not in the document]
Skyld: Extracting the body of the foreach to something like `getUri()` would be a good starting point :)

actions/form/class.CreateInstance.php
```diff
@@ -203,7 +206,49 @@ public function initElements()
...
+        $uris = [];
+
+        /** @var core_kernel_classes_Class $class */
+        foreach ($this->classes as $class) {
+            $uris[] = $class->getUri();
+        }
+
+        return $uris;
+    }
 }
```
Skyld: Please add missing empty line at end of file

actions/form/class.CreateInstance.php
```diff
@@ -203,7 +206,49 @@ public function initElements()
...
+        $signatureGenerator = ServiceManager::getServiceManager()->get(SignatureGenerator::class);
+
+        return $signatureGenerator->generate($this->getDataToSign());
+    }
+
+    protected function getDataToSign()
+    {
+        $uris = [];
+
+        /** @var core_kernel_classes_Class $class */
```
Skyld: You can use `core_kernel_classes_Class[]` as docblock on the class property, so you don't have to do it here.

actions/form/class.Users.php
```diff
@@ -33,7 +33,7 @@
  * @package tao
  
  */
-class tao_actions_form_Users extends tao_actions_form_Instance
+class tao_actions_form_Users extends \oat\tao\model\controller\SignedFormInstance
```
Skyld: Please import class

models/classes/accessControl/data/DataAccessControl.php
Skyld: This method doesn't belong here. Please move it into a url helper class

models/classes/accessControl/data/DataAccessControl.php
```diff
@@ -68,25 +76,24 @@ public function hasAccess(User $user, $controller, $action, $parameters) {
...
      * Gets the cleaned paramName from paramValue ($cleanName)
      *
      * @param string $paramName just for logging purposes
-     * @param string $cleanName param to be cleared
+     * @param string $decodedUri param to be cleared
      *
      * @return string
      */
-    private function getCleanName($paramName, $cleanName)
+    private function getEncodedUri($decodedUri)
```
Skyld: `getEncodedUri` should not return with `decodedUri`. Please fix

models/classes/security/SecurityException.php
Skyld: Hm?

models/classes/security/SignatureGenerator.php
Skyld: What I mean is do not hardcode the word salt here, use the constant value instead.

models/classes/security/SignatureValidator.php
Skyld: It's clean code principle to use meaningful names. Data is the worst what we can have :) Please try to find something more meaningful.

scripts/install/RegisterSignatureGenerator.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function __invoke($params)
+    {
+        $signatureGenerator = new SignatureGenerator([
+            SignatureGenerator::OPTION_SALT => uniqid(mt_rand(), true)
+        ]);
+
+        $this->getServiceManager()->register(SignatureGenerator::SERVICE_ID, $signatureGenerator);
+
+        return new \common_report_Report(\common_report_Report::TYPE_SUCCESS, 'SignatureGenerator service is registered');
```
Skyld: Please import classes

scripts/install/RegisterSignatureGenerator.php
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) 2018 (original work) Open Assessment Technologies SA
+ *
+ */
+
+namespace oat\tao\scripts\install;
+
+use oat\tao\model\security\SignatureGenerator;
+use oat\taoLti\scripts\install\InstallServices;
+
+class RegisterSignatureGenerator extends InstallServices
```
Skyld: It should be `SignatureGeneratorRegister`. Also, `InstallServices` class is in taoLti extension. I think it's not a good approach to depend on an extension in Core. What do you think?

scripts/update/Updater.php
```diff
@@ -959,5 +960,15 @@ public function update($initialVersion) {
         }
 
         $this->skip('27.4.0', '27.5.0');
+
+        if ($this->isVersion('27.5.0')) {
+            $signatureGenerator = new SignatureGenerator([
```
Skyld: Please use the register here to fetch the generator.

models/classes/accessControl/data/DataAccessControl.php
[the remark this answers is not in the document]
Skyld: Ping

scripts/install/RegisterSignatureService.php
Skyld: About the fixed salt part: ok

actions/class.RdfController.php
```diff
@@ -62,6 +78,35 @@ protected function getClassService()
...
+
+        $root = $this->getRootClass();
+
+        if ($instance->isClass()) {
+            $class = new core_kernel_classes_Class($instance->getUri());
+
+            if (!($class->isSubClassOf($root) || $class->equals($root))) {
+                throw new SecurityException('Security issue');
+            }
+        } else if (!$instance->isInstanceOf($root)) {
```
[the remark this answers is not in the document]
SergiiTao: remove `else`? why?

actions/class.RdfController.php
```diff
@@ -343,7 +397,14 @@ public function editClassLabel()
...
 
-        $editClassLabelForm = new \tao_actions_form_EditClassLabel($clazz,  $this->getRequestParameters());
+        /** @var SignatureGenerator $signatureGenerator */
+        $signatureGenerator = ServiceManager::getServiceManager()->get(SignatureGenerator::class);
+
+        $signature = $signatureGenerator->generate(
+            \tao_helpers_Uri::encode($this->getRequestParameter('classUri'))
+        );
+
+        $editClassLabelForm = new \tao_actions_form_EditClassLabel($clazz,  $this->getRequestParameters(), $signature);
```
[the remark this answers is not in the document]
SergiiTao: will it look better or work faster?

helpers/form/validators/ResourceSignatureValidator.php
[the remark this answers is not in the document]
SergiiTao: ok. i got your point.

models/classes/GenerisTreeFactory.php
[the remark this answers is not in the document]
SergiiTao: i disagree, but ok:)

actions/class.RdfController.php
```diff
@@ -62,6 +78,35 @@ protected function getClassService()
...
+    {
+        $instance = $this->getResource($uri);
+
+        $root = $this->getRootClass();
+
+        if ($instance->isClass()) {
+            $class = new core_kernel_classes_Class($instance->getUri());
+
+            if (!($class->isSubClassOf($root) || $class->equals($root))) {
+                throw new SecurityException('Security issue');
```
Skyld: Please provide more meaningful exception messages

models/classes/accessControl/data/DataAccessControl.php
SergiiTao: decodedUri is a string

scripts/install/RegisterSignatureGenerator.php
SergiiTao: ops. autocomplete issue :-)

SergiiTao: i thing RegisterSignatureGenerator to be consistant and it sounds more logical to me.

views/js/provider/resources.js
```diff
@@ -155,7 +155,7 @@ define([
              * @param {String} destinationClassUri - the destination class
              * @returns {Promise<Object>} resolves with the data of the new resource
              */
-            copyTo : function copyTo(uri, destinationClassUri) {
+            copyTo : function copyTo(uri, destinationClassUri, signature) {
```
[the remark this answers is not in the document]
SergiiTao: fixed

views/js/layout/actions/common.js
```diff
@@ -620,7 +624,7 @@ define([
                             this.disable();
 
                             resourceProvider
-                                .moveTo(selectedUri, destinationClassUri)
+                                .moveTo(selectedData, destinationClassUri)
```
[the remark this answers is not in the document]
jsconan: Anyway, seems it is working fine.

PR #1909, opened 2018-12-11
Correct order of calling sort event and function to get proper sort params in the listeners

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-6428

The current implementation does not pass sort order value via the event.

### DIFF
--- a/views/js/ui/datatable.js
+++ b/views/js/ui/datatable.js
@@ -702,14 +702,16 @@ define([
          * @fires dataTable#sort.datatable
          */
         _sort: function($elt, sortBy, asc, sortType) {
+
+            var options = this._sortOptions($elt, sortBy, asc, sortType);
+
             /**
              * @event dataTable#sort.dataTable
              * @param {String} column - The name of the column to sort
              * @param {String} direction - The sort direction
              */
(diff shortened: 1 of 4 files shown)


views/js/ui/datatable.js
```diff
@@ -702,14 +702,16 @@ define([
...
+
+            var options = this._sortOptions($elt, sortBy, asc, sortType);
+
             /**
              * @event dataTable#sort.dataTable
              * @param {String} column - The name of the column to sort
              * @param {String} direction - The sort direction
              */
-            $elt.trigger('sort.' + ns, [sortBy, asc, sortType]);
+            $elt.trigger('sort.' + ns, [options.sortby, options.sortorder, options.sorttype]);
```
krampstudio: Could you please update the event documentation so we know what we expect in the parameters

krampstudio:
I've verified the unit test and if the sort is still working correctly in TAO.
@tikhanovichA could you please check it works in NCCER's env ?

tikhanovichA:
Yes, will check it.

views/js/ui/datatable.js
Alroniks: Fixed

PR #1915, opened 2018-12-19
Fix/tao 7281/fix client detector location

siwane:
This is a copy from client diagnostic services


models/classes/clientDetector/detector/WebBrowserDetector.php
```diff
@@ -0,0 +1,84 @@
...
+namespace oat\tao\model\clientDetector\detector;
+
+use oat\generis\model\OntologyAwareTrait;
+use Sinergi\BrowserDetector\Browser;
+
+class WebBrowserDetector extends AbstractDetector
+{
+    use OntologyAwareTrait;
+
+    const MAKE_CLASS = 'http://www.tao.lu/Ontologies/TAODelivery.rdf#BrowserMake';
```
zagovorichev: I know that this is legacy, but why `http://www.tao.lu/Ontologies/TAODelivery.rdf`?  `http://www.tao.lu/Ontologies/tao.rdf` or `http://www.tao.lu/Ontologies/taoDiagnosticTool.rdf`

zagovorichev: but it will be not easy to change on the existing environment, maybe later?

zagovorichev:
- [ ] @jbout Could you confirm, please, that we need to have this code in TAO core?
- [] PHPUnit tests
- [] RDF migration (from taoDiagnosticTool)

PR #1916, opened 2018-12-20
blacklist for actions

zagovorichev:
@jbout This is the very first step of this task implementation - implemented new AnnotationRoute service to control such annotations: `@RouteAnnotation(action = "NotFound");`
the second step - implementation of NotFound action for all of the listed actions
and the last one - migrate phpDoc Annotation reader to Doctrines annotation reader


models/classes/routing/RouteAnnotationService.php
```diff
@@ -0,0 +1,101 @@
...
+     * @return bool
+     */
+    public function hasAccess($className, $methodName = '')
+    {
+        $access = true;
+        try {
+            $annotations = $this->getAnnotations($className, $methodName);
+            if (array_key_exists(AnnotationReaderService::PROP_SECURITY, $annotations)) {
+                foreach ($annotations[AnnotationReaderService::PROP_SECURITY] as $rule) {
+                    switch ($rule) {
```
SashaTAO: So we need to handle situations when someone writes ``` @security("Deny")``` or ```@security("DENY")```?

models/classes/routing/RouteAnnotationService.php
```diff
@@ -0,0 +1,101 @@
...
+                        case 'hidden':
```
SashaTAO: Typo: 'hide' instead of 'hidden'

models/classes/routing/RouteAnnotationService.php
zagovorichev: not sure that we need to make them case insensitive. I mean this is like a constant, like a syntax, only programmers can change or use it, so let them do it right.

zagovorichev:
@SashaTAO Update

models/classes/routing/RouteAnnotationService.php
SashaTAO: ok

PR #1918, opened 2018-12-20
TAO-7261 - Removed local namespace references from RDF export

martijn-tao:
The local namespace references could not always be found (eg. original item was removed, or local namespace doesn't match).


helpers/data/class.GenerisAdapterRdf.php
```diff
@@ -138,6 +136,9 @@ private function addResource(EasyRdf_Graph $graph, core_kernel_classes_Resource
             if (!empty($triple->lg)) {
                 $graph->addLiteral($triple->subject, $triple->predicate, $triple->object, $triple->lg);
             } elseif (common_Utils::isUri($triple->object)) {
+                if (strpos($triple->object, LOCAL_NAMESPACE) !== false) {
```
martijn-tao: This if statement is the added functionality. The other changes is code cleanup based on code inspection

helpers/data/class.GenerisAdapterRdf.php
```diff
@@ -113,13 +111,13 @@ public function export(core_kernel_classes_Class $source = null)
     private function addClass(EasyRdf_Graph $graph, core_kernel_classes_Class $resource)
     {
         $this->addResource($graph, $resource);
-        foreach ($resource->getInstances(false) as $instance) {
```
siwane: You are going to break the platform with these 3 lines

models/classes/export/class.RdfExporter.php
```diff
@@ -92,30 +93,29 @@ public function export($formValues, $destination)
...
         } elseif (count($xmls) > 1) {
 
-            //merge the xml of each instances...
-
             $baseDom = new DomDocument();
             $baseDom->formatOutput = true;
             $baseDom->loadXML($xmls[0]);
 
-            for ($i = 1; $i < count($xmls); $i++) {
+            for ($i = 1, $iMax = count($xmls); $i < $iMax; $i++) {
```
siwane: Oo?

helpers/data/class.GenerisAdapterRdf.php
martijn-tao: The default value for the `$recusrsive` param in these methods is set to false, so I don't see how this would impact the current functionality

models/classes/export/class.RdfExporter.php
martijn-tao: Using a `count()` call every loop can get quite expensive. This prevents that by storing the value of `count($xmls)` in `$iMax`, and using that variable for each loop.

helpers/data/class.GenerisAdapterRdf.php
siwane: my bad

martijn-tao:
All logic has been moved into the tao-core extension

PR #1921, opened 2018-12-31
Feature/tao 7016 enhance waiting dialog

oatymart:
A part of the work for https://oat-sa.atlassian.net/browse/TAO-7016

Enables a secondary button to be shown in the `waitingDialog` component (currently only used inside the Test Runner).


views/js/ui/waitingDialog/waitingDialog.js
```diff
@@ -142,23 +161,44 @@ define([
...
                 });
 
-                $button = $('[data-control="waiting"]', this.dialog.getDom());
                 $content = $('.content', this.dialog.getDom());
+                $button = $('[data-control="waiting"]', this.dialog.getDom());
+                if (this.config.showSecondary) {
+                    this.dialog.getDom().addClass('has-secondary');
+                    $secondaryButton = $('[data-control="secondary"]', this.dialog.getDom());
+                    $secondaryButton.on('click', function() {
+                        self.trigger('click.secondary');
```
krampstudio: You shouldn't trigger events with namespaces and I'd name differently `secondaryaction` for example ? We don't name event's based on the DOM event but based on what they're supposed to do : an external action ? an additionnal action ? etc.

views/scss/inc/_components.scss
```diff
@@ -200,6 +200,37 @@ ul.listbox {
...
+        .message {
+            font-weight: bold;
+        }
+        .buttons {
+            // override traditional button layout
+            width: 100%;
+            text-align: center;
+            button.waiting:disabled {
+                // override traditional disabled button style
+                background-color: transparent !important;
```
krampstudio: We consider `!important` as a bad practice, (except to override a `!important`, but then everything will become  `!important`).
If not required, please always try a to use the CSS specifier weights

views/scss/inc/_components.scss
```diff
@@ -200,6 +200,37 @@ ul.listbox {
...
+                color: $textColor !important;
+                opacity: 1 !important;
+                font-size: 1.6rem !important;
```
krampstudio: There's a mixin : `@include font-size(16)` (please check the exact syntax)

views/scss/inc/_components.scss
```diff
@@ -200,6 +200,37 @@ ul.listbox {
     right: 0;
     overflow: hidden;
 
+    &.has-secondary {
```
krampstudio: Is the stil only bound to the `waitingDialog` component ? If yes, could you please move the rules into a dedicated file `tao/views/js/ui/waitingDialog/{css,scss}` (and add it to the build file too)

views/scss/inc/_components.scss
oatymart: I don't like it but unfortunately the `button.waiting` already has several `!important` properties, and the design requires that I override them. I will look again if there's any way to work around it.

views/js/ui/waitingDialog/waitingDialog.js
oatymart: Ok

views/scss/inc/_components.scss
oatymart: Moved.

views/scss/inc/_components.scss
krampstudio: unfortunately there are no way to get around `!important` The issue is on the original css.

views/js/ui/waitingDialog/waitingDialog.js
```diff
@@ -142,23 +170,50 @@ define([
...
+                $button = $('[data-control="waiting"]', this.dialog.getDom());
+                if (this.config.showSecondary) {
+                    this.dialog.getDom().addClass('has-secondary');
+                    $secondaryButton = $('[data-control="secondary"]', this.dialog.getDom());
+                    $secondaryButton.on('click', function() {
+                        self.trigger('secondaryaction');
+                    });
+                    if (this.config.buttonSeparatorText) {
+                        $betweenButtonsText = $('<span>')
+                            .html(__('or'))
```
krampstudio: as a good practice when you insert text, please update only the text node : 
`.text(__('or'))` in jQuery or `node.textContent = __('or')` using the DOM API.
This will prevent possible XSS when the content contains tags, so as a convention : when you insert text, allow only to insert text...

PR #1923, opened 2019-01-09
Refactor/tao 7352/login component

uncleempty:
This PR introduces login component.
It was done so to move some code and DOM manipulations from BE side to FE. As the way it is currently done on the BE side - just provides a lot of mess.

### DIFF
--- a/actions/class.Main.php
+++ b/actions/class.Main.php
@@ -224,38 +224,6 @@ class tao_actions_Main extends tao_actions_CommonModule
 			}
 		}
 
-//        $renderedForm = $form->render();
-//
-//        // replace the login form by a fake form that will delegate the submit to the real form
-//        // this will allow to prevent the browser ability to cache login/password
-//        if ($disableAutoComplete) {
-//            // make a copy of the form and replace the form attributes
-//            $fakeForm = preg_replace('/<form[^>]+>/', '<div class="form loginForm fakeForm">', $renderedForm);
-//            $fakeForm = str_replace('</form>', '</div>', $fakeForm);
-//
-//            // replace the password field by a text field in the actual form,
-//            // so the browser won't detect it and won't be able to cache the credentials
-//            $renderedForm = preg_replace('/type=[\'"]+password[\'"]+/', 'type="text"', $renderedForm);
-//
-//            // hide the actual form,
-//            // it will be submitted through javascript delegation
-//            $renderedForm = preg_replace_callback('/<form([^>]+)>/', function($matches) {
-//                $str = $matches[0];
-//                if (false !== strpos($str, ' style=')) {
-//                    $str = preg_replace('/ style=([\'"]+)([^\'"]+)([\'"]+)/', ' style=$1$2;display:none;$3', $str);
-//                } else {
-//                    $str = '<form' . $matches[1] . ' style="display:none;">';
-//                }
-//                return $str;
-//            }, $renderedForm);
-//
-//            // the fake form will be displayed instead of the actual form,
-//            // it will behave like the actual form
-//            $renderedForm .= $fakeForm;
-//        }
-//
-//        $this->setData('form', $renderedForm);
-
         $this->setData('title', __("TAO Login"));
 
         $this->setData('autocompleteDisabled', (int)$disableAutoComplete);
--- a/actions/class.Main.php
+++ b/actions/class.Main.php
@@ -221,7 +221,14 @@ class tao_actions_Main extends tao_actions_CommonModule
                 } catch (core_kernel_users_Exception $e) {
                     $this->setData('errorMessage', __('Invalid login or password. Please try again.'));
                 }
-			}
+			} else {
+                foreach ($form->getElements() as $formElement) {
+                    $fieldError = $formElement->getError();
+                    if ($fieldError) {
+                        $this->setData('fieldMessages_' . $formElement->getName(), $fieldError);
+                    }
+                }
+            }
 		}
 
         $this->setData('title', __("TAO Login"));
--- a/actions/class.Main.php
+++ b/actions/class.Main.php
@@ -50,10 +50,10 @@ class tao_actions_Main extends tao_actions_CommonModule
 {
     use LoggerAwareTrait;
 
-	/**
-	 * First page, when arriving on a system
-	 * to choose front or back office
-	 */
+    /**
+     * First page, when arriving on a system
+     * to choose front or back office
+     */
     public function entry()
     {
         $this->defaultData();
@@ -64,26 +64,26 @@ class tao_actions_Main extends tao_actions_CommonModule
             }
         }
 
-	    if (empty($entries)) {
-	        // no access -> error
-	        if (common_session_SessionManager::isAnonymous()) {
+        if (empty($entries)) {
+            // no access -> error
+            if (common_session_SessionManager::isAnonymous()) {
                 /* @var $urlRouteService DefaultUrlService */
                 $urlRouteService = $this->getServiceLocator()->get(DefaultUrlService::SERVICE_ID);
                 $this->redirect($urlRouteService->getLoginUrl());
-	        } else {
-	            common_session_SessionManager::endSession();
+            } else {
+                common_session_SessionManager::endSession();
                 return $this->returnError(__('You currently have no access to the platform'), true, 403);
-	        }
-	    } elseif (count($entries) == 1 && !common_session_SessionManager::isAnonymous()) {
-	        // single entrypoint -> redirect
-	        $entry = current($entries);
-	        return $this->redirect($entry->getUrl());
-	    } else {
-	        // multiple entries -> choice
-	        if (!common_session_SessionManager::isAnonymous()) {
-	            $this->setData('user', $this->getSession()->getUserLabel());
-	        }
-    	    $this->setData('entries', $entries);
+            }
+        } elseif (count($entries) == 1 && !common_session_SessionManager::isAnonymous()) {
+            // single entrypoint -> redirect
+            $entry = current($entries);
+            return $this->redirect($entry->getUrl());
+        } else {
+            // multiple entries -> choice
+            if (!common_session_SessionManager::isAnonymous()) {
+                $this->setData('user', $this->getSession()->getUserLabel());
+            }
+            $this->setData('entries', $entries);
             $naviElements = $this->getNavigationElementsByGroup('settings');
             foreach($naviElements as $key => $naviElement) {
                 if($naviElement['perspective']->getId() !== 'user_settings') {
@@ -101,8 +101,8 @@ class tao_actions_Main extends tao_actions_CommonModule
             $this->setData('current-section', $this->getRequestParameter('section'));
             $this->setData('content-template', array('blocks/entry-points.tpl', 'tao'));
             $this->setView('layout.tpl', 'tao');
-	    }
-	}
+        }
+    }
 
     /**
      * Authentication form,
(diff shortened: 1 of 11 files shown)


views/js/ui/login/login.js
```diff
@@ -0,0 +1,215 @@
...
+            /**
+             *
+             * @returns {*}
+             */
+            createFakeForm : function createFakeForm() {
+                var $fakeFormDom = this.getElement().clone();
+
+                var $form = $fakeFormDom.find('form').clone();
+
+                var $fakeForm = $form.replaceWith('<div class="form loginForm fakeForm">' + $form.html() + '</div>');
```
jsconan: Could you use a Handlebars template instead?

`fakeForm.tpl`:
```html
<div class="form loginForm fakeForm">{{{form}}}</div>
```

then:
```javascript
$fakeFormDom.html(fakeFormTpl({form: $fakeFormDom.find('form').html()});
```

views/js/ui/login/login.js
```diff
@@ -0,0 +1,215 @@
...
+            },
+
+            manipulateFormDom : function manipulateFormDom() {
+                var $form, $pwdInput, $pwdLabel;
+
+                if (this.isAutocompleteDisabled()) {
+                    $form = this.getFakeForm();
+                } else {
+                    $form = this.getElement().find('form');
+                }
```
jsconan: Indeed, this code could be factored.
```javascript
getForm: function getForm() {
    var $form;
    if (this.isAutocompleteDisabled()) {
        $form = this.getFakeForm();
    } else {
        $form = this.getElement().find('form');
    }
    return $form;
}
```

views/js/ui/login/login.js
```diff
@@ -0,0 +1,215 @@
...
+
+                $pwdInput = $form.find('input[type=password]');
+                $pwdLabel = $form.find('label[for=' + $pwdInput.attr('name') + ']');
+
+                $pwdInput.replaceWith('<span class="viewable-hiddenbox">'
+                    + $pwdLabel[0].outerHTML
+                    + $pwdInput[0].outerHTML
+                    + '<span class="viewable-hiddenbox-toggle" tabindex="0"><span class="icon-preview"></span><span class="icon-eye-slash" style="display: none;"></span></span></span>'
+                );
+                $pwdLabel.remove();
```
jsconan: Could you use a Handlebars template instead?

views/js/ui/login/login.js
```diff
@@ -0,0 +1,215 @@
...
+                } else {
+                    $form = this.getElement().find('form');
+                }
+
+                $pwdInput = $form.find('input[type=password]')[0];
+                $inputToggle = $form.find('.viewable-hiddenbox-toggle');
+                $viewIcon = $form.find('span.icon-preview');
+                $hideIcon = $form.find('span.icon-eye-slash');
+
+                show = function() {
```
jsconan: do you expect to change the assigment later? if no, please use regular function declaration. That introduces some mess in the code, and since the function is anonymous, it is not properly named while debugging or tracing errors.

```javascript
function show() {
    ...
}
function hide() {
    ...
}
function autoHide() {
    ...
}
```

or eventually assign the function at declaration time, with the name:
```javascript
var show = function show() {
    ...
};
```

views/js/ui/login/login.js
```diff
@@ -0,0 +1,215 @@
...
+                    $viewIcon.hide();
+                    $hideIcon.show();
+
+                    $pwdInput.type = 'text';
+                    $pwdInput.autocomplete = 'off';
+
+                    window.addEventListener('mousedown', autoHide);
```
jsconan: Be careful when dealing with global events like that. You should always use namespace, and ensure a proper context clean on time

views/js/ui/login/login.js
```diff
@@ -0,0 +1,215 @@
...
+                };
+
+                hide = function() {
+                    $hideIcon.hide();
+                    $viewIcon.show();
+
+                    $pwdInput.type = 'password';
+                    $pwdInput.autocomplete = self.isAutocompleteDisabled() ? 'off' : 'on';
+
+                    window.removeEventListener('mousedown', autoHide);
```
jsconan: The same here, be careful when dealing with global events like that. You should always use namespace, and ensure a proper context clean on time

actions/class.Main.php
```diff
@@ -224,39 +224,11 @@ public function login()
...
-            // the fake form will be displayed instead of the actual form,
-            // it will behave like the actual form
-            $renderedForm .= $fakeForm;
-        }
-
-        $this->setData('form', $renderedForm);
         $this->setData('title', __("TAO Login"));
 
+        $this->setData('autocompleteDisabled', (int)$disableAutoComplete);
+        $this->setData('passwordRevealEnabled', (int)$enablePasswordReveal);
```
jsconan: Is it actually an integer? The component is stating it is a boolean...

views/js/controller/login.js
```diff
@@ -41,71 +50,76 @@ define([
...
+                    $fakeForm.find(':input').each(function () {
+                        var $field = $(this);
+                        $loginForm.find('input[name="' + $field.attr('name') + '"]').val($field.val());
+                    });
 
-            versionWarning.init();
+                    // just submit the real form as if the user did it
+                    loadingBar.start();
+                    $loginForm.submit();
+                }
```
jsconan: Should it not be part of the login component? It is weird to have such business logic outside of the component.

Eventually, the component could trigger events to stick on in order to animate the loading bar and other side effects

views/js/controller/login.js
```diff
@@ -41,71 +50,76 @@ define([
...
+                 */
+                function displayMessages() {
+                    var $fields = $context.find(':input');
+                    _.forEach(messages, function (message, level) {
+                        if (message) {
+                            feedback().message(level, message).open();
+                            $fields.addClass(level);
+                        }
+                    });
+                }
```
jsconan: The same here, I have the feeling this should be handled by the component itself

views/js/controller/login.js
```diff
@@ -41,71 +50,76 @@ define([
...
-            $fakeForm
-                .find('input[type="submit"], button[type="submit"]')
-                .off('click').on('click', function (e) {
-                    e.preventDefault();
-                    submitForm();
-                });
+                // empty $fields sent
+                if (!messages.error && $context.find('.form-error').length) {
+                    messages.error = __('All fields are required');
+                }
```
jsconan: This too should be managed by the  component

views/js/controller/login.js
```diff
@@ -41,71 +50,76 @@ define([
...
+
+                // submit the form when the user hit the submit button inside the fake form
+                $fakeForm
+                    .find('input[type="submit"], button[type="submit"]')
+                    .off('click').on('click', function (e) {
                     e.preventDefault();
                     submitForm();
-                }
-            });
+                });
```
jsconan: Definitely, this should lie in the component

views/js/controller/login.js
```diff
@@ -41,71 +50,76 @@ define([
...
 
-            $loginBtn.removeAttr('disabled')
-                     .removeClass('disabled');
+                // submit the form when the user hit the ENTER key inside the fake form
+                $fakeForm.on('keypress', function (e) {
+                    if (e.which === 13) {
+                        e.preventDefault();
+                        submitForm();
+                    }
+                });
```
jsconan: And that too :)

views/js/controller/login.js
```diff
@@ -41,71 +50,76 @@ define([
...
 
-            loadingBar.stop();
+                $loginBtn.removeAttr('disabled')
+                    .removeClass('disabled');
```
jsconan: This too should be in the component.

views/js/ui/login/login.js
```diff
@@ -0,0 +1,306 @@
...
+
+            /**
+             * Attaches events to password reveal options
+             */
+            attachPasswordRevealEvents : function attachPasswordRevealEvents() {
+                var $form, $pwdInput, $inputToggle, $viewIcon, $hideIcon;
+
+                var self = this;
+
+                var autoHide = function(event) {
```
jsconan: could you name the function?

views/js/ui/login/login.js
```diff
@@ -0,0 +1,306 @@
...
+                    if (!event.target.isSameNode($pwdInput) && !event.target.isSameNode($hideIcon[0]) && !event.target.isSameNode($inputToggle[0])) {
+                        hide();
+                    }
+                };
+
+                var show = function() {
```
jsconan: same here, could you name the function?

views/js/ui/login/login.js
```diff
@@ -0,0 +1,306 @@
...
+
+                    $pwdInput.type = 'text';
+                    $pwdInput.autocomplete = 'off';
+
+                    window.addEventListener('mousedown.' + ns, autoHide);
+
+                    $pwdInput.focus();
+                };
+
+                var hide = function() {
```
jsconan: and here too, could you name the function?

views/js/ui/login/login.js
```diff
@@ -0,0 +1,306 @@
...
+                 */
+                var submitForm = function submitForm() {
+                    // if the fake form exists, copy all fields values into the real form
+                    $fakeForm.find(':input').each(function () {
+                        var $field = $(this);
+                        $loginForm.find('input[name="' + $field.attr('name') + '"]').val($field.val());
+                    });
+
+                    // just submit the real form as if the user did it
+                    self.trigger('submit.' + ns);
```
jsconan: Caution, if you trigger the event with the namespace, we cannot listen to `submit` but will have to listen to `submit.login` only, which is too restrictive and might prevent to use proper namespace later. Namespace should be use only on listeners (`before`, `on`, `after`, `off`), not on actioners, unless strong reason.

Also, could you document the event? in the function annotation something likes `@fires submit when the form is submitted`, while triggering something like:
```
/**
 * @event submit
 * @param ...
 */
```

views/js/ui/login/login.js
```diff
@@ -0,0 +1,306 @@
...
+                    $fakeForm = this.createFakeForm();
+
+                    this.hide();
+                    this.getElement().find('form').attr('id', 'loginForm');
+                    this.getContainer().prepend($fakeForm);
+
+                    // submit the form when the user hit the submit button inside the fake form
+                    $fakeForm
+                        .find('input[type="submit"], button[type="submit"]')
+                        .off('click').on('click', function (e) {
```
jsconan: caution, without namespace all click listeners will go away

views/js/ui/login/login.js
```diff
@@ -0,0 +1,306 @@
...
+                            e.preventDefault();
+                            submitForm();
+                        });
+
+                    // submit the form when the user hit the ENTER key inside the fake form
+                    $fakeForm.on('keypress', function (e) {
+                        if (e.which === 13) {
```
jsconan: could you use the `key` api instead? `which` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which)

views/js/ui/login/tpl/login.tpl
```diff
@@ -0,0 +1,29 @@
+<div class="xhtml_form login-component">
+    <form action="/tao/Main/login" method="post" id="{{name}}" name="{{name}}">
```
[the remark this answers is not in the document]
uncleempty: Yup, would move it

actions/class.Main.php
```diff
@@ -221,42 +221,21 @@ public function login()
...
-
-            // hide the actual form,
-            // it will be submitted through javascript delegation
-            $renderedForm = preg_replace_callback('/<form([^>]+)>/', function($matches) {
-                $str = $matches[0];
-                if (false !== strpos($str, ' style=')) {
-                    $str = preg_replace('/ style=([\'"]+)([^\'"]+)([\'"]+)/', ' style=$1$2;display:none;$3', $str);
-                } else {
-                    $str = '<form' . $matches[1] . ' style="display:none;">';
+			} else {
```
jsconan: Caution, you added tabs in the file, that violates [PSR-2 rules](https://www.php-fig.org/psr/psr-2/). Please check your IDE settings.

actions/class.Main.php
```diff
@@ -221,42 +221,21 @@ public function login()
...
                 }
-                return $str;
-            }, $renderedForm);
-
-            // the fake form will be displayed instead of the actual form,
-            // it will behave like the actual form
-            $renderedForm .= $fakeForm;
-        }
+            }
+		}
```
jsconan: The same here, tabs instead of spaces.

PR #1930, opened 2019-01-11
Fix/tao 6637/log 500 exceptions

SergiiTao:


scripts/update/Updater.php
```diff
@@ -875,6 +875,9 @@ public function update($initialVersion) {
             $this->setVersion('21.5.0');
         }
 
+<<<<<<< HEAD
+        $this->skip('21.5.0', '23.0.0');
```
siwane: *biiiip*

models/classes/mvc/error/ExceptionInterpretor.php
```diff
@@ -107,7 +106,7 @@ protected function interpretError() {
     }
     
     public function getTrace() {
-        return $this->trace;
+        return $this->exception ? $this->exception->getMessage() : '';
```
Skyld: Why would a method called `getTrace` return an exception message? Please either rename the method to fulfill it's purpose or return the trace

models/classes/mvc/error/ExceptionInterpretor.php
```diff
@@ -130,11 +129,11 @@ public function getResponse() {
         $class = $this->getResponseClassName();
         /*@var $response ResponseAbstract */
         $response = new $class;
-        $response->setServiceLocator($this->getServiceLocator())
-                ->setException($this->exception)
-                ->setHttpCode($this->returnHttpCode)
-                ->trace($this->trace);
+        $response->setServiceLocator($this->getServiceLocator());
```
Skyld: This seems weird to me. How is it ensured that exception was set here through `setException` method? Why do we need the `getResponseClassName` method? It should be namespace prefixed before already. This should be refactored.

models/classes/mvc/error/ExceptionInterpretor.php
```diff
@@ -74,7 +74,6 @@ public function setException(Exception $exception){
      * and http status to return
      */
     protected function interpretError() {
```
Skyld: Please fix all the conventional errors in the class

models/classes/mvc/error/ResponseAbstract.php
```diff
@@ -129,9 +129,11 @@ public function send()
     /**
```
Skyld: Please fix all the conventional errors in this class

Skyld:
Please resolve conflicts with develop

models/classes/mvc/error/ExceptionInterpretor.php
SergiiTao: had the same question.
it should be used in one place only, but have no idea about forks.
if you want to rename it we should depreacte it, create the new one, wait some time...

models/classes/mvc/error/ExceptionInterpretor.php
SergiiTao: it is weird indeed. but here `$response->setException($this->exception)` we're using the setter again.
it means (should mean) that this is not important value and if we set it to null nothing bad happens.

models/classes/mvc/error/ExceptionInterpretor.php
SergiiTao: I can do it as a separate PR. To make a life of reviewer easier ;)

models/classes/mvc/error/ExceptionInterpretor.php
Skyld: Okay I see it was not you, leave it as it is then, it's not in scope now.

models/classes/mvc/error/ExceptionInterpretor.php
Skyld: Where? You just removed the `setException` call from here. What do you mean?

Skyld: Ignore it, as discussed

PR #1932, opened 2019-01-14
TAO-7417 web installer

handleman:
Task: https://oat-sa.atlassian.net/browse/TAO-7417
Description:
When you are installing tao application with web installer you re getting into jQuery crush when return to the "Licensing Agreement" step from last step.
screenshot 2019-01-13 at 19 51 19

With approve from Bertrand side, was decided to wrap jquery-ui lib into AMD module manually. It would be proper temporary solution since we are planning to remove jquery-ui in future


handleman:
@thomasgarrard  could you check?

PR #1944, opened 2019-01-22
Fix/tao 7082/fixed deque without queue

zagovorichev:
dequeue action calls each time to run an action, even if the user wasn't queued


models/classes/actionQueue/implementation/InstantActionQueue.php
```diff
@@ -151,12 +151,15 @@ protected function dequeue(QueuedAction $action, User $user)
     {
         $key = $this->getQueueKey($action);
         $positions = $this->getPositions($action);
-        unset($positions[$user->getIdentifier()]);
+        if (array_key_exists($user->getIdentifier(), $positions)) {
+            // now we sure that this user has been queued
```
boajer: Looks like this condition should wrap also changing trend `if` statement,  that would allow to minimise writes to the persistence.

uncleempty: Changes applied

PR #1949, opened 2019-01-22
Update CommonRestModule to be able to work with classes

Babacooll:
This PR aims at being able to work with "classes" as well as "resources" in the CommonRestModule.

The CommonRestModule currently implements post(), put(), delete(), get() that are able to work with resources. Those methods are now deprecated (will be removed in the future) and 4 facades have been added on top of them restPostResource(), restPutResource(), restDeleteResource(), restGetResource(), they have the same signature and when the 4 previous methods will be deprecated the logic will be reimplemented inside those new ones.

The old index() method gives access to the original API (for resources) without any breaking change.
The new resources() method gives access to the original API (for resources but with breaking changes (the ACL is more extensive).
The new classes() method gives access to a new API that works with classes.

There are also 3 new methods restPostClass(), restDeleteClass() and restGetClass().


siwane (approved):

zagovorichev:
1. Do we need the required parameters (at least `label`) for class creation?
2. PHP Warning during the class creation

2df7195574766b3e74f9aeafe38f2737b51f6b44670227d2b7332117334bab82

PR #1950, opened 2019-01-22
Remove Easter Egg from item's stylesheets

start-admin:
https://oat-sa.atlassian.net/browse/TAO-7458

replace the comment by a /* Do not edit */


helpers/CssHandler.php
```diff
@@ -26,25 +26,7 @@ class CssHandler{
 
     public static function buildWarning() {
         return <<<EOF
```
dietertaotesting: Since this is now a on liner there is no point in using [heredoc ](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) anymore. Please use `return '/* Do not edit */';` instead.

PR #1957, opened 2019-01-25
Fix/TAO-6103 Horizontal orientation is not displayed properly in Preview mode

bziondik:


views/js/ui/waitForMedia.js
```diff
@@ -71,13 +74,13 @@ define(['jquery'], function($){
                 return;
             }
 
-            $img.each(function(){
-                if(this.complete){
-                    //the image is already loaded by the browser
-                    imageLoaded.call(this);
+            $allSource.each(function(){
+                if(this.tagName === "IMG" && this.complete || this.tagName === "LINK" && this.sheet){
```
uncleempty: Hello, although it will work because of operator precedence, but could you add grouping with `()` for better readability?

uncleempty: Also, are you sure that relying on attribute with [experimental status](https://developer.mozilla.org/en-US/docs/Web/API/LinkStyle#Browser_compatibility) and approved compatibility in Edge only is the only way to check if css is loaded?

PR #1964, opened 2019-01-30
Feature/tao 7419 dialog heading & checkbox

oatymart:
UI parts for https://oat-sa.atlassian.net/browse/TAO-7419

Just adding a checkbox template and the possibility to display a heading in the dialog template.

See https://github.com/oat-sa/extension-tao-testqti/pull/1427 for the main work which requires these changes.


views/js/ui/dialog/tpl/checkbox.tpl
```diff
@@ -0,0 +1,5 @@
+<label for="{{id}}">
+    <input type="checkbox" id="{{id}}" name="{{id}}" {{#if checked}}checked{{/if}} />
+    <span class="icon-checkbox"></span>
+    {{{text}}}
```
krampstudio: Could you please use `{{text}}` instead, since we target only to insert text, that'll prevent some xss cases.

views/js/ui/dialog/tpl/body.tpl
```diff
@@ -1,5 +1,9 @@
 <div class="preview-modal-feedback modal">
     <div class="modal-body clearfix">
+        {{#if heading}}
+        <h4 class="strong">{{{heading}}}</h4>
```
krampstudio: Could you please use `{{heading}}` instead, since we target only to insert text, that'll prevent some xss cases.

PR #1965, opened 2019-01-31
TAO-6785 - Message fix in feedback tpl

ferenckiss-tao:
Two issues has been fixed:

**1. Alert content currently is able to evaluate scripts**

Steps to reproduce:
1. Log in as admin 
2. Create a test with "<script>alert("test label alert")</script>" as label
3. Click on Publish on the action bar
4. Select destination
5. Publish should fail
6. Alert appears before the error message

Root cause:
tao/views/js/ui/report/feedback.tpl line 3 -> {{{message}}}
This way we don't escape html in html in message. Since this is normally a reason text message, I'm suggesting to escape html every time.

Suggestion for fixing:
{{message}}

**2. XSS on Result page**

Steps to reproduce:
1. Log in as admin 
2. Create a test
5. Create a Delivery
6. Do the test as guest or as test taker
7. Include xss script in any input or textarea field
8. Open Results page as admin
9. Export table
10. Click on Add all responses
11. Script will be evaluated from the table rows

Root cause:
tao/views/js/ui/datatable/tpl/layout.tpl at line 100

<td class="{{id}}">{{{property id ../../this}}}</td>
Suggestion for fixing:

<td class="{{id}}">{{property id ../../this}}</td>
IF there is no other case when we have to put html in it (like <i><b> etc)!!!


views/build/package-lock.json
```diff
@@ -1,5746 +0,0 @@
-{
-  "name": "tao-build",
```
krampstudio: why did you remove the `package-lock.json` ?

views/js/lib/dompurify/purify.min.js
```diff
@@ -0,0 +1 @@
...ply(t,n)});var v=Object.seal||function(e){return e},b=v(/\{\{[\s\S]*|[\s\S]*\}\}/gm),T=v(/<%[\s\S]*|[\s\S]*%>/gm),A=v(/^data-[\-\w.\u00B7-\uFFFF]/),x=v(/^aria-[\-\w]+$/),L=v(/^(?:(?:(?:f|ht)tps?|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i),S=v(/^(?:\w+script|data):/i),w=v(/[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205f\u3000]/g),k="function"==typeof Symbol&&"symbol"==typeof Symbol.iterator?function(e){return typeof e}:function(e){return e&&"function"==typeof Symbol&&e.constructor===Symbol&&e!==Symbol.prototype?"symbol":typeof e},M=("undefined"!=typeof Reflect&&Reflect).apply,E=Array.prototype.slice,O=Object.freeze,N=function(){return"undefined"==typeof window?null:window};M||(M=function(e,t,n){return e.apply(t,n)});var _=function(e,t){if("object"!==(void 0===e?"undefined":k(e))||"function"!=typeof e.createPolicy)return null;var n=null;t.currentScript&&t.currentScript.hasAttribute("data-tt-policy-suffix")&&(n=t.currentScript.getAttribute("data-tt-policy-suffix"));var r="dompurify"+(n?"#"+n:"");try{return e.createPolicy(r,{createHTML:function(e){return e}})}catch(e){return console.warn("TrustedTypes policy "+r+" could not be created."),null}};return r()});
```
krampstudio: Could you please use the non minified version. Most of the time using minified version works but we already had some issues when bundling them and optimizing already optimized files.
Could you please used instead : https://github.com/cure53/DOMPurify/blob/master/dist/purify.js ?

views/js/ui/report/feedback.tpl
```diff
@@ -1,6 +1,6 @@
 <div class="feedback-{{type}} {{#if hasChildren}}hierarchical{{else}}leaf{{/if}}">
     <span class="icon-{{type}}{{#if hasChildren}} hierarchical-icon{{/if}}"></span>
-    <div class="message">{{{message}}}</div>
+    <div class="message">{{message}}</div>
```
krampstudio: You should use the `dompurify` helper also there. 
The message property seems to contain some formatting in some situation

views/js/lib/dompurify/purify.min.js
ferenckiss-tao: yes

views/js/ui/report/feedback.tpl
ferenckiss-tao: ok

PR #1966, opened 2019-01-31
Fix/tao 7450/check class uri

tikhanovichA:
https://oat-sa.atlassian.net/browse/TAO-7450


actions/class.RestController.php
```diff
@@ -66,6 +66,12 @@ protected function getClassFromRequest(\core_kernel_classes_Class $rootClass)
 
         if ($this->hasRequestParameter(self::CLASS_URI_PARAM)) {
             $class = $this->getClass($this->getRequestParameter(self::CLASS_URI_PARAM));
+
+            if ($class === null || !$class->exists()) {
```
SashaTAO: If class-uri is empty another type of exception is thrown (can not create class), maybe we could handle this case as well.
In the end of this method we already check ```if ($class === null || !$class->exists())``` and assign $root class in that case. I think we should check only  ```if($class === null)``` there.

tikhanovichA:
Fixed

PR #1967, opened 2019-01-31
Fix/si 254/fix js url builder

alexeyvy:
Fixes this: https://oat-sa.atlassian.net/browse/NCSUP-182


views/js/util/url.js
```diff
@@ -176,7 +176,17 @@ define([
                         if(!_.isEmpty(acc) || hasQueryString){
                             acc += '&';
                         }
-                        acc += encodeURIComponent(key) + '=' + encodeURIComponent(value);
+                        if (typeof value === "object") {
+                            for (var objectProperty in value) {
```
hutnikau: You can use `_.forOwn(obj, function(value, key) { } );` to iterate through object properties.

views/js/util/url.js
```diff
@@ -176,7 +176,14 @@ define([
                         if(!_.isEmpty(acc) || hasQueryString){
                             acc += '&';
                         }
-                        acc += encodeURIComponent(key) + '=' + encodeURIComponent(value);
+                        if (typeof value === "object") {
+                            _.forOwn(value, function(parameterName, parameterValue) {
+                                parameterName = encodeURIComponent(key) + "[" + encodeURIComponent(parameterName) + "]";
```
krampstudio: do not reassign the `parameterName` variable : 
```
_.forOwn(value, function(parameterName, parameterValue) {
    acc += encodeURIComponent(key) + "[" + encodeURIComponent(parameterName) + "]=" + encodeURIComponent(parameterValue) + "&";
});
```

alexeyvy: done

views/js/util/url.js
alexeyvy: done

PR #1970, opened 2019-02-01
TAO-5856 pagestatus component tests vere failed

handleman:
one test has been failing http://qa.test.it.taocloud.org:8080/job/build-tao-js/lastCompletedBuild/testReport/http_..127.0.0.1_8082.tao.ui/pageStatus/

the problem was in wrong timing for asynchronous assertion checks.


krampstudio:
@handleman did you find what was wrong ?

handleman:
@krampstudio fixed issue with absence of blur and focus events on headless chrome. Could you check one more time

views/js/ui/pageStatus.js
```diff
@@ -122,14 +122,14 @@ define([
                 if(e.target === win){
                     pageStatus.trigger('statuschange', 'blur', e.timeStamp);
                 }
-            }, 200, {leading : true, trailing: false} ));
+            }, {leading : true, trailing: false} ));
```
krampstudio: Why did you remove the debounce wait ?

views/js/test/ui/pageStatus/test.js
```diff
@@ -47,40 +48,51 @@ define([
...
-        var popup = window.open('/','test','width=300,height=300,visible=none');
+        var popup = window.open('/tao/views/js/test/ui/pageStatus/blank.html','test','width=300,height=300,visible=none');
+        var secondPopup;
 
         var pageStatus = pageStatusFactory({
             window :  popup
         });
-
-        var counter = 0;
+        if (/HeadlessChrome/.test(window.navigator.userAgent)) {
```
krampstudio: i'd prefer to either : 
 - disable the test in headless mode 
 - have one test for browser/one for headless
Because I don't understand by reading this test which case is supposed to produce what.
Having a dynamic number of expectation makes the result uncertain

views/js/test/ui/pageStatus/test.js
```diff
@@ -47,40 +48,51 @@ define([
...
+            QUnit.expect(4);
+        }else{
+            QUnit.expect(6);
+        }
 
         pageStatus
-            .on('statuschange', function(status){
+            .on('statuschange', _.once(function(status){
```
krampstudio: Why do you restrict the handler ? 
Do you expect it to be triggered multiple time but want to assert it has been called only once ? 
You can direclty use `.one('statuschange'` or `on/off`

views/js/ui/pageStatus.js
handleman: did not found reasons to have fixed time delay.
By default _.debounce will differ execution till next system "tick" in other words will wait for clean call-stack. It would be enough to eliminate event duplication made synchronously within one AMD module.

views/js/test/ui/pageStatus/test.js
handleman: Agree, will separate them

views/js/test/ui/pageStatus/test.js
handleman: The main goal of this particular test is to check if our module can read all window events. 
To check some events like , blur, focus, load we have to duplicate some others which are related.

for example, If we need to check **hide** event on new popup window, we will invoke these events in this particular order:
1) focus
2) unload
3)show
4) load
5) unload
6) hide

As you can see ** unload** event is duplicated because popup window needs to unload default blank content before loading given .html document.

And there could be other nuances related to order of events since them are asynchronous.

handleman: So I am checking ability to catch such events, neither than amount of fired events. Amount and order could be different across the browsers and call-stack business

views/js/ui/pageStatus.js
krampstudio: If I remember well on some browser the event is called multiple times in some situations, like the fullscreen notification on OSX.  I think the debounce wait was set on purpose.
Since we are using this feature to kick user out of a test when we detect a lost of focus, or we don't when we detect the blur/focus was inside allowed elements, I'd suggest to not change it without testing all the use cases (we actually had a lot to handle since the API seems to be inconsistent across browsers and OSs.

handleman: ok, sure.

krampstudio:
If it is too difficult to have a consistent behaviour between browsers and headless, I've no problem to deactivate some cases.

handleman:
@krampstudio 

what is done ( according your feedback)

* have one test for browser/one for headless
* got back" the debounce wait"

need your aprove / disaprove  on my answer below your "Why do you restrict the handler" in the comment above.

>If it is too difficult to have a consistent behaviour between browsers and headless, I've no problem to deactivate some cases.

I am not sure that I got the idea of what exactly should be deactivated. 
Now tests are executed as expected, all window events are checked for desktop, and all  except `blur`, `focus` for headless.

PR #1973, opened 2019-02-04
tao/actions/class.CommonRestModule.php spaces indents

siarhei-baradzin:
Needed for https://oat-sa.atlassian.net/browse/TAO-7390

GIVEN the class structure of Tao
WHEN the test takers are to be placed in an existing class folder OR in a new class folder
THEN the API allows the ability to have the test takers put into the specified location


siarhei-baradzin:
https://github.com/oat-sa/tao-core/pull/1977 created instead of it

PR #1983, opened 2019-02-07
Fix/tao 7517/fix csvexporter

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/TAO-7517

Issue: if there are many entries CsvExporter raised out of memory error.
Fix: Removed unnecessary file layer and now csv goes directly to stdout.

HOWTO Test: go to EventLog extension, push button "Export log entries"


tikhanovichA:
Was fixed by https://github.com/oat-sa/extension-tao-eventlog/pull/93

PR #1985, opened 2019-02-08
Feature/tao 7378/calculator integrate nthroot

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-7378

Here is the implementation of the `nthRoot` feature into the scientific calculator.
In first place, a forked version of `expr-eval` is now in use. This was required to ease the implementation of the `nthRoot`. The reason why is `nthrt` is a function accepting 2 parameters, which is not easily manageable for the display. With the forked version, the function can be used as a binary operator, simply by prefixing it with a `@`. So, now instead of `nthrt(8, 4)` we have `4 @nthrt 8`. An important change though was to invert the parameters orders in the existing function to reflect the operands order of a binary operator: previously `nthrt(value, level)`, now `nthrt(level, value)`.

Regarding the changes made on the calculator, here is a summary:
- as discussed in the previous PR, the CSS is now isolated for this sole component
- the lib `express-eval` is updated with the forked version
- the calculator component directly loads the CSS
- the main component (`calculator/core/board.js`) has been updated to take care of prefixed operator functions, and the space handling has also been fixed
- the tokenizer has been updated to take care of prefixed functions
- the terms labels have been extracted and put in a dedicated list. That brings more consistency across display, this way the keyboard can use exactly the same labels as the displayed expression, and all keyboard can rely on a shared base of labels.
- the term definition now takes care of the exponent notation
- the tokens renderer is now able to apply exponent notation, on both sides of an operator, so either `nthRoot` or `power of` operators are taken into account

Can be tested with the unit tests, from `tao/views/js/test/ui/maths/calculator/*`, or by playing with the calculator widget, like `/tao/views/js/test/ui/maths/calculator/scientificCalculator/test.html`. 

On the unit test side, the following have been impacted:
- `/tao/views/js/test/util/mathsEvaluator/test.html`
- `/tao/views/js/test/ui/maths/calculator/scientificCalculator/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/board/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/labels/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/terms/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/tokenizer/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/tokens/test.html`
- `/tao/views/js/test/ui/maths/calculator/plugins/screen/simpleScreen/test.html`

You can run them locally using this command:
`npx grunt connect:dev qunit:single --test=/path/to/test.html`

~*Note*: UX did not yet validate the visual. Slight changes could occur on the display.~


views/js/ui/maths/calculator/core/terms.js
```diff
@@ -13,357 +13,444 @@
...
+
+    /**
+     * Defines the terms that can be tokenized from an expression
+     * @type {term[]}
+     */
     return {
         // Digits definition
         NUM0: {
-            label: '0',
+            label: labels.NUM0,
```
krampstudio: I'm not sure to understand why the "labels" are separated from the "terms". Especially since they're also somehow linked in the template.

views/js/ui/maths/calculator/core/tpl/backspace.tpl
```diff
@@ -0,0 +1,6 @@
+<svg class="backspace-outline" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="32" height="32" viewBox="0 0 24 24">
```
krampstudio: Could you please optimize inline svg, size matter...
For example, `svgo` with default options leads to 37% of weight loss : 
```
<svg class="backspace-outline" xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 24 24"><path d="M19 21H9c-1.436 0-3.145-.88-3.977-2.046l-2.619-3.667c-.629-.881-1.157-1.621-1.188-1.661-.246-.344-.249-.894-.008-1.241.018-.025.561-.783 1.204-1.686L5.02 7.046C5.855 5.879 7.566 5 9 5h10c1.654 0 3 1.346 3 3v10c0 1.654-1.346 3-3 3zM3.229 12.999l.806 1.125 2.618 3.667C7.104 18.424 8.223 19 9.001 19h10c.552 0 1-.45 1-1.001V8c0-.551-.448-1-1-1h-10c-.776 0-1.897.576-2.351 1.209l-2.608 3.652-.813 1.138z"/><path d="M13.707 13l2.646-2.646a.502.502 0 0 0 0-.707.502.502 0 0 0-.707 0L13 12.293l-2.646-2.646a.5.5 0 0 0-.707.707L12.293 13l-2.646 2.646a.5.5 0 0 0 .707.708L13 13.707l2.646 2.646a.5.5 0 1 0 .708-.706L13.707 13z"/></svg>
```
It is important because the content inside template is left untouched

krampstudio: Same for the other inlined SVGs

views/js/ui/maths/calculator/plugins/keyboard/templateKeyboard/defaultTemplate.tpl
```diff
@@ -1,37 +1,37 @@
 <div class="calculator-keyboard">
     <div class="row">
-        <button class="key operator" data-command="term" data-param="SQRT"><span>&radic;</span></button>
-        <button class="key operator" data-command="term" data-param="LPAR"><span>(</span></button>
-        <button class="key operator" data-command="term" data-param="RPAR"><span>)</span></button>
-        <button class="key command" data-command="clear"><span>C</span></button>
-        <button class="key command" data-command="clearAll"><span>AC</span></button>
+        <button class="key operator" data-command="term" data-param="SQRT"><span>{{{labels.SQRT}}}</span></button>
```
krampstudio: Could you confirm there's no way to set a label dynamically ?

krampstudio: (to prevent ways to inject code using `{{{html}}}`

views/js/ui/maths/calculator/scss/calculator.scss
```diff
@@ -0,0 +1,357 @@
...
+    $calculatorSwitchPadding: 4px;
+    $calculatorScreenPadding: 2px;
+    $calculatorBaseFontSize: 12px;
+
+    position: relative;
+    font-family: monospace, sans-serif;
+    font-size: $calculatorBaseFontSize;
+    width: 100%;
+    height: 100%;
+    display: flex;
```
krampstudio: You still need to use the prefixes or the dirty prefixer mixin (for IE mostly and some webkits)

views/js/ui/maths/calculator/scss/calculator.scss
```diff
@@ -0,0 +1,357 @@
+/* Style for the calculator component */
+@import "inc/bootstrap";
+
+/* font size mixin specific to calculator to use em instead of rem unit */
+@function calc-em-size($em) {
+    @return ($em) * 1em;
+}
+@mixin calc-font-size($em) {
+    font-size: calc-em-size($em) !important;
```
krampstudio: No way to work this around ?
We should ban `!important` as much as possible (only keep it when the rule to override is using `!important`)

views/js/ui/maths/calculator/core/tpl/backspace.tpl
dietertaotesting: FYI, https://jakearchibald.github.io/svgomg/ is pretty good

views/js/ui/maths/calculator/scss/calculator.scss
dietertaotesting: ref: https://github.com/oat-sa/tao-core/blob/master/views/scss/inc/_functions.scss#L68

views/js/ui/maths/calculator/scss/calculator.scss
dietertaotesting: I hadn't seen this on the earlier PRs but `font-size` should be in `rem` rather than `em` unless there is a good reason for that.

views/js/ui/maths/calculator/core/board.js
```diff
@@ -57,6 +57,12 @@ define([
         position: 0
     };
 
+    /**
+     * Regex that matches the prefixed function operators
+     * @type {RegExp}
+     */
+    var rePrefixedTerm = /^@[a-zA-Z_][a-zA-Z_0-9]*$/;
```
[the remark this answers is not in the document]
dietertaotesting: `[a-zA-Z_0-9]` can be written as `[\w]`

views/js/ui/maths/calculator/core/board.js
```diff
@@ -445,20 +455,21 @@ define([
...
+                        // prepend space when either the term to add or the previous term is an identifier
+                        if (position && needSpace) {
                             value = ' ' + value;
                         }
 
-                        // append space when the either the term to add or the previous term is an identifier
-                        if (position < expression.length && (tokensHelper.isIdentifier(nextToken) || (isIdentifier && !tokensHelper.isSeparator(nextToken)))) {
+                        // append space when either the term to add or the next term is an identifier
+                        if ((!position && needSpace) || (position < expression.length && tokenNeedSpace(nextToken))) {
                             value += ' ';
```
dietertaotesting: I don't mean to nitpick and wouldn't let the PR fail because of this but a variable with the meaning _something needs space_ should be called `needsSpace`, not `needSpace` (3rd person vs 1st person)

views/js/ui/maths/calculator/core/board.js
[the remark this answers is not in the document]
jsconan: This is `@`, followed by letter or underscore, and optionally followed by any alphanum or underscore.
Yes, the second part could be written with a shorten notation. I choose to keep the long notation for better readability. Should I change for the short notation?

views/js/ui/maths/calculator/core/terms.js
jsconan: The reason why is to share a consistent content all across template and terms. This avoid to copy & paste svg content for instance.

Labels can also be slightly different depending on the place there are used. The button won't have always the same display as the expression. (for instance `nthroot` is displayed <sup>y</sup>&radic;x on the button, while it is displayed only &radic; in the result)

At the end, this separation is just to simplify the use of the labels.

views/js/ui/maths/calculator/core/tpl/backspace.tpl
jsconan: You are right. I will apply the inlining.
BTW, this is also a benefit of having separated list of labels, this way we can apply the inlining only at one place and make sure every template is using it.

views/js/ui/maths/calculator/plugins/keyboard/templateKeyboard/defaultTemplate.tpl
jsconan: The template is managed by the `templateKeyboard` plugin. There is no way to inject dynamically a label.

views/js/ui/maths/calculator/scss/calculator.scss
jsconan: Yes, there is a good reason: the font-size is reflecting the container's size. This is why the font-size needs to be relative to its parent. This is the trick applied to be able to resize the calculator and have the content growing.

jsconan: The `!important` is also required since the default TAO style is opposing its own style with the same flag.

jsconan:
I updated the PR with fixes regarding comments, and also the UX concerns.

views/js/ui/maths/calculator/core/board.js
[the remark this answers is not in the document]
dietertaotesting: I personally would change it because it's a very common expression. Then again this is your code, you do what makes most sense to you.

views/js/ui/maths/calculator/scss/calculator.scss
dietertaotesting: Thanks for the clarification

dietertaotesting: The `!important` in above context is optional https://github.com/oat-sa/tao-core/blob/ff757bf222bc8171febb3a301af5b42914bfe7f9/views/scss/inc/_functions.scss#L223

jsconan: Indeed this is an option, but it is in use in the overridden classes.
The same for the button: https://github.com/oat-sa/tao-core/blob/94536918798346d82e4a9cd9a482328e271d7066/views/scss/inc/_buttons.scss#L4

PR #1986, opened 2019-02-08
Fix/tao 7450/check class label

tikhanovichA:


actions/class.RestController.php
```diff
@@ -104,7 +98,9 @@ protected function getClassFromRequest(\core_kernel_classes_Class $rootClass)
             }
         }
         if ($class === null || !$class->exists()) {
```
SashaTAO: I think if $class null we should still use $rootClass. Otherwise it won't be possible to create a new class in root folder

PR #1993, opened 2019-02-13
TAO-7391 User API

Alroniks:
https://oat-sa.atlassian.net/browse/TAO-7391

Requires https://github.com/oat-sa/extension-tao-funcacl/pull/65


SashaTAO:
Can we use this router ```\oat\tao\model\routing\AbstractApiRoute``` for new controller to have nicer URL (it seems to be a new  approach)?

actions/class.UserAPI.php
```diff
@@ -0,0 +1,258 @@
...
+            $this->validateUris($parameters);
+
+            /** @var core_kernel_classes_Resource $user */
+            $user = $userService->addUser($login, $password, $this->getResource(array_shift($roles)));
+
+            foreach ($roles as $role) {
+                $userService->attachRole($user, $this->getResource($role));
+            }
+
+            $user->setPropertiesValues($parameters);
```
SashaTAO: Before we set properties maybe it would be useful to validate language parameter value. We can do it in a similar way like in TestTaker API:
https://github.com/oat-sa/extension-tao-testtaker/blob/develop/models/CrudService.php#L138

actions/class.UserAPI.php
```diff
@@ -0,0 +1,258 @@
...
+
+            return $user;
+        } catch (common_Exception $e) {
+            throw new common_exception_RestApi($e->getMessage());
```
SashaTAO: It may be not really safe in case we have PDO exception or something else with sensitive information in the message. 
I think it would be better to throw correct exceptions with specific messages where we need it.

actions/class.UserAPI.php
```diff
@@ -0,0 +1,258 @@
...
+    {
+        return [
+            'post' => ['login', 'password', 'userLanguage', 'roles']
+        ];
+    }
+
+    /**
+     * @return array
+     */
+    protected function getMandatoryUriParameters()
```
SashaTAO: This method name is a bit confusing because ```defaultLanguage``` is not really mandatory parameter.

actions/class.UserAPI.php
Alroniks: It's already done by checking URI. Or you mean to check if language really exist?

actions/class.UserAPI.php
Alroniks: I would like to leave it a mandatory parameter. I know that tt API can accept a string with language code but according to our other APIs, it is not correct.

PR #2001, opened 2019-02-18
Feature/tao 7628 qunit 2

handleman:
*definition* : Migrate to Qunit version 2+ for unit testing for frontend code. (https://qunitjs.com/upgrade-guide-2.x/) 

*task* : https://oat-sa.atlassian.net/browse/TAO-7628


views/js/lib/qunit/qunit-parameterize.js
```diff
@@ -2,132 +2,171 @@
...
-
-		var createTest = function(methodName, title, expected, callback, parameters) {
-			QUnit[methodName](
-				title,
-				expected,
-				function(assert) { return callback.call(this, parameters, assert); }
-			);
-		};
+	cases: (function() {
+		'use strict';
```
krampstudio: Wouldn't it be better to wrap the whole file ?  
```js
(function() {
  'use strict';

	window.QUnit.extend(window.QUnit, {
         //...
```

views/js/lib/qunit/qunit-parameterize.js
```diff
@@ -2,132 +2,171 @@
...
-				for (var i = 0; i < length; ++i) {
-					var currentCaseI = getItem(currentCases, i);
-					var dataI = getItem(addData, i);
-					var newCase = mix(currentCaseI, dataI);
-
-					if (newCase) newCases.push(newCase);
-				}
-				currentCases = newCases;
 
+			init: function(testCasesList) {
```
krampstudio: Is it a requirement ? I'd have preferred to keep the same API

views/js/test/controller/app/test.js
```diff
@@ -18,122 +18,121 @@
...
-define([
-    'jquery',
-    'controller/app',
-    'core/historyRouter',
-    'core/logger',
-    'ui/feedback'
-], function ($, appController, historyRouterFactory, loggerFactory, feedback) {
-    'use strict';
+define( [
+    "jquery",
```
krampstudio: Does the QUnit migration tool rewrite the quotes ?

views/js/test/controller/app/test.js
```diff
@@ -18,122 +18,121 @@
...
+        { title: "off" },
+
+        // Statifier
+        { title: "getState" },
+        { title: "setState" }
+    ];
 
-    QUnit.module('API', {
-        teardown: function() {
+    QUnit.module( "API", {
```
krampstudio: Is it also the QUnit migration tool that adds the extra space ?

views/build/grunt/test.js
```diff
@@ -32,7 +32,7 @@ module.exports = function(grunt) {
                 puppeteer : {
                     ignoreHTTPSErrors: true,
                     timeout: 30000,
-                    args: [ "--no-sandbox", "--disable-gpu", "--disable-popup-blocking" ],
+                    args: [ "--no-sandbox", "--disable-gpu", "--disable-popup-blocking" , "--autoplay-policy=no-user-gesture-required"],
```
krampstudio: could you please explain me this option

views/js/lib/qunit/qunit-parameterize.js
handleman: I didn't write this lib manually, I got it from its repo https://github.com/AStepaniuk/qunit-parameterize/blob/master/qunit2-parameterize.js
I forked it actually, so maybe it is not worth to rewrite original authors idea.
As for me, using strict mode is questionable idea in general.

Need you aproveme to make these changes.

views/js/lib/qunit/qunit-parameterize.js
handleman: It is its official API for Qunit 2
https://github.com/AStepaniuk/qunit-parameterize/blob/master/qunit2-parameterize.js

Need your approve, If it worth to rewrite original lib

views/js/test/controller/app/test.js
handleman: Unfortunately yes

views/js/test/controller/app/test.js
handleman: Unfortunately yes

views/build/grunt/test.js
handleman: it is related to `tao/views/js/test/ui/mediaplayer/test.js`
some asserts were failed because of headless chrome do not autoplay video without real user interaction.

Strange thing here is that test was passed right with QUnit 1...
My suggestion is that it is a question of inner puppeteer dependency management regarding of what test framework is invoking it. So Qunit 2 allows puppeteer to use latest version of Chrome engine, and Qunit 1 require some older patch/version of it

views/js/lib/qunit/qunit-parameterize.js
krampstudio: Ok then keep it without modifications then

views/js/test/controller/app/test.js
krampstudio: No way to workaround this ?
It isn't the most important but we will loose a huge part of the history and our conventions tend to favor single over double quotes. 
Could you check if you can't replace this within the migration tool ?

views/js/test/controller/app/test.js
krampstudio: ok

views/build/grunt/test.js
krampstudio: ok, it's related to the new media playing policy introduced recently. I'm fine with that.

views/js/lib/qunit/qunit-parameterize.js
krampstudio: If this is the official plugin API, then I'm fine.

krampstudio:
Another point, could you please verify `blanket` is still working. If not we will consider removing it in favor of a better code coverage tool (maybe as part of another story)

views/js/test/controller/app/test.js
handleman: replaced double quotes to ordinal

krampstudio:
I ran the test to compare the results of the `develop` branch and that one. 

 - `develop` : 
```
$> cd views/build && npx grunt connect:test taotest
...
>> 3514 tests completed with 0 failed, 0 skipped, and 0 todo. 
>> 16380 assertions (in 137852ms), passed: 16380, failed: 0
```

 - `feature/TAO-7628-qunit-2` : 
```
$> cd views/build && npx grunt connect:test taotest
...
>> 3514 tests completed with 0 failed, 0 skipped, and 0 todo. 
>> 16382 assertions (in 110788ms), passed: 16382, failed: 0
```

@handleman Any idea where those 2 assertions come from ? 
From the mediaplayer test ?

manifest.php
```diff
@@ -49,7 +49,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '27.4.1',
+    'version' => '27.4.2',
```
krampstudio: this is a breaking change.

handleman:
>Any idea where those 2 assertions come from ? From the mediaplayer test ?

 Hm, It is not from media player definitely, because I did not add any new assertions there. 
I can assume that it could be connected with different exception processing between `QUnit 1` and `QUnit 2` 

*The case is* Qunit number 2 runs all tests inside `try{}catch{}` block, and previous version did not! 

So, while I was fixing broken tests I met problem with some test were fallen because of raised exceptions inside dependent scripts, especially where `expect()` wasn't used. 
I managed to fix these tests without changing the code, because mostly it was a question of timing inside async tests.
I suppose that several assertions wasn't ever checked on previous version of Qunit, because of *skipping assertions* when exception was risen inside dependent files.

views/js/test/core/eventifier/test.html
```diff
@@ -33,8 +34,10 @@
                 //load the test
                 require(['test/core/eventifier/test'], function(){
 
-                    //Tests loaded, run tests
-                    QUnit.start();
+                    //waiting till .getConfig() (/tao/ClientConfig/config) request done;
+                    setTimeout(function () {
+                        QUnit.start();
+                    }, 10);
                 });
```
handleman: this fix related to situation where '/tao/views/js/test/core/eventifier/test.html' fails on direct access without grunt command. Ounce of problem was that `eventifier` and `logger`  were initialised before controller's `.getConfig()`  request was executed, and providers were unset.

handleman: this fix is related to situations where `/tao/views/js/test/core/eventifier/test.html` was fallen on direct access (from browser without grunt command) .
The source of problem was that `eventifier` and `logger` were initialised before controller's `getConfig()` request finished (`/tao/ClientConfig/config`) and so  `loggerFactory.providers` were empty at that time.

handleman:
@krampstudio Could you check again?

PR #2006, opened 2019-02-19
[WIP] Feature/tao 7305 frontend xsrf token handler

oatymart:
The ticket is https://oat-sa.atlassian.net/browse/TAO-7305
Backend ticket is https://oat-sa.atlassian.net/browse/TAO-7304

Related PRs: https://github.com/oat-sa/extension-tao-testqti/pull/1439

Changes done so far:
- added `core/request` module, mixing old functionality of `core/dataProvider/request` and `qtiServiceProxy`
- converted `core/dataProvider/request` to thin wrapper returning the response body
- converted `taoQtiTest/runner/proxy/qtiServiceProxy` to work with this
- converted `core/tokenHandler` to manage a store of multiple tokens (fully async, Promise-based)
- added `core/tokenStore` to encapsulate the store functionality (fully async, Promise-based)
- used the `core/store` backend to ensure a single source of truth for the token pool
- async unit tests for all these modules

To Do:
- ~~feedback message / UX~~
- ~~expand test coverage~~
- ~~remove failing tests in `dataProvider/request`, `communicator/poll` (and `taoQtiTest/runner/proxy/qtiServiceProxy`) expecting token to be returned in the body~~
- disable/remove token expiry functions if not needed
- ~~remove logging~~


views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+         */
+        return {
+
+            /**
+             * Get the oldest token from the queue
+             * Remove its store entry as well
+             *
+             * @returns {Promise<Object>} the token object
+             */
+            get: function get() {
```
krampstudio: Could you please rename it according to the FIFO model ?
Is it more a [queue](https://en.wikipedia.org/wiki/Queue_(abstract_data_type)) or a [Stack](https://en.wikipedia.org/wiki/Stack_(abstract_data_type)) 
so `pop` or `dequeue` ?
because `get` often means getting something without any mutation.

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+                        value: token,
+                        receivedAt: Date.now()
+                    };
+                }
+                return getStore().then(function(storage){
+                    return storage.setItem(token.value, token);
+                })
+                .then(function(updated){
+                    if (updated) {
+                        return self.enforceMaxSize().then(true);
```
krampstudio: what's the meaning of `.then(true)`  ?

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+                    if (result) {
+                        return getStore()
+                            .then(function(storage){
+                                return storage.removeItem(key);
+                            })
+                            .then(function(removed) {
+                                return removed;
+                            });
+                    }
+                    return Promise.resolve(false);
```
krampstudio: since you are in a `Promise` you can just `return false`

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+             * @returns {Promise<Boolean>} resolves once removed
+             */
+            remove: function remove(key) {
+                return this.has(key).then(function(result) {
+                    if (result) {
+                        return getStore()
+                            .then(function(storage){
+                                return storage.removeItem(key);
+                            })
+                            .then(function(removed) {
```
krampstudio: this part is not necessary

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+            clear: function clear() {
+                return getStore().then(function(storage){
+                    return storage.clear();
+                });
+            },
+
+            /**
+             * Log queue contents & store contents
+             */
+            log: function log(msg) {
```
krampstudio: I guess it's for debugging purpose ? 
Please use `core/logger` and logging should be implicit (not part of the API)

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+
+            /**
+             * Checks one token and removes it from the store if expired
+             * @param {Object} token - the token object
+             * @returns {Promise<Boolean>}
+             */
+            checkExpiry: function checkExpiry(token) {
+                var self = this;
+                if (Date.now() - token.receivedAt > config.tokenTimeLimit) {
+                    return self.remove(token.value).then(function(removed) {
```
krampstudio: same here the `.then` clause is not necessary

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+            /**
+             * Add a new token object to the queue
+             * Add an entry to the store as well
+             *
+             * @param {Object} token - the token object
+             * @param {String} token.value - long alphanumeric string
+             * @param {Number} token.receivedAt - timestamp
+             * @returns {Promise<Boolean>} - true if added
+             */
+            add: function add(token) {
```
krampstudio: is it a `push` or `enqueue` ?

views/js/core/tokenHandler.js
```diff
@@ -13,45 +13,179 @@
...
+                    configStore.setItem('clientConfigFetched', true);
+                });
+
+                return Promise.resolve(tokens).then(function(newTokens) {
+                    // Add the fetched tokens to the store, synchronously:
+                    // Chaining the promises using Array.prototype.reduce is necessary
+                    // to manage token addition & deletion correctly
+                    return newTokens.reduce(function(previousPromise, nextToken) {
+                        return previousPromise.then(function() {
+                            return self.setToken(nextToken);
```
krampstudio: I think you can achieve this with the module `core/promiseQueue` : 
```js
vat setTokenQueue = promiseQueue();
_.forEach(tokens, function(token){
    promiseQueue.serie(function(){
           return self.setToken(nextToken);
    })
});
return promiseQueue.then(function() {
                        return true;
                    });
```
(to be verified, otherwise we should have the ability to run sequential promises as an helper)

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,277 @@
...
+                return this.getIndex().then(function(latestIndex) {
+                    var keysToRemove;
+                    var excess = latestIndex.length - config.maxSize;
+                    if (excess > 0) {
+                        keysToRemove = latestIndex.slice(0, excess);
+                        return Promise.all(_.map(keysToRemove, function(key) {
+                            return self.remove(key);
+                        }));
+                    }
+                    return Promise.resolve(true);
```
krampstudio: since you're already in a promise chain, you can `return true`

actions/class.CommonModule.php
```diff
@@ -68,6 +71,47 @@ public function initialize()
...
+
+        /** @var TokenService $tokenService */
+        $tokenService = $this->getServiceLocator()->get(TokenService::SERVICE_ID);
+
+        try {
+            $newToken = $tokenService->validateToken($csrfToken);
+        } catch (common_exception_Unauthorized $e) {
+            $this->logCsrfFailure($e->getMessage(), $csrfToken);
+        } catch (common_Exception $e) {
+            $this->logCsrfFailure($e->getMessage(), $csrfToken);
```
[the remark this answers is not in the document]
martijn-tao: Joel mentioned to not catch this exception instead, so I'll go with that approach for now

actions/class.CommonModule.php
```diff
@@ -354,4 +398,31 @@ protected function getServiceManager()
...
+    /**
+     * @param string $exceptionMessage
+     * @param string|null $token
+     * @throws common_exception_Unauthorized
+     * @throws common_exception_Error
+     */
+    private function logCsrfFailure($exceptionMessage, $token = null)
+    {
+        /** @var UserService $userService */
+        $userIdentifier = $this->getSession()->getUser()->getIdentifier();
```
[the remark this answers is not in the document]
martijn-tao: If no session is available the `getSession()` call will generate a new `common_session_AnonymousSession` session

helpers/form/elements/xhtml/class.Token.php
```diff
@@ -30,12 +30,12 @@ class tao_helpers_form_elements_xhtml_Token extends tao_helpers_form_elements_xh
 {
     public function render()
     {
-        $tokenService = $this->getServiceManager()->get(TokenService::SERVICE_ID);
-
-        //always add a new token
-        $newToken = $tokenService->createToken();
-        $this->setValue($newToken);
-        Context::getInstance()->getResponse()->setCookie($tokenService->getTokenName(), $newToken, null, '/');
+//        $tokenService = $this->getServiceManager()->get(TokenService::SERVICE_ID);
```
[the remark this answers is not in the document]
martijn-tao: This will be replaced with the new method as part of TAO-7306

models/classes/security/xsrf/TokenService.php
```diff
@@ -123,6 +123,50 @@ public function checkToken($token)
...
+        }
+
+        if ($isValid !== true) {
+            if ($expired) {
+                $this->revokeToken($token);
+            }
+            throw new \common_exception_Unauthorized();
+        }
+
+        $this->revokeToken($token);
```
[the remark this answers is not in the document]
martijn-tao: In this case it wouldn't make sense to check, since we already know the token exists in the pool (which is what would cause `false` as a return  value of `revokeToken()`

actions/class.CommonModule.php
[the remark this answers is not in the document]
krampstudio: ok

helpers/form/elements/xhtml/class.Token.php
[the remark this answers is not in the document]
krampstudio: How are you planning to proceed then ?

views/js/core/dataProvider/request.js
```diff
@@ -26,85 +26,49 @@
...
+     * @param {Boolean} [noToken = false] - to disable the token
      * @returns {Promise} that resolves with data or reject if something went wrong
      */
-    return function request(url, data, method, headers, background){
-        return new Promise(function(resolve, reject){
-
-            if(_.isEmpty(url)){
-                return reject(new TypeError('At least give a URL...'));
+    return function request(url, data, method, headers, background, noToken) {
+        return coreRequest({
```
krampstudio: The promise handling can be simplified , since you are in a `then` clause, the initial promise is already resolved,  the then is used to to update the resolved value in that case, no need to add a new promise in the chain.
```js
return coreRequest({}).
.then(function(response) {
    if(!_.isUndefined(response)){
        if(response.success){
           return response.success;
        } 

        //Promise.reject works too, but usually it is used when throw doesn't (in another level of async function)
        throw response.data;
   }
});
//no need to catch since you return the promise 
```

views/js/core/dataProvider/request.js
```diff
@@ -26,85 +26,49 @@
...
-                    response = xhr.responseText;
-                }
-
-                return reject(createError(response, xhr.status + ' : ' + xhr.statusText, xhr.status));
-            });
+            else if (response.success) {
+                return Promise.resolve(response.data);
+            }
+            else {
+                return Promise.reject(response.data);
```
krampstudio: uually we ensure the thrown error is of type `Error`

views/js/core/request.js
```diff
@@ -0,0 +1,223 @@
...
+                                if (response && response.success === true) {
+                                    // there's some data
+                                    resolve(response);
+                                }
+
+                                //the server has handled the error
+                                reject(createError(response, __('The server has sent an empty response'), xhr.status));
+                            })
+                            .catch(function(error) {
+                                logger.error(error);
```
krampstudio: do not forget to reject to the upper promise (remember the chain shouldn't be broken) or you want tot ignore the errors from `setToken`  ?

views/js/core/request.js
```diff
@@ -0,0 +1,223 @@
...
+                            type: 'error',
+                            message: errorThrown || __('An error occurred!')
+                        });
+
+                        setTokenFromXhr(xhr)
+                            .then(function () {
+                                reject(createError(response, xhr.status + ' : ' + xhr.statusText, xhr.status));
+                            })
+                            .catch(function(error) {
+                                logger.error(error);
```
krampstudio: same as above

helpers/form/elements/xhtml/class.Token.php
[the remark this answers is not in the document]
martijn-tao: We're separating this over 2 branches. This branch is meant for the core functionality (generating a pool, and the ability to validate and sync tokens). The other branch will be a refactor to replace the existing implementations with the new one.

actions/class.CommonModule.php
```diff
@@ -68,6 +71,47 @@ public function initialize()
...
     }
 
+    /**
+     * Validate the current request using the CSRF token header.
+     *
+     * @return string
+     * @throws common_exception_Error
+     * @throws common_exception_Unauthorized
+     */
+    public function validateCsrf()
```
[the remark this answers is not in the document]
martijn-tao: FYI: Checked this with Joel as well, and moved the functionality to a trait. Nice catch!

views/js/core/tokenHandler.js
```diff
@@ -13,45 +13,166 @@
...
+                        tokenStore.log('tokenHandler.getToken');
+                        return added;
+                    });
+            },
+
+            /**
+             * Extracts tokens from the Client Config which should be received on every page load
+             * @returns {Promise<Boolean>} - resolves true when completed
+             */
+            getClientConfigTokens() {
```
krampstudio: Please don't use the shorthand notation on ES5 code (only for ES2015+) since it needs to be transpiled to work on old browsers like IE11: 

```js
getClientConfigTokens : function getClientConfigTokens()
```

helpers/form/elements/xhtml/class.Token.php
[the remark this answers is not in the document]
krampstudio: ok, but we can't merge into develop commented code

models/classes/security/xsrf/CsrfValidatorTrait.php
```diff
@@ -0,0 +1,105 @@
...
+     * @throws common_exception_Error
+     */
+    private function  logCsrfFailure($exceptionMessage, $token = null)
+    {
+        $userIdentifier = $this->getSession()->getUser()->getIdentifier();
+        $requestMethod  = $this->getRequestMethod();
+        $requestUri     = $this->getRequestURI();
+        $requestHeaders = $this->getRequest()->getHeaders();
+
+        $this->logWarning('Failed to validate CSRF token. The following exception occurred: ' . $exceptionMessage);
```
krampstudio: Where does the `logWarning` method come from ?

models/classes/security/xsrf/TokenService.php
```diff
@@ -123,6 +123,57 @@ public function checkToken($token)
...
+     */
+    private function isExpired($token, $revokeIfExpired = false)
+    {
+        $expired = false;
+        $actualTime = microtime(true);
+        $timeLimit  = $this->getTimeLimit();
+
+        if (($timeLimit > 0) && $token['ts'] + $timeLimit < $actualTime) {
+            $expired = true;
+            !$revokeIfExpired ?: $this->revokeToken($token);
```
krampstudio: I wouldn't have bound the revokation inside the method itself : "do one thing" principle, I'd have used this method only to verify if the token is expired, not to also revoke it : 
```
if ($this->isExpired($token)) {
    $this->revokeToken($token)
}  
```

views/js/core/communicator/poll.js
```diff
@@ -120,80 +121,90 @@ define([
...
-                        // receive optional security token
-                        if (response.token) {
-                            tokenHandler.setToken(response.token);
+                    tokenHandler.getToken().then(function(token) {
+                        if (token) {
+                            headers['X-CSRF-Token'] = token;
                         }
 
+                        // send messages to the remote service
+                        $.ajax({
```
krampstudio: I'm wondering if we can't also replace this direct call by `core/request` ?

krampstudio: The ultimate goal would be to have only one module in charge of requests.

views/js/core/request.js
```diff
@@ -0,0 +1,224 @@
...
+
+            /**
+             * Fetches a security token and appends it to headers, if required
+             * @returns {Promise<Object>} - resolves with headers object
+             */
+            var computeHeaders = function computeHeaders() {
+                var headers = _.extend({}, options.headers);
+                if (!options.noToken) {
+                    return tokenHandler.getToken().then(function(token) {
+                        headers['X-CSRF-Token'] = token || 'none';
```
krampstudio: We could move the header name into a dedicated variable, ie. `var tokenHeaderName = 'X-CSRF-Token';`

views/js/core/tokenHandler.js
```diff
@@ -13,45 +13,168 @@
...
+             * Internally, old tokens are deleted to keep queue within maximum pool size
              * @param {String} newToken
-             * @returns {Object} - this
+             * @returns {Promise<Boolean>} - resolves true if successful
              */
             setToken: function setToken(newToken) {
-                token = newToken;
-                return this;
+                return tokenStore.enqueue(newToken)
+                    .then(function(added) {
```
krampstudio: this clause isn't  necessary : 
```js
setToken: function setToken(newToken) {
   return tokenStore.enqueue(newToken);
}
```

views/js/core/tokenHandler.js
```diff
@@ -13,45 +13,168 @@
...
+                            return tokenStore.dequeue().then(function(currentToken) {
+                                return currentToken.value;
+                            });
+                        }
+                        else if (!clientConfigFetched) {
+                            // Client Config allowed! (first and only time)
+                            return self.getClientConfigTokens()
+                                .then(function() {
+                                    return tokenStore.dequeue().then(function(currentToken) {
+                                        if (currentToken) {
```
krampstudio: For readability and to ensure the behavior is consitent, I'd factorize this in a function : 

```js
getToken: function getToken() {
    //...
    var getFirstTokenValue  = function getFirstTokenValue(){
        return tokenStore.dequeue().then(function(currentToken) {
            if(currentToken){
                return currentToken.value;
            }
            return null;
        });
    };
 
    //...
 
    // Token available, use it
     return getFirstTokenValue();

    //...
   // Client Config allowed! (first and only time)
         return self.getClientConfigTokens().then(getFirstTokenValue);
                
```

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,260 @@
...
+                    });
+            },
+
+            /**
+             * Checks one token and removes it from the store if expired
+             * @param {Object} token - the token object
+             * @returns {Promise<Boolean>}
+             */
+            checkExpiry: function checkExpiry(token) {
+                var self = this;
```
krampstudio: you don't need `self` here since you stay in the same scope

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,260 @@
...
+            /**
+             * Checks all the tokens in the store to see if they expired
+             * @returns {Promise<Boolean>} - resolves to true
+             */
+            expireOldTokens: function expireOldTokens() {
+                var self = this;
+                return self.getTokens().then(function(tokens) {
+                    // Check each token's expiry, synchronously:
+                    return Object.values(tokens).reduce(function(previousPromise, nextToken) {
+                        return previousPromise.then(() => {
```
krampstudio: Please do not use arrow function until we setup Babel

views/js/core/tokenStore.js
```diff
@@ -0,0 +1,260 @@
...
+
+            /**
+             * Checks all the tokens in the store to see if they expired
+             * @returns {Promise<Boolean>} - resolves to true
+             */
+            expireOldTokens: function expireOldTokens() {
+                var self = this;
+                return self.getTokens().then(function(tokens) {
+                    // Check each token's expiry, synchronously:
+                    return Object.values(tokens).reduce(function(previousPromise, nextToken) {
```
krampstudio: Please do not use `Object.values` until we setup Babel

krampstudio: you can use 
```js
_.reduce(tokens, function( previousPromise, nextToken) {
    //...
}, Promise.resolve())
```

helpers/form/elements/xhtml/class.Token.php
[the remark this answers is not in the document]
martijn-tao: Good point, will revert this change

models/classes/security/xsrf/CsrfValidatorTrait.php
martijn-tao: This is fixed in the "follow-up" branch of this PR. Implemented the same changes for that here (abstraction of methods)

actions/class.CommonModule.php
```diff
@@ -304,4 +305,58 @@ public function setServiceLocator(ServiceLocatorInterface $serviceLocator)
...
+
+    /**
+     * Validate a CSRF token, based on the CSRF header.
+     *
+     * @throws common_Exception
+     * @throws common_exception_Unauthorized
+     */
+    protected function validateCsrf()
+    {
+        $csrfTokenHeader = $this->getPsrRequest()->getHeader(TokenService::CSRF_TOKEN_HEADER);
```
siwane: Could u check if the header exist? (hasHeader)

actions/class.CommonModule.php
```diff
@@ -304,4 +305,58 @@ public function setServiceLocator(ServiceLocatorInterface $serviceLocator)
...
+        $csrfToken = current($csrfTokenHeader);
+
+        if (empty($csrfToken)) {
+            $this->logCsrfFailure('Missing X-CSRF-Token header.');
+        }
+
+        /** @var TokenService $tokenService */
+        $tokenService = ServiceManager::getServiceManager()->get(TokenService::SERVICE_ID);
```
siwane: This method is deprecated could u please is the serviceLocator instead?

PR #2011, opened 2019-02-21
Fix/tao 7150 regression

handleman:
**description**

regression with tooltip styling was found (see the screenshot)

**related PR** 
https://github.com/oat-sa/extension-tao-itemqti/pull/1240


handleman:
@krampstudio could you look at it?

views/scss/inc/_tooltip.scss
```diff
@@ -19,7 +19,7 @@ $dark : #404040;
 }
 
 .popper,
-.tooltip {
+.tooltip[role="tooltip"] {
```
handleman: the problem was that inside taoQtiItem extension there were lines
```
.qti-item div{
  padding:0;
}
```
So I have isolated tooltip container that way. `[role=tooltip]` is used only on tooltip bubble element as expected.

PR #2031, opened 2019-03-01
TAO-7152 - CSP header configuration UI

martijn-tao:
Added new page under settings called "Security". On this page you will find a form that allows you to set the CSP header policy for the platform.

Also added a new key value persistence for the settings. This was done to avoid the usage of RDF for reading and writing simple platform settings.


views/js/controller/routes.js
```diff
@@ -53,6 +53,11 @@ define(function(){
                 'index' : 'controller/users/index',
                 'add'   : 'controller/users/add'
             }
-        }
+        },
+        'Security' : {
+            'actions' : {
+                'index' : 'form/cspHeader'
+            }
```
oatymart: A more suitable place to put the Javascript file we are going to call would be:
```
'Security' : {
            'actions' : {
                'index' : 'controller/security/cspHeaderForm'
            }
}
```

views/templates/security/view.tpl
```diff
@@ -0,0 +1,94 @@
...
+        <?= get_data('cspHeaderForm'); ?>
+    </div>
+    <?php endif;?>
+</div>
+
+<?php if (get_data('cspHeaderFormSuccess')): ?>
+<div id="csp-header-success" class="hidden"><?= get_data('cspHeaderFormSuccess') ?></div>
+<?php endif;?>
+
+<script>
```
oatymart: The contents of the `<script>` block should come out of this template and into the controller we defined in the routes (`/tao/views/js/controller/security/cspHeaderForm.js`)

actions/form/class.CspHeader.php
```diff
@@ -0,0 +1,198 @@
...
+     */
+    private function setFormData()
+    {
+        $currentSetting = $this->getSettings();
+        $listSettings = [];
+        if ($currentSetting === 'list') {
+            $listSettings = $this->getListSettings();
+        }
+
+        if (!isset($_POST[self::SOURCE_RADIO_NAME]) && $currentSetting) {
```
alexeyvy: Can we avoid using global variables here? Maybe $this->data or something?

config/default/settingsStorage.conf.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
alexeyvy: The service id is 'SettingStorage' while this file is 'settingsStorage' (one more 's')

helpers/form/validators/CspHeaderValidator.php
```diff
@@ -0,0 +1,171 @@
...
+ */
+
+namespace oat\tao\helpers\form\validators;
+
+/**
+ * Validates the given CSP headers
+ *
+ * @author Martijn Swinkels <m.swinkels@taotesting.com>
+ */
+class CspHeaderValidator extends \tao_helpers_form_Validator
```
alexeyvy: This class contains much logic in it, therefore I think should be covered with some tests (more precisely, evaluate() as a single public function and an entry point to all the logic is a target for comprehensive testing :) )

models/classes/settings/CspHeaderSettingsInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace oat\tao\model\settings;
+
+/**
+ * Interface CspHeaderSettingsInterface
+ *
+ * @author Martijn Swinkels <m.swinkels@taotesting.com>
+ */
+interface CspHeaderSettingsInterface
```
alexeyvy: maybe use interface inheritance here? I mean `CspHeaderSettingsInterface` abstraction always extends `SettingsStorageInterface`

actions/form/class.CspHeader.php
martijn-tao: I believe the posted form data is currently only available through the `$_POST` varaible. I could modify the form to add the post data to a class field however. What are your thoughts @alexeyvy?

models/classes/settings/CspHeaderSettingsInterface.php
martijn-tao: I'm only using this interface to define the constants used by this specific configuration. The `\oat\tao\model\settings\SettingsStorageInterface` actually defines the implementation of the settingsStorage

PR #2032, opened 2019-03-04
Feature/tao 7660/decorated dialog with icon

oatymart:
Frontend part needed for https://oat-sa.atlassian.net/browse/TAO-7660
Design mockup: https://projects.invisionapp.com/share/SUQNO7U3AJE#/screens/348763320

This extends the `ui/dialog` component with a `class` param which gets applied to the root element of the modal. By applying the included CSS rules, this allows any modal dialog to be "decorated" with extra space in its left-hand side, where any TAO icon can appear.

To test: trigger a modal dialog on any part of TAO. Inspect its HTML and add the following classes to the root (class="modal") element:
* `icon-{any valid tao icon}`
* (optional) `success|info|warning|danger|error` to override default colour

For example: `icon-warning` or `icon-result-ok info`


dietertaotesting:
Please update the description accordingly

PR #2033, opened 2019-03-04
Feature/tao 7233/request reponse handling

siwane:


models/classes/http/HttpFlowTrait.php
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * Copyright (c) 2019 (original work) Open Assessment Technologies SA;
+ *
+ */
+
+namespace oat\tao\model\http;
+
+use common_http_Request;
+use Context;
+use GuzzleHttp\Psr7\Uri;
```
Skyld: I don't know, for me it's weird that you are using Guzzle as foundation, which is a HTTP client for HTTP APIs. Rather than building on top of Guzzle I would take this opportunity to integrate with an already existing framework. Since we are moving towards Symfony we could use Symfony HTTP foundation, so in the later steps of refactoring the conversion will be less painful. Another solution might be to use Slim (http://www.slimframework.com/docs/v3/concepts/value-objects.html), it's very lightweight and would fit in our needs. What is your opinion?

siwane: IMO, the goal is to go middleware oriented solution for easier maintainability and to isolate some business code to unit. Http PSR7 implementation will allow us to respect a standard that can be reused in Slim context. However, this `flow` process has to be handled by the middleware engine and not from the controller. The purpose of this trait is to isolate these interactions in order to remove them. I mark this conversation as resolved and will come with a proper plan to make TAO bootstrapping better

jbout:
While not perfect a good step in the right direction

PR #2037, opened 2019-03-11
Fix/tao 1764/media player stop when leaving

start-admin:
Added check for authoring mode

https://oat-sa.atlassian.net/browse/TAO-1764

if autostart is enabled then player should not play in authoring mode.

### DIFF
--- a/views/js/ui/mediaplayer.js
+++ b/views/js/ui/mediaplayer.js
@@ -1891,7 +1891,7 @@ define([
             this.mute(!!this.startMuted);
             if (this.autoStartAt) {
                 this.seek(this.autoStartAt);
-            } else if (this.autoStart && this.section!="authoring") {
+            } else if (this.autoStart && this.section !== 'authoring') {
                 this.play();
             }
         },
--- a/views/js/ui/mediaplayer.js
+++ b/views/js/ui/mediaplayer.js
@@ -1877,6 +1877,9 @@ define([
             this._setState('canseek', this.config.canSeek);
             this._setState('loading', false);
 
+            this.url = new URL(window.location.href);
+            this.section = this.url.searchParams.get("section");
+
             /**
              * Triggers a media ready event
              * @event mediaplayer#ready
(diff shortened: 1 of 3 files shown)


start-admin:
@oatymart Could you check ?

views/js/ui/mediaplayer.js
```diff
@@ -1877,6 +1877,9 @@ define([
             this._setState('canseek', this.config.canSeek);
             this._setState('loading', false);
 
+            this.url = new URL(window.location.href);
+            this.section = this.url.searchParams.get("section");
+
```
oatymart: This approach of scraping the page URL is not very robust. If the URL scheme was changed in some way in the future (I'm not saying it's likely but it might happen), this fix would fail.

A better approach would be to find the place where the previewer mediaplayer is initialised, and overwrite its config with `{autoStart: false}`. See if you can find some way in `tao/ui/previewer` to do that.

start-admin: @oatymart, but we need autostart not to work only on authoring. Maybe in Tao is there any right way to get the current route?

views/js/ui/mediaplayer.js
```diff
@@ -1888,7 +1891,7 @@ define([
             this.mute(!!this.startMuted);
             if (this.autoStartAt) {
                 this.seek(this.autoStartAt);
-            } else if (this.autoStart) {
+            } else if (this.autoStart && this.section!="authoring") {
```
ferenckiss89: Space, strict equal and single quotes for strings please to keep consistency

this.section !== 'authoring'

views/js/ui/mediaplayer.js
jsconan: We must avoid to introduce coupling between unrelated components. Basically a component cannot know about its container. Especially, a component does not have to take care of a particular context that is not related to it, and this should not be the duty of a component to know what happen outside: a component is providing a configurable feature, that is all it can do.

Then to react to the context, this should be the duty of the container, or the client that is integrating the component.

For instance, from the item authoring, in the MediaInteraction creator widget you could ensure the autostart is disabled when instantiating the component.

PR #2041, opened 2019-03-14
Feature/tao 7617/persist highlighted text

oatymart:
It's a small part of https://oat-sa.atlassian.net/browse/TAO-7617

The main change is to enable the list of blacklisted containers to be extended with a parameter passed to the highlighter. A change to the logic was necessary because the blacklisted containers had not been properly implemented.

I've added a test to cover a passed-in blacklist parameter.


krampstudio:
Could you please describe the changes you did ?
Are they covered by the tests ?

views/js/ui/highlighter.js
```diff
@@ -36,7 +36,7 @@ define([
      * Children of those nodes types cannot be highlighted
      * @type {string[]}
      */
-    var containersBlackList = [
+    var _defaultBlackList = [
```
krampstudio: Even if you saw it somewhere, we had a discussion regarding this convention (to prefix private variable by an underscore) and we came up to the conclusion that's not what we wanted. Variable names should be explicit enough to not used this kind of convention (and here the name you've selected is sufficient : `defaultBackList`)
Could you please unprefix your variable.

views/js/ui/highlighter.js
```diff
@@ -47,11 +47,21 @@ define([
      * @param {Object} options
      * @param {Object} options.className - name of the class that will be used by the wrappers tags to highlight text
      * @param {Object} options.containerSelector - allows to select the root Node in which highlighting is allowed
+     * @param {Object} [options.containersBlackList] - additional blacklist selectors to be added to module instance's blacklist
+     * @param {String} [options.id] - key for identification/storage
```
krampstudio: This entry has been added by mistake ?

views/js/ui/highlighter.js
oatymart: ok, done.

views/js/ui/highlighter.js
oatymart: Was used for debugging, I've removed it.

PR #2042, opened 2019-03-14
Fix/tao 7512/tooltips on item authoring

handleman:
**Task** https://oat-sa.atlassian.net/browse/TAO-7512

**summary** tooltips are not shown randomly after test item being saved in back-office.

**What was fixed** Investigation took a bit longer than expected, but I found out, that the problem was in tooltip factory function, which has been taking its params in form of DOM element *by link* instead of *shallow copy* of original DOM element, so there was collision when Tooltip factory function was executed several times with same params given.

+ Proper unit test checking collision was added.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -49,7 +49,7 @@ return array(
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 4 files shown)


oatymart:
This seems fine. Good solution and tests pass.

PR #2052, opened 2019-03-22
Small phpdoc fix

siarhei-baradzin:
Add "|\JsonSerializable" to phpdoc

### DIFF
--- a/actions/class.CommonModule.php
+++ b/actions/class.CommonModule.php
@@ -223,7 +223,7 @@ abstract class tao_actions_CommonModule extends LegacyController implements Serv
     /**
      * Return json response.
      *
-     * @param array $data
+     * @param array|\JsonSerializable $data
      * @param int $httpStatus
      */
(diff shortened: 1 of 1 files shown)


actions/class.CommonModule.php
```diff
@@ -223,7 +223,7 @@ protected function getClientTimeout()
     /**
      * Return json response.
      *
-     * @param array $data
+     * @param array|\JsonSerializable $data
      * @param int $httpStatus
      */
     protected function returnJson($data, $httpStatus = 200)
```
julien-sebire: Why not also add array as typehint if we're sure we'll always receive an array?

siarhei-baradzin: The point is that $data could be not only array, but any type which can be encoded by json_encode(), and object implementing \JsonSerializable is one of common cases. This change  removes warnings in many places where we call returnJson() passing  \JsonSerializable

PR #2059, opened 2019-03-26
TAO-7830. Pass owner (current user) as a parameter to ImportByHandler…

alexeyvy:
https://oat-sa.atlassian.net/browse/TAO-7830.

This fix enables us to pass current user (sessionManager->getUser) to the task queue as a parameter. We want it because we can't let the task queue use sessions to retrieve users, as its context is unaware of those, which currently leads to the bug.

To put it simply, before creating a task, we take current session, remember its user ID and put it inside the task properties as a parameter that can simply be read in the context of background task instead of calling SessionManager->getUser() which is missing in the task queue context.


models/classes/task/ImportByHandler.php
```diff
@@ -52,6 +53,6 @@ public function __invoke($params)
 
         $this->propagate($importer);
 
-        return $importer->import($this->getClass($params[self::PARAM_PARENT_CLASS]), $params[self::PARAM_FORM_VALUES]);
+        return $importer->import($this->getClass($params[self::PARAM_PARENT_CLASS]), $params[self::PARAM_FORM_VALUES], $params[self::PARAM_OWNER]);
```
martijn-tao: Should the attached interface (`\tao_models_classes_import_ImportHandler`) be updated as well. The import method there is only using 2 params. Even though the changes in https://github.com/oat-sa/extension-tao-mediamanager/pull/136 should work fine with this, it might cause unexpected behaviour if the interface or other importers get changed at a later stage

PR #2061, opened 2019-03-28
Fix/tao 6125/misplaced class signature enforcer

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-6125

Due to a change made to support security signatures in the backoffice actions (see JIRA ticket), the `resources selector` was broken in any other places it is in use.
The reason why is a misplaced enforcer to ensure the existence of the signature for both the resource and its parent.

This PR aims to fix the issue by moving the enforcer in a proper place.

No unit test is impacted by the change.

To test the issue:
* go to the test authoring
* see the UI is broken somehow as it is impossible to add items to the test

To test the back office actions:
- check with and without the `resources selector`
- play with the actions in several places: items, tests, ...

To activate the `resources selector`:
* open the config file `config/tao/client_lib_config_registry.conf.php`
* add or modify the following lines:
```php
        'layout/tree/loader' => array(
            'treeProvider' => 'resource-selector'
        )
```

Expected behavior:
- the test authoring is working properly as before
- the back office actions still work as expected (create resources or classes, delete, move, copy, etc.)

~~Note: for now the class deletion still not work. So far I don't know if this is due to the frontend or the backend, so I still need to investigate this point.~~

Note 2: during the merge of #2033 the manifest file has been changed back from `Sprint98` to `Sprint97`, so I also include this little fix here.


jsconan:
Finally, I managed to fix the class deletion

PR #2063, opened 2019-04-02
Feature/tao 7762/Improve the handling of the last result in the calculator

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-7762

Rework the way to manage the last result variable.

Previously, to be able to properly support the sign change without having weird display, the result variable was replaced by it immediate value when the sign change was applied. Unfortunately, that introduced a loss of precision.

And when the expression was recorded in the history, the last result variable was also lost after having replaced it by its value. And again that introduced loss of precision (for instance: `1/3`, then `ans*3`, and the result was `0.999999999`).

In order to bring a better user experience, without loosing precision, a new approach has been suggested: always keep the variable in the expression, but display it properly with a clear distinction.

This PR implements a new display, the variable being displayed within parenthesis and with a different background color. When the expression is added to the history, the current variables are also stored in order to keep consistency in computation. And to improve the display, the variable is displayed rounded when necessary, with a ellipsis to show the internal precision is bigger than the displayed value.

A few changes have been made in the calculator engine:
- a new helper take care of the expression: `ui/maths/calculator/core/expression`. Most of its content has been moved from the `tokens` helper.
- the `sign` plugin has been updated to remove the special applied on the last variable
- the `remind` plugin has been reworked to take care of the context variables
- a fix has been applied on the tokenizer to better recognize the numbers (an issue has been found with exponential notation like in the number `42e12`), to do so 2 lexers are working together, one for the numbers and one for the whole expression
- the variables API is also more consistent now

Several unit tests are impacted by this PR:
- `/tao/views/js/test/ui/maths/calculator/scientificCalculator/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/board/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/expression/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/tokenizer/test.html`
- `/tao/views/js/test/ui/maths/calculator/core/tokens/test.html`
- `/tao/views/js/test/ui/maths/calculator/plugins/core/history/test.html`
- `/tao/views/js/test/ui/maths/calculator/plugins/modifiers/sign/test.html`
- `/tao/views/js/test/ui/maths/calculator/plugins/screen/simpleScreen/test.html`
- `/tao/views/js/test/util/mathsEvaluator/test.html`

To run them:
```
cd tao/views/build
npx grunt connect:dev qunit:single --test=${testPath}
```

Or in a browser: 
`http://YOUR_HOST/tao/views/js/test/ui/maths/calculator/scientificCalculator/test.html`


views/js/ui/maths/calculator/plugins/core/history.js
```diff
@@ -135,10 +153,7 @@ define([
                      * @event history
                      * @param {Array} history - The current history list
                      */
-                    calculator.trigger('history', _.pluck(history, 'value'));
-                })
-                .on(nsHelper.namespaceAll('destroy', pluginName), function () {
-                    reset();
+                    calculator.trigger('history', _.pluck(history, 'expression'));
```
ricproenca-tao: Why `._pluck`  instead of .`_map`?

jsconan: Indeed, `_.pluck` is not supported anymore in further versions of lodash. I will change it.

PR #2065, opened 2019-04-03
Fix/tao 7427/header menu styles bug

oatymart:
After the HTML change to the TAO BackOffice header menu in https://github.com/oat-sa/tao-core/pull/2048, one more line of SCSS is needed to keep our previous styles intact.

Specifically, the `background: inherit` property needs to be kept at each level of the hierarchy, otherwise its value will be unset at lower levels. Which looks like this:

Screen Shot 2019-04-03 at 16 59 20


ferenckiss89:
Looks good. Thank you

PR #2078, opened 2019-04-12
Feature/nex-41/item-preview

handleman:
**task**
https://oat-sa.atlassian.net/browse/NEX-41

**main PR**

https://github.com/oat-sa/extension-tao-testqti-previewer/pull/18

**short summary**

original feature required change inside /tao/views/js/ui/transformer.js

**description**

Y axis optional parameter was added to the `transformer.setTransformOrigin()` method, because of the need to  set that axis for mobile device emulator frame ( see original task).


views/js/ui/transformer.js
```diff
@@ -323,12 +323,13 @@ define([
          * Set the transformation origin to another value
          *
          * @param $elem
-         * @param value
+         * @param {Number} valueX
+         * @param {Number} [valueY]
          * @private
          */
-        setTransformOrigin: function($elem, value) {
+        setTransformOrigin: function($elem, valueX, valueY) {
```
krampstudio: Could you please add the method name : `setTransformOrigin: function setTransformOrigin($elem, valueX, valueY) {`

PR #2080, opened 2019-04-16
Update date range constraints when emptying the fields

krampstudio:
Related to https://oat-sa.atlassian.net/browse/TAO-8096

When the value is empty of the `ui/dateRange/dateRange` is empty, we update also the date picker constraints. Previously the min/max constraints were kept when removing the value.

This can be tested on the date filters from the "Export result" page, or in the "NCCER Test Report" section


krampstudio:
I've made some refactoring to support some changes made in the middle...
The tests have also been updated : 
```
cd tao/views/build
npx grunt connect:test qunit:single --test=/tao/views/js/test/ui/datetime/picker/test.html
npx grunt connect:test qunit:single --test=/tao/views/js/test/ui/dateRange/test.html
```

PR #2082, opened 2019-04-18
Fix/tao 8113/defensive approach for filesystem aware trait

siwane:


models/classes/taskQueue/Task/FilesystemAwareTrait.php
```diff
@@ -49,7 +49,11 @@ trait FilesystemAwareTrait
      */
     protected function saveFileToStorage($localFilePath, $newFileName = null)
     {
-        if (!file_exists($localFilePath)) {
+        if (is_array($localFilePath) && isset($localFilePath['path'])) {
```
martijn-tao: Since we can now handle an array as well, it would make sense to update the type hint for `$localFilePath` to something like `@param string|array $localFilePath`

PR #2085, opened 2019-04-23
Replace jqueryui dialog by ui/modal in extension manager

krampstudio:
Regression found in https://oat-sa.atlassian.net/browse/TAO-7149
 
 - The dependency to `jqueryui` wasn't declared by the module itself. It was working by chance (another module had to import it previously)
 - This PR brings a quick fix to use `ui/modal` instead of jqueryui's `dialog` 

How to reproduce : 
 - Ensure your instance is in `DEBUG_MODE` (in `config/generis.conf.php`)
 - Open the `Extension Manager` from `Settings`
 - Select extensions and click the `Install` button

What's expected : 
 - A modal popup is displayed, you can confirm or cancel the installation, follow up the installation progress

### DIFF
--- a/views/js/controller/settings/extensionManager.js
+++ b/views/js/controller/settings/extensionManager.js
@@ -22,11 +22,10 @@
 define([
     'jquery',
     'i18n',
-    'context',
-    'helpers',
+    'util/url',
     'ui/feedback',
     'ui/modal'
-], function($, __, context, helpers, feedback){
+], function($, __, urlUtil, feedback){
     'use strict';
 
     var ext_installed = [];
@@ -67,11 +66,11 @@ define([
         progressConsole(__('Installing extension %s...').replace('%s', ext));
         $.ajax({
             type: "POST",
-            url: helpers._url('install', 'ExtensionsManager', 'tao'),
+            url: urlUtil.route('install', 'ExtensionsManager', 'tao'),
             data: 'id='+ext,
             dataType: 'json',
-            success: function(data) {
-                helpers.loaded();
+            success: function success(data) {
+
                 if (data.success) {
                     progressConsole(__('> Extension %s succesfully installed.').replace('%s', ext));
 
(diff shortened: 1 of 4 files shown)


views/js/controller/settings/extensionManager.js
```diff
@@ -183,36 +217,33 @@ define(['jquery', 'i18n', 'context', 'helpers', 'ui/feedback'], function($, __,
...
+                $('[data-control=confirm]', $modalContainer).on('click', function(e){
+                    e.preventDefault();
+                    progressConsole(__('Preparing installation...'));
+                    $('.buttons', $modalContainer).remove();
+                    installError = 0;
+                    indexCurrentToInstall = 0;
+                    installNextExtension();
+                });
+
+                $modalContainer.modal({
```
oatymart: Any particular reason you've chosen `ui/modal` here rather than `ui/dialog` which has the required button templates and callback functionality built in?

views/js/controller/settings/extensionManager.js
```diff
@@ -1,4 +1,33 @@
...
+ * Copyright (c) 2013-2019 (original work) Open Assessment Technologies SA ;
+ */
+
+/**
+ * Extension manager controller
+ */
+define([
+    'jquery',
+    'i18n',
+    'context',
```
oatymart: I don't see `context` being used anywhere, is it needed?

views/templates/extensionManager/view.tpl
```diff
@@ -86,8 +86,18 @@ use oat\tao\helpers\Template;
...
-<div id="installProgress" title="<?= __('Installation...') ?>">
-    <div class="progress"><div class="bar"></div></div>
-    <p class="status">...</p>
-    <div class="console"></div>
+<div id="installProgress" class="modal" title="<?= __('Installation...') ?>">
+    <div class="modal-body">
+        <div class="progress"><div class="bar"></div></div>
+        <p class="status">...</p>
+        <div class="console"></div>
+        <div class="buttons rgt">
```
oatymart: See above comment about `ui/dialog`, I think it's mostly fine in this case because the page is a one-off, but the extra button templating could have been avoided.

views/js/controller/settings/extensionManager.js
krampstudio: Yes, the main reason is the amount of necessary refactoring to use `ui/dialog` which is a `ui/component` against `ui/modal` which still works as a jquery plugin by transforming an existing element to a modal.

PR #2090, opened 2019-04-26
[TAO-7936] WK-1280 PG - TCiaB - tool movement/resizing enhancement for touchscreen

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-7936
 
Improve movement/resizing behaviour of dynimacComponent, to increase user experience for touch screen devices users 
  
#### How to test
 
Create an ACT delivery with enabled calculator and scratchpad tools
Log in as a test taker
Run the delivery and try to resize/move calculator and scratchpad tools
   
#### Dependencies

Companion PR :
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1489
 - [ ] https://github.com/oat-sa/extension-tao-act/pull/1171
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1492
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1494


views/js/ui/dynamicComponent.js
```diff
@@ -359,11 +376,55 @@ define([
...
+                    } else if (width > config.maxWidth) {
+                        width = config.maxWidth;
+                    }
 
-                    var width  = e.rect.width < config.minWidth ? config.minWidth : e.rect.width > config.maxWidth ? config.maxWidth : e.rect.width;
-                    var height = e.rect.height < config.minHeight ? config.minHeight : e.rect.height > config.maxHeight ? config.maxHeight : e.rect.height;
+                    // make sure that component is inside the container
+                    if (elementOffset.left + width > parentOffset.left + $parent.width()) {
+                        if (config.proportionalResize) {
+                            return;
```
jsconan: In such complex code, in-the-middle `return` should be avoided.
Moreover, the code complexity should be reduced somehow. For instance, it could be splitted by smaller functions. One of the benefit would be to ease the unit testing.

views/js/ui/dynamicComponent.js
```diff
@@ -359,11 +376,55 @@ define([
...
+                    }
+
+                    // make sure that component is inside the container
+                    if (elementOffset.top + height > parentOffset.top + $parent.height()) {
+                        if (config.proportionalResize) {
+                            return;
+                        } else {
+                            height -= (height + elementOffset.top) - (parentOffset.top + $parent.height());
+                        }
+                    }
```
jsconan: Is it not possible to reduce the complexity of this code?
Having a simpler approach would be beneficial, and would be less error-prone.

handleman:
@jsconan  you said:
>I understand the UX required some visual changes. But wouldn't it be better to keep compatibility, by managing options?

What does it mean exactly? Should I refactor `dynamicComponent.js` and make it possible to switch between old and new design by passing some options? like `{theme: 'modern'}` ?

I propose to discuss with UX team do we really need to switch back to old version since it was not only about changing styles but solving some distinct problems which were in old version.

jsconan:
@handleman, regarding the compatibility issue I suggested in my review, after having discussed that point with the UX team it appears this is ok, the new design should be the reference now. So there is nothing to change here.

But the other comments are still accurate.

handleman:
@jsconan I have fixed all requested, could you do another round of review?

PR #2092, opened 2019-04-30
Feature/act 322 raise log messages level in worker

hutnikau:
Some of tasks stuck on production workers and never gets executed.
To improve stability of task queue and to facilitate investigation of support issues I added a few log messages and catching `Error`.

### DIFF
--- a/models/classes/taskQueue/Worker/AbstractWorker.php
+++ b/models/classes/taskQueue/Worker/AbstractWorker.php
@@ -79,7 +79,7 @@ abstract class AbstractWorker implements WorkerInterface
                 // execute the task
                 $taskReport = $task();
 
-                $this->logWarning('Task ' . $task->getId() . ' has been processed.', $this->getLogContext());
+                $this->logInfo('Task ' . $task->getId() . ' has been processed.', $this->getLogContext());
 
                 if (!$taskReport instanceof Report) {
                     $this->logWarning('Task ' . $task->getId() . ' should return a report object.', $this->getLogContext());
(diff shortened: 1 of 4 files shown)


models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -79,6 +79,8 @@ public function processTask(TaskInterface $task)
                 // execute the task
                 $taskReport = $task();
 
+                $this->logWarning('Task ' . $task->getId() . ' has been processed.', $this->getLogContext());
```
gyszucs: Is there any specific reason for using warning here?

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -90,6 +92,9 @@ public function processTask(TaskInterface $task)
             } catch (\Exception $e) {
                 $this->logError('Executing task ' . $task->getId() . ' failed with MSG: ' . $e->getMessage(), $this->getLogContext());
                 $report = Report::createFailure(__('Executing task %s failed', $task->getId()));
+            } catch (\Error $e) {
```
gyszucs: Have you tried this on PHP 5.6 also? Is it working fine?

hutnikau: We use polyfill for php 5:
https://github.com/oat-sa/tao-core/blob/master/composer.json#L55

It declares Error class for old php versions.

gyszucs: okay, good to know ;)

PR #2093, opened 2019-05-02
Fix/tao 8208 missaligned item per page selector

anshul119:
### Ticket: 

https://oat-sa.atlassian.net/browse/TAO-8208

### Summary

When External User logs in to the system and clicks on the Proctor Screen, on the right side of the page you can see a dropdown box with the text "25 items per page" and it's showing misaligned in the page.

### How to reproduce the issue?

Log in to the testing system as an external user and select a child organization (TU, ATEF, ATU). click the Proctor button. The page will display datatable with missaligned items per page selector

image

### Expected

image

### How did I fix the issue?

I changed the dom structure a bit. Instead of absolute positioning the page selector which is very prone to errors, I used flexbox.


views/scss/inc/_datatable.scss
```diff
@@ -1,4 +1,9 @@
 .datatable-wrapper {
+    .datatable-header {
```
krampstudio: Unfortunately, until _autoprefixer_ get integrated you still need to use this ugly `vendor-prefix` function, especially for IE:
```
@include vendor-prefix(display, flex, value, (-ms-, -webkit-, ''));
@include vendor-prefix(flex-direction, column, property, (-ms-, -webkit-, ''));
@include vendor-prefix(align-items, center, property, (-ms-, -webkit-, ''));
@include vendor-prefix(justify-content, space-between, property, (-ms-, -webkit-, ''));
```

views/js/ui/datatable/tpl/layout.tpl
```diff
@@ -7,17 +7,19 @@
...
     {{/if}}
 
-    {{#with options.status}}
-    <h2>
-        <span class="empty-list hidden">{{#if empty}}{{empty}}{{else}}{{../../options.labels.empty}}{{/if}}</span>
-        <span class="available-list hidden"><span class="text">{{#if available}}{{available}}{{else}}{{../../options.labels.available}}{{/if}}</span>: <span class="count">{{count}}</span></span>
-        <span class="loading">{{#if loading}}{{loading}}{{else}}{{../../options.labels.loading}}{{/if}}...</span>
-    </h2>
-    {{/with}}
+    <div class="datatable-header">
```
krampstudio: Could you please add one assertion into https://github.com/oat-sa/tao-core/blob/master/views/js/test/ui/datatable/test.js#L1235, to assert this wrapper exists, so if someone change the template the test will fail

views/scss/inc/_datatable.scss
anshul119: added

views/js/ui/datatable/tpl/layout.tpl
anshul119: added

scripts/update/Updater.php
```diff
@@ -1032,7 +1032,7 @@ public function update($initialVersion) {
             /** @var TokenStore $tokenStore */
             $tokenStore = $tokenService->getOption(TokenService::OPTION_STORE);
             $tokenStore->removeTokens();
-            $this->setVersion('34.0.0');
+            $this->setVersion('34.0.1');
```
krampstudio: If someone is in version `34.0.0` he will be stuck in limbs...
```
        $this->skip('31.1.0', '33.6.0');
        if ($this->isVersion('33.6.0')) {
             //...
             $this->setVersion('34.0.0');
        }
        $this->skip('34.0.0', '34.0.1');
```

views/scss/inc/_datatable.scss
```diff
@@ -1,4 +1,13 @@
...
+        @include vendor-prefix(display, flex, value, (-ms-, -webkit-, ''));
+        @include vendor-prefix(flex-direction, column, property, (-ms-, -webkit-, ''));
+        @include vendor-prefix(align-items, center, property, (-ms-, -webkit-, ''));
+        @include vendor-prefix(justify-content, space-between, property, (-ms-, -webkit-, ''));
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
```
krampstudio: The function `vendor-prefix` takes care of adding the version without prefixes.
So please remove them to prevent duplicates

anshul119: Perfect, fixed

scripts/update/Updater.php
anshul119: done

PR #2095, opened 2019-05-02
Fix/TAO-7112/media interaction disappears after changing video size

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-7112
 
Initial values of width and height is number, but after resizing they turned into string.
Was changed:
1. To checking `options.hasSize` added option that `options.width` and `options.height` can be string
2. Property `start` which is used  for noUiSlider was converted to number

 
#### Steps to reproduce
- Add Media interaction
- Add either uploaded video from Media or YouTube address (e.g. https://www.youtube.com/watch?v=Do3oFnaoYCM)
- Change video size
- Click "Done" button on the media interaction or click anywhere on the item authoring screen
- Click anywhere on the media interaction to activate "Interaction properties"

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -1035,6 +1035,6 @@ class Updater extends \common_ext_ExtensionUpdater {
             $this->setVersion('34.0.0');
         }
 
-        $this->skip('33.6.0', '33.6.1');
+        $this->skip('34.0.0', '33.6.1');
     }
 }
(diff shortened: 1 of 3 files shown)


scripts/update/Updater.php
```diff
@@ -1034,5 +1034,7 @@ public function update($initialVersion) {
             $tokenStore->removeTokens();
             $this->setVersion('34.0.0');
         }
+
+        $this->skip('33.6.0', '33.6.1');
```
Skyld: I think this is wrong, version 34.0.0 is already assigned above. Please check

PR #2097, opened 2019-05-06
Fix/tao 8180 items previously selected on brs

rkarolis:
Created `getVisibleNodes()` to get visible nodes, based on parent nodes status (more like a work around oposed by getting direct tree structured data).
Also on change nodesCount based on visible nodes, for switching the selectAll checkbox.

https://oat-sa.atlassian.net/browse/TAO-8180


views/js/ui/resource/selectable.js
```diff
@@ -229,11 +230,25 @@ define([
...
+            /**
+             * Select only all visible nodes.
+             * @returns {Object[]} nodes
+             */
+            selectVisible: function selectVisible(){
+                var $component = this.getElement();
+                var $elements = $component.find('[data-uri]').filter(function () {
+                    return $(this).parents('.closed').length === 0;
+                });
+                this.select(_.map($elements, function (element) {
```
IgorSalevan: Please, change the `_.map()' to `es6` `Array.prototype.map().

views/js/ui/resource/selectable.js
```diff
@@ -265,15 +280,14 @@ define([
...
-
             /**
              * Select all nodes.
              * @returns {selectable} chains
              * @fires selectable#change
              */
             selectAll : function selectAll(){
-                return this.select(_.keys(nodes));
-            },
+                return this.selectVisible();
```
krampstudio: I'd have kept those 2 methods separated (selectAll vs selectVisible)

views/js/ui/resource/selector.js
```diff
@@ -103,7 +103,7 @@ define([
         return _(resources)
             .filter({ type : nodeTypes.class })
             .map(function(resource){
-                var classNode = _.pick(resource, ['uri', 'label']);
+                var classNode = _.pick(resource, ['uri', 'label', 'signature', 'state']);
```
krampstudio: `signature` is not always there and is an artifact. The component has no knowledge of it.

views/js/ui/resource/tree.js
```diff
@@ -258,20 +258,28 @@ define([
...
-                            self.query({ classUri : $class.data('uri') });
-                        }  else {
-                            $class.removeClass('closed');
-                        }
+                    var node = self.getNode($class.data('uri'));
+
+                    if(!$class.children('ul').children('li').length){
+                        self.query({ classUri : $class.data('uri') });
                     }
+                    if(node) {
```
krampstudio: Could you please explain a bit the logic here ? 
Since the flow is `event` -> `query` -> `update` you can have here 2 updates (one after the query and one if the node is there)

views/js/ui/resource/tree.js
```diff
@@ -258,20 +258,28 @@ define([
...
 
                 /**
                  * Close a class node
                  * @param {jQueryElement} $class
                  */
                 var closeClass = function closeClass($class){
+                    var node = self.getNode($class.data('uri'));
+                    if(node) {
+                        node.state = 'closed';
+                        self.update([node]);
```
krampstudio: same than here, please check the data flow

views/js/ui/resource/tree.js
rkarolis: I added update change for node.state because it doesn't reflect closed status.

rkarolis: Otherwise the status of open classes will be closed. Only css class will reflect the status of class.

PR #2099, opened 2019-05-08
Fix/tao 7149 tmp dialog issue jquery file download

krampstudio:
The library `jquery.fileDownload` was using an illegal dependency to jquery-ui, based on some options.

This PR contains a quick fix, waiting to replace this library.

How to test : 
Try to download some results with or without this PR
image


oatymart:
Verified that this fixes the file download error in Results.

PR #2100, opened 2019-05-08
open eslint task to multiple report format

krampstudio:
Enable multiple reports format for the ESLint task  in TAO (those formats are expected for the CI)

How to test this pull request :  
```
cd views/build
npm run lint -- --extension=taoItems --format=checkstyle
npm run lint -- --extension=taoItems --format=json
```
This should have created the following reports : 
```
./views/build/reports/CHECKSTYLE-taoItems.xml
./views/build/reports/JSON-taoItems.json
```

The test reports have also been improved in order to contain meaningfull suite and test names (based on namespaces)
to test them :  
```
cd views/build
npx grunt connect:test qunit_junit taoitemstest
```
Check the reports `./views/build/reports/TEST-*`


views/build/grunt/test.js
```diff
@@ -63,19 +63,20 @@ module.exports = function(grunt) {
...
 
-                fileNamer : function(url){
+                fileNamer : function fileNamer(url){
                     return url
-                    .replace(testUrl + '/', '')
-                    .replace('/test.html', '')
-                    .replace(/\//g, '.');
+                        .replace(baseUrl + '/', '')
+                        .replace('views/js/test/', '')
+                        .replace('/test.html', '')
```
btamas: This solution looks a little bit unsafe for me. I prefer something like this: `path.dirname(path.relative('/view/js/test', new URL(baseUrl).pathname))`

btamas:
It works, but I have a small modification request to have cleaner and more comprehensible code.

views/build/grunt/test.js
krampstudio: Yes I've replaced some of the expression by pathname/dirname, but `relative` doesn't help in that situation (I need the first token of the path). 
I've then extracted the replacement in a function and used `split/join` instead.
I also figure out the module was missing in the reports so I fix it also

views/build/grunt/test.js
```diff
@@ -62,21 +115,15 @@ module.exports = function(grunt) {
...
-                        .replace('views/js/test/', '')
-                        .replace('/test.html', '')
-                        .replace(/\//g, '.');
+                fileNamer : url => testUrlToNamespace(url, '-'),
+                classNamer : (moduleName, url) => {
+                    //moduleName is a sentence so make it camelcase
+                    const moduleNameSuffix = moduleName
+                        .toLowerCase()
+                        .replace(/ (.)/g, match => match.toUpperCase())
+                        .replace( / /g, '' );
```
btamas: You can solve this with one `replace`.
`'foo bar baz'.replace(/ (.)/g, (fullMatch,firstMatch) => firstMatch.toUpperCase() )`;

PR #2106, opened 2019-05-10
[WIP] Fix/tao 8180 brs previously selected items

rkarolis:
Based on: https://github.com/oat-sa/tao-core/pull/2097

### DIFF
--- a/views/js/ui/resource/tree.js
+++ b/views/js/ui/resource/tree.js
@@ -13,7 +13,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 (original work) Open Assessment Technologies SA ;
+ * Copyright (c) 2017-2019 (original work) Open Assessment Technologies SA ;
  */
 
 /**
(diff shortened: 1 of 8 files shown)


views/js/ui/resource/tree.js
```diff
@@ -258,20 +258,28 @@ define([
...
-                            $class.removeClass('closed');
-                        }
+                    var node = self.getNode($class.data('uri'));
+
+                    if(!$class.children('ul').children('li').length){
+                        self.query({ classUri : $class.data('uri') });
                     }
+                    if(node) {
+                        node.state = 'open';
+                        self.update([node]);
```
krampstudio: Do you only need to update the state ? Because here you'll re-render the node and it's children .
If you only need the state to be updated, you can simply call 
```js
     node.state = 'open';
     $class.removeClass('closed')
               .addClass('open');
```
no ?

views/js/ui/resource/tree.js
```diff
@@ -258,20 +258,28 @@ define([
...
 
                 /**
                  * Close a class node
                  * @param {jQueryElement} $class
                  */
                 var closeClass = function closeClass($class){
+                    var node = self.getNode($class.data('uri'));
+                    if(node) {
+                        node.state = 'closed';
+                        self.update([node]);
```
krampstudio: Same here

PR #2112, opened 2019-05-15
Feature/action serializer

hutnikau:


hutnikau:
To be moved to generis ext.

PR #2117, opened 2019-05-16
Fix/tao 8315/disable csrf for messages

oatymart:
https://oat-sa.atlassian.net/browse/TAO-8315

The short term fix for this bug will be to disable CSRF tokens for the "messages" feature used by some customers.

I rewrote the polling request to use the new `core/request` module, so that the requests will be guaranteed sequential and it's easy to turn the tokens on/off later.

Also rewrote the unit test and made a necessary improvement to `core/request`.

Should be tested in the Test Runner on ACT and NFER envs.

### DIFF
--- a/views/js/core/communicator/poll.js
+++ b/views/js/core/communicator/poll.js
@@ -126,7 +126,7 @@ define([
                         dataType: 'json',
                         contentType: 'application/json',
                         sequential: true,
-                        noToken: true, // TODO: tokens disabled here as temporary fix for TAO-8315
+                        noToken: false,
                         timeout: config.timeout
                     })
                     .then(function(response) {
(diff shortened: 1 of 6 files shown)


views/js/core/communicator/poll.js
```diff
@@ -121,90 +118,51 @@ define([
...
-                            self.trigger('receive', response);
+                    coreRequest({
+                        url: config.service,
+                        method: 'POST',
+                        headers: headers,
+                        data: JSON.stringify(req),
+                        dataType: 'json',
+                        contentType: 'application/json',
+                        sequential: true,
+                        noToken: true, // TODO: tokens disabled here as temporary fix for TAO-8315
```
krampstudio: please consider using the token see https://github.com/oat-sa/extension-tao-testqti/pull/1501#issuecomment-493081733

oatymart: I've now re-enabled them based on your suggestion. Thanks!

PR #2121, opened 2019-05-21
Breaking/nex 89 integrate tao core sdk

btamas:
Integrate tao core sdk

Move libraries to other position:

core/areaBroker -> ui/areaBroker
core/validator -> ui/validator
core/taskQueue -> ui/taskQueue


btamas:
> * Could you please update the `views/package-lock.json` with  `@oat-sa/tao-core-sdk@0.1.1`
> * Could you please use the following [.Jenskinsfile](https://github.com/oat-sa/tao-core/blob/d9df766050c5f228822c819115fdd92ff8588d0c/.Jenkinsfile) because the docker image that run the CI now requires to have also `npm`

`package-lock` fixed
`.Jenkins` file fixed as well, but the problem was not the missing `npm`, instead `npm` didn't have right to create directory into `/`. I set the `HOME` to `.` with an environment variable.

PR #2122, opened 2019-05-21
Fix/tao 8041/modal window is not keyboard accessible

start-admin:
https://oat-sa.atlassian.net/browse/TAO-8041

**Steps to reproduce:**
- Open item authoring.
- Make some changes in item.
- Click on "Manage items" to see the modal window
- Try to use Tab to switch buttons
_or_
- Open test in _secure mode_
- Wait for the modal window to appear.
- Try to use Tab to switch buttons

**Expected result:** When the modal is loaded, the focus is placed on the first interactive element inside it. Buttons can be switched using Tab.

**Actual result:** When the modal is loaded, the focus does not placed on the first interactive element inside it. The close button is not available for Tab.

**cause of the problem:** The list of buttons available for Tab was created earlier than the close button and therefore the close button was not available for Tab.

**solution to the problem:** Changed the logic for creating a list of buttons available for Tab. The close button is created earlier and added to the list of buttons. Added new styles for elements in focus.


views/js/ui/dialog.js
```diff
@@ -411,8 +412,21 @@ define([
...
+                    disableClosing: this.disableClosing,
+                    disableEscape: this.disableEscape
+                }).on('closed' + _scope, function() {
+                    if (self.autoDestroy) {
+                        self.destroy();
+                    }
+                });
                 $buttons = this.$buttons.find('button');
 
+                $buttons.push($(_scope).find('.modal-close')[0]);
```
zagovorichev: what if there are no `.modal-close` at all?

views/js/ui/modal.js
```diff
@@ -100,7 +100,7 @@ define(['jquery', 'core/pluginifier', 'core/dataattrhandler'], function ($, Plug
 
                 //Initialize the close button for the modal dialog
                 if ($('.' + options.modalCloseClass, $modal).length === 0 && !options.disableClosing) {
-                    $('<div class="' + options.modalCloseClass + '"><span class="icon-close"></span></div>').appendTo($modal);
+                    $('<button class="' + options.modalCloseClass + '"><span class="icon-close"></span></button>').appendTo($modal).focus();
```
zagovorichev: so, I can see 2 options:
1. It's possible that we don't have a close button
2. that we have a close button but with any class name (not `.modal-close`)

views/scss/inc/_components.scss
```diff
@@ -206,11 +206,20 @@ ul.listbox {
...
         font-weight: bold;
-        color: $textColor;
+        color: $textColor !important;
+        background-color: transparent;
         @include font-size(20);
         &:hover{
             opacity: 0.6;
         }
+        &:focus{
+            box-shadow: 0 0 0px 3px #003664;
```
zagovorichev: we need a variable for that color `#003664`

zagovorichev: and why do we need px near the 0?

start-admin:
@zagovorichev could you check ?

views/js/ui/dialog.js
```diff
@@ -408,10 +408,27 @@ define([
...
+                    disableClosing: this.disableClosing,
+                    disableEscape: this.disableEscape
+                }).on('closed' + _scope, function() {
+                    if (self.autoDestroy) {
+                        self.destroy();
+                    }
+                });
+
                 $buttons = this.$buttons.find('button');
+                closeButton = $(_scope).find('#modal-close-btn')[0];
```
zagovorichev: one more time: are you sure that `modal-close-btn` exists? Element is not required

PR #2125, opened 2019-05-21
Fix/tao 7852/clipboard

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-7852

New Clipboard tool (to control it)


zagovorichev:
@start-admin Update

PR #2126, opened 2019-05-21
Fix/tao 7326/decode htmlchars in password

yuryfilippovich:


zagovorichev:
@yuryfilippovich Provide a full description and test case: why do we need this and for what you did it, please

yuryfilippovich:
@zagovorichev 
For example, on https://tciab-test.act.taocloud.org we have the password which contains & (1234& for example).
Steps:
1. Go to Remote Environments settings and set up this password, login, host .
2 Go to Data Synchronization and try to sync data. 
We will get 401 response from the remote server (https://tciab-test.act.taocloud.org).
This password 1234& save as 1234&amp;

zagovorichev:
@yuryfilippovich please, copy this to the very first comment of this PR (there is `No description provided.` now)

jbout (changes requested):
Not clear what the purpose is, no unit tests to highlight the issue

PR #2129, opened 2019-05-22
Feature/nex 93/monostyle switch

oatymart:
Frontend support for https://oat-sa.atlassian.net/browse/NEX-93

I'm tweaking the `ui/switch` component with a `monoStyle` parameter which colours the off state, making it more useful for toggling between two options of similar importance (not just disabled/enabled).

Without param:
switch-old
With param:
switch-monostyle

You can run the unit test in the browser to play with the result.


views/js/ui/switch/switch.js
```diff
@@ -67,6 +67,7 @@ define([
      * @param {Boolean} [config.off.active = true] - the default state
```
ferenckiss89: Update the license pls

views/js/test/ui/switch/test.js
```diff
@@ -259,4 +259,25 @@ define(['jquery', 'ui/switch/switch'], function($, switchFactory) {
                 ready();
```
ferenckiss89: update the license pls

PR #2133, opened 2019-05-23
Fix/nex 88 use random port for frontend tests

krampstudio:
Use a random port if not set for the Frontend tests.
This is mainly for CI when the same test suite runs multiple times on the same node (for example it runs once for the PR commit and one for the merge commit)

How to test: 
```
cd views/build
npm i
npx grunt connect:test taotest
```


views/build/Gruntfile.js
```diff
@@ -8,18 +29,18 @@ module.exports = function(grunt) {
...
-    var reportOutput        = grunt.option('reports') || 'reports';                     //where reports are saved
-    var buildOutput         = grunt.option('output')   || 'output';
-    var testUrl             = grunt.option('testUrl') || '127.0.0.1';                   //the port to run test web server, override with "--testPort value" to CLI if needed
-    var testPort            = parseInt(grunt.option('testPort'), 10) || 8082;                         //the port to run test web server, override with "--testPort value" to CLI if needed
-    var livereloadPort      = parseInt(grunt.option('livereloadPort'), 10) || true;     //the livereload port, override with "--livereloadPort 35729" to CLI if needed
-
-    var sassTasks   = [];
-    var bundleTasks = [];
-    var testTasks   = [];
+    const root             = path.resolve('../../../').replace(/\\/g, '/');      //tao dist root
```
btamas: `path` module resolves path based on the OS. So if it is macOs or linux, it will be `/` and `\\` in windows. I do not understand why `.replace/\\/g, '/')` is necessary. I do not have windows to test it.

views/build/Gruntfile.js
```diff
@@ -45,26 +65,26 @@ module.exports = function(grunt) {
 
     extensionHelper.getExtensions().forEach(function(extension){
 
-        var extensionKey = extension.toLowerCase();
-        var gruntDir = root + '/' + extension + '/views/build/grunt';
+        const extensionKey = extension.toLowerCase();
+        const gruntDir = path.join(root, extension, '/views/build/grunt');
```
btamas: I like that you used `path.join` to concatenate path, but it is not the proper way, because if this code will run in windows, it won't work because of `/` separators. It should be : `path.join(root, extension, 'view', 'build', 'grunt')`

krampstudio: I think path.join normalize the final path, ie. `path.join('foo', 'bar/baz')` creates `foo\bar\baz` on windows

views/build/Gruntfile.js
krampstudio: I'll check the history

krampstudio: This has been added by https://github.com/oat-sa/tao-core/pull/538 If I remember because the `root` variable is used in both paths and URLs 
For sure there is more elegant way to convert paths to URL

PR #2134, opened 2019-05-23
Fix/tao 8041/fixed qunit tests

zagovorichev:
Fixed qunit tests for the dialogs components


zagovorichev:
@jsconan Could you check this PR, I know that would be better to rewrite dialog as a component, but now it could be enough (as a quick fix)

PR #2136, opened 2019-05-23
Fix/TAO-8318/Proctor terminated session

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-8318

Bug related to these changes in request.js https://github.com/oat-sa/tao-core/commit/260e99238934420d583953d4e5fb1e7f2176e970#diff-2e6dd4d72bb11ed226ad289bbade954cR164

Request URL: https://nccersso.taocloud.org/taoQtiTest/Runner/move
Status Code: **200 OK**
Response in this case: 
```js
{**"success":false**,"type":"TestState","code":4,"message":"The assessment has been terminated. You cannot interact with it anymore.","messages":[{"channel":"teststate","message":{"type":"close","code":4,"message":"The assessment has been terminated. You cannot interact with it anymore. Please contact your proctor if required."}}]}
```

In case xhr.status 200 and `response.success === false` response will be put in new Error by function `createError`
https://github.com/oat-sa/tao-core/blob/a8dcde9b87d14e89a2f8db2422d907cbb31d6ec4/views/js/core/request.js#L181
https://github.com/oat-sa/tao-core/blob/a8dcde9b87d14e89a2f8db2422d907cbb31d6ec4/views/js/core/request.js#L62-L76

**and then in proxy message will not be send to channel teststate**
because of condition if (response.data && response.data.messages)
https://github.com/oat-sa/extension-tao-test/blob/06d14c4083fa9594d6dfa718ab80e4ff804a38c2/views/js/runner/proxy.js#L599-L611

on this step, we have such response:
image

and in testState plugin test will not be closed
https://github.com/oat-sa/extension-tao-testqti/blob/7a2fc0e7fd5f398249164b8a55e27d23b05ec318/views/js/runner/plugins/controls/testState/testState.js#L75-L84

**Solution 1**
**Case with code 200 and response.success === false is considered as a correct response** and promise will be resolved (like it was previously)
Revert changes done by https://github.com/oat-sa/tao-core/commit/260e99238934420d583953d4e5fb1e7f2176e970#diff-2e6dd4d72bb11ed226ad289bbade954cR164

Another fact for this solution: this case ( response.success=false ) will be handled in qtiServiceProxy https://github.com/oat-sa/extension-tao-testqti/blob/7a2fc0e7fd5f398249164b8a55e27d23b05ec318/views/js/runner/proxy/qtiServiceProxy.js#L98-L102

**Solution 2**
**Case with code 200 and response.success === false is considered as an error** and promise will be rejected (like it was now)
To change condition in proxy.js to process case with error, for example:
```
var messages = response.data && (response.data.messages || response.data.response && response.data.response.messages);
```

#### Steps to reproduce
1. Start a test and authorize by proctor
2. Pass some items and terminate session by proctor
3. Click next as TT

### DIFF
--- a/views/js/core/request.js
+++ b/views/js/core/request.js
@@ -172,7 +172,7 @@ define([
                                     return reject(createError(response, xhr.status + ' : ' + xhr.statusText, xhr.status, xhr.readyState > 0));
                                 }
 
-                                if (xhr.status === 200 || response && response.success === true) {
+                                if (xhr.status === 200 || ( response && response.success === true)) {
                                     // there's some data
                                     return resolve(response);
                                 }
--- a/views/js/core/request.js
+++ b/views/js/core/request.js
@@ -172,7 +172,7 @@ define([
                                     return reject(createError(response, xhr.status + ' : ' + xhr.statusText, xhr.status, xhr.readyState > 0));
                                 }
 
-                                if (response && response.success === true) {
+                                if (xhr.status === 200 || response && response.success === true) {
                                     // there's some data
                                     return resolve(response);
                                 }


oatymart:
I think you may be right. The funny thing is that we never checked `xhr.status === 200` in the old code (as far as I can tell). It was only present in an unreleased commit.

However, if we want the `{success: false}` response to be handled (rejected) in `qtiServiceProxy.js` like before, we do indeed need to resolve it in `core/request`.

So maybe the following would work:
```
if (response) {
    return resolve(response);
}
```
But I have not tested this yet. @krampstudio could you add your view?

krampstudio:
Sorry for the late response.

In term of error management we need to differentiate multiple things :  
 - Something went wrong, it's unexpected and we should (in theory) never have it. It's a 500 error code, with an error message, but this is a system failure. It's the worst situation for the user.
 - Handled or _controlled_ error, for precise situations like timeout, missing resource, authentication issue, an error code different, with status codes in the `20x` range or the `40x` range. Those cases can occur but hardly ever, and they are handled correctly by the system and the user experience is only affected a bit.
 - The 3rd type concern the user feedback : a wanted behavior, like in form validation, etc. This is part of the system, we handle it. It's not really an error but a behavior. The status code is 200 (everything is ok) but the response contains some informations to handle it correctly. This is the current case.

This middleware, and using the exception system to manage a real behavior **is an architecture mistake**. We shouldn't have implemented this behavior like this and using the exception system to handle a valid behavior is a very bad idea. Unfortunately this was to support the backend that throws an exception in those situations.

The ideal solution would be that the server send back something like : 
```
{
  "success" : true,
  "data" : {
      "testContext" : {
           "state" : 5 //I don't remember the code for terminated...
       }
   }
```
and to handle it by the test runner without all those complicated middlewares.

But to be safe, I'd suggest to use the Solution 1. 
@martijn-tao  Could you please if any of the other implementations liek `core/dataProvider/request` were also resolving the promise in case of 200 even with `success : false`

oatymart:
Previously, both `core/dataProvider/request` and `taoQtiTest/runner/proxy/qtiServiceProxy.js` (the 2 old generalised request makers I am aware of) both used the same basic logic:
```
if(response && response.success){
    return resolve(response);
}
else {
    return reject(response)
}
```
So `200 {success: false}` would have been rejected.

Now, both those modules call on `core/request` to make the request. Are we saying that `core/request` should now **resolve** the `200 {success: false}` so that its caller can then **reject** its Promise? Isn't the result the same?

krampstudio:
@martijn-tao https://github.com/oat-sa/tao-core/commit/260e99238934420d583953d4e5fb1e7f2176e970#diff-2e6dd4d72bb11ed226ad289bbade954cR164
shows that is was resolving it if it had either a status at  `200` or `success : true`

PR #2141, opened 2019-05-24
Bugfix/tao 8224/fix generis tree

anshul119:
concerned ticket: https://oat-sa.atlassian.net/browse/TAO-8224

### Issue

There is a miscalculation in the number of displayed class instances and current position in `tao/views/js/generis.tree.select.js`

It does not takes into account the number of classes when updating the position in the tree when using the 'next 10' pagination button.

### Fix

I calculate the number of classes in the array of class instances and subtract it before updating the position.


views/js/generis.tree.select.js
```diff
@@ -269,23 +269,34 @@ define(['jquery', 'lodash', 'i18n', 'context', 'generis.tree', 'helpers', 'ui/fe
...
                     DATA[i].attributes['class'] = instance.options.instanceClass+" node-instance node-draggable";
                     if (!$('#'+DATA[i].attributes['id'], $(TREE_OBJ.container)).length) TREE_OBJ.create(DATA[i], TREE_OBJ.get_node(NODE[0]));
                     // If the check all options. Add the incoming nodes to the list of node to check
                     if (options.checkedNodes === "*") {
                         instance.checkedNodes.push(DATA[i].attributes.id);
                     }
                 }
+
+                // Counting the number of classes in the data array
+                for (j; j < DATA.length; j++) {
```
btamas: `DATA.reduce((amount, {type}) => type === 'class' ? amount+1 : amount, 0)` or
`DATA.filter(({type}) => type === 'class').length`

views/js/generis.tree.select.js
```diff
@@ -269,23 +269,34 @@ define(['jquery', 'lodash', 'i18n', 'context', 'generis.tree', 'helpers', 'ui/fe
...
+
                 if(instance.checkResourcePermissions){
                     DATA = instance.convertDataWithPermissions(DATA);
                 }
                 //Hide paginate options
                 instance.hidePaginate(NODE, TREE_OBJ);
                 //Display incoming nodes
-                for (var i=0; i<DATA.length; i++) {
+                for (i; i<DATA.length; i++) {
                     DATA[i].attributes['class'] = instance.options.instanceClass+" node-instance node-draggable";
```
siarhei-baradzin: If you already have this loop you can revert all other changes and do not iterate by DATA again, simply
```js
countClass += DATA[j].type === 'class'; 
```
in existing loop

siarhei-baradzin:
1. Looks like this fix will work, but did you understand why in `generis.tree.js:122` we do exactly the same calculations of position (substracting classes count) and save them to meta `position` too, but looks like they don't affect to future pagination (different/wrong nodes?), don't we have some kind of logic duplication?
2. Did you consider to use layout/tree if generis.tree is deprecated?

views/js/generis.tree.select.js
anshul119: Good catch, we don't really need this new loop.

views/js/generis.tree.select.js
anshul119: will not need this anymore :')

views/js/generis.tree.select.js
```diff
@@ -269,23 +269,28 @@ define(['jquery', 'lodash', 'i18n', 'context', 'generis.tree', 'helpers', 'ui/fe
             options = $.extend(options, pOptions);
 
             $.post(this.dataUrl, options, (function(instance) {return function(DATA) {
+                var countClass = 0;
+                var i = 0;
```
siarhei-baradzin: why did you decide to move the definition of `i` variable outside the loop?

anshul119: It's a concept in javascript called hoisting. All the variables should be moved on top of the current scope.

krampstudio: @siarhei-baradzin it's also to match the coding guidelines (ESLint should have reported it to you)

anshul119:
> 1. Looks like this fix will work, but did you understand why in `generis.tree.js:122` we do exactly the same calculations of position (substracting classes count) and save them to meta `position` too, but looks like they don't affect to future pagination (different/wrong nodes?), don't we have some kind of logic duplication?
> 2. Did you consider to use layout/tree if generis.tree is deprecated?

1. So, extractMeta function inside generis.tree.js runs only on the initial request. When you click on paginate button the paginateInstances function runs in generis.tree.select.js and not the other one.

2. Layout tree looks very different in implementation than the generis.tree. I feel it would be a big change to make the whole refactoring which might affect multiple clients.

krampstudio:
Just a side node for the next PR, please be careful to the branch naming (`bugfix` instead of `fix`)

views/js/generis.tree.js
```diff
@@ -200,13 +200,11 @@ define(['jquery', 'i18n', 'class'], function($, __, Class) {
 		},
 
 		/**
-		 * Hide paginate options
-		 * @param NODE
-		 * @param TREE_OBJ
-		 * @private
+		 * Hide pagination
+		 * @param nodeId
```
krampstudio: Could you please document the type you are expecting  : 
```js
/**
 * Hide the pagination
 * @param {String} nodeId - the identifier of the pagination node
 */

File: views/js/generis.tree.js [hunk 5]
```diff
@@ -200,13 +200,11 @@ define(['jquery', 'i18n', 'class'], function($, __, Class) {
...
 		 */
-		hidePaginate: function(NODE, TREE_OBJ) {
-			$(NODE).find('.paginate').each(function(){
+		hidePaginate: function(nodeId) {
```
krampstudio: Could you please use named function expression : `hidePaginate: function hidePaginate(nodeId) {`
I know it's old code, but the coding guidelines are to fix what we change.

views/js/generis.tree.js
anshul119: added

views/js/generis.tree.js
anshul119: agreed :)

PR #2147, opened 2019-05-27
Fixed float rounding issue in unit test.

julien-sebire:
This PR fixes the test for Token timestamp float rounding. Although there are other tests failing, Jenkins integration shows us that this one is passing.


julien-sebire:
Replaced by https://github.com/oat-sa/tao-core/pull/2174.

PR #2148, opened 2019-05-28
FIX/TAO-8396/Secure-mode-doesn't-work

start-admin:
https://oat-sa.atlassian.net/browse/TAO-8396
https://oat-sa.atlassian.net/browse/TAO-8397

**Steps to reproduce**:

Open test in secure mode
Wait for the modal window to appear.

**Expected result**: The test is launch in secure mode successfully.

**Actual result**: Error message 'Error: Cannot read property 'shortcut' of undefined' is appeared, an user is not able to proceed the test.

**reason**: change logic for tests. https://github.com/oat-sa/tao-core/pull/2134/files


krampstudio:
please check if the tests are still running

start-admin:
@krampstudio tests is OK . already checked.

PR #2149, opened 2019-05-28
Fix/tao 8411/charset dbal mysql

tikhanovichA:


tikhanovichA:
Moved all logic with default charset to generis

PR #2153, opened 2019-05-29
FIX/TAO-8396/Secure-mode-doesn't-work

start-admin:
https://oat-sa.atlassian.net/browse/TAO-8396
https://oat-sa.atlassian.net/browse/TAO-8397

**Steps to reproduce**:

Open test in secure mode
Wait for the modal window to appear.

**Expected result**: The test is launch in secure mode successfully.

**Actual result**: Error message 'Error: Cannot read property 'shortcut' of undefined' is appeared, an user is not able to proceed the test.

**reason**: change logic for tests. https://github.com/oat-sa/tao-core/pull/2134/files


scripts/update/Updater.php
```diff
@@ -1063,7 +1063,7 @@ public function update($initialVersion) {
                 $rdf->remove($triple);
                 $rdf->add($triple);
             }
-            $this->setVersion('35.8.2');
+            $this->setVersion('35.8.3');
```
zagovorichev: sorry, you can't do it like that, please use `skip` method

zagovorichev:
Now I don't see a focus on the close button by default

views/js/ui/dialog/alert.js
```diff
@@ -32,12 +32,13 @@ define(['lodash', 'ui/dialog'], function (_, dialog) {
...
             message: message,
             buttons: 'ok',
-            autoDestroy: true
+            autoDestroy: true,
+            autoRender: true
         }).on('create.dialog', function(){
             if (onCreateDialog) {
                 onCreateDialog();
             }
-        }).render();
```
zagovorichev: Now I don't see a focus on the close button by default

SashaTAO:
A lot of unit tests are failed in this branch - http://qa.kitchen.it.taocloud.org/blue/organizations/jenkins/Github%20Pull%20Requests%20for%20oat-sa%2Ftao-core/detail/FIX%2FTAO-8396%2FSecure-mode-doesnt-work/8/pipeline/48

Was it caused by these changes?

PR #2158, opened 2019-06-03
Breaking/nex 90 extract ui modules

btamas:
Remove UI modules and extract them to https://github.com/oat-sa/tao-core-ui-fe

Jira: https://oat-sa.atlassian.net/browse/NEX-90

Rename `ui.js` to `uikitLoader.js`

Dependency:
- [x] https://github.com/oat-sa/tao-core-ui-fe/pull/1
- [x] https://github.com/oat-sa/tao-core-ui-fe/pull/2
- [x] add released npm package version into package.json

Related:
- [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1516
- [ ] https://github.com/oat-sa/extension-tao-devtools/pull/161
- [x] https://github.com/oat-sa/extension-tao-datauri/pull/27


krampstudio:
I did a quick fix on the case of `uikitLoader` vs `uikitloader` which will works on mac and windows but not if the file is served on linux (case sensitive fs)

.Jenkinsfile
```diff
@@ -89,7 +89,7 @@ mkdir -p tao/views/locales/en-US/
                 stage('Frontend Tests') {
                     agent {
                         docker {
-                            image 'alekzonder/puppeteer'
+                            image 'btamas/puppeteer-git'
```
ferenckiss89: @krampstudio should not we put these under oat-sa or somewhere that is more general than specific git users?

krampstudio: yes,  those images are only temporary (we will even have our own registry with the correct images once stables)

PR #2159, opened 2019-06-03
file_path is now optional in web installer.

julien-sebire:
We move the other way around: make the "file_path" parameter optional in both CLI and web installer.
To test it, try to install without specifying a path in the last left box on the second screnn of installer (see attachment in https://oat-sa.atlassian.net/browse/TAO-8162).


chazotte:
The only issue that I faced was related to permissions not properly checked for the data folder.
If I create a new data folder as sudo (with no chown to www-data), it will pass but will fail at the very end of the install.

PR #2161, opened 2019-06-03
Fix/puppeteer bridge timeout

krampstudio:
It looks like the api has changed and the migration between qunit 1 and 2 didn't handle that point.
This PR brings back timeouts in test suites


views/build/config/chrome-bridge.js
```diff
@@ -79,11 +79,11 @@
     /**
      * The test case is done
      */
-    QUnit.testDone(logs => {
-        const testName = logs.name;
-        if(runningTestsTimeouts.has(testName)){
-            clearTimeout(runningTestsTimeouts.get(testName));
-            runningTestsTimeouts.delete(testName);
+    QUnit.testDone(function(logs) {
```
anshul119: I'm wondering why arrow function cant be used here, any reason?

PR #2167, opened 2019-06-06
PR to adapt notification and task log table schema to spanner.

julien-sebire:
This seems to be the schema used for installation, instead of the on in oat-sa/generis/core/kernel/persistence/smoothsql/install/SmoothRdsModel.php.
(thanks @bugalot for pointing this out).


install/utils/class.DbalDbCreator.php
```diff
@@ -134,47 +134,44 @@ public function listDatabases(){
...
-     * @author "Lionel Lecaque, <lionel@taotesting.com>"
+     * @author 'Lionel Lecaque, <lionel@taotesting.com>'
      */
     private function createStatementsSchena(){
-    	$table = $this->schema->createTable("statements");
-    	$table->addColumn("modelid", "integer",array("notnull" => true,"default" => 0));
-    	$table->addColumn("subject", "string",array("length" => 255,"default" => null));
-    	$table->addColumn("predicate", "string",array("length" => 255,"default" => null));
+    	$table = $this->schema->createTable('statements');
+        $table->addColumn('id', 'string', ['length' => 25, 'notnull' => true]);
```
ekkinox: space to remove, beginning of line

install/utils/class.DbalDbCreator.php
```diff
@@ -134,47 +134,44 @@ public function listDatabases(){
...
-//     	$table->addColumn("stread", "string",array("length" => 255,"default" => null,"notnull" => false));
-//     	$table->addColumn("stedit", "string",array("length" => 255,"default" => null,"notnull" => false));
-//     	$table->addColumn("stdelete", "string",array("length" => 255,"default" => null,"notnull" => false));
-    	$table->setPrimaryKey(array("id"));
-    	$table->addOption('engine' , 'MyISAM');
-    	$table->addColumn("epoch", "string" , array("notnull" => null));
+        $table->addColumn('l_language', 'string', ['length' => 255]);
+
+    	$table->addColumn('author', 'string', ['length' => 255]);
+        $table->addColumn('epoch', 'string' , ['notnull' => true]);
```
ekkinox: space to remove, beginning of line

install/class.Installator.php
```diff
@@ -181,7 +184,16 @@ public function install(array $installData)
 				$dbConnectionParams['portability'] = \Doctrine\DBAL\Portability\Connection::PORTABILITY_ALL;
 				$dbConnectionParams['fetch_case'] = PDO::CASE_LOWER;
 			}
-				
+            if($installData['db_driver'] == 'gcp-spanner') {
+                $dbConnectionParams = [
+                    'dbname' => $installData['db_name'],
+                    'instance' => $installData['db_host'],
+                    'driverClass' => SpannerDriver::class,
+//                    'wrapperClass' => SpannerConnection::class,
```
SergiiTao: do we need to keep this comment?

install/utils/class.DbalDbCreator.php
```diff
@@ -63,24 +54,47 @@ public function __construct($params){
...
+        if (isset($this->dbConfiguration['driver'])) {
+            return $this->dbConfiguration['driver'];
+        }
+
+        $driverNames = [
+            SpannerDriver::class => 'gcp-spanner',
+        ];
+
+        if (isset($this->dbConfiguration['driverClass'])) {
+            $driverClass = $this->dbConfiguration['driverClass'];
```
SergiiTao: do we need to check if this class exists?

install/utils/class.DbalDbCreator.php
```diff
@@ -63,24 +54,47 @@ public function __construct($params){
...
+
     /**
      * @author "Lionel Lecaque, <lionel@taotesting.com>"
      * @param string $dbName
      */
     public function dbExists($dbName){
         $sm = $this->getSchemaManager();
-		common_Logger::d('Check if database with name \'' .$dbName. '\' exists for driver ' . $this->dbConfiguration['driver']);
-        if($this->dbConfiguration['driver'] == 'pdo_oci'){
+		common_Logger::d('Check if database with name \'' .$dbName. '\' exists for driver ' . $this->driverName);
```
SergiiTao: `sprintf('Check if database with name "%s" exists for driver %s', $dbName, $this->driverName)`
looks more readable

install/utils/class.DbalDbCreator.php
```diff
@@ -129,57 +143,54 @@ public function listDatabases(){
...
+    		common_Logger::d('driver is ' . $this->driverName);
+    	   	if($this->driverName != 'pdo_sqlsrv'
+    	   			&& $this->driverName != 'pdo_oci'){
+    	   		$table->addIndex(['predicate', 'object'], 'k_po');
     	   	}
-    	} 	
+    	}
 
-    
+        $table->addOption('engine' , 'MyISAM');
```
SergiiTao: it should be inside `if($this->driverName != 'pdo_mysql'){` block. no?

SergiiTao: i see we have `createKeyValueStoreTable` function. should it be used here?

install/utils/class.DbalDbCreator.php
```diff
@@ -222,58 +233,69 @@ private  function buildSchema(){
...
      */
     public function initTaoDataBase(){
     	$platform = $this->connection->getDatabasePlatform();
     	$queries = $this->schema->toSql($platform);
     	foreach ($queries as $query){
-    	   	$this->connection->executeUpdate($query);
+    	   	$this->connection->exec($query);
     	}
-    	if($this->dbConfiguration['driver'] == 'pdo_mysql'){
+    	if($this->driverName == 'pdo_mysql'){
```
SergiiTao: I see that we check it multiple times. does it make sense to create a dedicated function for this check?

models/classes/notification/implementation/RdsNotification.php
```diff
@@ -215,6 +216,13 @@ public function notificationCount($userId)
...
-}
\ No newline at end of file
+    /**
+     * Generates a unique, not auto-increment based, primary key.
+     *
+     * @return string
+     */
+    public function getUniquePrimaryKey()
+    {
+        return strrev(uniqid('', true));
```
SergiiTao: why do we need strrev here?

install/utils/class.DbalDbCreator.php
julien-sebire: > it should be inside `if($this->driverName != 'pdo_mysql'){` block. no?
I think the `engine` option applies to mysql only, so it should be in a `else` instead...

> i see we have createKeyValueStoreTable function. should it be used here?
I don't think it's the same matter here. `createKeyValueStoreTable` is about key value storage, and this is about triple storage, right?

install/utils/class.DbalDbCreator.php
julien-sebire: You're right, this whole file should be refactored. But it's not the time for this in my opinion. The very goal of all this is to confirm that we can run tao with a spanner database, and see how it performs...

models/classes/notification/implementation/RdsNotification.php
julien-sebire: The use of strrev allows use to have non-consecutive uuids (base on currenttime) to be given to spanner as primary key of the table, on which spanner bases the spreading over nodes.
Without the strrev, we would end up with sequenced PK and all would be saved to the same spanner node, creating hotspot and loosing the sharding feature of spanner.

install/utils/class.DbalDbCreator.php
```diff
@@ -129,57 +143,54 @@ public function listDatabases(){
...
      * @author "Lionel Lecaque, <lionel@taotesting.com>"
      */
     private function createModelsSchema(){
-	    $table = $this->schema->createTable("models");
-	    $table->addColumn('modelid', "integer",array("notnull" => true,"autoincrement" => true));
-	    $table->addColumn('modeluri', "string", array("length" => 255,"default" => null));
-	    $table->addOption('engine' , 'MyISAM');
-	    $table->setPrimaryKey(array('modelid'));
+	    $table = $this->schema->createTable('models');
+	    $table->addColumn('modelid', 'string', ['length' => 25, 'notnull' => true]);
```
ekkinox: again, 23 since uniqid('', true) length = 23

models/classes/notification/implementation/RdsNotification.php
```diff
@@ -78,10 +78,11 @@ public function sendNotification(NotificationInterface $notification)
         $platform = $this->getPersistence()->getPlatForm();
 
         $sqlQuery    = 'INSERT INTO ' . self::NOTIF_TABLE .
-                        ' (' . $this->getAllFieldString() . ') 
+            ' (' . self::NOTIF_FIELD_ID . ' , ' . $this->getAllFieldString() . ') 
                             VALUES ( ? , ? , ? , ? , ? , ? , ? , ? )';
```
ekkinox: for later / safety: we can generate the `?, ?, ...` string based on number of fields in getAllFieldString (avoid surprises)

models/classes/notification/implementation/RdsNotification.php
ekkinox: fair node dispatch of data on spanner

models/classes/taskQueue/TaskLog/Broker/RdsTaskLogBroker.php
```diff
@@ -257,7 +260,7 @@ public function search(TaskLogFilter $filter)
 
             $filter->applyFilters($qb);
 
-            $collection = TaskLogCollection::createFromArray($qb->execute()->fetchAll());
+            $collection = TaskLogCollection::createFromArray($qb->execute()->fetchAll(), $this->getPersistence()->getPlatForm()->getDateTimeTzFormatString());
```
ekkinox: can reuse already got `$platform ` here as `$platform-> getDateTimeTzFormatString()`

models/classes/notification/implementation/RdsNotification.php
peetya: I think an additional `?` is missing from here.

install/class.Installator.php
```diff
@@ -181,7 +184,16 @@ public function install(array $installData)
 				$dbConnectionParams['portability'] = \Doctrine\DBAL\Portability\Connection::PORTABILITY_ALL;
 				$dbConnectionParams['fetch_case'] = PDO::CASE_LOWER;
 			}
-				
+            if($installData['db_driver'] == 'gcp-spanner') {
```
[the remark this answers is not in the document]
julien-sebire: Fixed.

install/class.Installator.php
julien-sebire: Right! Removed! ;)

models/classes/notification/implementation/RdsNotification.php
julien-sebire: Yes, or even better, use the insert method with a key=>value array... Done!

models/classes/taskQueue/TaskLog/Broker/RdsTaskLogBroker.php
julien-sebire: It's not in the same method actually, but it's ok, it's more readable. Done.

install/utils/class.DbalDbCreator.php
julien-sebire: The idea here is that DBAL asks for EITHER the `driver` OR the `driverClass` parameter. It first tests if `driver` exists, in which case it returns a standard dbal driver, referenced by its string name. It then tries to find a driver class for an exotic driver. I used an array to allow the function to handle another custom driver the same way. If you think it's overkill, I can get your simplification...

install/utils/class.DbalDbCreator.php
julien-sebire: Ok, done.

install/utils/class.DbalDbCreator.php
```diff
@@ -63,24 +54,47 @@ public function __construct($params){
...
+            }
+        }
    		catch(Exception $e){
    			$this->connection = null;
             common_Logger::e($e->getMessage() . $e->getTraceAsString());
    			throw new tao_install_utils_Exception('Unable to connect to the database ' . $params['dbname'] . ' with the provided credentials: ' . $e->getMessage());
    		}
    	}
 
+   	protected function findDriverName() {
```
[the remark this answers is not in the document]
julien-sebire: Yeah right. And also phpdoc.

install/utils/class.DbalDbCreator.php
julien-sebire: Done

scripts/install/InstallNotificationTable.php
```diff
@@ -44,15 +44,15 @@ public function __invoke($params)
...
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_STATUS       , "integer"  ,array("default" => 0 , "notnull" => false,"length" => 255));
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_TITLE        , "string"   ,array("length" => 255));
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_MESSAGE      , "text"     ,array("default" => null));
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_SENDER       , "string"   ,array("default" => null , "notnull" => false,"length" => 255));
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_SENDER_NANE  , "string"   ,array("default" => null , "notnull" => false,"length" => 255));
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_CREATION     , "datetime" ,array("notnull" => true));
-            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_UPDATED      , "datetime" ,array("notnull" => true));
+            $queueTable->addColumn(RdsNotification::NOTIF_FIELD_ID           , 'string'   , ['length' => 25, 'notnull' => true,]);
```
ekkinox: imo its 23

jbout:
Please split into:
* notification service
* modelfactory

do not use ServiceManager::getServiceManager() as it breaks the ability to mock classes
augment the version number in manifest and provide update script

julien-sebire:
PR split into:
https://github.com/oat-sa/tao-core/pull/2305 (version bump)
https://github.com/oat-sa/tao-core/pull/2300
https://github.com/oat-sa/tao-core/pull/2301
https://github.com/oat-sa/tao-core/pull/2306
https://github.com/oat-sa/tao-core/pull/2307

PR #2178, opened 2019-06-12
Refactor/cleanup sas

jbout:


manifest.php
```diff
@@ -53,7 +53,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '36.3.1',
+    'version' => '36.3.2',
```
krampstudio: Considering one deprecation and the change of the `abstract` method I'd recommand updating the minor version `36.4.0`, no ?

PR #2181, opened 2019-06-14
Fix/tao 8461/russian locale

tikhanovichA:
https://oat-sa.atlassian.net/browse/TAO-8461


tikhanovichA:
https://github.com/oat-sa/tao-core/pull/2183

PR #2187, opened 2019-06-19
trigger change on clearing picker field

anshul119:
Ticket: https://oat-sa.atlassian.net/browse/TAO-8477

Currently if you clear the date picker manually with keyboard, it does not trigger change and the data table is not refreshed.

I fixed it by triggering change event in any case the datepicker is changed.


krampstudio:
I'll close this PR since your changes should have been made against https://github.com/oat-sa/tao-core-ui-fe

PR #2194, opened 2019-06-26
set default header to self

hutnikau:
Looks like on some environments default value of `CSP_HEADER_SETTING` was not set.

Also `SetDefaultCSPHeader` script sets it's value to `'self'` but in the updater script (line 1001) it's set to `*` but looks like it should be `'self'`

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -53,7 +53,7 @@ return array(
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '37.1.1',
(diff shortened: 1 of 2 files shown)


martijn-tao:
@hutnikau regarding this comment:

> Also SetDefaultCSPHeader script sets it's value to `'self'` but in the updater script (line 1001) it's set to `*` but looks like it should be `'self'`

This was done on purpose. We wanted to use `*` for existing environments so that it wouldn't break the current functionality. For new environment we want to use `'self'`however, since this is better in terms of security

PR #2195, opened 2019-06-26
Feature/NEX-91/Scanner to fetch npm paths from extensions for grunt test task

oatymart:
This helper is potentially useful for any extension which will be partially vendorised. Often there will be unit tests in the `js/test` folder which rely on code in a `views/node_modules/@oat-sa/**/dist` folder. The requirejs paths need to be defined to make that work, and it would be best not to do that centrally.

With this PR, an extension can contain an optional file `views/build/grunt/paths.json`
Example contents:
```
{
    "taoQtiItem/qtiCommonRenderer": "../../../taoQtiItem/views/node_modules/@oat-sa/tao-item-runner-qti/dist/qtiCommonRenderer",
    "taoQtiItem/qtiItem"          : "../../../taoQtiItem/views/node_modules/@oat-sa/tao-item-runner-qti/dist/qtiItem",
    "taoQtiItem/qtiRunner"        : "../../../taoQtiItem/views/node_modules/@oat-sa/tao-item-runner-qti/dist/qtiRunner",
    "taoItems/assets"             : "../../../taoItems/views/node_modules/@oat-sa/tao-item-runner/dist/assets"
}
```
These are a copy of the same paths that will be loaded into the client config registry under a normal TAO install.
The paths will be scanned in and added to the requirejs paths when a command like `npx grunt connect:dev taoqtiitemtest` is run.


views/build/tasks/helpers/extensions.js
```diff
@@ -43,23 +43,43 @@ module.exports = function(grunt, root){
...
+                }
+            });
+            return paths;
+        },
 
+        // parse a 'paths.json' file in each extension, and if it exists,
+        // append its contents to a flat object
+        getExtensionsExtraPaths : function getExtensionsExtraPaths(extensions){
+            var self = this;
+            var npmPaths = {};
```
jsconan: I'm a little confused here. Some part of the code is using ES6, and some is still in ES5. Not sure ES6 is fully allowed in grunt tasks.
However, if this is ok why not applying the same flavor everywhere? (at least in the new code)

example:
```javascript
getExtensionsExtraPaths(extensions) {
    const npmPaths = {};
    extensions = extensions || self.getExtensions(true);
    extensions.forEach(extension => {
        try {
            const paths = require(`${this.getExtensionPath(extension)}/views/build/grunt/paths.json`);
            Object.entries(paths).forEach(([key, value]) => npmPaths[key] = value);
        }
        catch (e) {
            // file doesn't exist: ok
        }
    });
    return npmPaths;
}
```

ferenckiss89: Since it is run by node the es6 syntax is allowed inside but not for export and import. I'm suggesting to change only the parts which we included now and changed.

views/build/tasks/helpers/extensions.js
```diff
@@ -43,23 +43,43 @@ module.exports = function(grunt, root){
...
+            extensions = extensions || self.getExtensions(true);
+            extensions.forEach(function(extension){
```
btamas: I do not like the `npmPaths` variable name, because later it can contain any custom or extra paths.

```
return extensions.reduce((customPaths, extension) => {
  try {
    return {...customPaths, ...require(path.join(this.getExtensionPath(extension), 'views', 'build', 'grunt', 'paths.json'))};
  } catch(e) {
    return customPaths;
  }
}, {});
```

views/build/grunt/test.js
```diff
@@ -134,14 +134,19 @@ module.exports = function(grunt) {
                 port: testPort,
                 base: root,
                 middleware: function(connect, options, middlewares) {
-
+                    var extraPaths;
```
krampstudio: you can use the ES2015 syntax if you wish, grunt runs over node and we support version > 8.12.0 that are compliant with our coding guide

krampstudio:
Ok for me.
@jsconan  this approach is only used for testing and bundling. At runtime the dynamic paths are served by the lib registry since you can't define in tao-core paths to other extensions.
The ideal solution would be refactor everything (all modules names everywhere) to not have aliases but this would be impossible 
Regarding IDE, you would have the issue anyway (as soon as you have some node_modules) may be you can configure the scanning only on the following pattern : `*/views/node_modules/@oat-sa/*/dist`

views/build/tasks/helpers/extensions.js
krampstudio: agree with the `npmPaths` missleading name

PR #2197, opened 2019-06-27
Fix/tao 8214/psr response exporter

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/TAO-8214 (fix of nccer reports is in https://github.com/oat-sa/extension-tao-nccer-delivery/pull/908)

We started to use Psr Responses instead of directly writing to the output but AbstractFileExporter and CsvExporter conflict with it producing php warnings in output.

All changes don't break compatibility with old code, but the new recommended way to download file is to call `$csvExporter->exportFileResponse()` instead of `export(..., true, ...)` (which is still works for compatibility).

Interface `PsrResponseExporter` extending `Exporter` interface added with `exportFileResponse` method. `CsvExporter` implements it, but `AbstractFileExporter` doesn't, because I didn't want to add new abstract method to `AbstractFileExporter` which is extended in many extensions (it would make all extending classes broken)

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -53,7 +53,7 @@ return array(
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '37.1.1',
+    'version' => '37.2.1',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => array(
         'generis' => '>=11.3.0',
--- a/manifest.php
+++ b/manifest.php
@@ -53,7 +53,7 @@ return array(
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '37.2.1',
+    'version' => '37.2.0',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => array(
         'generis' => '>=11.3.0',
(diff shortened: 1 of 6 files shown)


scripts/update/Updater.php
```diff
@@ -1066,6 +1066,6 @@ public function update($initialVersion) {
             $this->setVersion('35.8.2');
         }
 
-        $this->skip('35.8.2', '37.1.1');
+        $this->skip('35.8.2', '37.2.1');
```
tikhanovichA:
```suggestion
        $this->skip('35.8.2', '37.2.0');
```

manifest.php
```diff
@@ -53,7 +53,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '37.1.1',
+    'version' => '37.2.1',
```
tikhanovichA:
```suggestion
    'version' => '37.2.0',
```

models/classes/export/PsrResponseExporter.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @method string|ResponseInterface export()
+ */
+interface PsrResponseExporter extends Exporter
+{
+    /**
+     * @param ResponseInterface|null $originResponse base result on response
+     * @return ResponseInterface
+     */
+    public function exportFileResponse(ResponseInterface $originResponse = null);
```
hutnikau: This method returns response, so it's obviously getter.
So this method is used to _get response to export file_
What it rename it to `getFileExportResponse`?
Because current name assumes that given response (`$originResponse`) will export it somewhere.

siarhei-baradzin: Absolutely agree, fixed.

P.S. I just tried to name it similar to existing method

PR #2198, opened 2019-06-27
TAO-8589 - unique-field-validator

yuryfilippovich:
https://oat-sa.atlassian.net/browse/TAO-8589

Modified unique field validator (add parameters: current property name and property id) to be possible using this validator to check property value for unique
Refactoring cloneInstance method in models/classes/GenerisServiceTrait.php


helpers/form/validators/class.Unique.php
```diff
@@ -26,10 +26,17 @@
  * @author Joel Bout, <joel@taotesting.com>
  * @package tao
  */
-class tao_helpers_form_validators_Unique
-    extends tao_helpers_form_Validator
+
+use oat\oatbox\validator\ExtendedValidatorInterface;
+
+class tao_helpers_form_validators_Unique extends tao_helpers_form_Validator implements ExtendedValidatorInterface
```
augustas: There seems to be an integration test for unique validator, https://github.com/oat-sa/tao-core/blob/master/test/integration/ValidatorTest.php#L572 Maybe it could be updated with the case for current entity update.

augustas (commented):
The updated unique validator works well, the approach should be approved by an architect

PR #2200, opened 2019-06-28
Based date formatting on the platform specific DateTimeTzFormatString.

julien-sebire:
This PR requires https://github.com/oat-sa/generis/pull/643/.
For now, dates are formatted with default `Y-d-m H:i:s` for insertion into db datetime type.
Unfortunately, this doesn't work for every db engine, so I made it rely on the platform specific date format.
Added unit testing to cover changed code.


test/unit/model/taskQueue/TaskLog/Broker/RdsTaskLogBrokerTest.php
```diff
@@ -20,11 +20,47 @@
...
 {
+    /**
+     * @var RdsTaskLogBroker
+     */
+    protected $subject;
+
+    public function setUp()
+    {
+        $persistenceId = 'rds_result_storage_test';
+        $databaseMock = $this->getSqlMock('rds_result_storage_test');
```
bugalot: Is it an apporpriate name ?

julien-sebire: Fixed...

PR #2201, opened 2019-06-28
Feature/tao 8195/method not allowed handling

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/TAO-8195

Depends on: https://github.com/oat-sa/generis/pull/644

Handle common_exception_MethodNotAllowed in a proper way 
(405 status code and 'Allow' header according to RFC
https://tools.ietf.org/html/rfc7231#section-6.5.5

### DIFF
--- a/models/classes/mvc/error/ExceptionInterpretor.php
+++ b/models/classes/mvc/error/ExceptionInterpretor.php
@@ -24,6 +24,7 @@ use Exception;
 use common_exception_MissingParameter;
 use common_exception_BadRequest;
 use common_exception_ResourceNotFound;
+use Slim\Http\StatusCode;
 use tao_models_classes_MissingRequestParameterException;
 use Zend\ServiceManager\ServiceLocatorAwareInterface;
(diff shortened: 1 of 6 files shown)


models/classes/mvc/error/ExceptionInterpretor.php
```diff
@@ -82,36 +88,43 @@ protected function interpretError() {
...
+                break;
             case 'ActionEnforcingException':
             case 'tao_models_classes_FileNotFoundException':
             case common_exception_ResourceNotFound::class:
                 $this->returnHttpCode    = 404;
                 $this->responseClassName = 'MainResponse';
-            break;
+                break;
+            case common_exception_MethodNotAllowed::class:
+                $this->returnHttpCode    = 405;
```
hutnikau: Can we use `Slim\Http\StatusCode` constants ?

siarhei-baradzin: Done

PR #2206, opened 2019-07-02
Covered RdsNotification::sendNotification

julien-sebire:
Covered RdsNotification::sendNotification in tests for future changes.
**No fonctionnal change happened.**

Had to change 
```
$persistence = \common_persistence_Manager::getPersistence(RdsNotification::DEFAULT_PERSISTENCE);
```
to
```
$persistence = $this->getServiceLocator()->get(\common_persistence_Manager::SERVICE_ID)->getPersistenceById(RdsNotification::DEFAULT_PERSISTENCE);
```
Thanks @jbout for this, wouldn't have figured out.

Also had to change `getServiceManager()` to `getServiceLocator()` in `InstallNotificationTable` to be able to mock it correctly.

PR #2207, opened 2019-07-02
Changed autoincrement primary key to uuid.

julien-sebire:
This PR requires https://github.com/oat-sa/tao-core/pull/2206.

Not to be merged to develop until Spanner Benchmark is finished and the data migration policy is architecturally defined.

Only added uuid primary key.

### DIFF
--- a/models/classes/notification/implementation/RdsNotification.php
+++ b/models/classes/notification/implementation/RdsNotification.php
@@ -78,10 +78,11 @@ class RdsNotification
         $platform = $this->getPersistence()->getPlatForm();
 
         $sqlQuery    = 'INSERT INTO ' . self::NOTIF_TABLE .
-                        ' (' . $this->getAllFieldString() . ') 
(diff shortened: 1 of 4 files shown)


models/classes/notification/implementation/RdsNotification.php
```diff
@@ -78,10 +78,11 @@ public function sendNotification(NotificationInterface $notification)
         $platform = $this->getPersistence()->getPlatForm();
 
         $sqlQuery    = 'INSERT INTO ' . self::NOTIF_TABLE .
```
mike-ionut-mihai-sandu-tao: is it possible to use here the query builder?

julien-sebire: That could be done for the sake of readability, but it's actually part of another ticket, so the idea here was to go quick with the minimal change

julien-sebire:
Take in account in https://github.com/oat-sa/tao-core/pull/2167

PR #2211, opened 2019-07-04
Fix/tao 8587/migrate security settings

SashaTAO:
Requires: https://github.com/oat-sa/generis/pull/649

The goal of this PR is to configure SettingsStorage service to use 'default_kv' persistence and migrate existing security settings


yuryfilippovich:
We get error if run script with next parameters 
php index.php "oat\tao\scripts\tools\MigrateSecuritySettings" default  --wet

PR #2213, opened 2019-07-10
Fix/tao 8631/sync csrf token config

jsconan:
## Context
Related to: https://oat-sa.atlassian.net/browse/TAO-8631

Requires: 
 - [x] https://github.com/oat-sa/tao-core-sdk-fe/pull/24

## Summary
During an assessment, if the test taker spends more than 24 minutes on the same item without triggering any request to the server, the following error will be raised: `No tokens available. Please refresh the page.`. This event will occur no matter the duration of the server session.

The reason why is the anti-CSRF protection hardcodes the TTL of the token to 24 minutes, on the frontend side, no matter what is set on the platform.

To fix this wrong behavior, we need to address the issue on both sides:
- the frontend should take care of the platform configuration
- the backend should provide the plartform configuration to the frontend

The companion PR aims to address the frontend issue.

This PR rework the way the config is supplied to the client. Previously, only the list of predefined tokens was given to the client. Now, the relevant token store configuration should be provided.

## Way to improve
By default, the time limit for the anti-CSRF token is set to unlimited. An improvement would be to align that with the PHP session time limit.

## Important when merging
Once the companion PR is merged, please update accordingly the file `views/package.json` to reflect the version.

## Unit Test
Unit tests have been updated.

To unit test:
```
vendor/phpunit/phpunit/phpunit tao/test/unit
```

To scope the unit test:
```
vendor/phpunit/phpunit/phpunit tao/test/unit/model/security
```

## Steps to reproduce
To reproduce the issue:
- take a test
- stay more than 24 minutes on the same item, then try to navigate

In order to reduce the time to wait, you can also change the default TTL in `tao/views/node_modules/@oat-sa/tao-core-sdk/dist/core/tokenHandler.js`, by modifying the value of `tokenTimeLimit`, at line 29:
```javascript
    var defaults = {
      maxSize: 6,
      tokenTimeLimit: 1000 * 60 * 24  // change this value to a smaller one ;-)
    };
```

### DIFF
--- a/actions/class.ClientConfig.php
+++ b/actions/class.ClientConfig.php
@@ -45,7 +45,7 @@ class tao_actions_ClientConfig extends tao_actions_CommonModule
 
         /** @var TokenService $tokenService */
         $tokenService = $this->getServiceLocator()->get(TokenService::SERVICE_ID);
-        $this->setData('tokens', json_encode($tokenService->getClientConfig()));
+        $this->setData(TokenService::JS_DATA_KEY, json_encode($tokenService->getClientConfig()));
 
         //get extension paths to set up aliases dynamically
         $extensionsAliases = ClientLibRegistry::getRegistry()->getLibAliasMap();
--- a/actions/class.ClientConfig.php
+++ b/actions/class.ClientConfig.php
@@ -45,15 +45,7 @@ class tao_actions_ClientConfig extends tao_actions_CommonModule
 
         /** @var TokenService $tokenService */
         $tokenService = $this->getServiceLocator()->get(TokenService::SERVICE_ID);
-        $tokenPool = $tokenService->generateTokenPool();
-        $jsTokenPool = [];
-        foreach ($tokenPool as $key => $token) {
-            if ($key !== TokenService::FORM_POOL) {
-                $jsTokenPool[] = $token->getValue();
-
-            }
-        }
-        $this->setData('tokens', json_encode([TokenService::JS_TOKEN_KEY => $jsTokenPool]));
+        $this->setData('tokens', json_encode($tokenService->getClientConfig()));
 
         //get extension paths to set up aliases dynamically
         $extensionsAliases = ClientLibRegistry::getRegistry()->getLibAliasMap();
--- a/models/classes/security/xsrf/TokenService.php
+++ b/models/classes/security/xsrf/TokenService.php
@@ -53,6 +53,7 @@ class TokenService extends ConfigurableService
 
     const CSRF_TOKEN_HEADER = 'X-CSRF-Token';
     const FORM_POOL = 'form_pool';
+    const JS_DATA_KEY = 'tokenHandler';
     const JS_TOKEN_KEY = 'tokens';
     const JS_TOKEN_POOL_SIZE_KEY = 'maxSize';
     const JS_TOKEN_TIME_LIMIT_KEY = 'tokenTimeLimit';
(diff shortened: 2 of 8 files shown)


views/package.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
     "@oat-sa/tao-core-libs": "~0.1.0",
-    "@oat-sa/tao-core-sdk": "~0.5.0",
+    "@oat-sa/tao-core-sdk": "oat-sa/tao-core-sdk-fe#fix/TAO-8631/accept-token-lifetime-from-config",
```
jsconan: Reminder: this line must be updated once the companion PR is merged

actions/class.ClientConfig.php
```diff
@@ -45,15 +45,7 @@ public function config()
...
-        $tokenPool = $tokenService->generateTokenPool();
-        $jsTokenPool = [];
-        foreach ($tokenPool as $key => $token) {
-            if ($key !== TokenService::FORM_POOL) {
-                $jsTokenPool[] = $token->getValue();
-
-            }
-        }
-        $this->setData('tokens', json_encode([TokenService::JS_TOKEN_KEY => $jsTokenPool]));
+        $this->setData('tokens', json_encode($tokenService->getClientConfig()));
```
tikhanovichA:
```suggestion
        $this->setData(TokenService::JS_TOKEN_KEY, json_encode($tokenService->getClientConfig()));
```

models/classes/security/xsrf/TokenService.php
```diff
@@ -253,20 +255,24 @@ protected function invalidate($pool)
...
 
-        $store = $this->getStore();
-        $pool = $store->getTokens();
+        if ($withForm) {
+            $store = $this->getStore();
+            $pool = $store->getTokens();
 
-        if ($poolSize> 0 && isset($pool[self::FORM_POOL])) {
-            $poolSize++;
+            if ($poolSize> 0 && isset($pool[self::FORM_POOL])) {
```
tikhanovichA:
```suggestion
            if ($poolSize > 0 && isset($pool[self::FORM_POOL])) {
```

actions/class.ClientConfig.php
jsconan: good catch

models/classes/security/xsrf/TokenService.php
jsconan: hawk eye! :)

actions/class.ClientConfig.php
jsconan: Well I'm about to apply the change, but I feel uncomfortable with the business logic: `JS_TOKEN_KEY` contains the key of the list of tokens in the dataset, this is not the name of the dataset itself. A better approach would be to have a proper constant for that purpose. But then, is it the responsability of the engine to take care of the template specific? I mean should the engine manage how the controller is linked to the template to hydrate it?
If so, I shouls use the same PHP constant there, shouldn't? https://github.com/oat-sa/tao-core/blob/8a62ae3ca922efc5ad7ab40599800e213f6d4fd9/views/templates/client_config.tpl#L16

tikhanovichA: Thanks for your response.
About a new constant, I will say yes. It can be very useful for business logic. 
About the tpl, yes it will be better to use constant there

tikhanovichA:
The fix works like expected. Good job!

actions/class.ClientConfig.php
jsconan: Done

PR #2214, opened 2019-07-12
Feature/tao 8511/refactor basic auth type

tikhanovichA:
https://oat-sa.atlassian.net/browse/TAO-8511


test/unit/auth/AuthTypeTest.php
```diff
@@ -0,0 +1,75 @@
...
+            ->expects($this->once())
+            ->method('send')
+            ->with($requestMock, ['auth'=> ['testLogin', 'testPassword'], 'verify' => false]);
+
+        $authType->setClient($clientMock);
+
+        $authType->call($requestMock);
+    }
+
+    public function testFaildValidationBasicAuthType()
```
alexeyvy: Faild?

models/classes/auth/AbstractCredentials.php
```diff
@@ -0,0 +1,55 @@
...
+
+    protected function validate($properties)
+    {
+        foreach ($properties as $key => $value) {
+            $validatedProperties = $this->getProperties();
+            if (!in_array($key, $validatedProperties, false)) {
+                throw new \common_exception_ValidationFailed($key);
+            }
+        }
+        return true;
```
alexeyvy: This line makes little sense, as there's an exception thrown instead of `return false`

models/classes/auth/AbstractAuthType.php
```diff
@@ -94,4 +100,37 @@ public function getInstance()
...
+        if (!$withKeys) {
+            $data = array_values($data);
+        }
+
+        return $data;
+
+    }
+
+    /**
+     * @return AbstractCredentials
```
alexeyvy: Isn't it a 'string'?

models/classes/auth/AbstractAuthType.php
```diff
@@ -94,4 +100,37 @@ public function getInstance()
...
+     */
+    abstract public function getCredentialsClassName();
```
alexeyvy: Instead of a class name here, can we expect an object initialized? This would make instantiating more explicit which leads to clearer DI

models/classes/auth/AbstractCredentials.php
```diff
@@ -0,0 +1,55 @@
...
+     * AbstractCredentials constructor.
+     * @param array $properties
+     * @throws \common_exception_ValidationFailed
+     */
+    public function __construct($properties = [])
+    {
+        $this->validate($properties);
+        foreach ($properties as $key => $value) {
+            if (method_exists($this, $key)) {
+                $this->{$key} = $value;
```
alexeyvy: An implicit call, could we get rid of one?

models/classes/auth/AbstractAuthType.php
tikhanovichA: It should be an object. I will fix this method. Thanks.

models/classes/auth/AbstractAuthType.php
tikhanovichA: Yes, agree. I will fix this method. Thanks.

test/unit/auth/AuthTypeTest.php
tikhanovichA: will fix

models/classes/auth/AbstractCredentials.php
tikhanovichA: Thanks, will check it

alexeyvy: I suggest moving filed management to the credential object side entirely

models/classes/auth/AbstractCredentials.php
tikhanovichA: yes, agree. Will fix it

models/classes/auth/AbstractCredentials.php
tikhanovichA: ok

models/classes/auth/AbstractCredentials.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * @param $properties
+     * @throws \common_exception_ValidationFailed
+     */
+    protected function validate($properties)
+    {
+        foreach ($properties as $key => $value) {
+            $validatedProperties = array_keys($this->getProperties());
```
alexeyvy: we can call getProperties only once, above the loop

models/classes/auth/BasicAuthType.php
```diff
@@ -114,10 +115,36 @@ protected function loadCredentials()
...
+    public function getCredentialsClass($parameters = [])
+    {
+        return new BasicAuthCredentials($parameters);
+    }
+
+    /**
+     * @param $clientOptions
+     * @return mixed
+     */
+    protected function getClient($clientOptions = [])
```
bartlomiejmarszal: Shouldn't we create some configurable service for that?

models/classes/auth/AbstractCredentials.php
```diff
@@ -0,0 +1,58 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2019  (original work) Open Assessment Technologies SA;
+ *
+ */
+
+namespace oat\tao\model\auth;
+
+abstract class AbstractCredentials
```
bartlomiejmarszal: Can we have some phpdocs here?

models/classes/auth/BasicType.php
```diff
@@ -0,0 +1,101 @@
...
+     * @param $clientOptions
+     * @return mixed
+     */
+    protected function getClient($clientOptions = [])
+    {
+        return new Client($clientOptions);
+    }
+
+    /**
+     * @deprecated
```
alexeyvy: can we deprecate getInstance & setInstance at the level of the abstract class?

models/classes/auth/BasicAuthType.php
tikhanovichA: No, this method just for the unit test

models/classes/auth/AbstractAuthType.php
```diff
@@ -73,8 +76,8 @@ public function __toPhpCode()
     {
         return 'new '.get_class($this).'()';
     }
-
     /**
+     * @deprecated
```
bartlomiejmarszal: Please write description what should be used instead

models/classes/auth/AbstractAuthType.php
```diff
@@ -85,6 +88,15 @@ public function setInstance(\core_kernel_classes_Resource $instance = null)
...
     /**
+     * @param array $credentials
+     */
+    public function setCredentials($credentials = [])
+    {
+        $this->credentials = $credentials;
+    }
+
+    /**
+     * @deprecated
```
bartlomiejmarszal: Please write description what should be used instead

test/unit/auth/BasicTypeTest.php
```diff
@@ -0,0 +1,75 @@
...
+        $clientMock = $this->createMock(Client::class);
+
+        $clientMock
+            ->expects($this->once())
+            ->method('send')
+            ->with($requestMock, ['auth'=> ['testLogin', 'testPassword'], 'verify' => false]);
+
+        $authType->setClient($clientMock);
+
+        $authType->call($requestMock);
```
bartlomiejmarszal: This should return some result. Why don't we assert this result?

PR #2220, opened 2019-07-15
Fix/tao 8612/list property dialog

oatymart:
https://oat-sa.atlassian.net/browse/TAO-8612

The jQuery dialog plugin was removed some months ago. This means we are currently unable to open the item class properties dialog, which contains a list editor.

- I refactored `$dialogElm.dialog({...})` to replace it with our in-house `ui/dialog`
- I added a new template for the dialog content
- Some small CSS changes were needed to make the list editor and its context menu plugin look acceptable
- Missing icons used in the context menu (were giving 404s before)

You can see the old functionality by installing package-tao branch `release-3.3-rc02`.

To test:
1. Log into BackOffice.
2. Go to Items, click on an item class
3. Click on "Manage Schema" button in the content action bar  
4. Click on "Add Property"
5. Edit the newly added property
6. In the "Type" field, select any list type.
7. Click on the green plus sign under "List values" (or select "+ Add / Edit Lists" in dropdown)
8. Interact with the list editor (open, close, create, rename, delete, close, reopen)


views/js/uiForm.js
```diff
@@ -572,236 +577,233 @@ define([
...
+                var dialogData;
+                var classUri;
+
+                /**
+                 * Creates the jsTree list manager and attaches behaviours
+                 * @param {String} treeId
+                 * @returns {jQuery}
+                 */
+                function createListsTree(treeId) {
+                    var url = context.root_url + 'taoBackOffice/Lists/';
```
ferenckiss89: es6 maybe? (I assume it was a copy mostly from an existing file)

anshul119:
Why are we using png images for the icon? Cant we add some rich icons from icomoon and use those insteaD?

oatymart:
> Why are we using png images for the icon? Cant we add some rich icons from icomoon and use those instead?

It's a good idea for a future improvement, for sure. At the moment the aim is just to restore the functionality we already had.

views/js/uiForm.js
```diff
@@ -572,236 +577,233 @@ define([
...
-                                    url: dataUrl
-                                }
-                            },
-                            types: {
-                                "default": {
-                                    renameable: true,
-                                    deletable: true,
-                                    creatable: true,
-                                    draggable: false
+                var rangeId;
```
anshul119: If you wanna be a good lad and have some time on your hand you could convert this file to es6. :D

anshul119:
checked the functionality, it works as expected. We just discovered a bug which already existed in the old dialog box.

PR #2233, opened 2019-07-26
Read variables from .env file in project dir if it is provided.

julien-sebire:
Reads variables from .env file in project dir if it is provided. If no `.env` file is provided, just does nothing.
This allows us to then access variables values in `$_ENV` global array.
Works for CLI, Apache, Nginx, ...

Example:

.env file:
```
GENERIS_CACHE_URI=http://my-redis.com/blah/blah/blah
```

.php file:
```
echo $_ENV['GENERIS_CACHE_URI']; // => http://my-redis.com/blah/blah/blah
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
         "oat-sa/composer-npm-bridge": "^0.3.0",
-        "symfony/dotenv": "^4.0"
+        "symfony/dotenv": "~4.3"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,8 @@
         "pimple/pimple": "^3.0",
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
-        "oat-sa/composer-npm-bridge": "^0.3.0"
+        "oat-sa/composer-npm-bridge": "^0.3.0",
+        "symfony/dotenv": "^5.0@dev"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
         "oat-sa/composer-npm-bridge": "^0.3.0",
-        "symfony/dotenv": "^5.0@dev"
+        "symfony/dotenv": "^4.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -56,7 +56,8 @@
         "pimple/pimple": "^3.0",
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
-        "oat-sa/composer-npm-bridge": "^0.3.0"
+        "oat-sa/composer-npm-bridge": "^0.3.0",
+        "symfony/dotenv": "^5.0@dev"
```
Babacooll: Why do we need ^5.0 version ?

julien-sebire: That's right, I just picked the version that popped out when `composer requiring` the repo... Fixed to the last stable version.

composer.json
```diff
@@ -56,7 +56,8 @@
         "pimple/pimple": "^3.0",
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
-        "oat-sa/composer-npm-bridge": "^0.3.0"
+        "oat-sa/composer-npm-bridge": "^0.3.0",
+        "symfony/dotenv": "^4.0"
```
Babacooll: Shouldn't it be ^4.3 ?

models/classes/mvc/Bootstrap.php
```diff
@@ -89,6 +90,13 @@ class Bootstrap implements ServiceManagerAwareInterface
      */
     public function __construct($configuration)
     {
+        $envDir = __DIR__ . '/../../../../';
+        $envFile = $envDir. '.env';
+        if (file_exists($envFile)) {
+            $dotenv = new Dotenv();
+            $dotenv->load($envFile);
```
peetya: What do you think about using `loadEnv` instead of `load`?

julien-sebire: You're right. Fixed.

composer.json
julien-sebire: I even went to `~4.3` so that we could include higher 4.x version also.

models/classes/mvc/Bootstrap.php
```diff
@@ -89,6 +90,13 @@ class Bootstrap implements ServiceManagerAwareInterface
      */
     public function __construct($configuration)
     {
+        $envDir = __DIR__ . '/../../../../';
+        $envFile = $envDir. '.env';
```
ekkinox: - would merge in one variable
- would use `DIRECTORY_SEPARATOR` to avoid windows issues

julien-sebire: Fixed in PR https://github.com/oat-sa/tao-core/pull/2238.

PR #2235, opened 2019-07-29
Feature/tao 8420/configurable webhooks

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/TAO-8420

Main approaches:

**Supported events**
List of events which could be used to perform webhooks now is empty and could be extended only by developers (**approved by Joel**). 
To do it they:
1. Implement `WebhookSerializableInterface` in event to separate a way how data gets serialized in the event from `jsonSerialize` method (**approved by Joel**)
2. In Updater/install script add the event to the list of supported events in `EventWebhooksService` by `$eventWebhookService->registerEvent('eventName', $eventManager);` which updates `EventWebhooksService` configuration and also attaches service method as a handler for event. We have to reregister both `EventWebhooksService` and `EventManager` after this. Passing EventManager to `registerEvent()` emphasizes the fact that it gets changed and needs to be reregistered
3. Also in the future list of supported events could be used to present them in admin panel

**No support for bunching events to perform only one webhook request**
Approved by Joel

**EventWebhookConfigRepositoryInterface as an abstraction**
Another implementation of it could be provided later to store webhooks configuration in RDS/KV storage (**approved by Joel**). No support for `set` methods for now, only read operations are available

**EventWebhookConfigFileRepository implementation**
Simplified impl of repository needed for MVP. Reads configuration from it's config which will be modified/replaced on update/install by devops (because we already have such flow of providing configuration). No runtime/scripts modifications. The configuration consists of:
1. webhooks with unique ids
2. events with an assigned list of webhooks ids
(**approved by Joel**)

**Typed DTO for representing config in repo interface**
`models/classes/webhooks/ConfigEntity` dir contains DTOs to represent data stored in config (in form of arrays) to expose it in a safe way to the outside (for another repo implementation data could be stored in KV and have another representation than PHP arrays)

**Passing information to the task**
Joel recommended not to pass auth details to a task but retrieve it when the task runs by webhook config id.

_Of course it could lead to accessing to unexisting webhook conf at the moment when Task is processed if config was changed. Probably we have to decide how to handle it later (from discussion with @tikhanovichA)_

I decided to pass to a task only serialized data from an event and webhook id. To describe the structure of a data which will be passed to a task I decided to introduce `class WebhookTaskMetadata extends ArrayObject` with consts for keys which is Traversable and could be directly used as task metadata.

**WebhookTaskServiceInterface**
Created to complete `EventWebhooksService` class and provide tests for it. Not implemented at the moment, will create and enqueue task with passed metadata (will be done in the following task)


config/default/eventWebhookConfigRepository.conf.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Default config header created during install
+ */
+
+return new oat\tao\model\webhooks\EventWebhookConfigFileRepository(array(
```
bartlomiejmarszal: Can you explain this long name? Why not name it EventWebhookRepository.conf.php

models/classes/webhooks/ConfigEntity/Webhook.php
```diff
@@ -0,0 +1,126 @@
...
+     */
+    private $auth;
+
+    /**
+     * Create object from array generated by toArray() method
+     *
+     * @param array $array
+     * @return self
+     */
+    public static function fromArray($array)
```
bartlomiejmarszal: Why you are using static function here?

models/classes/webhooks/ConfigEntity/WebhookAuth.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @var array
+     */
+    private $properties;
+
+    /**
+     * @param array $array
+     * @return WebhookAuth
+     */
+    public static function fromArray(array $array)
```
bartlomiejmarszal: Why u are using static function here?

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,166 @@
...
+    {
+        $eventName = $event->getName();
+
+        if (!$this->isEventRegistered($eventName)) {
+            $this->logError("Event '$eventName' is not supported by " . self::class);
+            return false;
+        }
+
+        if (!($event instanceof WebhookSerializableInterface)) {
+            $this->logError(sprintf('Event "%s" passed to "%s" is not "%s"',
```
bartlomiejmarszal:
```suggestion
            $this->logError(sprintf('Event "%s" passed to "%s" is not implementing "%s"',
```

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,166 @@
...
+
+        return true;
+    }
+
+    /**
+     * @param Event|WebhookSerializableInterface $event
+     * @param string[] $webhookConfigIds
+     */
+    protected function createTasksForEvent(Event $event, $webhookConfigIds) {
+        $eventData = $event->serializeForWebhook();
```
bartlomiejmarszal: Interface `Event` does not have method `serializeForWebhook`

models/classes/webhooks/ConfigEntity/Webhook.php
siarhei-baradzin: In config file we store arrays and want to create objects (with typed properties which implement WebhookInterface) to return them in Repository getters.

Do you suggest to create a separate factory class for that purpose?

models/classes/webhooks/EventWebhooksService.php
siarhei-baradzin: That's why I add `@param Event|WebhookSerializableInterface $event` to phpdoc (ok for phpstorm). Looks a little strange, but check for WebhookSerializableInterface was performed before. Do you suggest to add one more runtime check in this method?

config/default/eventWebhookConfigRepository.conf.php
siarhei-baradzin: it should be equal to SERVICE_ID in `EventWebhookConfigRepositoryInterface`.

_Event_ - probably we can have webhooks not only for events but another webhooks later
_Config_ - stores only **config** which is needed to implement webhook, because webhook itself is smth different (a way of delivering information or compiled object with all needed information for request maybe)
_Repository_ - It's a class to read/write/store webhook configs. It's not a config itself

It was the idea. All my task was about **configs** for webhooks. Probably in next tasks we can use "webhook" word without "config". Not 100% sure it's the best naming. If you insist I can rename it

models/classes/webhooks/ConfigEntity/WebhookAuth.php
siarhei-baradzin: To use it in the previous static function, check my comment

models/classes/webhooks/ConfigEntity/Webhook.php
bartlomiejmarszal: It will be difficult for unit test purposes when some other service would like to use this method

models/classes/webhooks/EventWebhooksService.php
bartlomiejmarszal: As you calling method that is in WebhookSerializableInterface, I think this param should point to this interface. Consider extending this interface with Event interface

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,182 @@
...
+    public function isEventRegistered($eventName)
+    {
+        $supportedEvents = $this->getRegisteredEvents();
+        return isset($supportedEvents[$eventName]);
+    }
+
+    public function handleEvent(Event $event)
+    {
+        if (!$this->checkEventIsSupported($event)) {
+            return;
```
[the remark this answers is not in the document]
siarhei-baradzin: It writes error to the log instead. I try not to throw exceptions in the event handler because they can stop the execution of a code which triggers events. If we have an error in handler/posting task/task serialization I don't want to occasionally stop delivery execution creation with 500 error

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        /** @var WebhookSerializableEventInterface $event */
+
+        $webhookConfigIds = $this->getEventWebhookConfigRepository()->getWebhookConfigIds($event->getName());
+        if (count($webhookConfigIds) === 0) {
+            return;
```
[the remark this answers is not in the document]
siarhei-baradzin: no, it's just a case when we have supported (registered) event but don't have connected handlers in config

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Option value is array of ['eventName' => true, ... ] of supported events
+     * Such array structure is needed to perform quick search by key
+     */
+    const OPTION_SUPPORTED_EVENTS = 'supportedEvents';
+
+    /**
+     * @inheritDoc
+     */
+    public function registerEvent($eventName, EventManager $eventManager)
```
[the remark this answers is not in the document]
siarhei-baradzin: Yes, we can. I added it as an argument to emphasize the fact that method changes EventManager's config and caller (update/install script) have to re-register EventManager after this call (so the caller has $eventManager variable anyway)

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $supportedEvents = $this->getRegisteredEvents();
+        $supportedEvents[$eventName] = true;
+        $this->setOption(self::OPTION_SUPPORTED_EVENTS, $supportedEvents);
+
+        $eventManager->attach($eventName, $this->getEventHandlerCallback());
```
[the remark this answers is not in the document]
siarhei-baradzin: const defined in `EventWebhooksServiceInterface`

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        $this->createTasksForEvent($event, $webhookConfigIds);
+    }
+
+    /**
+     * @param Event $event
+     * @return bool
+     */
+    protected function checkEventIsSupported(Event $event)
```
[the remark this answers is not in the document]
siarhei-baradzin: Changed to private

models/classes/webhooks/EventWebhooksService.php
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        return true;
+    }
+
+    /**
+     * @param WebhookSerializableEventInterface $event
+     * @param string[] $webhookConfigIds
+     */
+    protected function createTasksForEvent(WebhookSerializableEventInterface $event, $webhookConfigIds) {
```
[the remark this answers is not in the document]
siarhei-baradzin: Actually, I'm not sure task creation is the scope of this task. I added this code to show how I see the feature code structure and to complete this class to have the ability to write tests for it. When we start implementing tasks it could be changed. 

The approach of this method purpose is from tao/models/classes/taskQueue/QueueDispatcher.php (which will be probably used) where task gets created and enqueued in the same 'createTask' method

> This method is not creating tasks, it's just preparing WebhookTaskMetadata 
It calls `WebhookTaskServiceInterface::createTask()` to create tasks

test/unit/webhooks/EventWebhooksServiceTest.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    public function testUnregisterEvent()
+    {
+        $eventName = 'TestEvent';
+        $this->eventManagerMock->expects($this->once())
+            ->method('detach')
+            ->with($eventName, [EventWebhooksService::SERVICE_ID, 'handleEvent']);
+
+        $service = $this->getService([
```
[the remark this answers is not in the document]
siarhei-baradzin: fixed

models/classes/webhooks/EventWebhooksService.php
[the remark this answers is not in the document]
bartlomiejmarszal: I wouldn't emphasize this fact. I would retrieve it using getter and keep only event name in param.

test/unit/webhooks/EventWebhookConfigFileRepositoryTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+    public function testGetWebhookConfigIds() {
+        $this->repository = new EventWebhookConfigFileRepository([
+            EventWebhookConfigFileRepository::OPTION_EVENTS => [
+                'TestEvent' => ['wh1']
+            ],
+            EventWebhookConfigFileRepository::OPTION_WEBHOOKS => ['wh1' => []]
+        ]);
+
+        $this->assertEquals(['wh1'], $this->repository->getWebhookConfigIds('TestEvent'));
```
bartlomiejmarszal:
```suggestion
        $result = $this->repository->getWebhookConfigIds('TestEvent')
        $this->assertEquals(['wh1'], $result);
```

test/unit/webhooks/EventWebhookConfigFileRepositoryTest.php
```diff
@@ -0,0 +1,104 @@
...
+        $this->assertEquals([], $this->repository->getWebhookConfigIds('AnotherEvent'));
```
bartlomiejmarszal: Extract this to new test method

models/classes/webhooks/EventWebhookConfigRepositoryInterface.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace oat\tao\model\webhooks;
+
+use oat\tao\model\webhooks\ConfigEntity\WebhookInterface;
+
+/**
+ * Represents webhooks config entries storage. We will add 'set' methods here if we are going to implement
+ * dynamic configuration for webhooks (admin web panel)
+ */
+interface EventWebhookConfigRepositoryInterface
```
bartlomiejmarszal:
```suggestion
interface WebhookRegistryInterface
```

models/classes/webhooks/EventWebhookConfigFileRepository.php
```diff
@@ -0,0 +1,80 @@
...
+ *      value: array of string ids of connected webhooks
+ *
+ * `webhooks` is array with:
+ *      key: webhook unique id
+ *      value: array representation of ConfigEntry\Webhook
+ *
+ * See tao/test/unit/webhooks/WebhookConfigFileRepositoryTest.php
+ * for example of configuration
+ */
+class EventWebhookConfigFileRepository extends ConfigurableService implements EventWebhookConfigRepositoryInterface
```
bartlomiejmarszal:
```suggestion
class WebhookFileRegistry extends ConfigurableService implements EventWebhookConfigRepositoryInterface
```

config/default/eventWebhookConfigRepository.conf.php
bartlomiejmarszal:
```suggestion
return new oat\tao\model\webhooks\WebhookFileRegistry(array(
```

PR #2237, opened 2019-07-30
Fix/tao 7833/update property labels

ferenckiss89:
Jira: https://oat-sa.atlassian.net/browse/TAO-7833

- JS check on regular property labels to keep them up to date
- unit test for the new method

### DIFF
--- a/views/js/form/post-render-props.js
+++ b/views/js/form/post-render-props.js
@@ -391,7 +391,7 @@ define([
                 var $parentHeadingLabel = $(this).find('.property-heading-label');
                 var $editBlockLabel = $(this).find('.property-edit-container input[name$="_label"]');
 
-                if (editBlockLabel.val() !== '') {
+                if ($editBlockLabel.val() !== '') {
                     $parentHeadingLabel.text($editBlockLabel.val());
                 }
             }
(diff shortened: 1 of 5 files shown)


views/js/form/post-render-props.js
```diff
@@ -380,6 +381,23 @@ define([
...
+     * Checks and updates property labels 
+     * @param {Object} $properties - properties object
+     */
+    function _checkRegularPropertyLabels($properties) {
+        $properties.each(function() {
+            if($(this).hasClass('regular-property')) {
+                var $parentHeadingLabel = $(this).find('.property-heading-label');
+                var $editBlockLabel = $(this).find('.property-edit-container input[name$="_label"]');
+
+                if ($parentHeadingLabel.text() !== $editBlockLabel.val()) {
```
oatymart: I would say no need to even check for inequality. Just set it.

ferenckiss89: true

views/js/form/post-render-props.js
```diff
@@ -380,6 +381,23 @@ define([
...
+     * Checks and updates property labels 
+     * @param {Object} $properties - properties object
+     */
+    function _checkRegularPropertyLabels($properties) {
+        $properties.each(function() {
+            if($(this).hasClass('regular-property')) {
+                var $parentHeadingLabel = $(this).find('.property-heading-label');
+                var $editBlockLabel = $(this).find('.property-edit-container input[name$="_label"]');
+
+                if (editBlockLabel.val() !== '') {
```
oatymart: $editBlockLabel

ferenckiss89: yep

PR #2242, opened 2019-08-05
Fix/cli error msg

jbout:


boajer:
Nice catch!

models/classes/routing/CliController.php
```diff
@@ -23,7 +23,6 @@
 use oat\oatbox\action\ActionService;
 use oat\oatbox\action\ResolutionException;
 use common_report_Report as Report;
-use oat\oatbox\action\Help;
```
boajer: I have feeling  you've removed the only occurrence of  `oat\oatbox\action\Help`  usage across codebase.
https://github.com/search?l=PHP&p=7&q=org%3Aoat-sa+oat%5Coatbox%5Caction%5CHelp&type=Code

Should be consider risky step and remove class itself?  @hutnikau @jbout

bugalot: It seems to me also it's the only one in the code base. In favour to remove the class.

jbout: Agreed, but it is in generis

boajer:
Merged, thank you all.

PR #2246, opened 2019-08-13
Tao 8665/file helper extract archive

hutnikau:
copied from https://github.com/oat-sa/extension-tao-mediamanager/blob/master/model/ZipImporter.php#L143
After merge of that PR I will update `ZipImporter` in order to use `tao_helpers_File` to extract archive


helpers/class.File.php
```diff
@@ -481,6 +484,68 @@ public static function addFilesToZip(ZipArchive $zipArchive, $src, $dest, $withE
...
+            fclose($originalPackage);
+            fclose($tmpFile);
+
+            $archiveFile = $tmpFilePath;
+        }
+
+        $archiveDir = \tao_helpers_File::createTempDir();
+        $archiveObj = new \ZipArchive();
+
+        if ($archiveFile instanceof File) {
```
augustas: This code block is very similar to the one above, copying the zip file to temporary location and setting the path to it.

helpers/class.File.php
```diff
@@ -481,6 +484,68 @@ public static function addFilesToZip(ZipArchive $zipArchive, $src, $dest, $withE
...
+            $originalPackage = $archiveFile->readStream();
+            stream_copy_to_stream($originalPackage, $tmpFile);
+            fclose($originalPackage);
+            fclose($tmpFile);
+
+            $archiveFile = $tmpFilePath;
+        }
+
+        $archiveDir = \tao_helpers_File::createTempDir();
+        $archiveObj = new \ZipArchive();
```
augustas: Creation of these variables could be moved closer to their usage.

helpers/class.File.php
```diff
@@ -481,6 +484,68 @@ public static function addFilesToZip(ZipArchive $zipArchive, $src, $dest, $withE
...
+
+        if ($archiveFile instanceof File) {
+            // get a local copy of zip
+            $tmpName = \tao_helpers_File::concat([\tao_helpers_File::createTempDir(), $archiveFile->getBasename()]);
```
augustas: References to `\tao_helpers_File::` could be replaced with `self::`

test/integration/helpers/FileHelperTest.php
```diff
@@ -200,4 +200,15 @@ public function testGetAllZipNames()
         $this->assertEquals(['path/to/data/text.txt', 'path/to/log.log'], tao_helpers_File::getAllZipNames($zipArchive));
         $zipArchive->close();
     }
+
+    public function testExtractArchive()
+    {
+        $path = realpath('./samples/samples.zip');
```
augustas: `samples.zip` is not available and this test case results in an error.

augustas (changes requested):
I understand this code was taken from other extension, but I think there are a few ways to improve it

PR #2259, opened 2019-08-26
Feature/tao 8421/retry mechnism

bartlomiejmarszal:
### Retry Mechnism in task queue
https://oat-sa.atlassian.net/browse/TAO-8421

After invoking, when task return Report::Error, retry mechanism will check if task has retryMax configured and will create new child task that will be executed until status will change to completed or `maxRetry` count reach configured level


models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +213,49 @@ private function getRemoteStatus(TaskInterface $task)
...
+            }
+
+        }
+    }
+
+    /**
+     * @param TaskInterface $task
+     * @return bool
+     */
+    private function isTaskCancelled(TaskInterface $task): bool
```
siarhei-baradzin: Pls pay attention, method name doesn't match its behavior: it checks if "task is **NOT** canceled"

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +213,49 @@ private function getRemoteStatus(TaskInterface $task)
     {
         return $task instanceof CallbackTaskInterface ? $task->getCallable()->getRemoteStatus() : $task->getRemoteStatus();
     }
+
+
+    private function retryMechanism(CallbackTaskInterface $task)
+    {
+        $callableTask = $task->getCallable();
+        if ($retryMax = $task->getParameter('retryMax')) {
```
siarhei-baradzin: If the existence of `retryMax` and `retryCount` params is a part of task contract (and changes task processing) consider moving correspondent string literals to TaskInterface as consts.

Also, it's good to be sure that these params are not in use in existing tasks in our extensions and mention them in `tao/models/classes/taskQueue/README.md`

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -151,6 +154,10 @@ public function processTask(TaskInterface $task)
                 }
             }
 
+            if ($status === TaskLogInterface::STATUS_FAILED && $task instanceof CallbackTaskInterface) {
```
siarhei-baradzin: What if task extends `AbstractTask` and implements only `TaskInterface`?
What if we don't reach retryMax number but want to stop retrying for some reasons (for example, webhook endpoint answers with http 400, which means further retries don't make sense opposite to timeout error)?

models/classes/webhooks/WebhookRegistryInterface.php
```diff
@@ -40,4 +40,9 @@ public function getWebhookConfig($id);
      * @return string[]
      */
     public function getWebhookConfigIds($eventName);
+
+    /**
+     * @return int
+     */
+    public function getMaxRetries($webhookConfigId);
```
siarhei-baradzin: Are you sure we need to extend interface by additional method if such information could be gotten from `getWebhookConfig(...)->getMaxRetry()`?

models/classes/webhooks/task/WebhookTaskParams.php
```diff
@@ -71,4 +72,12 @@ public function getWebhookConfigId()
     {
         return $this[self::WEBHOOK_CONFIG_ID];
     }
+
+    /**
+     * @return string
```
siarhei-baradzin: looks like it should be int

models/classes/webhooks/WebhookRegistryInterface.php
bartlomiejmarszal: I used this method when implementing configurable service based on this interface.

models/classes/taskQueue/Worker/AbstractWorker.php
bartlomiejmarszal: We should specify errors that will trigger retry mechanism. Right now it will be dependent from generated common report. We could add there some attribute that will be responsible for deciding if we wish to retry process.

siarhei-baradzin:
Some architectural notes:

1. Pls check out `updateParent()` in `tao/models/classes/taskQueue/TaskLog.php` logic. In case of 1 unsuccessful retry and then 1 successful, the main task will stay with FAILED status.
Possible solution: 
1.1 return an error in a child task only for last unsuccessful retry
1.2 add new `updateRetryableParent()` method which will set STATUS_COMPLETED for parent if at least one child is completed.

2. You can identify a task as retryable not by searching for specific params but (more OOP way)_ as introducing `RetryableTaskInterface` (as it did with `ChildTaskAwareInterface`)

bartlomiejmarszal:
> Some architectural notes:
> 
> 1. Pls check out `updateParent()` in `tao/models/classes/taskQueue/TaskLog.php` logic. In case of 1 unsuccessful retry and then 1 successful, the main task will stay with FAILED status.
>    Possible solution:
>    _1.1_ return an error in a child task only for last unsuccessful retry
>    _1.2_ add new `updateRetryableParent()` method which will set STATUS_COMPLETED for parent if at least one child is completed.
> 2. You can identify a task as retryable not by searching for specific params but (more OOP way)_ as introducing `RetryableTaskInterface` (as it did with `ChildTaskAwareInterface`)

I had this approach until [this commit]( https://github.com/oat-sa/tao-core/pull/2259/commits/3dd7e11792ac3ee6f8e4d4a2bea1d2c8f674676e)

I will consider to move this interface back. For now I would like to focus on main task queue operations used in introduced mechanism.

models/classes/taskQueue/Worker/AbstractWorker.php
siarhei-baradzin: Anyway, task could be just TaskInterface (extended from AbstractTask) and it will not pass your check `$task instanceof CallbackTaskInterface`

models/classes/webhooks/WebhookRegistryInterface.php
siarhei-baradzin: I see, but why we need a separate method to get individual param from webhook config, if we have `getWebhookConfig()` method? It will complicate implementations of other WebhookRegistryInterface implementations in the feature..

bartlomiejmarszal: I find that more readable. Am I wrong?

siarhei-baradzin:
Now we use child/parent tasks not quite for its intended purpose. The original purpose is to auto-update parent status to COMPLETED when all children are done and to FAILED when at least one failed. Child tasks represent parts of work in the scope of parent task as I understand it.

I have doubts now about using parent/child relations for retries. Also, consider the case in the feature when we send several events in one request (one of the features originally requested) and some of them have "accepted" status and some of them have "error" status. Repeating the whole request would be wrong in this case

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +214,60 @@ private function getRemoteStatus(TaskInterface $task)
...
     }
+
+
+    private function retryMechanism(CallbackTaskInterface $task)
+    {
+        $callableTask = $task->getCallable();
+        if ($retryMax = $task->getParameter(WebhookTaskParams::RETRY_MAX)) {
+            $retryCount = $this->setRetryCountParam($task);
+            if ($retryMax >= $retryCount) {
+                $this->getQueueDispatcher()->createTask($callableTask, $task->getParameters(), $task->getLabel(), $task, true);
```
bopoda: Short strings look better. If the method has a lot of arguments, it's easier to read the code when each argument is situated on the next line:

```
$this->getQueueDispatcher()
    ->createTask(
        $callableTask,
        $task->getParameters(),
        $task->getLabel(),
        $task,
        true
    );
```
But, ofcourse, it depends on conventions inside the project.

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +214,60 @@ private function getRemoteStatus(TaskInterface $task)
     {
         return $task instanceof CallbackTaskInterface ? $task->getCallable()->getRemoteStatus() : $task->getRemoteStatus();
     }
+
+
+    private function retryMechanism(CallbackTaskInterface $task)
+    {
+        $callableTask = $task->getCallable();
+        if ($retryMax = $task->getParameter(WebhookTaskParams::RETRY_MAX)) {
+            $retryCount = $this->setRetryCountParam($task);
```
bopoda: looks that method `setRetryCountParam` is getter. Actually, the method is `set` and `get` at the same time what is not good. Maybe it makes sense to name the method like `increaseRetryCount` or like that. And after that use `$retryCount = $task->getParameter(WebhookTaskParams::RETRY_COUNT)` as above.

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +214,60 @@ private function getRemoteStatus(TaskInterface $task)
...
+    private function setRetryCountParam(CallbackTaskInterface $task)
+    {
+        $retryCount = $task->getParameter(WebhookTaskParams::RETRY_COUNT);
+        if ($retryCount === null) {
+            $retryCount = 1;
+            $task->setParameter(WebhookTaskParams::RETRY_COUNT, $retryCount);
+        } else {
+            $retryCount++;
+            $task->setParameter(WebhookTaskParams::RETRY_COUNT, $retryCount);
+        }
```
bopoda:
```
$retryCount = $task->getParameter(WebhookTaskParams::RETRY_COUNT);
$task->setParameter(WebhookTaskParams::RETRY_COUNT, ++$retryCount);
```
increment should work perfect for `null` too

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -151,6 +155,10 @@ public function processTask(TaskInterface $task)
                 }
             }
 
+            if ($status === TaskLogInterface::STATUS_FAILED && $task instanceof CallbackTaskInterface) {
```
gyszucs: Why is it important to check whether it is an instance of `CallbackTaskInterface`?

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +214,60 @@ private function getRemoteStatus(TaskInterface $task)
     {
         return $task instanceof CallbackTaskInterface ? $task->getCallable()->getRemoteStatus() : $task->getRemoteStatus();
     }
+
+
+    private function retryMechanism(CallbackTaskInterface $task)
+    {
+        $callableTask = $task->getCallable();
```
gyszucs: I don't think you need the internal callable. You can simply operate on the current task object.

models/classes/taskQueue/Worker/AbstractWorker.php
gyszucs: And if you only operate on the current task object, you can simply store a new task like it is done few lines above:
`$this->queuer->enqueue(clone $task, $task->getLabel());` So you don't have to get the `QueueDispatcher` service from the service locator again, it is already injected into `$this->queuer`.

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -206,4 +214,60 @@ private function getRemoteStatus(TaskInterface $task)
...
+     */
+    private function getUserFactoryService()
+    {
+        return $this->getServiceLocator()->get(UserFactoryServiceInterface::SERVICE_ID);
+    }
+
+    /**
+     * @return QueueDispatcherInterface
+     */
+    private function getQueueDispatcher()
```
gyszucs: So as I mentioned above, you won't need this method.

models/classes/webhooks/task/WebhookTask.php
```diff
@@ -32,10 +34,11 @@
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
-class WebhookTask extends AbstractAction implements TaskAwareInterface
+class WebhookTask extends AbstractAction implements TaskAwareInterface, ChildTaskAwareInterface
```
gyszucs: This task doesn't really need `ChildTaskAwareInterface` at all. This interface was mainly introduced for linking remote tasks to the local task which is not the case here.

models/classes/taskQueue/Worker/AbstractWorker.php
gyszucs: Furthermore, I think we should have a more strict condition here to clearly identify such tasks which has this retry behaviour otherwise any other type of tasks failed will be fall into it. Maybe a separate interface which even can define some related method, for example `getRetryCount`?

bartlomiejmarszal: > Why is it important to check whether it is an instance of `CallbackTaskInterface`?

I have access to WebhookTask through getCallable

Screenshot 2019-08-27 at 09 32 05

models/classes/taskQueue/Worker/AbstractWorker.php
bartlomiejmarszal: If I would like to determine that my WebhookTask is instance of RetriableInterface and that it will contain specific methods used to store/retrieve max retries count and current retry count ie.

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -67,12 +69,13 @@ public function __construct(QueuerInterface $queuer, TaskLogInterface $taskLog)
      */
     public function processTask(TaskInterface $task)
     {
-        if ($this->taskLog->getStatus($task->getId()) != TaskLogInterface::STATUS_CANCELLED) {
+        if (!$this->isTaskCancelled($task)) {
```
agalstyan: may be it would easier to have a following interface?
```php
if (!$task->isCancelled()) {
 ...
}
```

PR #2268, opened 2019-09-10
TAO-8740. Implementation base of tus protocol.

Yury-Razhkou:


models/classes/tusUpload/TusUploadServerService.php
```diff
@@ -0,0 +1,547 @@
...
+            return $this->prepareResponseData(null, self::HTTP_NOT_FOUND);
+        }
+
+        $isDeleted = $this->getCachePersistance()->delete($key);// todo: find how delete on KV_Storage works
+
+        if (!$isDeleted || !file_exists($resource)) {
+            return $this->prepareResponseData(null, self::HTTP_GONE);
+        }
+
+        unlink($resource);
```
zagovorichev: You're using file storage service here https://github.com/oat-sa/tao-core/pull/2268/files#diff-ca054dcc8678b3fa06c8687b47927faaR308, but you're trying to delete the file directly

PR #2282, opened 2019-09-20
Fix/tao 9154/rest api user label

yuryfilippovich:
Fix post api post request for creating uers (user creating with 2 labels)
Steps to reproduce send post request to /tao/api/users
Check-in created user in the database
 he will have 2 record with http://www.w3.org/2000/01/rdf-schema#label values 
1 from request and 2 'User' + login param

Expected behavior: 
user has only 1 label - from a request (if it was not in request then 'User' + login)


controller/api/Users.php
```diff
@@ -219,6 +220,12 @@ public function post()
                 $userService->attachRole($user, $this->getResource($role));
             }
 
+            if (array_key_exists(OntologyRdfs::RDFS_LABEL, $parameters)) {
```
zagovorichev: I guess that `$userService` have to be responsible for updating data in the `$user` model

models/classes/class.UserService.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 use oat\generis\model\GenerisRdf;
```
zagovorichev: could you move the header (with License) to the top of the file, please?

zagovorichev:
Everything works, good job, thank you :)

PR #2285, opened 2019-09-24
Fix/port explode

jbout:
https://oat-sa.atlassian.net/browse/NEX-334


serge-salamanka-1pt:
tested on Playground. @julien-sebire please merge

PR #2288, opened 2019-10-02
Fix/tao 9282/webhook task logging

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/TAO-9282

During testing of salesforce authorization I noticed our log messages not always informative enough.

`OAuth2AuthType` for example, throws guzzle's `RequestException` which isn't properly handled in `WebhookTask`. Also it throws `common_exception_ValidationFailed` without message (while `getUserMessage()` has one)

After CodeClimate complaints (methods count exceeded) I had to split WebhooksTask class and I extracted logging and reporting logic to separate WebhookTaskReports class

### DIFF
--- a/models/classes/webhooks/task/WebhookTask.php
+++ b/models/classes/webhooks/task/WebhookTask.php
@@ -224,7 +224,8 @@ class WebhookTask extends AbstractAction implements TaskAwareInterface
     /**
      * @return WebhookTaskReports
      */
-    private function getWebhookTaskReports() {
+    private function getWebhookTaskReports()
+    {
         /** @noinspection PhpIncompatibleReturnTypeInspection */
         return $this->getServiceLocator()->get(WebhookTaskReports::class);
     }
--- a/models/classes/webhooks/task/WebhookTask.php
+++ b/models/classes/webhooks/task/WebhookTask.php
@@ -26,7 +26,6 @@ use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Exception\ServerException;
 use GuzzleHttp\Psr7\Request;
 use oat\oatbox\extension\AbstractAction;
-use oat\oatbox\log\LoggerAwareTrait;
 use oat\tao\model\taskQueue\Task\TaskAwareInterface;
 use oat\tao\model\taskQueue\Task\TaskAwareTrait;
 use oat\tao\model\webhooks\configEntity\WebhookAuthInterface;
@@ -100,10 +99,13 @@ class WebhookTask extends AbstractAction implements TaskAwareInterface
     {
         try {
             $response = $this->getWebhookSender()->performRequest($request, $authConfig);
+        }
+        catch (ServerException $connectException) {
+            $this->retryTask();
+            return $this->getWebhookTaskReports()->reportBadResponseException(
+                $this->getTaskContext(), $connectException
+            );
         } catch (BadResponseException $badResponseException) {
-            if ($badResponseException instanceof ServerException) {
-                $this->retryTask();
-            }
             return $this->getWebhookTaskReports()->reportBadResponseException(
                 $this->getTaskContext(), $badResponseException
             );
--- a/models/classes/webhooks/task/WebhookTask.php
+++ b/models/classes/webhooks/task/WebhookTask.php
@@ -97,26 +97,32 @@ class WebhookTask extends AbstractAction implements TaskAwareInterface
      */
     private function performRequest(RequestInterface $request, WebhookAuthInterface $authConfig = null)
     {
+        $errorReport = $response =null;
         try {
             $response = $this->getWebhookSender()->performRequest($request, $authConfig);
-        }
-        catch (ServerException $connectException) {
+        } catch (ServerException $connectException) {
             $this->retryTask();
-            return $this->getWebhookTaskReports()->reportBadResponseException(
+            $errorReport = $this->getWebhookTaskReports()->reportBadResponseException(
                 $this->getTaskContext(), $connectException
             );
         } catch (BadResponseException $badResponseException) {
-            return $this->getWebhookTaskReports()->reportBadResponseException(
+            $errorReport = $this->getWebhookTaskReports()->reportBadResponseException(
                 $this->getTaskContext(), $badResponseException
             );
         } catch (ConnectException $connectException) {
             $this->retryTask();
-            return $this->getWebhookTaskReports()->reportConnectException($this->getTaskContext(), $connectException);
+            $errorReport = $this->getWebhookTaskReports()->reportConnectException(
+                $this->getTaskContext(), $connectException
+            );
         } catch (RequestException $requestException) {
-            return $this->getWebhookTaskReports()->reportRequestException($this->getTaskContext(), $requestException);
+            $errorReport = $this->getWebhookTaskReports()->reportRequestException(
+                $this->getTaskContext(), $requestException
+            );
         }
 
-        return $this->handleResponse($response);
+        return $response
+            ? $this->handleResponse($response)
+            : $errorReport;
     }
 
     /**
--- a/models/classes/webhooks/task/WebhookTask.php
+++ b/models/classes/webhooks/task/WebhookTask.php
@@ -31,7 +31,6 @@ use oat\tao\model\taskQueue\Task\TaskAwareInterface;
 use oat\tao\model\taskQueue\Task\TaskAwareTrait;
 use oat\tao\model\webhooks\configEntity\WebhookAuthInterface;
 use oat\tao\model\webhooks\configEntity\WebhookInterface;
-use oat\tao\model\webhooks\log\WebhookEventLogInterface;
 use oat\tao\model\webhooks\WebhookRegistryInterface;
 use oat\tao\model\webhooks\WebhookTaskServiceInterface;
 use Psr\Http\Message\RequestInterface;
@@ -39,7 +38,6 @@ use Psr\Http\Message\ResponseInterface;
 
 class WebhookTask extends AbstractAction implements TaskAwareInterface
 {
-    use LoggerAwareTrait;
     use TaskAwareTrait;
 
     /**
@@ -60,8 +58,7 @@ class WebhookTask extends AbstractAction implements TaskAwareInterface
             $request = $this->prepareRequest($webhookConfig);
             return $this->performRequest($request, $webhookConfig->getAuth());
         } catch (\Exception $exception) {
-            $this->logException($exception);
-            return \common_report_Report::createFailure($this->getExceptionMessage($exception));
+            return $this->getWebhookTaskReports()->reportInternalException($this->getTaskContext(), $exception);
         }
     }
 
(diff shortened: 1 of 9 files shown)


models/classes/webhooks/task/WebhookTask.php
```diff
@@ -326,12 +222,11 @@ private function getWebhookTaskService()
     }
 
     /**
-     * @return WebhookEventLogInterface
+     * @return WebhookTaskReports
      */
-    private function getWebhookEventLog()
-    {
+    private function getWebhookTaskReports() {
```
alexeyvy:
```suggestion
    private function getWebhookTaskReports()
    {
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+    public function setUp()
+    {
+        $this->webhookEventLogMock = $this->createMock(WebhookEventLogInterface::class);
+        $this->loggerMock = $this->createMock(LoggerInterface::class);
+    }
+
+    public function testReportInternalException()
+    {
+        $reports = new WebhookTaskReports();
+        $reports->setServiceLocator($this->getServiceLocatorMock([
```
bartlomiejmarszal: please move creating `ServiceLocator` mock into `setUp` method

models/classes/webhooks/task/WebhookTaskReports.php
```diff
@@ -0,0 +1,258 @@
...
+        $this->getWebhookEventLog()->storeInvalidHttpStatusLog($taskContext, $statusCode);
+        return $this->reportError($taskContext, "Response status code is $statusCode", $response);
+    }
+
+    /**
+     * @param WebhookTaskContext $taskContext
+     * @param ResponseInterface $response
+     * @return \common_report_Report
+     */
+    public function reportInvalidBodyFormat(WebhookTaskContext $taskContext, $response)
```
alexeyvy:
```suggestion
    public function reportInvalidBodyFormat(WebhookTaskContext $taskContext, ResponseInterface $response)
```

models/classes/webhooks/task/WebhookTaskReports.php
```diff
@@ -0,0 +1,258 @@
...
+            $badResponseException->getResponse()
+        );
+    }
+
+    /**
+     * @param WebhookTaskContext $taskContext
+     * @param ResponseInterface $response
+     * @return \common_report_Report
+     */
+    public function reportInvalidStatusCode(WebhookTaskContext $taskContext, $response)
```
alexeyvy:
```suggestion
    public function reportInvalidStatusCode(WebhookTaskContext $taskContext, ResponseInterface $response)
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
+use oat\generis\test\TestCase;
+use oat\tao\model\webhooks\log\WebhookEventLogInterface;
+use oat\tao\model\webhooks\task\WebhookResponse;
+use oat\tao\model\webhooks\task\WebhookTaskContext;
+use oat\tao\model\webhooks\task\WebhookTaskParams;
+use oat\tao\model\webhooks\task\WebhookTaskReports;
+use Psr\Log\LoggerInterface;
+
```
bartlomiejmarszal:
```suggestion
use PHPUnit_Framework_MockObject_MockObject;

```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+class WebhookTaskReportsTest extends TestCase
+{
+    /**
+     * @var WebhookEventLogInterface|\PHPUnit_Framework_MockObject_MockObject
```
bartlomiejmarszal:
```suggestion
     * @var WebhookEventLogInterface|PHPUnit_Framework_MockObject_MockObject
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+     */
+    private $webhookEventLogMock;
+
+    /**
+     * @var LoggerInterface|\PHPUnit_Framework_MockObject_MockObject
```
bartlomiejmarszal:
```suggestion
     * @var LoggerInterface|PHPUnit_Framework_MockObject_MockObject
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+
+    public function testReportInternalException()
+    {
+        $reports = new WebhookTaskReports();
+        $reports->setServiceLocator($this->getServiceLocatorMock([
+            WebhookEventLogInterface::SERVICE_ID => $this->webhookEventLogMock,
+        ]));
+        $reports->setLogger($this->loggerMock);
+
+        /** @var WebhookTaskParams|\PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```
bartlomiejmarszal:
```suggestion
        /** @var WebhookTaskParams|PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+
+    public function testReportConnectException()
+    {
+        $reports = new WebhookTaskReports();
+        $reports->setServiceLocator($this->getServiceLocatorMock([
+            WebhookEventLogInterface::SERVICE_ID => $this->webhookEventLogMock,
+        ]));
+        $reports->setLogger($this->loggerMock);
+
+        /** @var WebhookTaskParams|\PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```
bartlomiejmarszal:
```suggestion
        /** @var WebhookTaskParams|PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+        $taskParamsMock = $this->createMock(WebhookTaskParams::class);
+        $taskParamsMock->method('getEventId')->willReturn('eventId');
+
+        /** @var WebhookTaskContext|\PHPUnit_Framework_MockObject_MockObject $taskContextMock */
```
bartlomiejmarszal:
```suggestion
        /** @var WebhookTaskContext|PHPUnit_Framework_MockObject_MockObject $taskContextMock */
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+
+    public function testRequestExceptionWithResponse()
+    {
+        $reports = new WebhookTaskReports();
+        $reports->setServiceLocator($this->getServiceLocatorMock([
+            WebhookEventLogInterface::SERVICE_ID => $this->webhookEventLogMock,
+        ]));
+        $reports->setLogger($this->loggerMock);
+
+        /** @var WebhookTaskParams|\PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```
bartlomiejmarszal:
```suggestion
        /** @var WebhookTaskParams|PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,430 @@
...
+
+    public function testRequestExceptionWithoutResponse()
+    {
+        $reports = new WebhookTaskReports();
+        $reports->setServiceLocator($this->getServiceLocatorMock([
+            WebhookEventLogInterface::SERVICE_ID => $this->webhookEventLogMock,
+        ]));
+        $reports->setLogger($this->loggerMock);
+
+        /** @var WebhookTaskParams|\PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```
bartlomiejmarszal:
```suggestion
        /** @var WebhookTaskParams|PHPUnit_Framework_MockObject_MockObject $taskParamsMock */
```

models/classes/webhooks/task/WebhookTask.php
```diff
@@ -149,37 +123,26 @@ private function handleBadResponseException(BadResponseException $badResponseExc
...
+        if (!$this->isAcceptableResponseStatusCode($response->getStatusCode())) {
             $this->retryTask();
-            return $this->reportError("Response status code is $statusCode", $response);
+            return $this->getWebhookTaskReports()->reportInvalidStatusCode($this->getTaskContext(), $response);
         }
 
         $parsedResponse = $this->getWebhookResponseFactory()->create($response);
         if ($parsedResponse->getParseError()) {
-            $this->getWebhookEventLog()->storeInvalidBodyFormat($this->getTaskContext(), $response->getBody());
+            return $this->getWebhookTaskReports()->reportInvalidBodyFormat($this->getTaskContext(), $response);
```
alexeyvy: >return 

so now a parse error means there'll no further status recognition? It looks good because the previous implementation was a bit confusing

test/unit/webhooks/task/WebhookTaskReportsTest.php
siarhei-baradzin: done

models/classes/webhooks/task/WebhookTask.php
```diff
@@ -103,44 +101,20 @@ private function performRequest(RequestInterface $request, WebhookAuthInterface
         try {
             $response = $this->getWebhookSender()->performRequest($request, $authConfig);
         } catch (BadResponseException $badResponseException) {
-            return $this->handleBadResponseException($badResponseException);
+            if ($badResponseException instanceof ServerException) {
```
bartlomiejmarszal: Why we can't catch this error and perform this action?

test/unit/webhooks/task/WebhookTaskReportsTest.php
```diff
@@ -0,0 +1,357 @@
...
+        $this->assertContains(__FILE__, $report->getMessage());
+    }
+
+    public function testReportConnectException()
+    {
+        $reports = new WebhookTaskReports();
+        $reports->setServiceLocator($this->serviceLocatorMock);
+        $reports->setLogger($this->loggerMock);
+
+        $exception = new ConnectException('e_msg', new Request('POST', 'uru'));
```
bartlomiejmarszal: Is that mean that this is an integration test?

siarhei-baradzin: I hope no. You mean I should use mock instead?

models/classes/webhooks/task/WebhookTask.php
siarhei-baradzin: 1. that way you have to create temp variable, record report, retryTask(), return variable (less readable, can cause CodeClimate complaining about lines count in method)
2. in case of async queue (which should be used with webhooks) the order doesn't matter
3. in case of sync queue we perform retryTask() before returning report anyway

But I can change it

test/unit/webhooks/task/WebhookTaskReportsTest.php
bartlomiejmarszal: I think. Because if you are creating two objects and you assessing that they are working with each other, it's called integration test.

models/classes/webhooks/task/WebhookTask.php
bartlomiejmarszal: I don't understand this. 
1. Why do you have to create temporary variables? What kind of variables? Don't get it. 
```suggestion
          catch(ServerException $exception) {
              $this->retryTask();
          }
```

test/unit/webhooks/task/WebhookTaskReportsTest.php
siarhei-baradzin: I use Guzzle exception as a value object, no big difference from using mocks I think. Moreover, it's problematic to mock it (https://stackoverflow.com/a/35759802)

models/classes/webhooks/task/WebhookTask.php
siarhei-baradzin: no, it should be (as it was before) a branch in handling BadResponseException and `return $this->getWebhookTaskReports()->reportBadResponseException(...)` should still be called

bartlomiejmarszal: Why can't it be caught in another catch?

bartlomiejmarszal: This if looks silly here.

siarhei-baradzin: It can, I see no difference here

bartlomiejmarszal: As I said. this `if` statement looks not right here?

siarhei-baradzin: fixed

PR #2294, opened 2019-10-08
Fix/tao 9303/default settings storage values

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-9303

reason
Default CSP_HEADER_SETTING is not provided, that's why we have an issue on the tao installation

solution
CSP_HEADER_SETTING = 'self' by default

### DIFF
--- a/install/services/SetupSettingsStorage.php
+++ b/install/services/SetupSettingsStorage.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


install/services/SetupSettingsStorage.php
```diff
@@ -0,0 +1,36 @@
...
+use oat\oatbox\extension\InstallAction;
+use oat\tao\model\settings\CspHeaderSettingsInterface;
+use oat\tao\model\settings\SettingsStorageInterface;
+
+class SetupSettingsStorage extends InstallAction
+{
+    public function __invoke($params)
+    {
+        $settingsStorageService = $this->getServiceLocator()->get(SettingsStorageInterface::SERVICE_ID);
+        $settingsStorageService->set(CspHeaderSettingsInterface::CSP_HEADER_SETTING, 'self');
```
Skullsneeze: Please make sure that just providing self is sufficient. According to the CSP standard the keyword self needs to be wrapped in quotes, so this might need to be something like `"'self'"`

Skullsneeze: It's been a while since I worked on this. As long as you don't see any errors in your console it's all good ;)

PR #2295, opened 2019-10-09
Feature/nec 20/add lithuanian language

AndreyNegrov:
add new language (lt-LT)


locales/lt-LT/lang.rdfr
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tikhanovichA: I think we don't need that file

AndreyNegrov: fixed

PR #2296, opened 2019-10-09
TTD-18 Tests extends from `generis\test\TestCase`. All mock objects e…

Yury-Razhkou:
Related to: https://oat-sa.atlassian.net/browse/TTD-18
depends on: https://github.com/oat-sa/generis/pull/678
All tests extends `generis\test\TestCase` instead of `\PHPUnit_Framework_TestCase`
All Mock objects instances of `generis\test\MockObject` instead of `\PHPUnit_Framework_MockObject_MockObject`. Mocks was used only in phpdoc.
Possible way to test:
1) run unit tests.
2) update code.
3) run unit tests again. result should be the same as in 1.


manifest.php
```diff
@@ -54,10 +54,10 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '39.1.0',
+    'version' => '39.2.0',
```
bartlomiejmarszal: Major version have to be updated

bartlomiejmarszal: The required version doesn't change the major version so we change only minor version.

Yury-Razhkou: From all discussions there was a summary. If requered extension changes major version then we change major version too. If not then we can change minor version.
In this extension generis version changed from 12.3 to 12.5

PR #2300, opened 2019-10-14
Fixed database configurator.

julien-sebire:
This PR adds the option for Spanner as a DB engine in the installer.


julien-sebire:
Thanks Mike for pointing this out! ;-) Fixed.

PR #2302, opened 2019-10-16
added translations of phrases for the Lithuanian language

AndreyNegrov:
added translations of phrases for the Lithuanian language


manifest.php
```diff
@@ -54,7 +54,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '39.3.0',
+    'version' => '39.4.0',
```
tikhanovichA: Do we really need to up the minor version? Maybe just a patch? Because you just fixed some translations

AndreyNegrov: fixed

PR #2308, opened 2019-10-22
Fix/tao 7465/fix sync models

yuryfilippovich:


scripts/tools/rdf/LangRdfModelFix.php
```diff
@@ -0,0 +1,59 @@
...
+use oat\generis\model\kernel\persistence\file\FileIterator;
+use oat\oatbox\extension\AbstractAction;
+use tao_install_utils_Exception;
+use tao_install_utils_ModelCreator;
+
+
+/**
+ * sudo -u www-data php index.php 'oat\tao\scripts\tools\rdf\LangRdfModelFix' [--notLocal]
+ **/
+class LangRdfModelFix extends AbstractAction
```
tikhanovichA: Maybe we need to extend ScriptAction?

scripts/tools/rdf/LangRdfModelFix.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param $params
+     * @return Report
+     * @throws tao_install_utils_Exception
+     */
+    public function __invoke ($params)
+    {
+        $models = (new tao_install_utils_ModelCreator(LOCAL_NAMESPACE))->getLanguageModels();
+        $rdf = ModelManager::getModel()->getRdfInterface();
+        $total = 0;
+        $forceModelId = (isset($params[0]) && $params[0] == '--notLocal') ? null : 1;
```
tikhanovichA: Could you please change it to `dryrun` option? Like we did for other scripts

scripts/update/Updater.php
```diff
@@ -1241,5 +1241,11 @@ public function update($initialVersion) {
             $this->setVersion('39.1.0');
         }
         $this->skip('39.1.0', '39.3.1');
+
+        if ($this->isVersion('39.3.1')) {
+            OntologyUpdater::syncModels();
+            $this->addReport(new Report(Report::TYPE_WARNING, 'Run php index.php \'oat\tao\scripts\tools\rdf\LangRdfModelFix\''));
```
tikhanovichA: Can we add the same code into the Updater?

scripts/tools/rdf/LangRdfModelFix.php
yuryfilippovich: changed

scripts/tools/rdf/LangRdfModelFix.php
yuryfilippovich: added

scripts/update/Updater.php
yuryfilippovich: moved

PR #2312, opened 2019-10-29
NEX-487 - Error displayed after deleting a role, even though role is deleted

martijn-tao:
The response missed the `success => true` parameter, which is used by the front-end to verify if the request was successful.

### DIFF
--- a/actions/class.Roles.php
+++ b/actions/class.Roles.php
@@ -130,7 +130,6 @@ class tao_actions_Roles extends tao_actions_RdfController
      * @throws UserErrorException
      * @throws common_exception_BadRequest
      * @throws common_exception_MissingParameter
-     * @return void
      */
     public function delete()
     {
@@ -138,24 +137,25 @@ class tao_actions_Roles extends tao_actions_RdfController
             throw new common_exception_BadRequest('wrong request mode');
         }
 
-        if (!$this->getRequestParameter('uri')) {
-            $this->returnJson([
+        $parsedBody = $this->getPsrRequest()->getParsedBody();
+        if (!isset($parsedBody['uri'])) {
+            return $this->returnJson([
                 'deleted' => false,
                 'success' => false
             ]);
-            return;
         }
 
         $role = $this->getCurrentInstance();
+        $roleUri = $role->getUri();
 
-        if (in_array($role->getUri(), $this->forbidden)) {
-            throw new UserErrorException($role->getLabel() . ' could not be deleted');
+        if (in_array($roleUri, $this->forbidden)) {
+            throw new UserErrorException(sprintf('You are not allowed to delete the role "%s".', $role->getLabel()));
         }
 
         // Gather users associated with the role.
         $userClass = $this->getClass(GenerisRdf::CLASS_GENERIS_USER);
         $users = $userClass->searchInstances(
-            [GenerisRdf::PROPERTY_USER_ROLES => $role->getUri()],
+            [GenerisRdf::PROPERTY_USER_ROLES => $roleUri],
             ['recursive' => true, 'like' => false]
         );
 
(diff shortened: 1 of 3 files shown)


actions/class.Roles.php
```diff
@@ -125,42 +125,45 @@ public function assignUsers()
...
 
     /**
-     * Delete a group or a group class
+     * Delete a role.
+     *
      * @throws UserErrorException
      * @throws common_exception_BadRequest
-     * @throws common_exception_Error
      * @throws common_exception_MissingParameter
      * @return void
```
julien-sebire: Return type should be `string|bool`.

martijn-tao: Actually the response is set, but not returned as part of this method

actions/class.Roles.php
```diff
@@ -125,42 +125,50 @@ public function assignUsers()
...
-                        throw new UserErrorException(__('This role is still given to one or more users. Please remove the role to these users first.'));
-                    }
-                } else {
-                    throw new UserErrorException($role->getLabel() . ' could not be deleted');
-                }
-            }
+        }
 
-            $this->returnJson(array('deleted' => $deleted));
+        if (!$this->getRequestParameter('uri')) {
```
mike-ionut-mihai-sandu-tao: this is deprecated 
please try to use the data from '$this->getPsrRequest()' instad of  `$this->getRequest()`

actions/class.Roles.php
```diff
@@ -125,42 +125,50 @@ public function assignUsers()
...
+            $this->returnJson([
+                'deleted' => false,
+                'success' => false
+            ]);
+            return;
```
mike-ionut-mihai-sandu-tao: I think we can use
`return $this->returnJson` , 
I can see that `returnJson` has a  `void return type`
but the code does not look ok like this

actions/class.Roles.php
```diff
@@ -125,42 +125,50 @@ public function assignUsers()
...
         }
+
+        $role = $this->getCurrentInstance();
+
+        if (in_array($role->getUri(), $this->forbidden)) {
```
mike-ionut-mihai-sandu-tao: we are checking that the role has a URI that is inside the forbidden array and then we throw a `UserErrorException`
... is sounds strange

actions/class.Roles.php
```diff
@@ -125,42 +125,50 @@ public function assignUsers()
...
+            throw new UserErrorException($role->getLabel() . ' could not be deleted');
```
mike-ionut-mihai-sandu-tao: the message is `$role->getLabel() . ' could not be deleted'` but i think is should be different 
and expose the fact that the current role should not be deleted because the forbidden array (business rule) does not allow it

actions/class.Roles.php
```diff
@@ -125,42 +125,50 @@ public function assignUsers()
...
+
+        if (!empty($users)) {
+            throw new UserErrorException(__('This role is still given to one or more users. Please remove the role to these users first.'));
+        }
+
+        $deleted = $this->getClassService()->removeRole($role);
+
+        $this->returnJson([
+            'deleted' => $deleted,
+            'success' => $deleted
```
mike-ionut-mihai-sandu-tao: i think this is the only required change from the full development, so maybe because this is tao_core and it might impact other sides, let's just add this as a one-line change, and after refactor the function

martijn-tao: True, this is the actual fix, but I think we should follow the boy scout rule right?

mike-ionut-mihai-sandu-tao: totally agree, and I think that's a perfect approach,  but I would like to point out that `codeclimate` has the report that we have the cognitive complexity of 13, also by refactoring the code your ownership of the code will be self implied and mandatory,
also that the current refactored solution it is still complex and by definition a half baked solution

actions/class.Roles.php
julien-sebire: I think that if it's in the forbidden array means that it's forbidden to delete it...

actions/class.Roles.php
martijn-tao: The boy scout rule doesn't say fix everything. It says leave it better than it was. I think the changes that have been made improve the maintainability and readability of this piece of code. So while it might not fix the entire implementation, I still think we should keep it as is.

PR #2315, opened 2019-11-05
Adapt OauthService to be used in Lis auth

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/NEX-320

In order to reuse existing oauth request verification code in new Lis Outcomes Service I need some changes in oat\tao\model\oauth\OauthService:

1. Support of Psr requests added
2. Support of extracting oauth params from 'Authorization' header added
3. Added interface for DataStores which could be used in different implementations to formalize requirements (because despite the fact IMS library doesn't have such interface, it requires a list of methods from DataStore implementations)


models/classes/oauth/OauthService.php
```diff
@@ -154,11 +178,18 @@ private function buildAuthorizationHeader($signatureParameters) {
      */
     private function getOauthRequest(common_http_Request $request) {
         $params = array();
+
+        // In LRI launches oauth params are passed as POST params, but in LIS requests
```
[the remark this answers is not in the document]
siarhei-baradzin: fixed

models/classes/oauth/OauthService.php
```diff
@@ -108,24 +111,45 @@ public function sign(common_http_Request $request, \common_http_Credentials $cre
     /**
      * Validates the signature of the current request
      *
-     * @access protected
+     * @param \common_http_Request request
```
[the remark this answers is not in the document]
siarhei-baradzin: done

models/classes/oauth/OauthService.php
```diff
@@ -108,24 +111,45 @@ public function sign(common_http_Request $request, \common_http_Credentials $cre
...
 
+    /**
+     * Wrapper over parent validate method to support PSR Request object
+     *
+     * @param ServerRequestInterface $request
+     * @param \common_http_Credentials|null $credentials
+     * @return array [OAuthConsumer, token]
+     * @throws \common_http_InvalidSignatureException
+     */
+    public function validatePsrRequest(ServerRequestInterface $request, \common_http_Credentials $credentials = null)
```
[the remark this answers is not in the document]
siarhei-baradzin: There is no existing test for the whole class. And I don't see a good way to add tests without rewriting and separating this class, because it directly instantiates classes from IMS library (OAuthServer, for example, where the most of the logic located). Not sure such changes will be backward-compatible and are in the scope of my task

models/classes/oauth/ImsOauthDataStoreInterface.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
siwane: Could u specify behaviour/how to use methods through PHP doc? It allows to understand clearly how intefarce works and describe args usages

models/classes/oauth/OauthService.php
```diff
@@ -154,11 +180,17 @@ private function buildAuthorizationHeader($signatureParameters) {
      */
     private function getOauthRequest(common_http_Request $request) {
         $params = array();
+
+        // In LTI launches oauth params are passed as POST params, but in LIS requests
```
siwane: I do not like this code for special case, but required without new implementation

models/classes/oauth/ImsOauthDataStoreInterface.php
mike-ionut-mihai-sandu-tao: if they are self-described, or self implied please don't do that

siwane: agree it's why I'm asking, some arguments can be interpreted differently. Also it is a good practice & standart

PR #2317, opened 2019-11-05
fix retrieving user in lock srvice

hutnikau:


models/classes/user/implementation/UserLocksService.php
```diff
@@ -136,15 +136,18 @@ private function increaseLoginFails($login)
...
      */
     public function lockUser(User $user)
     {
-        $currentUser = UserHelper::getUser(tao_models_classes_UserService::singleton()->getCurrentUser());
+        $currentUser = tao_models_classes_UserService::singleton()->getCurrentUser();
 
         if (!$currentUser) {
             $currentUser = $user;
+        } else {
+            $currentUser = UserHelper::getUser(tao_models_classes_UserService::singleton()->getCurrentUser());
```
Alroniks, line 150: Seems `tao_models_classes_UserService::singleton()->getCurrentUser()` already assigned to `$currentUser`, so no need to fetch it again.  Solution would be `UserHelper::getUser($currentUser);`

But in this case, it makes sense to improve the whole condition.

models/classes/user/implementation/UserLocksService.php
```diff
@@ -233,12 +236,15 @@ public function isLockable(User $user)
     public function getLockoutRemainingTime($login)
     {
         $lastFailure = $this->getLockout()->getLastFailureTime($login);
+        $result = false;
```
Alroniks, line 239: It is possible to remove `$result` variable by inverting the condition and using return statements.

Alroniks:
Line it not changes, so it is a bit out of scope, but why in this line login retrieved from passed login, but identifier - from the current user?
```php
$this->getLockout()->setLockedStatus(UserHelper::getUserLogin($user), $currentUser->getIdentifier());
```

AndreyNegrov (approved):

PR #2319, opened 2019-11-07
Feature/tao 9579/non legacy controller

jbout:


models/classes/http/HttpFlowTrait.php
```diff
@@ -105,7 +104,7 @@ public function forwardUrl($url)
 
         //resolve the given URL for routing
         $resolver = new Resolver($request);
-        $this->propagate($resolver);
+        $resolver->setServiceLocator($this->getServiceLocator());
```
siwane: Why not using propagate? Is it to avoid smth that I'm not aware of?

bugalot: Same concern.

bugalot: I think I have my anwser @jbout. Is it because we have no guarantee that `$this->propagate` will be available? There is indeed a guarantee that `$this->getServiceLocator()` will be there.

models/classes/http/LegacyController.php
```diff
@@ -382,4 +383,17 @@ protected function logDeprecated($function = null)
         \common_Logger::i($message);
     }
 
+    /**
+     * Ensure the template is rendered as part of the response
+     * {@inheritDoc}
+     * @see \oat\tao\model\http\Controller::getPsrResponse()
+     */
+    public function getPsrResponse()
```
siwane: By doing that, you are forcing controllers to use LegacyController implementation. I understand it is isolating renderer from enforcer, does it the purpose of this PR?

PR #2322, opened 2019-11-11
OauthService: fix params involved in signature check

siarhei-baradzin:
https://oat-sa.atlassian.net/browse/NEX-320

The previous implementation had an issue with signature validating due to including too many params to base string of signature (got from `$request->getServerParams()`). According to https://tools.ietf.org/html/rfc5849#section-3.4.1.3.1 OAuth 1.0 spec we have to include only oauth params from Authorization header (excluding oauth_signature itself), query params (done in OAuthRequest internally) and POST params if content-type is `application/x-www-form-urlencoded`

We performed a test (together with @boajer) and checked signatures compatibility with deliver-be. We found they are compatible except 2 extra params used in deliver-be which we believe should be removed because they couldn't be found in request (lti_version and lti_message_type) and suitable only for LTI launch calls.


models/classes/oauth/OauthService.php
```diff
@@ -142,23 +142,39 @@ public function validate(common_http_Request $request, common_http_Credentials $
...
     }
 
+    /**
+     * @param ServerRequestInterface $request
+     * @return common_http_Request
+     */
+    private function buildCommonRequestFromPsr(ServerRequestInterface $request)
+    {
+        $body = (string) $request->getBody();
+        // https://tools.ietf.org/html/rfc5849#section-3.4.1.3.1
```
mike-ionut-mihai-sandu-tao: nice

models/classes/oauth/OauthService.php
```diff
@@ -142,23 +142,39 @@ public function validate(common_http_Request $request, common_http_Credentials $
...
      */
     public function getDataStore() {
         return $this->getSubService(self::OPTION_DATASTORE);
     }
 
+    /**
+     * @param ServerRequestInterface $request
+     * @return common_http_Request
+     */
+    private function buildCommonRequestFromPsr(ServerRequestInterface $request)
```
mike-ionut-mihai-sandu-tao: are we able to have tests for this ?

models/classes/oauth/OauthService.php
```diff
@@ -142,23 +142,39 @@ public function validate(common_http_Request $request, common_http_Credentials $
...
+    {
+        $body = (string) $request->getBody();
+        // https://tools.ietf.org/html/rfc5849#section-3.4.1.3.1
+        $contentTypeHeaders = $request->getHeader('Content-Type');
+        $params = reset($contentTypeHeaders) === 'application/x-www-form-urlencoded'
```
mike-ionut-mihai-sandu-tao: i did not saw why we need to reset the array maybe you can provide some details

models/classes/oauth/OauthService.php
siarhei-baradzin: I still can't do with small changes :( due to direct OAuthUtil (and other classes from lti lib) dependency

models/classes/oauth/OauthService.php
siarhei-baradzin: `$request->getHeader(...)` returns an array

mike-ionut-mihai-sandu-tao: reset — Set the internal pointer of an array to its first element
so maybe because of your checking against `=== 'application/x-www-form-urlencoded'` 
i think you might need `current($contentTypeHeaders)` instead

siarhei-baradzin: `reset` also returns the first element while `current` returns an element based on internal pointer position (which is unknown for array created by `$request->getHeader()`, it could lead to undefined behavior)

PR #2324, opened 2019-11-12
Added support of .env file for standalone scripts (e.g. taoInstall.php).

julien-sebire:
This PR enables reading of optional `.env` file in project root directory (see https://symfony.com/doc/current/components/dotenv.html for more details) in all standalone scripts requiring `tao/install/init.php`.
The same may be done for other parts of the application. .env is already available for the application as soon as it is bootstraped.


install/init.php
```diff
@@ -30,7 +33,13 @@
 require_once ($root . 'vendor/autoload.php');
 
 if (tao_install_utils_System::isTAOInstalled()) {
-    require_once ($root . '/config/generis.conf.php');
+    require_once ($root . 'config/generis.conf.php');
+}
+
+$envFile = $root . '.env';
+if (file_exists($envFile)) {
+    $dotenv = new Dotenv();
```
boajer: That is the 3rd place in the platform with env file loading. Might be we should reduce code duplication by wrapping it into class?

peetya: Agree. I'm not familiar with the infrastructure, but in best-case scenario we should define it only once. If TAO has multiple entrypoints (for instance web, cli, worker, etc...) we need to make a reusable class.

install/init.php
```diff
@@ -30,9 +33,11 @@
 require_once ($root . 'vendor/autoload.php');
 
 if (tao_install_utils_System::isTAOInstalled()) {
-    require_once ($root . '/config/generis.conf.php');
+    require_once ($root . 'config/generis.conf.php');
```
mike-ionut-mihai-sandu-tao: why did you remove the leading `/`

install/init.php
```diff
@@ -19,6 +19,9 @@
  *               2009-2012 (update and modification) Public Research Centre Henri Tudor (under the project TAO-SUSTAIN & TAO-DEV);
  *               2013-     (update and modification) Open Assessment Technologies SA;
  */
+
+use oat\tao\model\mvc\DotEnvReader;
```
mike-ionut-mihai-sandu-tao: should we use the service locator for this?

manifest.php
```diff
@@ -54,7 +54,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '39.5.2',
+    'version' => '39.6.0',
```
mike-ionut-mihai-sandu-tao: why is this a major?

julien-sebire: Afaik, in semver:
- Backward compatible breaking change => major
- New feature => minor
- Bug fix => release

install/init.php
julien-sebire: It's trailing in the `$root` variable already. This was doubling the `/`.

install/init.php
julien-sebire: I don't think so. This is very static indeed and has no effect if the file is not present.

models/classes/mvc/DotEnvReader.php
```diff
@@ -0,0 +1,45 @@
...
+{
+    /**
+     * Reads .env file into $_ENV.
+     *
+     * @param string $envFile (defaults to project .env file)
+     */
+    public function __construct($envFile = '')
+    {
+        if ($envFile === '') {
+            $envFile = __DIR__ . DIRECTORY_SEPARATOR . '..' . DIRECTORY_SEPARATOR . '..' . DIRECTORY_SEPARATOR
```
martijn-tao: To make perhaps make this a bit more readable, you could do something like:

```
__DIR__  . implode(DIRECTORY_SEPARATOR, array_fill(1, 4, '..')) . '.env'
```

install/init.php
hutnikau: Please use `DIRECTORY_SEPARATOR` constant in paths.

mike-ionut-mihai-sandu-tao:

PR #2325, opened 2019-11-12
prevent count warning on recent php>=7.2

krampstudio:
Related to https://oat-sa.atlassian.net/browse/TAO-9602

Prevents the PHP warning :
image

How to reproduce : 
 - PHP >= 7.2
 - Open a page using a `FormContainer` and check the error logs

### DIFF
--- a/helpers/form/class.FormContainer.php
+++ b/helpers/form/class.FormContainer.php
@@ -102,7 +102,6 @@ abstract class tao_helpers_form_FormContainer
 
         // set the values in case of default values
         if ($this->data && count($this->data) > 0) {
-        :qa
             $this->form->setValues($this->data);
         }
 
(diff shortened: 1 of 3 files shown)


helpers/form/class.FormContainer.php
```diff
@@ -101,7 +101,7 @@ public function __construct($data = [], $options = [])
         }
 
         // set the values in case of default values
-        if (count($this->data) > 0) {
+        if ($this->data && count($this->data) > 0) {
```
SergiiTao: as we don't have type hints i would do: 
`if (is_array($this->data) && !empty($this->data)) {...}`

or
```
$this->data = is_array($data) ? $data : [];
if (!empty($this->data)) {...}
```

krampstudio:
@SergiiTao can you please check again

PR #2326, opened 2019-11-14
Fix/tao 9550/added not disabled

tikhanovichA:
https://oat-sa.atlassian.net/browse/TAO-9550

For parsing elements from sub-classes in tree we need to the new callback for onopen event


AndreyNegrov (approved):

ricproenca-tao (changes requested):
Need to bump version again...

tikhanovichA:
> Need to bump version again...

Done

ricproenca-tao (approved):

PR #2341, opened 2019-11-26
ECS fixer applied

siwane:


actions/class.CommonRestModule.php
```diff
@@ -37,33 +38,32 @@ public function index()
     {
         try {
             $uri = null;
-            if ($this->hasRequestParameter("uri")) {
-                $uri = $this->getRequestParameter("uri");
-                if (!common_Utils::isUri($uri)) {
+            if ($this->hasRequestParameter('uri')) {
+                $uri = $this->getRequestParameter('uri');
+                if (! common_Utils::isUri($uri)) {
```
martijn-tao: Is it part of the PSR-12 standard to put a space after the `!` here? I don't think I've ever seen this before, and may need to be adjusted in the config

PR #2346, opened 2019-12-04
Fix/tao 8584/unit test dependency propagation

jbout:
* Use the service locator to be able to mock the user session
* Prevent use of static function to ensure ontology mock gets passed correctly


zagovorichev:
@jbout Could you please add a description and merge the `develop` branch here?

PR #2377, opened 2020-01-22
NFER-33 Remove text error before checkLogin for avoid conflicts

habemuscode:
https://oat-sa.atlassian.net/browse/NFER-33

Two conflicting text messages occur when you try to save a new test taker. Try to create a new test-taker with an existing login. Then try to create a new test-taker with a unique login and you will see 2 conflicting text messages below the login field.


views/js/users.js
```diff
@@ -16,7 +16,7 @@ define(['jquery', 'i18n', 'context'], function($, __, context){
                 // trim value
                 var trimmedValue = elt.val().replace(/^\s+/g,'').replace(/\s+$/g,'');
                 var value = elt.val();
-
+                $('input#http_2_www_0_tao_0_lu_1_Ontologies_1_generis_0_rdf_3_login ~ div.form-error').remove();
```
bziondik: @SergiiTao I am not sure that this id is the same in any instances of tao
Better to select by attribute `value="user"`

SergiiTao: `http://www.tao.lu/Ontologies/generis.rdf#login` definitely always the same

bziondik: thanks!

habemuscode: For me use value is tricky, because user can write the same in login and in password for example (and in repeat password), so you have 3 choices to manage. Another thing i think is use name attribute but has same value than id or use position in form (5 element) but i dont like because block in the future change the positions of input elements.

PR #2383, opened 2020-02-03
Added alphabetical sorting to the list of installed extensions

Alroniks:
Work done under task https://oat-sa.atlassian.net/browse/BOSA-23

It adds sorting to this list:
TAO 3 4 0-sprint121


actions/class.ExtensionsManager.php
```diff
@@ -148,6 +160,9 @@ public function disable()
     public function enable()
     {
         $this->assertIsDebugMode();
+
+        $this->getPsrRequest()->getQueryParams();
```
Yury-Razhkou: Why do we need this here?

Alroniks: Ops, forget to remove some experiments

Alroniks:
I've rerun checks manually in Jenkins and it works. Seems it was an issue when the commit was sent but now tests are passing.

actions/class.ExtensionsManager.php
```diff
@@ -35,24 +35,36 @@
...
+    /**
+     * @param array $extensions
+     *
+     * @return array
+     */
+    private function sortExtensionList(array $extensions = [])
+    {
+        usort($extensions, static function ($a, $b) {
+            /** @var common_ext_Extension $a */
+            /** @var common_ext_Extension $b */
```
wazelin: Better to set argument types of the anonymous function IMO.

wazelin: Oh, I was a couple of seconds late with that comment.

PR #2392, opened 2020-02-10
Fix/ncc 110/trigger select on items list loaded

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/NCC-110
**Description:**
On the "new assignment" when the user clicks on a sub-folder, event "change" is fired and handler subscribed on that event checks if folders has any selected items.
In case there is at least one selected item the button "Assign" change state to enabled.
The problem is on the first click on sub-folder there is no items and we do request to load them
When data for items is loaded the "change" event has been fired before items are rendered.
That's why we have to notify handler that items are loaded and rendered

**Steps to reproduce:** 
1. Log in as Sponsor
2. Go to Child Organization
3. Go to test Assignments
4. Click the "Add New assignment" button
5. Click the folder containing the desired battery

**Unit tests cli:** `npm run test`

### DIFF
--- a/views/js/generis.tree.select.js
+++ b/views/js/generis.tree.select.js
@@ -132,36 +132,26 @@ define([
                     //when a node is selected
                     onselect: function(NODE, TREE_OBJ) {
                         var servOptions = {};
+                        var $NODE = $(NODE);
+                        var $nodeParent = $(NODE)
+                            .parent()
+                            .parent();
+
                         if (instance.serverParameters.hasOwnProperty('order')) {
                             servOptions.order = instance.serverParameters.order;
                         }
                         if (instance.serverParameters.hasOwnProperty('orderdir')) {
                             servOptions.orderdir = instance.serverParameters.orderdir;
                         }
-                        if ($(NODE).hasClass('paginate-more')) {
-                            instance.paginateInstances(
-                                $(NODE)
-                                    .parent()
-                                    .parent(),
-                                TREE_OBJ,
-                                servOptions
-                            );
+                        if ($NODE.hasClass('paginate-more')) {
+                            instance.paginateInstances($nodeParent, TREE_OBJ, servOptions);
                             return;
                         }
-                        if ($(NODE).hasClass('paginate-all')) {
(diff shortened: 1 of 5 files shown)


views/js/generis.tree.select.js
```diff
@@ -225,234 +252,264 @@ define(['jquery', 'lodash', 'i18n', 'context', 'generis.tree', 'helpers', 'ui/fe
...
                             }
-                        });
-                    } else {
-                        instance.checkedNodes = options.checkedNodes;
+                        }
+
+                        instance.check(instance.checkedNodes);
+
+                        //Execute callback;
+                        if (instance.options.onChangeCallback) {
```
andrey-ponamarev: Here is the only one change in this file

views/js/generis.tree.select.js
```diff
@@ -13,133 +13,150 @@
...
+                    onchange: function(NODE, TREE_OBJ) {
+                        if (instance.options.onChangeCallback && !instance.isRefreshing) {
+                            instance.options.onChangeCallback(NODE, TREE_OBJ);
+                        }
+                    },
+                    //when a node is selected
+                    onselect: function(NODE, TREE_OBJ) {
+                        var servOptions = {};
+                        var $NODE = $(NODE);
+                        var $nodeParent = $(NODE)
```
btamas: Why do you select the node again?
`const $nodeParent = $NODE.parent().parent()`

views/js/generis.tree.select.js
```diff
@@ -225,234 +242,260 @@ define(['jquery', 'lodash', 'i18n', 'context', 'generis.tree', 'helpers', 'ui/fe
...
                             }
-                        });
-                    } else {
-                        instance.checkedNodes = options.checkedNodes;
+                        }
+
+                        instance.check(instance.checkedNodes);
+
+                        //Execute callback;
+                        if (instance.options.onChangeCallback) {
```
btamas:
```js
const { onChangeCallback } = instance.options;
if (onChangeCallback) {
  onChangeCallback();
}
```

PR #2394, opened 2020-02-12
Fix/cog 29/add text encoder polyfill to support ie

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/COG-29
**Description:** Polyfill for [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) is added to support IE
**Steps to reproduce:** 
1. Open IE11 
2. Login as ART admin (fosterabigail/Dm7ZOeu8W*)
3. Create a teacher
4. Create a student with the same district and school as a teacher's
5. Assign several tests to a student with the created teacher, also verify in Calendars that assessments are not expired
6. Log in as a teacher, go to Students tab and select your student

**Unit tests cli:** `npm run test`


views/js/lib/polyfill/text-encoder.js
```diff
@@ -0,0 +1,8 @@
+/* From https://github.com/anonyco/FastestSmallestTextEncoderDecoder
+ * @license Licensed under MIT license
+ * See https://raw.githubusercontent.com/jakearchibald/es6-promise/master/LICENSE
+ * @version 1.0.14
+*/
+'use strict';(function(f){function r(b){var a=b.charCodeAt(0)<<24,c=t(~a)|0,d=0,m=b.length|0,g="";if(5>c&&m>=c){a=a<<c>>>24+c;for(d=1;d<c;d=d+1|0)a=a<<6|b.charCodeAt(d)&63;65535>=a?g+=e(a):1114111>=a?(a=a-65536|0,g+=e((a>>10)+55296|0,(a&1023)+56320|0)):d=0}for(;d<m;d=d+1|0)g+="\ufffd";return g}function n(){}function u(b){var a=b.charCodeAt(0)|0;if(55296<=a&&56319>=a){var c=b.charCodeAt(1)|0;if(c===c&&56320<=c&&57343>=c){if(a=(a-55296<<10)+c-56320+65536|0,65535<a)return e(240|a>>>18,128|a>>>12&63,
```
jsconan: Caution, this code looks not safe as it is forcing the strict mode to the whole context, that might impact other scripts while concatenating (cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode).

Prefer to import source code instead of minified content when this content looks harmful. This will prevent silly issues due to the uglify/babel transformation upon bundling, as well as concatenation issues.
When necessary, you may also consider wrapping the content into a proper UMD.

PR #2395, opened 2020-02-13
Require trait instead of generated part of code

Alroniks:
Related to task https://oat-sa.atlassian.net/browse/NFER-31

Before the method was generated from the template, that isn't correct for several reasons.
1. Code on PHP written in template file, that is not correct, because it can be formatted and validated by IDE tools or other automated checks.
2. This method written in extension, that isn't required in the core, but code used in the core project. Better when code related to the application stored in the one codebase. Also, if someone will change this part of code it will be overridden during the next generation that may break the app. 

So, based on two points above I moved the method that generates icons to trait and I include it as part of the generated helper.


models/classes/iconBuilder/IconBuilderTrait.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param string $icon    name of the icon to display
+     * @param array  $options (optional) hashtable with HTML attributes, also allows to set element="almostAnyHtmlElement"
+     *
+     * @return string HTML element with icon
+     */
+    protected static function buildIcon($icon, array $options = [])
+    {
+        $dom = new DOMDocument();
+
+        $element = $options['element'];
```
wazelin: Since `$options` is an optional argument, `element` should be extracted with null coalescing `$element = $options['element'] ?? 'span';`

models/classes/iconBuilder/IconBuilderTrait.php
```diff
@@ -0,0 +1,49 @@
...
+        unset($options['element']);
+
+        $options['class'] = trim(implode(' ', [$options['icon'], $icon]));
```
wazelin: [Same](https://github.com/oat-sa/tao-core/pull/2395/files#r385202432) applies here.

models/classes/iconBuilder/IconBuilderTrait.php
Alroniks: Nice catch! Fixed

PR #2399, opened 2020-02-17
Feature/phpunit 7 ugrade

mike-ionut-mihai-sandu-tao:
Updated phpunit to version 7.0.3
tested on : PHP 7.2.24 (cli) (built: Oct 25 2019 04:29:43)

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -1,5 +1,7 @@
 <?php
 
+use oat\tao\scripts\update\Updater;
+
 /**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


test/unit/extensionManager/ExtensionManagerTest.php
```diff
@@ -2,6 +2,7 @@
 
 namespace oat\tao\test\unit\extensionManager;
 
+use oat\generis\test\PHPUnit_Framework_ExpectationFailedException;
```
julien-sebire: I don't get the idea of using a non-namespaced class here, and a namespaced on in the file below.

manifest.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use oat\tao\scripts\update\Updater;
```
julien-sebire: No change in this class is related to the subject of the PR.

PR #2402, opened 2020-02-19
[WIP][ACTP-310] Login Page - Fix HTML structure

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/ACTP-310
 
Update home link title and page title of login page
  
#### How to test
 
- prepare an instance of TAO with taoAct extension
- navigate to login page and check logo home link title and page title
- make sure to check with fresh installation and with update using updater script 
  
#### Dependencies

Companion PR:
 - [ ] https://github.com/oat-sa/extension-tao-act/pull/1306

### DIFF
--- a/config/default/login.conf.php
+++ b/config/default/login.conf.php
@@ -27,5 +27,6 @@ return [
      * @type bool
      */
     'disableAutocomplete' => false,
-    'enablePasswordReveal' => false
+    'enablePasswordReveal' => false,
+    'title' => 'TAO Login'
 ];
--- a/actions/class.Main.php
+++ b/actions/class.Main.php
@@ -232,7 +232,7 @@ class tao_actions_Main extends tao_actions_CommonModule
             }
         }
 
-        $this->setData('title', __("TAO Login"));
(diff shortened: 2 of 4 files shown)


config/default/login.conf.php
```diff
@@ -27,5 +27,6 @@
      * @type bool
      */
     'disableAutocomplete' => false,
-    'enablePasswordReveal' => false
+    'enablePasswordReveal' => false,
+    'title' => 'TAO Login'
```
jsconan: Unfortunately, this is not translatable. Could you use a service instead?

AntonTsymuk: yes, but i will pass this value later into i18n helper to get translation inside the login controller

AntonTsymuk: https://github.com/oat-sa/tao-core/pull/2402/files#diff-ce7f65685c6da98902fa941ad7f8e4b5R235

jsconan: A dedicated service would avoid possible translation issues. Anyway, it would require a backend dev.

PR #2404, opened 2020-02-19
Fix/nex-683/ontology lang sync

jbout:
Due to languages moving from model id 1 to 2 there is of losing language definitions when running OntologyUpdater::syncModels().
But we also need to ensure not to duplicate language definitions either


julien-sebire:
Reproduced steps:
- basic installation with all extensions on `develop` branches.
- `php index.php "oat\tao\scripts\SyncModels"`
- login breaks (500 error)
- get new code (pull `fix/ontology-lang-sync` branch of `tao-core`)
- `php tao/scripts/taoUpdate.php` => `Successfully updated tao to 40.9.6`
- login works

julien-sebire:
Checked with following steps:
- basic installation with all extensions on `develop` branches.
- `php index.php "oat\tao\scripts\SyncModels"`
- login breaks (500 error)
- get new code (pull `fix/ontology-lang-sync` branch of `tao-core`)
- `php tao/scripts/taoUpdate.php` => `Successfully updated tao to 40.9.6`
- login works

PR #2405, opened 2020-02-21
Fix/tbd 2/css fix

Yury-Razhkou:
Related to: https://oat-sa.atlassian.net/browse/TBD-2
Ways to test:
Open `TAO_URL/ltiDeliveryProvider/DeliveryRunner/runDeliveryExecution` back button will appears only if HTTP_REFERER is the same as TAO_URL.


views/templates/error/error500.tpl
```diff
@@ -10,7 +10,7 @@
 </head>
 
 <body>
-
+<?php xdebug_break();?>
```
Alroniks: Seems it should be removed from production code

models/classes/mvc/error/HtmlResponse.php
```diff
@@ -29,15 +29,19 @@
...
+
     public function send()
     {
         if (DEBUG_MODE) {
             $message = $this->exception->getMessage();
             $trace = $this->exception->getTraceAsString();
         }
+        $returnUrl = $_SERVER['HTTP_REFERER'];
+        if (parse_url($returnUrl, PHP_URL_HOST) != parse_url(ROOT_URL, PHP_URL_HOST)) {
+            $returnUrl = false;
```
Alroniks: Where this var is used? It is not passed anywhere.

Yury-Razhkou: It is the way how templates rendered here :`require Template::getTemplate('error/error'...`
This variable used in template:
```
<?php if (!empty($returnUrl)) : ?>
<li><a href="<?= $returnUrl ?>"><?=__('Go Back')?></a></li>
<?php endif; ?>
```

views/templates/error/error500.tpl
Yury-Razhkou: definitely yes! thanks.

models/classes/mvc/error/HtmlResponse.php
SergiiTao: IMHO looks a bit more readable. but not important

```php
$referer = $_SERVER['HTTP_REFERER'];

$returnUrl = (parse_url($referer, PHP_URL_HOST) === parse_url(ROOT_URL, PHP_URL_HOST))
    ? htmlentities($referer, ENT_QUOTES)
    : false;
```

PR #2409, opened 2020-02-27
create tao migrations script

hutnikau:
Crete new migration in taoQtiTest extension:
```
sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c generate -e taoQtiTest
```
Show migrations status
```
sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c status
```
Apply migrations
```
sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c migrate
```

Migrations are located in `migrations` folder of extension root.

See documentation page: https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/279318424/Tao+Updater+Migrations

requires: https://github.com/oat-sa/generis/pull/773


scripts/tools/Migrations.php
```diff
@@ -0,0 +1,223 @@
+<?php
+
+namespace oat\tao\scripts\tools;
+
+use Doctrine\DBAL\Connection;
+use oat\tao\scripts\tools\migrations\Configuration;
+use Doctrine\Migrations\Tools\Console\Command;
```
SergiiTao: I think we need to add Doctrine and Symfony stuff into composer

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,223 @@
...
+ * sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c generate -e taoAct
+ * sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c status
+ * sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c migrate
+ * ```
+ * @package oat\tao\scripts\tools
+ */
+class Migrations extends ScriptAction
+{
+
+    const MIGRATIONS_DIR = 'migrations';
```
SergiiTao: if we are using 7.2 we need to add a visibility modifier
as well as return types and declare strict_type if it is possible

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,223 @@
...
+
+    private $commands = [
+        'generate' => 'migrations:generate',
```
SergiiTao: it would be more convenient to use constants here and below.

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,223 @@
...
+        $helperSet = new HelperSet();
+        $helperSet->set(new QuestionHelper(), 'question');
+        $helperSet->set(new ConfigurationHelper($connection, $configuration));
+        $configuration->setMigrationsDirectory(ROOT_PATH);
+        $configuration->setMigrationsFinder(new TaoFinder(ROOT_PATH));
+        $configuration->setMigrationsNamespace('oat');
+        $this->execute($helperSet, $input);
+    }
+
+    private function migrate()
```
SergiiTao: status and migrate methods almost identical

scripts/tools/migrations/AbstractMigration.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+
+namespace oat\tao\scripts\tools\migrations;
+
+use Doctrine\Migrations\AbstractMigration as DoctrineAbstractMigration;
+
+abstract class AbstractMigration extends DoctrineAbstractMigration
```
SergiiTao: why do we need it?

migrations/Version20200225120234.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
SergiiTao: do we need it?

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,223 @@
...
+            case 'status':
+                $this->status();
+                break;
+            case 'migrate':
+                $this->migrate();
+                break;
+        }
+    }
+
+    private function generate()
```
SergiiTao: does it make sense to inject commands?

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,281 @@
...
+ * //show migrations status
+ * sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c status
+ * //apply all migrations
+ * sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c migrate
+ * //migrate to version
+ * sudo -u www-data php index.php '\oat\tao\scripts\tools\Migrations' -c migrate -v 202003120846502234_tao
+ * ```
+ * @package oat\tao\scripts\tools
+ */
+class Migrations extends ScriptAction
```
wazelin: This looks more like a CLI application, and potentially not something migration-specific.
What do you think of making it more generic and call `addCommand()` on it externally?

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /**
+     * This is an assumption
+     * @param common_ext_Extension $extension
+     * @return string
+     */
+    private function getExtensionNamespace(common_ext_Extension $extension)
+    {
+        return 'oat\\'.$extension->getId().'\\migrations';
```
wazelin: The namespace/location should probably be added to the README.

scripts/tools/Migrations.php
hutnikau: Do you mean that we could create an abstract command which could be used by other cli scripts?
I thinks this is out of scope of this task.

scripts/tools/Migrations.php
hutnikau: It is not something to be configured or defined manually.
Location/namespace will be the same for all extensions so there is no need to document it.

wazelin: Yes, and that's exactly why I suggest to document it :)
It is not modifiable and an extension author should know where to place the migration scripts.

scripts/tools/Migrations.php
wazelin: Can't argue with that. It's your call.

models/classes/extension/UpdateExtensions.php
```diff
@@ -50,17 +51,23 @@ public function __invoke($params)
         }
         $report = parent::__invoke($params);
 
-        // regenrate locals
+        $migrations = new Migrations();
```
SashaTAO: I'm not sure if it's safe to include migrations execution into Updater script by default. Our current agreement is that we execute db migrations after TAO update.

scripts/tools/Migrations.php
hutnikau: I'm not sure if readme file is correct place to put such technical details. Maybe it's better to create a documentation page in confluence related to new update process?

models/classes/extension/UpdateExtensions.php
hutnikau: This is not only for DB migrations. 
This feature supposed to replace current updater process

scripts/tools/migrations/AbstractMigration.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * Temporary helper until the service manager
+     * gets properly injected migration scripts
+     *
+     * @return ServiceManager
+     */
+    public function getServiceManager(): ServiceManager
+    {
+        return ServiceManager::getServiceManager();
```
jbout, line 19: this will make unit tests of migrations impossible

wazelin: It will complicate things, but won't make it impossible. One option is to use a partial mock of the migration class as a SUT. The other one is Mockery, which allows overloading static calls by mangling with autoloader a bit.

composer.json
```diff
@@ -57,7 +57,8 @@
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
         "oat-sa/composer-npm-bridge": "^0.3.0",
-        "symfony/dotenv": "~4.3"
+        "symfony/dotenv": "~4.3",
+        "doctrine/migrations" : "^2.0"
```
wazelin, line 61: Not a big deal, but `--sort-packages` is a nice option to use when running `composer require`.

BTW, it can also be applied by default if added in `composer.json`'s `config`.

scripts/tools/migrations/Configuration.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function generateVersionNumber(?DateTimeInterface $now = null) : string
+    {
+        if ($this->extension === null) {
+            throw new UnknownMigrationVersion('Unknown extension');
+        }
+        $version = parent::generateVersionNumber($now);
+        $intHash = crc32($this->extension->getId());
+        $intHash = substr($intHash,0,4);
+        return $version.$intHash.'_'.$this->extension->getId();
```
wazelin, line 45: I don't really understand why it's needed to have a part of a crc32, based on the extension's ID, and then use the extension ID in the key again.
`$intHash` may be omitted without any impact here IMO. I mean, it doesn't introduce any additional information, and won't help in avoiding collisions at all.

scripts/tools/migrations/AbstractMigration.php
jbout: Also, please do not test mock objects whenever possible

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,326 @@
...
+        $cli->setAutoExit(false);
+        $cli->setCatchExceptions(true);
+        $cli->setHelperSet($helperSet);
+        $cli->setCatchExceptions(false);
+        $cli->addCommands(array(
+            new migrations\commands\GenerateCommand(),
+            new Command\MigrateCommand(),
+            new Command\StatusCommand(),
+            new Command\ExecuteCommand(),
+            //new Command\DumpSchemaCommand(),
```
SergiiTao, line 244: we do not need those comments. right?

scripts/tools/migrations/Configuration.php
hutnikau: It's needed to ensure uniquness of the migration name. Doctrine migration tool omit letters and relay only on numbers to buld migratoin plan

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,365 @@
...
+    protected const COMMAND_ADD = 'add';
+
+    private $commands = [
+        self::COMMAND_GENERATE => 'migrations:generate',
+        self::COMMAND_STATUS => 'migrations:status',
+        self::COMMAND_MIGRATE => 'migrations:migrate',
+        self::COMMAND_EXECUTE => 'migrations:execute',
+        self::COMMAND_ROLLBACK => 'migrations:execute',
+        self::COMMAND_ADD => 'migrations:version',
+    ];
```
wazelin, line 61: Can be defined as a constant.

scripts/tools/migrations/TaoFinder.php
```diff
@@ -0,0 +1,52 @@
...
+        $dir = rtrim($this->getRealPath($directory), '\\/');
+        $root = rtrim($this->getRealPath(ROOT_PATH), '\\/');
+        if ($dir === $root) {
+            //global search
+            $pattern = $dir . '/*/migrations/Version*.php';
+        } else {
+            //search by extension
+            $pattern = $dir . '/migrations/Version*.php';
+        }
+        $files = glob($pattern);
```
wazelin, line 32: Is there a way to define migrations order?
It may happen that one will depend on another.

SashaTAO:
Based on latest discussions we need to have possibility to run taoUpdate without execution all migrations. As an option I would suggest to introduce migration categories or groups and execute only required and `safe` migrations by default. We can also add an additional parameter to taoUpdate script to be able to execute all migrations during update on our test instances on QA environments.

scripts/tools/migrations/TaoFinder.php
hutnikau: There was a long discussion about migrations order. They will be applied according to extensions dependencies and creation time.

hutnikau:
> Based on latest discussions I think we need to have possibility to run taoUpdate without execution all migrations. As an option I would suggest to introduce migration categories or groups and execute only required and `safe` migrations by default. We can also add an additional parameter to taoUpdate script to be able to execute all migrations during update on our test instances on QA environments.

I have strong opinion here that all migrations must be done automatically. All those changes are done to do one more step toward automation. Any manual steps must be avoided. Changelog notes should be used to warn operations that some updates may take long time. See: https://oat-sa.atlassian.net/wiki/spaces/~465844525/pages/340985162/Tao+Update+Process

hutnikau:
> Migration works during TAO update.
> But I think we have to add possibility to rollback executed migrations if something went wrong and also to execute `--dry-run`.

Rolling back changes functionality has been implemented. As applying migrations is going to be completely automated I do not see need to have ability to run in in `--dry-run` mode

.Jenkinsfile
```diff
@@ -45,14 +34,38 @@ registry.service.consul:4444/tao/dependency-resolver oat:dependencies:resolve --
...
+                        if (env.CHANGE_BRANCH != null) {
+                            branch = CHANGE_BRANCH
+                        } else {
+                            branch = BRANCH_NAME
+                        }
+                        env.branch = branch
+                        writeFile(file: 'composer.json', text: """
+                        {
+                            "require": {
+                                "oat-sa/extension-tao-devtools" : "dev-develop",
```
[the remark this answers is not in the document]
hutnikau: There is a tool which checks dependencies. But this check was not enabled yet for tao-core.
See example here: https://github.com/oat-sa/extension-tao-scheduler/blob/master/.Jenkinsfile#L106

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,385 @@
...
+    /**
+     * @param bool $rollback
+     * @return BufferedOutput
+     * @throws MigrationException
+     * @throws ScriptException
+     * @throws \common_ext_ExtensionException
+     */
+    private function execute(bool $rollback = false)
+    {
+        $input = [
```
gitromba, line 234: The command building part could be a good candidate for decoupling. It's being spread all across this class and spoiling its internal interface while introducing other code smells (control flag parameters, magic constants).

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,385 @@
...
+            mkdir($extension->getDir().self::MIGRATIONS_DIR);
+        }
+        $input = [
+            'command' => $this->commands[self::COMMAND_GENERATE],
+            '--namespace' => $this->getExtensionNamespace($extension)
+        ];
+        $configuration = $this->getConfiguration();
+        $configuration->setCustomTemplate(
+            __DIR__.DIRECTORY_SEPARATOR.'migrations'.DIRECTORY_SEPARATOR.'Template.tpl'
+        );
```
gitromba, line 165: Receiving this from configuration would be great.

gitromba, line 165: The devil doesn't sleep, an "is-readable" check would get a warm welcome.

scripts/tools/migrations/AbstractMigration.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\scripts\tools\migrations;
+
+use Doctrine\Migrations\AbstractMigration as DoctrineAbstractMigration;
+use oat\oatbox\service\ServiceManagerAwareInterface;
+use oat\oatbox\service\ServiceManagerAwareTrait;
+
+abstract class AbstractMigration extends DoctrineAbstractMigration implements ServiceManagerAwareInterface
```
gitromba, line 29: The children are generated, what's the point of having this abstract class around?

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+use common_ext_ExtensionsManager as ExtensionsManager;
+use common_ext_Extension as Extension;
+
+/**
+ * Class TaoComparator
+ * @package oat\tao\scripts\tools\migrations
+ */
+class TaoComparator implements Comparator
+{
+    private $extensionsManager;
```
gitromba, line 36: The type declaration is missing.

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param ExtensionsManager $extensionsManager
+     * @param \helpers_ExtensionHelper $extensionHelper
+     */
+    public function __construct(ExtensionsManager $extensionsManager, \helpers_ExtensionHelper $extensionHelper)
+    {
+        $this->extensionsManager = $extensionsManager;
+        $this->extensionHelper = $extensionHelper;
```
gitromba, line 45: The variable declaration is missing.

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function compare(Version $a, Version $b) : int
+    {
+        $merged = array_merge(
+            $this->extensionsManager->getInstalledExtensions(),
+            $this->getMissingExtensions()
+        );
+        $sortedExtensions = array_flip(array_keys($this->extensionHelper::sortByDependencies($merged)));
```
gitromba, line 54: There's no point injecting an instance if we are interested in class methods. I advise wrapping it around with a dynamic method. 
```
class Helper{
  public function sort($extensions) {
    return helpers_ExtensionHelper::sortByDependencies($extensions);
  }
}
```

P.S.: Helper is terrible name for classes, while I'm not familiar with the domain, I think an "OrderedExtensionCollection" class would be more expressive and declarative.

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+     * @param ExtensionsManager $extensionsManager
+     * @param \helpers_ExtensionHelper $extensionHelper
+     */
+    public function __construct(ExtensionsManager $extensionsManager, \helpers_ExtensionHelper $extensionHelper)
+    {
+        $this->extensionsManager = $extensionsManager;
+        $this->extensionHelper = $extensionHelper;
+    }
+
+    public function compare(Version $a, Version $b) : int
```
gitromba, line 48: I would delegate all this responsibility to a Version instance. It's a pity composition is impossible as there's no interface to implement. is given

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+        $versionA = (string) $a;
+        $versionB = (string) $b;
+        preg_match('/.*Version(\d+)_(.*)$/', $versionA, $matchesA);
+        preg_match('/.*Version(\d+)_(.*)$/', $versionB, $matchesB);
+        list($aClass, $aTime, $aExt) = $matchesA;
+        list($bClass, $bTime, $bExt) = $matchesB;
+        if ($aExt === $bExt) {
+            return ((int) $aTime) - ((int) $bTime);
+        }
+        return $sortedExtensions[$aExt] - $sortedExtensions[$bExt];
```
gitromba, line 64: Tremendous amount of code is working on to the prepare the prepare the data needed to comparison. This is begging for a class to be encapsulated and detached from here.

test/unit/scripts/tools/migrations/TaoComparatorTest.php
```diff
@@ -0,0 +1,112 @@
...
+ * Class TaoComparator
+ * @package oat\tao\scripts\tools\migrations
+ */
+class TaoComparatorTest extends TestCase
+{
+
+    public function testCompare()
+    {
+        $extensionsManagerMock = $this->getExtensionManagerMock();
+        $comporator = new TaoComparator($extensionsManagerMock, $this->getExtensionHelperMock());
```
gitromba, line 42: Typo. Correct one: ```comparator```

test/unit/scripts/tools/migrations/TaoComparatorTest.php
```diff
@@ -0,0 +1,112 @@
...
+        $versions = [
+            $versionFoo5 = new Version('Version5_foo'),
+            $versionFoo3 = new Version('Version3_foo'),
+            $extensionBar4 = new Version('Version4_bar'),
+            $extensionBar2 = new Version('Version2_bar'),
+            $extensionBaz1 = new Version('Version1_baz'),
+        ];
+
+        usort($versions, [$comporator, 'compare']);
```
gitromba, line 51: Here should be code directly related to ```Comparator```. 
Instead you have to deal with some unrelated preparation work, because the object responsibilities are not well defined.

test/unit/scripts/tools/migrations/TaoComparatorTest.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->assertEquals([
+            'Version3_foo',
+            'Version5_foo',
+            'Version1_baz',
+            'Version2_bar',
+            'Version4_bar',
+        ], $versionKeys);
+    }
+
+    private function getExtensionManagerMock()
```
gitromba, line 65: The size (and sole existence) of this method is a consequence of wrong dependencies. This should be in a test for another class, which is missing at the moment, but already evolved actually in a form of these code smells. Now it's time to reborn :)

models/classes/migrations/MigrationsService.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    /**
+     * Apply all migrations
+     * @return \common_report_Report
+     */
+    public function migrate()
+    {
+        $migrations = new Migrations();
+        $migrations->setServiceLocator($this->getServiceLocator());
+        return $migrations->__invoke(['-c', 'migrate']);
```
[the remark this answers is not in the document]
bugalot: Indeed, a mater of taste. IMHO it is more obvious it is a usage of an `invokable` with the original proposal. Let's think about rookies while writing code ;-) !

models/classes/migrations/MigrationsService.php
```diff
@@ -0,0 +1,62 @@
...
+    public function extensionInstalled(ExtensionInstalled $event)
+    {
+        $migrations = new Migrations();
+        $migrations->setServiceLocator($this->getServiceLocator());
+        if ($event->getExtension()->getId() === 'tao') {
+            //supposedly application initial install and generis just have been installed.
+            $migrations->__invoke(['-c', 'init']);
+            $migrations->__invoke(['-c', 'add', '-e', 'generis']);
+        }
+        $migrations->__invoke(['-c', 'add', '-e', $event->getExtension()->getId()]);
```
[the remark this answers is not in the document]
bugalot: @gitromba Long story short, the Migration class (below in PR) is a sub-class of ScriptAction (TAO Realm). A ScriptAction is an invokable action (with super awesome CLI argument management integrated) that can be either called via `CLI` or as an `invokable` e.g. 

1. `pwd` is root of TAO platform. `php index.php "oat\tao\scripts\tools\Migrations" -c init` or `php index.php "oat\tao\scripts\tools\Migrations" -c add -e generis`
2. In code as an invokable `$migrations->__invoke(['-c', 'init']);` or `$migrations->__invoke(['-c', 'add', '-e', 'generis']);`

That's a good way to expose software components to CLI as well in TAO.

Regarding the line `$migrations->setServiceLocator($this->getServiceLocator());`, it's just about manually injecting TAO's `ServiceLocator` in the `invokable`. Yes, we unfortunately, for historical reasons are using the `ServiceLocator` pattern but we want to get rid of it ;-) !

scripts/tools/migrations/AbstractMigration.php
bugalot: @gitromba Because instantiating an AbstractMigration would not make sense? :-)

scripts/tools/migrations/TaoClassNameGenerator.php
```diff
@@ -0,0 +1,64 @@
...
+declare(strict_types=1);
+
+namespace oat\tao\scripts\tools\migrations;
+
+use DateTimeImmutable;
+use DateTimeZone;
+use Doctrine\Migrations\Generator\ClassNameGenerator;
+use common_ext_Extension as Extension;
+
+class TaoClassNameGenerator extends ClassNameGenerator
```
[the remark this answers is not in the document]
bugalot: I would say that having a description of the class would be nice.

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,385 @@
...
+use oat\tao\scripts\tools\migrations\commands\GenerateCommand;
+use oat\tao\scripts\tools\migrations\TaoComparator;
+use oat\tao\scripts\tools\migrations\TaoMigrationRepository;
+use common_ext_Extension;
+use common_report_Report as Report;
+use common_ext_ExtensionsManager as ExtensionsManager;
+use common_ext_Extension as Extension;
+
+/**
+ * Class Migrations
```
bugalot, line 58: I would say that having a more detailed description of the class would be nice for comprehension.

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+use Doctrine\Migrations\Version\Comparator;
+use Doctrine\Migrations\Version\Version;
+use common_ext_ExtensionsManager as ExtensionsManager;
+use common_ext_Extension as Extension;
+
+/**
+ * Class TaoComparator
+ * @package oat\tao\scripts\tools\migrations
+ */
+class TaoComparator implements Comparator
```
bugalot, line 34: This nice class deserves a nice description.

scripts/tools/migrations/TaoMigrationRepository.php
```diff
@@ -0,0 +1,42 @@
...
+namespace oat\tao\scripts\tools\migrations;
+
+use Doctrine\Migrations\MigrationRepository;
+use \Doctrine\Migrations\Metadata\AvailableMigration;
+use oat\oatbox\service\ServiceManager;
+
+/**
+ * Class TaoMigrationRepository
+ * @package oat\tao\scripts\tools\migrations
+ */
```
bugalot, line 32: This class deserves a description and its connection with the others.

scripts/tools/migrations/Template.tpl
```diff
@@ -0,0 +1,35 @@
...
+use oat\oatbox\service\ServiceManagerAwareTrait;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class <className> extends AbstractMigration implements ServiceManagerAwareInterface
+{
+    use ServiceManagerAwareTrait;
+
+    public function getDescription() : string
```
bugalot, line 19: PSR-12 issue as detected above.

scripts/tools/migrations/commands/GenerateCommand.php
```diff
@@ -0,0 +1,90 @@
...
+
+use Doctrine\Migrations\Tools\Console\Command\DoctrineCommand;
+use Exception;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Class GenerateCommand
+ * @package oat\tao\scripts\tools\migrations\commands
```
bugalot, line 13: requires a class description

scripts/tools/migrations/commands/GenerateCommand.php
```diff
@@ -0,0 +1,90 @@
...
+ */
+class GenerateCommand extends DoctrineCommand
+{
+    /** @var string */
+    protected static $defaultName = 'migrations:generate';
+
+    protected function configure() : void
```
bugalot, line 20: PSR-12 issue. Other methods impacted.

scripts/update/Updater.php
```diff
@@ -1343,6 +1344,7 @@ public function update($initialVersion)
             $this->setVersion('42.0.4');
         }
 
-        $this->skip('42.0.4', '42.7.1');
+        $this->skip('42.0.4', '42.8.0');
```
bugalot, line 1347: Should be now `$this->skip('42.0.4', '42.9.0');`

composer.json
```diff
@@ -44,10 +44,12 @@
...
     },
     "require": {
-        "php": "^7.1",
-        "phpdocumentor/reflection-docblock": "2.*",
+        "oat-sa/generis": "^12.20.1",
         "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
+        "oat-sa/composer-npm-bridge": "^0.3.0",
         "oat-sa/jig": "~0.1",
+        "php": "^7.1",
+        "phpdocumentor/reflection-docblock": "2.*",
```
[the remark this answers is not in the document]
hutnikau: I think this is out of scope of this PR

composer.json
```diff
@@ -56,8 +58,8 @@
         "pimple/pimple": "^3.0",
         "imsglobal/lti": "^3.0",
         "paragonie/random_compat": "^2.0",
-        "oat-sa/composer-npm-bridge": "^0.3.0",
-        "symfony/dotenv": "~4.3"
+        "symfony/dotenv": "~4.3",
+        "doctrine/migrations" : "3.0.0-alpha1"
```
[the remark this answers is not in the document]
hutnikau: It was moved up

models/classes/migrations/MigrationsService.php
```diff
@@ -0,0 +1,62 @@
...
+ * Class MigrationsService
+ * @package oat\tao\model\migrations
+ */
+class MigrationsService extends ConfigurableService
+{
+    /**
+     * Apply all migrations
+     * @return \common_report_Report
+     */
+    public function migrate()
```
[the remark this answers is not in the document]
hutnikau: https://github.com/oat-sa/tao-core/pull/2409/files#diff-97e36923b5bd839786b3f709ae455952R55

models/classes/migrations/MigrationsService.php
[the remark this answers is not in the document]
hutnikau: When tao installed initially we have to mark all the existing migrations as skipped (using `add` method of doctrine migration tool) without their actual execution.

scripts/tools/Migrations.php
hutnikau: This check is done in doctrine migration tool (see `\Doctrine\Migrations\Generator\Generator::loadCustomTemplate`)

hutnikau: Usually I make something configurable when there is need to configure it appears.
Let's leave it as is as I do not see the need to use another template at the moment.

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @return BufferedOutput
+     * @throws MigrationException
+     * @throws ScriptException
+     * @throws \common_ext_ExtensionException
+     */
+    private function rollback()
+    {
+        return $this->execute(true);
```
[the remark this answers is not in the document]
hutnikau: Usually IDE does the trick:
image
But i can move it to the constant if you prefer

scripts/tools/migrations/AbstractMigration.php
hutnikau: Tao migrations, which extend this class, will have `ServiceManagerAwareTrait` mixed in it by default. Later we may add more tao specific functions in this class.

scripts/tools/migrations/TaoComparator.php
hutnikau: I'm not agree. Version is not responsible to compare itself with other versions.
And it will make impossible to change the comparison logic.

scripts/tools/migrations/TaoComparator.php
hutnikau: Such king of usage of static methods makes impossible to use mocks in the unit tests.

scripts/tools/migrations/TaoComparator.php
```diff
@@ -0,0 +1,78 @@
...
+        if ($aExt === $bExt) {
+            return ((int) $aTime) - ((int) $bTime);
+        }
+        return $sortedExtensions[$aExt] - $sortedExtensions[$bExt];
+    }
+
+    /**
+     * @return Extension[]
+     */
+    protected function getMissingExtensions()
```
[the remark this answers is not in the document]
hutnikau: Extensions which are not installed at time of execution of migration script, but must be installed as they are listed in the dependency list.

test/unit/scripts/tools/migrations/TaoComparatorTest.php
hutnikau: It was done in purpose. Usage `usort()` function better reflects it's real use case. The compare function will never be called directly.

test/unit/scripts/tools/migrations/TaoComparatorTest.php
hutnikau: ExtensionManager is directly used in TaoComparator and must be injected into constructor.

manifest.php
```diff
@@ -56,7 +57,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '42.7.1',
+    'version' => '42.8.0',
```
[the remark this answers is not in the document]
hutnikau: I will merge develop after merge approval.

scripts/tools/Migrations.php
hutnikau: I think this is the only responsibility of this class is to build command as it is only wrapper of doctrine migration tool.

scripts/tools/migrations/TaoComparator.php
hutnikau: The code is linear and straightforward here and serve the only purpose. I would like to leave it as is for better readability and easier debugging.

scripts/update/Updater.php
hutnikau: I will merge develop after PR merge approval

scripts/tools/Migrations.php
```diff
@@ -0,0 +1,324 @@
...
+
+    /**
+     * @return Connection
+     */
+    private function getConnection()
+    {
+        /** @var PersistenceManager $persistenceManager */
+        $persistenceManager = $this->getServiceLocator()->get(PersistenceManager::SERVICE_ID);
+        //todo: use migrations service to store persistence id in it's option
+        $persistence = $persistenceManager->getPersistenceById('default');
```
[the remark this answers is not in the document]
hutnikau: Done

scripts/update/Updater.php
```diff
@@ -1344,5 +1345,13 @@ public function update($initialVersion)
         }
 
         $this->skip('42.0.4', '42.2.1');
+
+        if ($this->isVersion('42.2.1')) {
```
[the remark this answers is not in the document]
hutnikau: fixed

.Jenkinsfile
```diff
@@ -2,32 +2,21 @@ pipeline {
     agent {
         label 'builder'
     }
+    parameters {
+        string(name: 'branch', defaultValue: '')
+    }
+    environment {
+        REPO_NAME='oat-sa/tao-core'
+        EXT_NAME='tao'
+        GITHUB_ORGANIZATION='oat-sa'
```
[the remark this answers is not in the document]
hutnikau: removed

.Jenkinsfile
```diff
@@ -2,32 +2,21 @@ pipeline {
     agent {
         label 'builder'
     }
+    parameters {
+        string(name: 'branch', defaultValue: '')
+    }
+    environment {
+        REPO_NAME='oat-sa/tao-core'
```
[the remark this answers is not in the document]
hutnikau: I think this is out of scope of this PR. I can create PRs in all the extensions where this option used to get this value from parameter.

.Jenkinsfile
```diff
@@ -2,32 +2,21 @@ pipeline {
     agent {
         label 'builder'
     }
+    parameters {
+        string(name: 'branch', defaultValue: '')
```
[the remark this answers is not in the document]
hutnikau: removed

.Jenkinsfile
```diff
@@ -45,14 +34,38 @@ registry.service.consul:4444/tao/dependency-resolver oat:dependencies:resolve --
...
+                        env.branch = branch
+                        writeFile(file: 'composer.json', text: """
+                        {
+                            "require": {
+                                "oat-sa/extension-tao-devtools" : "dev-develop",
+                                "${REPO_NAME}" : "dev-${branch}#${GIT_COMMIT}"
+                            },
+                            "minimum-stability": "dev",
+                            "require-dev": {
+                                "phpunit/phpunit": "~8.5"
```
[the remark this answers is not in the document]
hutnikau: done

scripts/tools/migrations/TaoComparator.php
hutnikau: fixed

scripts/tools/migrations/TaoComparator.php
hutnikau: fixed

test/unit/scripts/tools/migrations/TaoComparatorTest.php
hutnikau: fixed

migrations/Version202005081221002234_tao.php
```diff
@@ -0,0 +1,41 @@
...
+use oat\oatbox\service\ServiceManagerAwareTrait;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version202005081221002234_tao extends AbstractMigration implements ServiceManagerAwareInterface
+{
+    use ServiceManagerAwareTrait;
+
+    public function getDescription() : string
```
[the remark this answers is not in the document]
hutnikau: fixed both migration and template

scripts/tools/migrations/TaoMigrationRepository.php
hutnikau: done

scripts/tools/Migrations.php
hutnikau: done

scripts/tools/migrations/TaoComparator.php
hutnikau: done

scripts/tools/migrations/commands/GenerateCommand.php
hutnikau: fixed

composer.json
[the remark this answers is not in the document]
bugalot: @hutnikau These changes to the composer.json file are putting us in a conflicting state with project level composer.json and the extension composer.json. We should remove "oat-sa/generis" reference here to avoid conflicts. Putting dependencies on the composer extension level should happen in a second phase.

.Jenkinsfile
```diff
@@ -2,32 +2,17 @@ pipeline {
...
-changeBranch=$CHANGE_BRANCH
-TEST_BRANCH="${changeBranch:-$BRANCH_NAME}"
-echo "select branch : ${TEST_BRANCH}"
-docker run --rm  \\
--e "GITHUB_ORGANIZATION=${GITHUB_ORGANIZATION}" \\
--e "GITHUB_SECRET=${GIT_TOKEN}"  \\
-registry.service.consul:4444/tao/dependency-resolver oat:dependencies:resolve --main-branch ${TEST_BRANCH} --repository-name ${REPO_NAME} > build/composer.json
-                        '''
-                    )
-                }
```
bugalot, line 30: I'm not sure this should be removed! Do we really want to remove the CI right now if we decide to merge this in "silent mode"?

scripts/update/Updater.php
```diff
@@ -70,6 +70,7 @@
 use oat\tao\model\media\MediaService;
 use oat\tao\model\metadata\compiler\ResourceJsonMetadataCompiler;
 use oat\tao\model\metrics\MetricsService;
+use oat\tao\model\migrations\MigrationsService;
```
bugalot, line 73: I'm not sure this one is necessary

.Jenkinsfile
hutnikau: We do not need this step as dependency (the only one: generis) will be given from composer.json

composer.json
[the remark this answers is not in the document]
hutnikau: Fixed

scripts/tools/migrations/commands/GenerateCommand.php
hutnikau: done

.Jenkinsfile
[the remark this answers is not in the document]
gitromba: Your call, indeed this can be considered as an additional improvement.

models/classes/migrations/MigrationsService.php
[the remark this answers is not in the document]
gitromba: Thx!

models/classes/migrations/MigrationsService.php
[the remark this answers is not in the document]
gitromba: Thanks guys, that helped me a lot. 

Just to sum up:
Migrations' metadata storage need to be initialized manually by explicitly call the ```init``` command. Init has to happen only once thus it's tied to one specific component. I assume there's no other way to apply the "only once" behaviour so you have to rely on the assumption that the ```tao``` extension is always present and to trick it this way. 

There's something still blurry though with is the [```generis ``` extension](https://github.com/oat-sa/tao-core/blob/4affa0fdea5a823bcb36b74251fecadf55ded4cd/models/classes/migrations/MigrationsService.php#L58)
1. why it has to be skipped?
2. why here while dealing with the ```tao``` extension? What is the guarantee of having the ```generis``` extension available?

scripts/tools/Migrations.php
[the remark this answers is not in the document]
gitromba: Here I see only the value, I have no clue about what it's actually doing :(

And this is the first thing I turn off in my storm :)

scripts/tools/Migrations.php
gitromba: It does more than composing migration commands:
- gives a CLI API by inheriting ```ScriptAction``` and deals with actual execution
- generates migration files from template
- executes and manages migrations
- initializes the meta storage
- rollbacks migrations
- provides status on migrations
- configures the underlying Doctrine migrations layer
- initializes the DB connection
- resolves extensions
- defines the extension namespace

All these to be addressed in this PR is **not** my attention, rather to raise your awareness as there's lot more here than command-wrapping. These could be tackled step by step later on. The mindset here is to make the code better piece by piece, iteration by iteration.

scripts/tools/Migrations.php
gitromba: As a rule of thumb, ```__DIR__``` constants should be used only during bootstrap to configure the application. Then if you require this information, you could fetch it from the config/DIC. This gives you the benefit of having a more environment-indepent, robust application and single source of configuration.
Personally I think that ideally __DIR__ shouldn't be used at all directly in code, but rather be injected as an environment variable.

scripts/tools/migrations/AbstractMigration.php
gitromba: Maybe. Or YAGNI :)

scripts/tools/migrations/TaoComparator.php
gitromba: To stick with my example, in this case the ```Helper``` class needs to be mocked, which is pretty straightforward.

scripts/tools/migrations/TaoComparator.php
gitromba: The beauty of OOD that any kind of relations between 2 objects could be nicely represented.
```
if ($student->smarterThan($proctor)) { cheat(); }
if ($numberA->biggerThan($numberB) { echo "A is bigger."; }
```
The well-known ```equalTo``` akin methods is also a form of comparing 2 objects. 
When need arises, you could turn to strategy pattern (or polymorphism) to decouple nicely encapsulate different behaviours.

I have a thing on more "clever" objects, e.g. to decorate value objects with behaviour (as they are objects) and don't treat them as simple data bags or DTOs.

models/classes/migrations/MigrationsService.php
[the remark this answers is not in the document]
hutnikau: 1. Because migrations are used to update existing environment from state A to state B. When you install new instance of Tao it's already in state B.
2. tao depends on generis. Generis is the root of dependency tree. We need here this trick because generis is the only extension installed before tao-core

scripts/tools/migrations/TaoComparator.php
gitromba: Code is linear and straightforward. It's all about responsibilities. If you feel it's better here, than it is ;)

scripts/tools/migrations/TaoComparator.php
hutnikau: If you want to check who is smarter, you need to use a *test*
If you want to check who is taller, you need to use a *ruler*

Number istelf cannot compare itself with other number, it does not make sense. There should be another *object*, comparator, which encapsulates comparison rules and does the comparison. That's actually how docrtine migration tool implemented and that is what allows us to change the comparison rule from default to Tao specific.

scripts/tools/migrations/TaoComparator.php
hutnikau: Sorry, I do not get why we need to wrap helper in helper.
Does that mean that in for all other places where `helpers_ExtensionHelper::sortByDependencies` is used a wrapper must be created (one wrapper per method usage) ?

test/unit/scripts/tools/migrations/TaoComparatorTest.php
hutnikau: Added the test case (although i think this is just duplication of assertion which was done above).

scripts/update/Updater.php
hutnikau: removed

test/unit/scripts/tools/migrations/TaoComparatorTest.php
gitromba: Exactly. And this manager is just an extra layer between the real required information (extensions) and the comparator.
My thoughts as code snippet:
```
construct(Extenstions $extensions)
```
Declaring what are missing extensions in ```getMissingExtensions``` and being aware where are the extensions coming from due the direct ```ExtensionManager``` dependency is resulting in highly coupled code which is hard to maintain and to test (half of the test is only about mocking the manager).

scripts/tools/Migrations.php
hutnikau: All this actions are done by doctrine migration tool.
Two responsibilities of this class is to configure doctrine migration tool and proxy commands to it.

scripts/tools/Migrations.php
hutnikau: Good catch. Fixed

scripts/tools/Migrations.php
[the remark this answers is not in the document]
gitromba: Forgot to push? :)

test/unit/scripts/tools/migrations/TaoComparatorTest.php
hutnikau: This will make proper dependency injection impossible.
And this is contradicts abstraction principles and makes testing and customization more difficult. 

Object needs to have not data itself, but data providers, which will allow to retrieve any required data. Because this data (extensions list in our example) may be dynamic and next call of data provider will result with different set of extensions. 

While your approach (when we feed object at time of it's instantiation with all the necessary data) may make data in this object (list of missed extensions) obsolete.

scripts/tools/Migrations.php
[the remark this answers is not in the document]
hutnikau: This is outdated piece of code.

scripts/tools/migrations/TaoComparator.php
gitromba: I love your examples! I think they are very good ones, and I would like to challenge them ;)
But before that, I know how this part of migrations works (checked the code in Doctrine before my first comment) and I'm not challenging the actual solution. I'm fine with that. I'm challenging the "Doctrine-way" and doing it because this is how we learn and grow. 

It's all about the context to how you design your object hierarchy. Let me give an example.
In a specific domain it's better if I can tell you whether you are taller than me or not. And I don't want to disclose how I do it. I may have a ruler, or a page consisting all employees' height in TAO (that would explain why I want to keep it a secret ;)). I could google it on the Net, because you find everything there, people say :). I could make a phone call to someone knows this about you. Doesn't matter as long as the contract is respected.

In another context it seems more reasonable to expose what we are using to measure. You have a tape measure and you want to know the distance between 2 objects in the room? It's quite the contrary of previous example. Here the context shows a different demand against your model.

There's no good or bad design. I believe in context dependent design.

gitromba: If you put a like on my previous comment, I consider this resolved :) #noblackmailing

test/unit/scripts/tools/migrations/TaoComparatorTest.php
gitromba: Perhaps it's just me. There will be always personal preferences. Anyway, thx, I'll check it out right away.

gitromba: I'm fine with this. 
For the sake of transparency, I've had something else in mind. 
A super simple one, having only one assert at the end (which is asserting the result of ```compare``` method call). And I would feed them trough a data provider. ATM there's 1 test case, with a data provider you could have more cases covered (and it keeps the cases separated from the actual testing code).

test/unit/scripts/tools/migrations/TaoComparatorTest.php
gitromba: I'm curious on your opinion of proper dependency management but I think this might be too much in this PR so this stands here as my closing comment.

The scenario you have described makes me think, that the ```compare``` method is called multiple times during the upgrade process and the referenced objects could have different state during the execution. In this case, and if you are unable to instantiate the comparator object multiple times, indeed no other option left, just to keep this abstraction what you call data provider.

scripts/install/RegisterEvents.php
```diff
@@ -0,0 +1,48 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA
```
wazelin, line 18:
```suggestion
 * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
```

Protip, use `${YEAR}` in your `PHP File Header` template.

.Jenkinsfile
bugalot: @hutnikau In order to proceed step by step, but at higher velocity, I would suggest we remove the dependency to generis from the composer.json file totally. There are still unknown areas about linking extensions together in composer.json at this stage.

Could we please proceed baby-steps? I feel we don't master anything yet in terms of composer dependencies. As a plan I would suggest:

- We remove the generis dependency in composer (because the `*` will not give us the same dependency over time, better play safe please.
- We revert the changes in the JenkinsFile.
- We still use the dependency-resolver in Jenkins file for the moment.
- We tackle the dependency issue later on

bugalot: Playing that plan makes this PR mergeable and we can go next step.

.Jenkinsfile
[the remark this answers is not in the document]
bugalot: As stated above, I would not change Jenkins file for the moment. To be done in the next step. Let's make this mergeable now.

composer.json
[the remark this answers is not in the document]
bugalot: We need to make sure that the addition to composer will not break anything. Let's clarify

- generis to be removed (will be introduced next step). let's not break anything.
- npm bridge? Why do we add it? 
- phpdocumentor? Why do we add it?

gitromba:
Alpha packages must be allowed, @hutnikau please make sure **minimum-stability** is configured accordingly. 

Otherwise it could be breaking pipelines:
http://qa.kitchen.it.taocloud.org/blue/organizations/jenkins/Github%20Pull%20Requests%20for%20oat-sa%2Ftao-core/detail/TDR-21%2Fmigrations_tool/59/pipeline

manifest.php
```diff
@@ -120,6 +121,7 @@
             SetContainerService::class,
             RegisterResourceWatcherService::class,
             RegisterResourceEvents::class,
+            RegisterEvents::class,
```
wazelin, line 124: I'm probably missing a point here, but won't the existing migrations be applied upon a fresh installation?

PR #2413, opened 2020-03-02
UDI-76 fix(ACL): check for write permissions when editing class' label

wazelin:
Related to https://oat-sa.atlassian.net/browse/UDI-76

- UDI-76 fix(ACL): check for write permissions when editing class' label
- UDI-76 feat(form): allow disabling a whole form by providing `tao_helpers_form_FormContainer::IS_DISABLED` option
- UDI-76 feat(class-label): disable a whole class-label-editing form in case of insufficient permissions
- UDI-76 chore(version): bump `Updater` and `manifest` versions
- UDI-76 fix(class-label): respond with a proper error message when handling a request with insufficient permissions
- UDI-76 chore(class-label): remove deprecated `ServiceManager` usage from the `RdfController`
- UDI-76 chore(form): reorganize `tao_helpers_form_FormContainer` initialization conditions
- UDI-76 chore(form): fix `tao_helpers_form_FormContainer` comment typos
- UDI-76 chore(form-element): fix return type in `tao_helpers_form_FormElement` method docblocks
- UDI-76 chore(form-element): use strict comparison in `tao_helpers_form_FormElement::removeClass()`
- UDI-76 chore(form): extract form element disabling responsibility into `tao_helpers_form_FormElement`
- UDI-76 chore(dependency): breaking change – require `php ^7.1`
- UDI-76 chore(config): use `__DIR__` instead of `dirname(__FILE__)` in `manifest.php`
- UDI-76 chore(form): don't pre-render common actions such as `Save` button
- UDI-76 fix(form-element): disable a submit button when the whole form is disabled
- UDI-76 chore(form): don't pre-render common actions such as `Save` button
- UDI-76 fix(form-element): disable a submit button when the whole form is disabled
- UDI-76 chore(form): run simple test on `tao_helpers_form_FormContainer`


actions/class.RdfController.php
```diff
@@ -428,11 +428,16 @@ public function editClassLabel()
...
+        $classUri = $class->getUri();
+
         if ($myForm->isSubmited() && $myForm->isValid()) {
-            $class->setLabel($myForm->getValue(tao_helpers_Uri::encode(OntologyRdfs::RDFS_LABEL)));
+            if ($this->hasWriteAccess($classUri)) {
+                $class->setLabel($myForm->getValue(tao_helpers_Uri::encode(OntologyRdfs::RDFS_LABEL)));
+
+                $this->setData('selectNode', tao_helpers_Uri::encode($classUri));
+                $this->setData('message', __('%s Class saved', $class->getLabel()));
+            }
```
Yury-Razhkou: It is better to add here error message with explanation why it was not saved. Something like this:
`else{
$this->setData('errorMessage', __('Permission denied to write in the selected class'));
}`

wazelin: That's exactly what happens due to `$this->setData('reload', true);`.
image

Yury-Razhkou: I've made incorrect test and did not see this =)

wazelin:
Resolved a form disabling on a more global level. Now there's an option, disabling all form elements.
With actions it's trickier, so I wouldn't touch the `Save` button for now, as the actions contain pre-rendered values. It has to be refactored first so that the action sections contain objects and not raw HTML.

actions/class.RdfController.php
```diff
@@ -419,20 +419,27 @@ public function editClassLabel()
...
 
+        $classUri       = $class->getUri();
+        $hasWriteAccess = $this->hasWriteAccess($classUri);
+
         $editClassLabelForm = new tao_actions_form_EditClassLabel(
             $class,
             $this->getRequestParameters(),
             $signature,
-            [FormContainer::CSRF_PROTECTION_OPTION => true]
+            [FormContainer::CSRF_PROTECTION_OPTION => true, FormContainer::IS_DISABLED => !$hasWriteAccess]
```
SergiiTao: the input is disabled, but pushing button Save leads to 
image

wazelin: Fixed

helpers/form/class.Form.php
```diff
@@ -667,6 +666,20 @@ public function setValues($values)
         }
     }
 
+    /**
+     * Disables all elements of the form
+     */
+    public function disableElements()
+    {
+        foreach ($this->elements as $element) {
+            if ($element instanceof tao_helpers_form_elements_Hidden) {
```
SergiiTao: what do you think of to add method set/isEnabled to `tao_helpers_form_FormElement` ?

helpers/form/class.FormContainer.php
```diff
@@ -106,6 +107,10 @@ public function __construct($data = [], $options = [])
             $this->form->setValues($this->data);
         }
 
+        if ($this->form !== null && isset($options[self::IS_DISABLED]) && $options[self::IS_DISABLED]) {
```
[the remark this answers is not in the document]
wazelin: `composer.json` requires `"php": ">=5.6"`, so I cannot use the null-coalescing operator.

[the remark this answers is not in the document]
SergiiTao: please remove this restriction from composer. all clieant are on 7.2 + we have php>=7.1 in generis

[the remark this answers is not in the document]
wazelin: Technically it would be a breaking change then. I'd prefer not doing it now.

PR #2419, opened 2020-03-05
Fix/ACTP-293/Trigger userUpdated event for users created via API

shpran:
API Created User different from UI produced user

Related to: [https://oat-sa.atlassian.net/browse/ACTP-293](https://oat-sa.atlassian.net/browse/ACTP-293)

#### Changes
- added separate method to bind values and trigger `UserUpdatedEvent` with `hashForKey` value (if password exists);
- use this method for user creation/update via UI and for user creation via API to create necessary encryption key.

#### Steps to reproduce
- create user via API: `POST /tao/api/Users`;
- copy `uri` from response and find user data in database (`subject='[[ uri ]]'`);

**Bug (only when taoEncryption is installed)**:
There's no `#encryptionKey` for this user.

**Fix**:
User contain`#encryptionKey` field and value.


controller/api/Users.php
```diff
@@ -226,6 +227,12 @@ public function post()
 
             $userService->attachProperties($user, $parameters);
 
+            if (!array_key_exists(GenerisRdf::PROPERTY_USER_PASSWORD, $parameters)) {
+                $parameters[GenerisRdf::PROPERTY_USER_PASSWORD] = $user->getProperty(GenerisRdf::PROPERTY_USER_PASSWORD);
```
SashaTAO: I don't think that we have to add password from user object. It is excessive operation (password didn't change) and in this specific case it will not work since we store hashed password in db.
It would be better to check if `hashForKey` key exists in appropriate event handler where it's required.

controller/api/Users.php
```diff
@@ -226,6 +227,11 @@ public function post()
 
             $userService->attachProperties($user, $parameters);
 
+            $parameters[GenerisRdf::PROPERTY_USER_PASSWORD] = $user->getProperty(GenerisRdf::PROPERTY_USER_PASSWORD);
+            unset($parameters[UserRdf::PROPERTY_PASSWORD]);
```
SashaTAO: `GenerisRdf::PROPERTY_USER_PASSWORD` and `UserRdf::PROPERTY_PASSWORD` are the same. As we discussed it would be better to unset plain password a bit earlier in this method where we read it from request.

models/classes/class.UserService.php
```diff
@@ -544,4 +546,30 @@ public function checkCurrentUserAccess($roles)
...
+    /**
+     * @param core_kernel_classes_Resource $user
+     * @param array $values
+     * @param string|null $hashForKey
+     *
+     * @throws tao_models_classes_dataBinding_GenerisFormDataBindingException
+     *
+     * @return bool
+     */
+    public function triggerUpdatedEvent(core_kernel_classes_Resource $user, array $values, $hashForKey = null)
```
SashaTAO: Since we hash plain password before we call this method probably we could add it directly to values array.

models/classes/class.UserService.php
```diff
@@ -544,4 +545,25 @@ public function checkCurrentUserAccess($roles)
...
+     * @throws tao_models_classes_dataBinding_GenerisFormDataBindingException
+     *
+     * @return bool
+     */
+    public function triggerUpdatedEvent(core_kernel_classes_Resource $user, array $values)
+    {
+        $triggered = false;
+        $binder = new tao_models_classes_dataBinding_GenerisFormDataBinder($user);
+
+        if ($binder->bind($values)) {
```
SashaTAO: I found why we had to merge `hashForKey` to values array after we call `$binder->bind($values)`. If we have this field in $values array it will add `hashForKey` property to user object in database, which we don't need. Could you please return it to the previous version where you passed `hashForKey` as a third method parameter.

SashaTAO:
- [ ] Code is self documented (edited)

PR #2421, opened 2020-03-05
Fix/ACTP-294/Use PSR7 Request object in CommonRestModule

shpran:
Test taker API translates XML escaped characters in password.

Related to: [https://oat-sa.atlassian.net/browse/ACTP-294](https://oat-sa.atlassian.net/browse/ACTP-294)

#### Changes
XML entities will be replaced with related characters;

#### How to test
1) create new Test-taker via test-takers API: `POST /taoTestTaker/api/testTakers`;
**NOTE**: use some of these characters for password: `", ', &, <, >`. For example: `Password&1`
2) try to log in with this password.

**Bug**:
Error: `Invalid login or password. Please try again.`

**Fix**:
Successfully logged in.


models/classes/http/LegacyController.php
```diff
@@ -103,7 +103,12 @@ public function hasRequestParameter($name)
      */
     public function getRequestParameter($name)
     {
-        return $this->getRequest()->getParameter($name);
+        // Replace XML entities with related characters
+        return str_replace(
```
bartlomiejmarszal: Can we move this operation to separate method in controller and make unit test for this?

shpran: This logic is outdated.
Right now I am using PSR request to receive body parameters (instead of `clearfw` library).

PR #2425, opened 2020-03-06
Merge/nex 778/notfication object

bartlomiejmarszal:
Notification object moved the directory up.
The interface has required PHP docs.

Merge is a part of changes made in branch `feature/NEX-137/add-spanner-driver`

https://oat-sa.atlassian.net/browse/NEX-778

### DIFF
--- a/models/classes/notification/Notification.php
+++ b/models/classes/notification/Notification.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
  *
  */
 
--- a/models/classes/notification/Notification.php
+++ b/models/classes/notification/Notification.php
@@ -19,9 +19,7 @@
  *
  */
 
-namespace oat\tao\model\notification\implementation;
-
-use oat\tao\model\notification\NotificationInterface;
+namespace oat\tao\model\notification;
 
 class Notification implements NotificationInterface, \JsonSerializable
 {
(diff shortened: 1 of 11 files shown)


models/classes/notification/Notification.php
```diff
@@ -19,9 +19,7 @@
  *
  */
 
-namespace oat\tao\model\notification\implementation;
```
boajer: Why you are changing location of this class?

bartlomiejmarszal: implementation directory should serve only different persistence implementation and notification is only dummy DTO object that implements `NotificationInterface`. I would keep it next to the interface which it's implementing to keep consistency in implement directory.

mike-ionut-mihai-sandu-tao: so according to our discussion , let's please keep this inside the implementation folder so we comply with the existing status quo,  and please extend this thus allowing you to use the functionality into the newsql implementation properly

jbout: I agree on a conceptual level that Notification is not part of the implementation, it is shared between implementations and should be on the lower level. But if we want to move it we should consider backward compatibility

jbout:
By using the Notification as a shared object we no longer need the interface. Also the current NotificationServiceInterface should rely on the Notification object, current return types are incorrect

mike-ionut-mihai-sandu-tao:

test/unit/models/classes/notification/NotificationTest.php
```diff
@@ -0,0 +1,163 @@
+<?php declare(strict_types=1);
```
bartlomiejmarszal:
```suggestion
<?php
```

models/classes/notification/Notification.php
```diff
@@ -0,0 +1,159 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
```
boajer: 2017-2020? Or how should it looks like?

bartlomiejmarszal: I would say 2020

bartlomiejmarszal:
```suggestion
 * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
```

PR #2428, opened 2020-03-09
Fix/udi 97/export items user does not gave access

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/UDI-97

There was a possibility to fetch resources without having permissions to them.
I've created a service to fetch resources safely.


actions/class.Export.php
```diff
@@ -91,9 +98,9 @@ public function index()
                     $children = [];
```
wazelin: Should be defined as `[[]]`, otherwise `array_merge(...$children)` will fail in case of empty `$instances`.

models/classes/resources/SecureResourceService.php
```diff
@@ -0,0 +1,74 @@
...
+            $childrenIds[] = $child->getUri();
+        }
+
+        try {
+            $permissions = $permissionService->getPermissions(
+                $this->getUser(),
+                $childrenIds
+            );
+
+            foreach ($children as $child) {
```
wazelin: The loop shouldn't be within the `try` clause IMO. It's confusing that way.

wazelin: `array_filter` is a better fit, semantically speaking.
Just be looking at it, another developer won't have to think what exactly that loop's purpose is.

models/classes/resources/SecureResourceService.php
```diff
@@ -0,0 +1,74 @@
...
+    public function getAllChildren(core_kernel_classes_Class $resource): array
+    {
+        $result = [];
+        $children = $resource->getInstances(true);
+
+        $permissionService = $this->getPermissionProvider();
+
+        $childrenIds = [];
+
+        foreach ($children as $child) {
```
wazelin: `array_map` is a better fit for [the same reasons](https://github.com/oat-sa/tao-core/pull/2428/files#r390139041).

actions/class.Export.php
SergiiTao: fixed

models/classes/resources/SecureResourceService.php
SergiiTao: i don't think it looks better. especially `use` part
```php
            $result = array_filter(
                $children,
                static function (core_kernel_classes_Resource $child) use ($permissions) {
                    return in_array('READ', $permissions[$child->getUri()], true);
                }
            );
```

models/classes/resources/SecureResourceService.php
SergiiTao: IMHO array_map doesn't make things more readable but makes them slower :)

models/classes/resources/SecureResourceService.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
```
Yury-Razhkou: missing copyrights

PR #2429, opened 2020-03-11
TBD-7  Missing HTTP Security Headers

wazelin:
- TBD-7 chore: set strict types to `ActionProtector
- TBD-7 feat(DI): introduce `InjectionAwareService`, allowing to gradually switch to constructor-based dependency injection
- TBD-7 feat(settings): introduce `SecuritySettingsRepositoryInterface` and its implementation to perform query and persistence operations on security settings
- TBD-7 chore(settings): remove direct service locator usage from `CspHeader` form factory
- TBD-7 chore(settings): remove persistence logic fom `CspHeader` form factory
- TBD-7 chore(security): inject and use `SecuritySettingsRepositoryInterface` implementation in `ActionProtector`
- TBD-7 feat(security): add a default `X-Content-Type-Options: nosniff` to `ActionProtector`
- TBD-7 feat(settings): add a new `Force HTPPS on this platform` security setting
- TBD-7 chore(settings): remove a no-longer-relevant security settings form title
- TBD-7 feat: add `Strict-Transport-Security` response header with 1 year max age by default


test/unit/models/classes/security/DataAccess/Repository/SecuritySettingsRepositoryTest.php
```diff
@@ -0,0 +1,254 @@
...
+ */
+
+namespace oat\tao\test\unit\model\security\DataAccess\Repository;
+
+use oat\tao\model\security\Business\Domain\Setting;
+use oat\tao\model\security\Business\Domain\SettingsCollection;
+use oat\tao\model\security\DataAccess\Repository\SecuritySettingsRepository;
+use oat\tao\model\service\SettingsStorage;
+use PHPUnit\Framework\TestCase;
+use PHPUnit_Framework_MockObject_MockObject;
```
Yury-Razhkou: `use oat\generis\test\TestCase;`
`use oat\generis\test\MockObject;`
Needed for migration to next versions of phpunit. Here is more detailed explanation: https://oat-sa.atlassian.net/wiki/spaces/NEX/pages/127832340/Spike+PHPUnit+migration

wazelin: I wouldn't say it's really "needed" as the test doesn't seem to use any deprecated methods.

config/default/actionProtection.conf.php
```diff
@@ -1,7 +1,19 @@
 <?php
 
-return new oat\tao\model\security\ActionProtector([
-    'frameSourceWhitelist' => [
-        "'self'"
+use oat\tao\model\security\ActionProtector;
+use oat\tao\model\security\DataAccess\Repository\SecuritySettingsRepository;
+use oat\tao\model\service\SettingsStorage;
+
+return new ActionProtector(
```
wazelin: This is way too far from a good solution, but at least it allows to reverse dependencies.
As you can see it doesn't reuse an already instantiated dependency.

test/unit/models/classes/security/DataAccess/Repository/SecuritySettingsRepositoryTest.php
Yury-Razhkou: `PHPUnit_Framework_MockObject_MockObject` - deprecated. And extending from `generis\test\TestCase` for future. We don't know for sure what will be deprecated later. Or just for the same code style in the project. All tests in tao extended from `generis\test\TestCase`

wazelin: All right, thanks!

models/classes/security/Business/Domain/SettingsCollection.php
```diff
@@ -0,0 +1,92 @@
...
+            $settings
+        );
+    }
+
+    /**
+     * @return Traversable|Setting[]
+     */
+    public function getIterator(): Traversable
+    {
+        return new ArrayIterator(
```
SergiiTao: what do you think of `yield from`?

models/classes/security/Business/Domain/SettingsCollection.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->settings = array_combine(
+            array_map([$this, 'extractKey'], $settings),
+            $settings
+        );
+    }
+
+    /**
+     * @return Traversable|Setting[]
+     */
+    public function getIterator(): Traversable
```
SergiiTao: could you please add a test for this method?
actually i'd like to have for hasMetDependencies but it is private :)

models/classes/security/Business/Domain/SettingsCollection.php
wazelin: It would instantiate a `Generator`, which doesn't make a difference.

PR #2434, opened 2020-03-12
Fix/udi 73/autocomplete scroll dropdown

bziondik:
Related to: https://oat-sa.atlassian.net/browse/UDI-73

Related to: 

- [x] https://github.com/oat-sa/tao-core-ui-fe/pull/109

Updated jQuery-Autocomplete from 1.2.19 to 1.4.10

How to test:
1. Login
2. Navigate to either Items/Access Control or Tests/Access Control
3. Make sure Users and Roles are populated to the point where they get their own scrollbar
4. Search for users/roles and wait for the results dropdown to show
5. Scroll the parent page (not the scrollbars incorporated in the page)
6. The result dropdown should move with the search field.


views/js/lib/jquery.autocomplete/jquery.autocomplete.js
```diff
@@ -631,13 +626,13 @@
...
             that.selectedIndex = -1;
-            clearInterval(that.onChangeInterval);
+            clearTimeout(that.onChangeTimeout);
             $(that.suggestionsContainer).hide();
             that.signalHint(null);
         },
 
         suggest: function () {
-            if (this.suggestions.length === 0) {
+            if (!this.suggestions.length) {
```
habemuscode: Opinable (Not necessary to do): I like to avoid (performance, memory, blablabla) if conditionals with negative clauses inside, because are 2 operations instead of one. Going further, I like to return as fast as possible when you enter a function, this is called "guard" in Angular world (I think, that world change a lot xD)

My proposal:

`if (this.suggestions.length) return;`

Not only here, checking the others if conditionals

bziondik: It is 3rd part lib, and I think it doesn’t need to be updated by anyone except its authors
https://github.com/devbridge/jQuery-Autocomplete/blob/master/dist/jquery.autocomplete.js
Please, send this proposal to the official repo

PR #2438, opened 2020-03-17
Fix/validate empty resource list

SergiiTao:
validation fails when there is nothing to export

### DIFF
--- a/actions/class.Export.php
+++ b/actions/class.Export.php
@@ -105,7 +105,9 @@ class tao_actions_Export extends tao_actions_CommonModule
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


actions/class.Export.php
```diff
@@ -105,7 +105,9 @@ public function index()
                     }
                 }
 
-                $this->getSecureResourceService()->validatePermissions($exportData['instances'], ['READ']);
+                if (!empty($exportData['instances'])) {
+                    $this->getSecureResourceService()->validatePermissions($exportData['instances'], ['READ']);
```
wazelin: If there's a bug fix, the case should also be added to the tests, though.

PR #2440, opened 2020-03-18
Fix/nex 778/prevent override persistence

bartlomiejmarszal:
Installation fix to work with a spanner. 

https://oat-sa.atlassian.net/browse/NEX-778


scripts/update/Updater.php
```diff
@@ -1343,8 +1343,8 @@ public function update($initialVersion)
                 )
             );
 
-            $this->setVersion('41.2.0');
+            $this->setVersion('41.2.1');
```
mike-ionut-mihai-sandu-tao: why ?

.Jenkinsfile
```diff
@@ -53,6 +53,15 @@ registry.service.consul:4444/tao/dependency-resolver oat:dependencies:resolve --
                         label: 'Add phpunit',
                         script: 'composer require phpunit/phpunit:^8.5'
                     )
+                    sh(
```
mike-ionut-mihai-sandu-tao: this should be in a different PR

install/class.Setup.php
```diff
@@ -263,6 +213,16 @@ public function __invoke($params)
 
         $serviceManager = $installator->getServiceManager();
 
+        if (!isset($parameters['configuration']['generis']['persistences'])) {
+            throw new InvalidArgumentException('Your config should have a \'persistence\' key under \'generis\'');
```
mike-ionut-mihai-sandu-tao: is it a InvalidArgumentException or runtimeException ?

install/class.Setup.php
```diff
@@ -263,6 +213,16 @@ public function __invoke($params)
...
+        }
+        $persistences = $parameters['configuration']['generis']['persistences'];
+        if (isset($persistences['default'])) {
```
mike-ionut-mihai-sandu-tao: this can be set to '' and isset will return true
```
$persistences['default'] = '';

 if (isset($persistences['default'])) {
     echo  'breaking somthing';
 }
```
```suggestion
         if (!empty($persistences['default'])) {
```

install/class.Setup.php
```diff
@@ -327,4 +281,59 @@ public function __invoke($params)
...
+     * and then back into a persistence configuration to ensure backwards compatibility
+     * with the previous process
+     * @param array $persistences
+     * @return array
+     */
+    private function wrapPersistenceConfig($persistences)
+    {
+        $defaultPersistenceConfig = $persistences['default'];
+        $options = [];
+        if (isset($defaultPersistenceConfig['connection'])) {
```
mike-ionut-mihai-sandu-tao: a lot of complexity here
switch case could help us to make it a little bit more structured?

scripts/update/Updater.php
jbout: Looks like a bad merge

install/class.Setup.php
jbout: Unchanged move from line 150, all other checks also throw InvalidArgumentException, would not recommend changing the behaviour in the context of this PR.

install/class.Setup.php
jbout: This is not a validation but a test to see if persistences are defined as a service or as direct configuration parameters:
if $persistences['default'] is set => direct persistence configuration parameters
if $persistences['type'] is set => full service configuration
if neither is set => PROBLEM

else there would be many more tests required: Is is an array, does it contain a key db_driver and so on

install/class.Setup.php
```diff
@@ -327,4 +281,100 @@ public function __invoke($params)
...
     }
+
+    /**
+     * Transforms the seed persistence configuration into command line parameters
+     * and then back into a persistence configuration to ensure backwards compatibility
+     * with the previous process
+     * @param array $persistences
+     * @return array
+     */
+    private function wrapPersistenceConfig($persistences)
```
jbout: While I see how this config was difficult to read, this split makes it even harder for me to understand what is happening:
* we generate the command line parameters ($options) based on the seed config of the default persistence
* we then generate a new persistence config based on the command line parameters
* we replace the default persistence config in the seed, by the newly generated one

install/class.Setup.php
```diff
@@ -327,4 +281,72 @@ public function __invoke($params)
...
+    {
+        $seedConfiguration = $this->getCommandLineParameters($persistences['default']);
```
jbout: Much clearer now.
However calling the command line parameter array "seedConfiguration" is still misleading, as the seed config actually gets generated in line 297

bartlomiejmarszal: How would u called?
```suggestion
        $configuration = $this->getCommandLineParameters($persistences['default']);
```

bartlomiejmarszal:
```suggestion
        $parsedConfiguration = $this->getCommandLineParameters($persistences['default']);
```

mike-ionut-mihai-sandu-tao:

PR #2441, opened 2020-03-18
Fix/udi 97/have access to children under restricted class

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/UDI-97

Accessible children under unaccessible classes also will be hidden


Alroniks:
Seems good in general, but I have concerns about performance on large lists of items. Do we have a cache on any level?

SergiiTao:
> Seems good in general, but I have concerns about performance on large lists of items. Do we have a cache on any level?

yep. I'll check the possibility to cache

PR #2442, opened 2020-03-18
Cache implemented in SecureResourceService

SergiiTao:
Caching results

### DIFF
--- a/models/classes/resources/SecureResourceService.php
+++ b/models/classes/resources/SecureResourceService.php
@@ -23,6 +23,8 @@ declare(strict_types=1);
 
 namespace oat\tao\model\resources;
 
+use common_cache_Cache;
+use common_cache_NotFoundException;
 use common_exception_Error;
 use core_kernel_classes_Class;
 use core_kernel_classes_Resource;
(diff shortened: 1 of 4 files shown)


models/classes/resources/SecureResourceService.php
```diff
@@ -43,9 +49,19 @@ class SecureResourceService extends ConfigurableService
      *
      * @return core_kernel_classes_Resource[]
      * @throws common_exception_Error
+     * @throws common_cache_NotFoundException
      */
     public function getAllChildren(core_kernel_classes_Class $resource): array
     {
+        $user = $this->getUser();
+
+        $cacheKey = $this->getCacheKeyFactory()->create($resource, $user);
```
wazelin: Good job!

Would've been even better if the cached implementation existed as a separate service, though.
Single interface, which is implemented by the current service, and the caching one. The caching one could've rely on the current one as its dependency and primary data source, then the pre-existing implementation will stay intact, you'll just add another caching layer to it.

The current implementation breaks single responsibility and open/closed principles

PR #2444, opened 2020-03-20
Fix/udi 97/have access to children under restricted class

SergiiTao:
Ticked: https://oat-sa.atlassian.net/browse/UDI-97

hide resources without restrictions if they are under restricted class


models/classes/resources/ResourceAccessDeniedException.php
```diff
@@ -24,7 +24,16 @@
 namespace oat\tao\model\resources;
 
 use RuntimeException;
+use Throwable;
 
 class ResourceAccessDeniedException extends RuntimeException
 {
+    public function __construct($forbiddenResourceUri = '', $code = 0, Throwable $previous = null)
```
wazelin: Generally it's advised not to change a constructor of an exception class, just because when you instantiate an exception, you usually won't check it's constructor's interface. The recommendation here is to leave the constructor alone, and to introduce a static `create` method, accepting whatever custom arguments you need, and returning a new instance of `self`.

models/classes/resources/SecureResourceCachedService.php
```diff
@@ -0,0 +1,202 @@
...
+
+class SecureResourceCachedService extends InjectionAwareService implements SecureResourceServiceInterface
+{
+    /** @var User */
+    private $user;
+    /** @var SecureResourceService */
+    private $service;
+    /** @var string */
+    private $cacheServiceId;
+    /** @var int */
```
wazelin: `int|null`

models/classes/resources/SecureResourceCachedService.php
```diff
@@ -0,0 +1,202 @@
...
+                $data,
+                $cacheKey,
+                $this->ttl
+            );
+        }
+    }
+
+    private function getCache(): ?common_cache_Cache
+    {
+        $cacheEnabled = empty(trim($this->cacheServiceId));
```
wazelin: The variable name is confusing, but I suggest keeping it, and reversing its value (`!empty(...)`).

models/classes/resources/SecureResourceService.php
```diff
@@ -31,9 +31,9 @@
 use oat\oatbox\session\SessionService;
 use oat\oatbox\user\User;
 
-class SecureResourceService extends ConfigurableService
+class SecureResourceService extends ConfigurableService implements SecureResourceServiceInterface
 {
-    public const SERVICE_ID = 'tao/SecureResourceService';
+    private const UNDER_ROOT_URI = 'http://www.tao.lu/Ontologies/TAO.rdf#AssessmentContentObject';
```
wazelin: What did you mean by that name? Sorry, I'm a bit confused by it.

models/classes/resources/SecureResourceService.php
```diff
@@ -88,21 +129,88 @@ public function validatePermissions(array $resourceUris, array $permissionsToChe
...
+    }
+
+    private function getClass(core_kernel_classes_Resource $resource): core_kernel_classes_Class
+    {
+        // fetch parent class
+        if (!$resource->isClass()) {
+            return current($resource->getTypes());
+        }
+
+        if (get_class($resource) === core_kernel_classes_Class::class) {
```
wazelin: `$resource instanceof core_kernel_classes_Class` :)

models/classes/resources/SecureResourceService.php
```diff
@@ -88,21 +129,88 @@ public function validatePermissions(array $resourceUris, array $permissionsToChe
...
+        );
+
+        $this->validateResourceUriPermissions($parentUris, $permissionsToCheck);
+    }
+
+    private function getClass(core_kernel_classes_Resource $resource): core_kernel_classes_Class
+    {
+        // fetch parent class
+        if (!$resource->isClass()) {
+            return current($resource->getTypes());
```
wazelin: `reset`?
As you don't know the position of the internal pointer.

models/classes/resources/SecureResourceServiceAllChildrenCacheCollection.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+namespace oat\tao\model\resources;
+
+use common_exception_Error;
+use core_kernel_classes_Resource;
+
+class SecureResourceServiceAllChildrenCacheCollection
+{
+    /** @var core_kernel_classes_Resource[] */
```
wazelin: But how?
The property is initialized with `array_keys()` value.

models/classes/resources/SecureResourceServiceAllChildrenCacheCollection.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * @return array
+     * @throws common_exception_Error
+     */
+    public function getInstances(): array
+    {
+        $result = [];
+        if ($this->resources) {
```
wazelin: Redundant, the loop won't loop if the property is empty anyway.

models/classes/resources/SecureResourceServiceAllChildrenCacheCollection.php
```diff
@@ -0,0 +1,54 @@
...
+    /** @var core_kernel_classes_Resource[] */
+    private $resources;
+
+    public function __construct(array $resources)
+    {
+        $this->resources = array_keys($resources);
+    }
+
+    /**
+     * @return array
```
wazelin:
```suggestion
     * @return core_kernel_classes_Resource[]
```

models/classes/service/InjectionAwareService.php
```diff
@@ -45,9 +48,38 @@ private function getSerializedDependencies(): array
...
+                $message = sprintf(
+                    'Cannot find property "%s" in class %s. Please name properties exactly like constructor parameters, or overload %s',
+                    $parameterName,
+                    static::class,
+                    __METHOD__
+                );
+                throw new RuntimeException($message);
+            }
+
+            $classProperties = $class->getProperty($parameterName);
```
wazelin:
```suggestion
            $classProperty = $class->getProperty($parameterName);
```

models/classes/service/InjectionAwareService.php
```diff
@@ -45,9 +48,38 @@ private function getSerializedDependencies(): array
...
+
+            if ($classProperties->isPrivate() || $classProperties->isProtected()) {
+                $classProperties->setAccessible(true);
+            }
+
+            $dependencies[] = $classProperties->getValue($this);
```
wazelin:
```suggestion
            $classProperty = $class->getProperty($parameterName);

            if ($classProperty->isPrivate()) {
                $classProperty->setAccessible(true);
            }

            $dependencies[] = $classProperty->getValue($this);
```

scripts/update/Updater.php
```diff
@@ -1346,5 +1351,12 @@ public function update($initialVersion)
             $this->setVersion('41.2.0');
         }
         $this->skip('41.2.0', '41.4.0');
+
+        if ($this->isVersion('41.4.0')) {
+            $this->addReport(
+                Report::createInfo('To make SecureResourceService use cache please run \oat\tao\scripts\install\InstallSecureResourceCachedService script')
+            );
```
wazelin: $this->getServiceManager()->registerService(
            SecureResourceServiceInterface::SERVICE_ID,
            new SecureResourceCachedService(
                $this->getServiceManager()->get(SecureResourceServiceInterface::SERVICE_ID),
                new ValidatePermissionsCacheKeyFactory(),
                new GetAllChildrenCacheKeyFactory(),
                common_cache_Cache::SERVICE_ID,
                null
            )
        );

models/classes/service/InjectionAwareService.php
SergiiTao: agree with var name, but a protected property is not accessible. so we need to check `$classProperties->isProtected` also

scripts/update/Updater.php
SergiiTao: by default tao uses file cache. file cache does not support TTL.
so it should be configured properly before we start to cache everything forever :)

PR #2445, opened 2020-03-20
Fix/actp 291/add method to check if actionqueue activated

SashaTAO:
JIRA: https://oat-sa.atlassian.net/browse/ACTP-291

**Problem:** 
Existing method InstantActionQueue::getLimits() previously returned sum of limits, when we had just one restriction (number of active sessions/users). It's is not valid any more since we have multiple implementations of basic restriction https://github.com/oat-sa/tao-core/blob/master/models/classes/actionQueue/restriction/basicRestriction.php with different limits (number of active users, RDS CPU, etc.) and own configuration file for each restriction. But we still need a possibility to check if ActionQueue is enabled or not.

**Solution:** 
Check configuration for each restriction. Value `0` means there are no restrictions. If at least one restriction in config/tao/ActionQueue.conf.php has value different from `0` it means ActionQueue is enabled.

**How to test:** 
Call InstantActionQueue::isActionEnabled() method with different configurations for restrictions in config/tao/ActionQueue.conf.php

How to test via UI described in https://github.com/oat-sa/extension-lti-proctoring/pull/131


models/classes/actionQueue/implementation/InstantActionQueue.php
```diff
@@ -134,6 +136,25 @@ public function getLimits(QueuedAction $action)
...
+        $actionConfig = $this->getActionConfig($action);
+        foreach ($this->getRestrictions($actionConfig) as $restriction => $value) {
+            if ($value !== 0) {
+                $enabled = true;
+                break;
+            }
+        }
+
+        return $enabled;
+    }
```
shpran: I think there's no need to create property `$enabled` and use `break`:
```
public function isActionEnabled(QueuedAction $action): bool
{
    $actionConfig = $this->getActionConfig($action);

    foreach ($this->getRestrictions($actionConfig) as $value) {
        if ($value !== 0) {
            return true;
        }
    }

    return false;
}
```

models/classes/actionQueue/implementation/InstantActionQueue.php
```diff
@@ -237,7 +258,7 @@ protected function getPositions(QueuedAction $action)
      * @param array $actionConfig
      * @return array
      */
-    private function getRestrictions(array $actionConfig)
+    private function getRestrictions(array $actionConfig): array
     {
         return array_key_exists('restrictions', $actionConfig) ? $actionConfig['restrictions'] : [];
```
shpran: Nip: we can replace it with: `return $actionConfig['restrictions'] ?? []`;

SashaTAO: fixed

models/classes/actionQueue/implementation/InstantActionQueue.php
SashaTAO: fixed

models/classes/actionQueue/implementation/InstantActionQueue.php
```diff
@@ -125,6 +126,7 @@ public function getTrend(QueuedAction $action)
      * @param QueuedAction $action
      * @return int
      * @throws ActionQueueException
+     * @deprecated Use InstantActionQueue::iaActionEnabled to check if action has enabled restrictions
```
zagovorichev: typo: i**a**Action...

models/classes/actionQueue/implementation/InstantActionQueue.php
```diff
@@ -125,6 +126,7 @@ public function getTrend(QueuedAction $action)
...
      */
     public function getLimits(QueuedAction $action)
```
zagovorichev: how bool can replace int? not sure that I can use isActionEnabled instead. Probably you can change an interface and replace this condition: https://github.com/oat-sa/extension-lti-proctoring/blob/7535b1abe5f25e46209ee0c6e5a126b4e1dd9ecc/model/ActivityMonitoringService.php#L70

SashaTAO: This method was used before to check if there are limits for ActionQueue (if SUM of all limits in the configuration is > 0 it means Login Queue is enabled). Later these limits were moved to dedicated own configurations for some restriction implementations, like `config/taoDelivery/CapacityService.conf.php`. 
I didn't want to change the interface because even the current method name `getLimits` does not describe the purpose of how it's used - to check if LoginQueue is activated (limits > 0)

models/classes/actionQueue/implementation/InstantActionQueue.php
SashaTAO: fixed

PR #2448, opened 2020-03-20
TBD-13: Support custom copy-to-clipboard form fields

wazelin:
- TBD-13 feat: support `copy-to-clipboard` action on any form field with `copy-to-clipboard` class

Required by [taoLti extension](https://github.com/oat-sa/extension-tao-lti/pull/232).

### DIFF
--- a/views/js/form/post-render-props.js
+++ b/views/js/form/post-render-props.js
@@ -23,61 +23,87 @@ define([
 ], function ($, __, feedback) {
     'use strict';
 
+    function createCopyToClipboardHandler($field) {
+        var
+            successFeedback = $field.data('copy-success-feedback')
+                || __('Resource Identifier has been copied to the clipboard'),
+            failureFeedback = $field.data('copy-failure-feedback')
+                || __('Resource Identifier could not be copied to the clipboard');
+
+        return function () {
+            var success;
+            try {
+                $field.select();
+                success = document.execCommand('copy');
+                $field.blur();
+                if (success) {
+                    feedback().success(successFeedback);
+                } else {
+                    feedback().error(failureFeedback);
+                }
+            } catch (err) {
+                feedback().error(__('Your browser does not support copying to the clipboard'));
+            }
+        }
+    }
+
     /**
      * Add a field with URI of an item etc and a button to copy it to the clipboard
      * @param $container
      * @private
      */
-    function _idToClipboard($container) {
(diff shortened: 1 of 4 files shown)


views/js/form/post-render-props.js
```diff
@@ -23,61 +23,87 @@ define([
...
-                        feedback().error(__('Your browser does not support copying to the clipboard'));
-                    }
-                }),
-            $fieldBox = $('<span>', {class: 'uri-container'});
+            $fields = $container.find('#id, .copy-to-clipboard');
 
         // if the field has already been added
         if ($('.uri-container').length) {
             return;
         }
```
habemuscode: if ($('.uri-container').length) return;

Short-form to try to follow code climate rule of 25 lines per function

views/js/form/post-render-props.js
```diff
@@ -23,61 +23,87 @@ define([
...
 
-        if (!isInstanceForm || !$idField.length) {
+        if (!isInstanceForm) {
             return;
         }
```
habemuscode: Short-form to try to follow code climate rule of 25 lines per function

views/js/form/post-render-props.js
```diff
@@ -23,61 +23,87 @@ define([
...
-            .after($('<div>')
-                .append([$label, $fieldBox]));
-        $fieldBox.height($copyField.outerHeight());
-
-        $copyField.addClass('final')
-        // re-add value
-            .val(value);
+        if (!$fields.length) {
+            return;
+        }
```
habemuscode: Short-form to try to follow code climate rule of 25 lines per function

views/js/form/post-render-props.js
```diff
@@ -23,61 +23,87 @@ define([
...
+            } else {
+                $fieldBox.height($field.outerHeight());
+                $field.wrap($fieldBox);
+                $field.parent().append($button);
+            }
+
+            $button.on('click', createCopyToClipboardHandler($field));
+            $field.addClass('final')
+                // re-add value
+                .val(value);
```
habemuscode: $field.addClass('final').val(value); For 1 chain, we can put in same line

views/js/form/post-render-props.js
wazelin: I absolutely disagree with that approach and would like to discuss it on our retro.

PR #2450, opened 2020-03-23
UDI-71 Acccess control error on abort

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/UDI-71

**Description**
Access control for a class
Typing in the search field
Stops typing, just long enough for triggering the start of the search 
Starts typing again
--> then it is shown an error message "unknown error" 

**How to Reproduce**

- Go on the search field on ACL page
- Start typing a user name prefix which is expected to have many entries 
- Stop so the search is triggered
- Then start typing again before the search results are got from the API

**Solution**

- [x] Don’t show the message on canceled requests.


views/js/controller/backoffice.js
```diff
@@ -64,6 +64,9 @@ define([
             //intercept errors
             //TODO this should belongs to the Router
             $doc.ajaxError(function (event, request, settings) {
+                // Request was manually aborted, isn't a error
+                if (request.statusText === 'abort') return;
```
wazelin:
```suggestion
            $doc.ajaxError(function (event, request, settings, thrownError) {
                // Request was manually aborted, isn't a error
               if(thrownError == ‘abort’ || thrownError == ‘undefined’) return;
```

It should be safer, as `statusText` is not part of the API.

views/js/controller/backoffice.js
```diff
@@ -63,7 +63,10 @@ define([
 
             //intercept errors
             //TODO this should belongs to the Router
-            $doc.ajaxError(function (event, request, settings) {
+            $doc.ajaxError(function (event, request, settings, thrownError) {
+                // Request was manually aborted, isn't a error
+                if(!thrownError || thrownError === 'abort') return;
```
bziondik: In what cases `!thrownError` will be true? why we need this condition?

habemuscode: Jquery param, I don't know if can be empty in some case

wazelin: It's there just in case. Any kind of error will provide `thrownError`, but it's not really documented what goes there in case of XHR cancellation.

bziondik: but maybe we should handle `!thrownError` in the old way? For our case I received `thrownError === 'abort'`, but only in Chrome. 
I suggest `if(thrownError === 'abort') return;`
and test locally in IE11, Firefox, Safary
I can check in Safary

habemuscode: I check in firefox, and return 'abort'

bziondik: checked in Safary, return 'abort'

bziondik: checked in Firefox, return 'abort'

bziondik: checked in IE11, `thrownError === 'abort'`

habemuscode: Ok, I will implement @bziondik solution :)

PR #2466, opened 2020-04-01
TBD-8 changed column to bigint to store ip2long()

Yury-Razhkou:
Related to: TBD-8
Changed column type from integer to bigint to store ip2long()


wazelin:
Oh boy, here we go again...
Will you take care of the backport to 111?

wazelin:
Thanks for fixing the build, @Yury-Razhkou!

PR #2468, opened 2020-04-01
remove funcAcl dependency

hutnikau:
This PR add dependency of tao-code on funcAcl extension without mentioning it in manifest.php:
https://github.com/oat-sa/tao-core/pull/1568/files

This adds cyclic dependency (tao-code and funcAcl depends on each other).

Current PR removes this cyclic dependency.

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -26,7 +26,6 @@ use common_cache_Cache;
 use common_Exception;
 use common_report_Report as Report;
 use core_kernel_persistence_smoothsql_SmoothModel;
-use oat\funcAcl\models\ModuleAccessService;
 use oat\generis\model\data\event\ResourceCreated;
 use oat\generis\model\data\event\ResourceDeleted;
 use oat\generis\model\data\event\ResourceUpdated;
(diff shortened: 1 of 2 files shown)


scripts/update/Updater.php
```diff
@@ -589,11 +588,11 @@ public function update($initialVersion)
         $this->skip('14.4.1', '14.8.0');
 
         if ($this->isVersion('14.8.0')) {
-            $moduleService = ModuleAccessService::singleton();
-            $moduleService->remove(
+            AclProxy::revokeRule(new AccessRule(
```
bartlomiejmarszal, line 591: I don't understand this change here. It's some old version block change. Is this fix something? Maybe description of this PR is misleading.

jbout: This update script uses the implementation instead of the abstraction, breaking the dependency graph

hutnikau: Usage of `ModuleAccessService` in tao-core brings dependency of tao-code on `funcAcl` extension.
That cyclic dependency (makes them depend on each other).

PR #2476, opened 2020-04-07
use new Manifest class

hutnikau:
requires: https://github.com/oat-sa/generis/pull/764


.Jenkinsfile
```diff
@@ -13,21 +13,6 @@ pipeline {
                     label : 'Create build build directory',
                     script: 'mkdir -p build'
                 )
-
-                withCredentials([string(credentialsId: 'jenkins_github_token', variable: 'GIT_TOKEN')]) {
```
Babacooll, line 17: Why removing this ?

composer.json
```diff
@@ -40,14 +49,19 @@
...
+            ]
         }
     },
     "require": {
         "php": "^7.1",
         "phpdocumentor/reflection-docblock": "2.*",
         "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
         "oat-sa/jig": "~0.1",
+        "oat-sa/composer-npm-bridge": "^0.3.0",
+        "oat-sa/generis": "*",
```
Babacooll, line 64: You will change that version before merging right ?

.Jenkinsfile
hutnikau: We do not need resolve dependencies from manifest anymore (this command was used to give package names from extension id's listed in manifest.php and generate new fake composer.json file to install dependent extensions).

composer.json
hutnikau: Yes, it will be changed after https://github.com/oat-sa/generis/pull/764 will be merged and released

gitromba (approved):
LGTM. Unit tests are passing.

PR #2478, opened 2020-04-08
BOSA-82: Last selected tree item doesn't retain focus after performing an action on it

wazelin:
- BOSA-82 fix(tree): retain `lastSelected` item identifier
- BOSA-82 chore(tree): extract repeating node selection by ID into a method
- BOSA-82 chore(tree): reorganize and simplify `jstree.init.treeOptions.callback.onload.nodeSelection()` method
- BOSA-82 chore(version): bump a fix version

One way to reproduce the issue is to open a list of Tests, select a not-selected-by-default-test and publish it. The focus will jump back to the selected-on-render item.
This PR fixes it, and retains the focus on an item, selected last.

### DIFF
--- a/views/js/layout/tree/provider/jstree.js
+++ b/views/js/layout/tree/provider/jstree.js
@@ -371,13 +371,13 @@ define([
                         var treeState       = $container.data('tree-state') || {};
                         var selectNode      = treeState.selectNode || options.selectNode;
                         var nodeSelection   = function nodeSelection() {
-                            //after refreshing tree previously node will be already selected.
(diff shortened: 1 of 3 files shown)


views/js/layout/tree/provider/jstree.js
```diff
@@ -461,6 +455,8 @@ define([
                             signature: $node.data('signature')
                         };
 
+                        lastSelected = $node.attr('id');
```
andrey-ponamarev, line 458: The value is defined above by variable `nodeId`

wazelin: True, thanks!

PR #2479, opened 2020-04-08
Feature/tbd 18/implement key cloak api provider

SergiiTao:
it makes possible to configure `InjectionAwareService` in seed files
keys of options should reflect __constuctor parameters names
the order of options can be random. it will be properly sorted 

example
```php
"KeyCloakAuthParameters": {
    "type": "configurableService",
    "class": "oat\\taoUdir\\model\\Acl\\Api\\KeyCloakAuthParameters",
    "options": {
        "clientId": "temp",
        "username":  "user",
        "password": "password",
        "grantType": "password",
        "apiBaseUrl": "http://192.168.178.71:7777/auth/admin/realms/master/",
        "apiTokenBaseUrl": "http://192.168.178.71:7777/auth/realms/master/protocol/",
        "tokenUrl": "openid-connect/token"
    }
}
```
result
```php
return new oat\taoUdir\model\Acl\Api\KeyCloakAuthParameters(
    'openid-connect/token',
    'temp',
    'user',
    'password',
    'password',
    'http://192.168.178.71:7777/auth/admin/realms/master/',
    'http://192.168.178.71:7777/auth/realms/master/protocol/'
);
```


install/class.Setup.php
```diff
@@ -229,7 +229,11 @@ public function __invoke($params)
                     $className = $config['class'];
                     $params = $config['options'];
                     if (is_a($className, \oat\oatbox\service\ConfigurableService::class, true)) {
-                        $service = new $className($params);
+                        if (is_a($className, \oat\tao\model\service\InjectionAwareService::class, true)) {
```
wazelin, line 232:
```suggestion
                        if ($className instanceof \oat\tao\model\service\InjectionAwareService) {
```

install/class.Setup.php
```diff
@@ -282,6 +286,52 @@ public function __invoke($params)
...
+        }
+
+        // checking if there are gaps and default parameters can fill them
+        end($sortedParameters);
+        $lastKey = key($sortedParameters);
+        if (count($sortedParameters) - 1 < $lastKey) {
+            $listOfGaps = array_diff(range(0, $lastKey), array_keys($sortedParameters));
+
+            foreach ($constructParameters as $parameter) {
+                if (in_array($parameter->getPosition(), $listOfGaps, true)) {
```
wazelin, line 318:
```suggestion
            $gapMap = array_diff_key(array_fill(0, $lastKey + 1, true), $sortedParameters);

            foreach ($constructParameters as $parameter) {
                if (isset($gapMap[$parameter->getPosition()])) {
```

install/class.Setup.php
```diff
@@ -282,6 +286,52 @@ public function __invoke($params)
...
+
+        $constructParameters = $reflectionClass->getMethod('__construct')->getParameters();
+
+        $sortedParameters = [];
+
+        foreach ($constructParameters as $parameter) {
+            $parameterName = $parameter->getName();
+            if (isset($parametersToSort[$parameterName])) {
+                $sortedParameters[$parameter->getPosition()] = $parametersToSort[$parameterName];
+            }
```
wazelin, line 308:
```suggestion
            } elseif($parameter->isDefaultValueAvailable()) {
                $sortedParameters[$parameter->getPosition()] = $parameter->getDefaultValue();
            } else {
                throw new RuntimeException(
                    sprintf('There is no default parameter for `$%s` in %s::__contruct', $parameter->getName(), $class)
                );
            }
```
and the second loop can be removed I guess, or am I wrong here?

install/class.Setup.php
SergiiTao: instanceof doesn't work with class names.

install/class.Setup.php
SergiiTao: like. 
will check tomorrow

install/class.Setup.php
SergiiTao: the idea of the second loop to cover situation like
`SomeClass::__construct($a, $c='def', $d='def')`

`sortParameters('SomeClass', ['a'=>'val', 'd'=>'val'])`
in this case 'the sorter' adds `c` with default value

install/class.Setup.php
wazelin: Oh, right. Didn't notice it was a name.

install/class.Setup.php
SergiiTao: changed

install/class.Setup.php
wazelin: OK, but in that case `isset($parametersToSort[$parameterName])` will return `false` on a second argument (`$c`) and will add it with a default value to `$sortedParameters[1]`, if you apply my suggestion.

wazelin: I've an even better suggestion. See another comment.

install/class.Setup.php
```diff
@@ -282,6 +286,52 @@ public function __invoke($params)
...
+                    }
+
+                    $sortedParameters[$parameter->getPosition()] = $parameter->getDefaultValue();
+                }
+            }
+        }
+
+        ksort($sortedParameters);
+
+        return array_values($sortedParameters);
```
wazelin, line 332:
```suggestion
        foreach ($constructParameters as $parameter) {
            $parameterName = $parameter->getName();
            try {
                $sortedParameters[] = $parametersToSort[$parameterName] ?? $parameter->getDefaultValue();
            } catch (ReflectionException $exception) {
                throw new RuntimeException(
                    sprintf('There is no default parameter for `$%s` in %s::__contruct', $parameter->getName(), $class)
                );
            }
        }

        return $sortedParameters;
```

SergiiTao: i don't want to set all default values. only when it is necessary

wazelin: But what's the difference in the end?

install/class.Setup.php
wazelin: I mean, I see your point. But can that case even happen?
If the service was tried to be instantiated with arguments like `'a'=>'val', 'd'=>'val'`, it would've failed anyway.
I'll check deeper tomorrow, but seems to me, the default value setting may be omitted completely then.

install/class.Setup.php
SergiiTao: in the end if the last default parameter is not set it doesn't go to the config.
and if this default parameter will be changed in the class for some reason the behavior of the configurable class will be changed accordingly without the necessity to change the configuration.

install/class.Setup.php
SergiiTao: it won't fail. because of the second loop. it adds default value for the `$c`
this is just an edge case.
i hope it will never happen and all parameters including defaults will be configured

wazelin: No-no, I understand that. I'm just questioning the whole case here. How can it happen that we have values set for argument 1 and 3, but not for 2?

wazelin: And what I'm proposing on top of that is to fill up the set values only.

SergiiTao: if `2` and `3` have default values and `3` is important for me and `2` is not. so I can skip `2` in seed file and let Setup script fill it with a default value

wazelin: But how can you skip a second argument, and provide a third one?

SergiiTao: from a seed file
```php
"Service": {
    "type": "configurableService",
    "class": "ServiceClass",
    "options": {
        "first": "value1",
        "third":  "value3"
    }
}
```

install/class.Setup.php
wazelin: All right, I see what you mean now.
Updated the suggestion, kindly have a look.

install/class.Setup.php
```diff
@@ -301,35 +301,26 @@ private function sortParameters(string $class, array $parametersToSort): array
 
         $sortedParameters = [];
 
-        foreach ($constructParameters as $parameter) {
+        while($constructParameters) {
```
wazelin, line 304:
```suggestion
        while($constructParameters && $parametersToSort) {
```

PR #2485, opened 2020-04-15
UNO-195: Support unicode display name

wazelin:
- UNO-195 fix(sanitization): respect unicode string length in `tao_helpers_Display::textCleaner()` method
- UNO-195 feat(sanitization): allow unicode characters for `UTF*` encodings in `tao_helpers_Display::textCleaner()` method
- UNO-195 chore(version): bump a fix one

One way to reproduce the issue is to try and export any item/test/delivery with a multibyte unicode symbol in its label. The suggested file name name would be distorted.

### DIFF
--- a/helpers/class.Display.php
+++ b/helpers/class.Display.php
@@ -33,6 +33,7 @@ use oat\tao\model\service\ApplicationService;
  */
 class tao_helpers_Display
 {
+    private static $replacement = '_';
 
     /**
      * Enables you to cut a long string and end it with [...] and add an hover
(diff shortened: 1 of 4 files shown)


helpers/class.Display.php
```diff
@@ -153,11 +145,57 @@ public static function sanitizeXssHtml($input)
...
+    }
+
+    private static function replaceWith(array $matches, string $replacement): string
+    {
+        return str_repeat($replacement, mb_strlen(reset($matches), self::getEncoding()));
+    }
+
+    private static function getEncoding(): string
+    {
+        static $encoding;
```
SergiiTao, line 183: why here and not before methods?

wazelin: Just to keep it "super" private.

PR #2489, opened 2020-04-20
BSA-84: Render accessibility tools on top of .qti-content

wazelin:
- BSA-84 fix: increase `.action-bar`'s `z-index` in order to render sub-items such as accessibility tools over `.qti-content`, having a `z-index` value of `1`
- BSA-84 chore(version): bump a fix one


bziondik:
How to test? 
Some other tools in markup after `.action-bar`'s and no need to set z-index (calculator)
Some inside `.qti-content` but they handle z-index and can't be moved on `.action-bar`(masking, line reader)

wazelin:
> How to test?
> Some other tools in markup after `.action-bar`'s and no need to set z-index (calculator)
> Some inside `.qti-content` but they handle z-index and can't be moved on `.action-bar`(masking, line reader)

One way to test the fix is to install and use https://github.com/oat-sa/extension-tao-texthelp in a delivery.

The thing here is that the action bars, having items, should display those items on top of `.qti-content`, which has a `z-index` value of `1`.

wazelin:
> don't we need a version bump?

We do, it was merged without the conflicts T_T
https://github.com/oat-sa/tao-core/pull/2489/commits/82eb3a5250330b74c7d9279066645a09b1ea44ff

views/scss/inc/_action-bars.scss
```diff
@@ -5,6 +5,7 @@
     padding: 3px;
     overflow: hidden;
     color: whiten($info, .9);
+    z-index: 10;
```
bziondik, line 8: The fix should be done here
.text-to-speech {
z-index: 10;
}
https://github.com/oat-sa/extension-tao-texthelp/blob/114747a5d033a540be7026aaa5783bfdbfc5c269/views/js/runner/plugins/tools/textToSpeech/textToSpeech.scss#L3
It will be safer than change index in action-bar because we have a lot of tools and It can relay to some other issues

wazelin: No, I cannot agree with that.
The plugin shouldn't be aware of the context's styles whatsoever.

And since we've raised the `.qti-content`'s `z-index`, and `.action-bar` should actually be on top of it, it's only reasonable to have them raised higher. That will also help us in preventing issues like that with other plugins, having popovers.

bziondik: @wazelin ok, I agree with you. So maybe this fix https://github.com/oat-sa/tao-test-runner-qti-fe/pull/160/files was the wrong solution.
#qti-content {
        -webkit-overflow-scrolling: touch;
        max-width: map-get($widths, item-max-width) * 1px;
        width: 100%;
        margin: auto;
        **position: absolute;
        z-index: 1;**
}
And we should remove  z-index: 1; and fix in other way

wazelin: That was the first thing I tried to do, but failed and gave up.

Tried a different approach, and seem to have succeeded. I think it's a good solution, please have a look https://github.com/oat-sa/tao-test-runner-qti-fe/pull/176.

wazelin:
Replaced with https://github.com/oat-sa/tao-test-runner-qti-fe/pull/176.

PR #2496, opened 2020-04-29
UNO-205: Allow variadic arguments in InjectionAwareService implementations

SergiiTao:
skip checking parent constructor inherited from Configurable
added support of variadic parameters

### DIFF
--- a/models/classes/service/InjectionAwareService.php
+++ b/models/classes/service/InjectionAwareService.php
@@ -113,7 +113,7 @@ FACTORY;
 
             $parameter->isVariadic()
                 ? yield from $value
-                : yield;
+                : yield $value;
         }
     }
 
--- a/models/classes/service/InjectionAwareService.php
+++ b/models/classes/service/InjectionAwareService.php
@@ -21,7 +21,6 @@
 namespace oat\tao\model\service;
 
 use common_Utils;
-use oat\oatbox\Configurable;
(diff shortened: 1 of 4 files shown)


models/classes/service/InjectionAwareService.php
```diff
@@ -102,7 +109,11 @@ protected function iterateParameters(InjectionAwareService $service): iterable
                 $classProperty->setAccessible(true);
             }
 
-            yield $classProperty->getValue($service);
+            $value = $classProperty->getValue($service);
+
+            $parameter->isVariadic()
+                ? yield from $value
```
wazelin, line 115: Would you be so kind to add a test case for variadic arguments?

models/classes/service/InjectionAwareService.php
```diff
@@ -81,6 +82,12 @@ protected function iterateParameters(InjectionAwareService $service): iterable
     {
         $class = new ReflectionClass($service);
         $constructor = $class->getMethod('__construct');
+
+        // do not check parent constructor inherited from Configurable::class
+        if ($constructor->class === Configurable::class) {
```
wazelin, line 87: I'd say it's better to add an empty `__construct()` implementation to `InjectionAwareService` itself.

models/classes/service/InjectionAwareService.php
SergiiTao: added

PR #2501, opened 2020-05-04
Create standard way to responde Json

gabrielfs7:
The intent of this PR is create a standard way to build all the api responses without having to hardcode the payload.

- Easy to all the extensions follow the same standard.
- The developers do not need to hardcode response standard every time.
- If we want to extend or change the way we respond our api calls we can do it in a single place.
- The builder can add extra necessary headers to the response if needed.


scripts/update/Updater.php
```diff
@@ -1344,5 +1344,9 @@ public function update($initialVersion)
         }
 
         $this->skip('42.0.4', '42.0.5');
+
+        if ($this->isVersion('42.0.5')) {
```
mike-ionut-mihai-sandu-tao, line 1348: you don't need to do this ,  just increase the skipped version 
```suggestion
      $this->skip('42.0.4', '42.1.0');
```

models/classes/http/response/SuccessJsonResponse.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->data = $data;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function jsonSerialize(): array
+    {
+        return [
+            'success' => true,
```
mike-ionut-mihai-sandu-tao, line 39: - [ ] let's move this to constants please

models/classes/http/builder/ResponseBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2019 (original work) Open Assessment Technologies SA
```
boajer, line 18:
```suggestion
 * Copyright (c) 2020 (original work) Open Assessment Technologies SA
```

models/classes/http/response/SuccessJsonResponse.php
gabrielfs7: IMO we should not move this to constants. We do not want to externalise those parameters and this is part of the normalisation of the result. Can you please justify why should we have constant for this?

siwane:
After testing, I have found some unexpected behaviour:
TAO 3 4 0 sprint127

PR #2504, opened 2020-05-07
Check previous headers in ResponseEmitter

peetya:
Related JIRA tickets:
https://oat-sa.atlassian.net/browse/NEX-109
https://oat-sa.atlassian.net/browse/NEX-99

This PR provides fix for those cases when `ResponseEmitter` sets the header even if the headers were already sent by other services, like for example `tao_helpers_Http::returnStream` and `tao_helpers_Http::returnFile`, and it caused the following errors in the logs:
```
[2020-05-07 16:06:06] tao.DEBUG: Invoking oat\taoQtiTestPreviewer\controller\Previewer::asset by http://backoffice.docker.localhost/ontologies/tao.rdf#superUser ["GENERIS","CLEARRFW"] []
[2020-05-07 16:06:06] tao.ERROR: php error(2): Cannot modify header information - headers already sent by (output started at /var/www/html/tao/helpers/class.Http.php:381) {"0":"PHPERROR","file":"/var/www/html/tao/models/classes/http/ResponseEmitter.php","line":47,"trace":{"trace":[{"file":"/var/www/html/tao/models/classes/http/ResponseEmitter.php","line":47,"function":"header"},{"file":"/var/www/html/tao/models/classes/routing/ActionEnforcer.php","line":189,"function":"__invoke","class":"oat\\tao\\model\\http\\ResponseEmitter","type":"->"},{"file":"/var/www/html/tao/models/classes/routing/ActionEnforcer.php","line":163,"function":"execute","class":"oat\\tao\\model\\routing\\ActionEnforcer","type":"->"},{"file":"/var/www/html/tao/models/classes/routing/TaoFrontController.php","line":101,"function":"__invoke","class":"oat\\tao\\model\\routing\\ActionEnforcer","type":"->"},{"file":"/var/www/html/tao/models/classes/mvc/Bootstrap.php","line":341,"function":"__invoke","class":"oat\\tao\\model\\routing\\TaoFrontController","type":"->"},{"file":"/var/www/html/tao/models/classes/mvc/Bootstrap.php","line":181,"function":"mvc","class":"oat\\tao\\model\\mvc\\Bootstrap","type":"->"},{"file":"/var/www/html/tao/models/classes/mvc/Bootstrap.php","line":243,"function":"dispatchHttp","class":"oat\\tao\\model\\mvc\\Bootstrap","type":"->"},{"file":"/var/www/html/index.php","line":31,"function":"dispatch","class":"oat\\tao\\model\\mvc\\Bootstrap","type":"->"}]}} []
```

The easiest way to test it to preview an item in TAO and check the logs also (for this function bothh `returnStream` and `returnFile` are used). Also to avoid any regression TAO should get checked in general that all the endpoints renders the content properly. The expected behaviour now is to not get any error regarding the headers and also no regression got introduced.


models/classes/http/ResponseEmitter.php
```diff
@@ -44,13 +44,18 @@ public function __invoke(ResponseInterface $response)
             $response->getStatusCode(),
             $response->getReasonPhrase()
         );
-        header($http_line, true, $response->getStatusCode());
-        foreach ($response->getHeaders() as $name => $values) {
-            foreach ($values as $value) {
-                header("$name: $value", false);
+
+        if (!headers_sent()) {
```
gabrielfs7, line 48: IMO we should fix those extensions sending the headers before and not here. The headers are part of the response and in this case we have partial response emitter.

siwane: agree, it is a defensive approach to fix output from controller that has not been refactored yet

peetya: the best approach would be to refactor how we are handling the responses in general, because at the moment we are creating them on the fly, extending it with additional headers if needed. But there is no restriction that a Controller or an Action must return a Response object. If someone forget an `echo` in the code, it will throw error because of the double sent headers. This is a bugfix to prevent that.

PR #2508, opened 2020-05-12
Fix deprecated in cleanup

Yury-Razhkou:
Related to: BSA-119
Fixed deprecated class in cleanup script
How to test(example): `php index.php  '\oat\tao\scripts\tools\CleanClass' 'taoItems_models_classes_ItemsService' 'taoItems' 'http://sample/bosa.rdf#i5e284568c573919336809f37f03c7c0c9'`

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -56,7 +56,7 @@ return [
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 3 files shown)


ktsirkou:
> Related to: BSA-119
> Fixed deprecated class in cleanup script

It will be very helpful if you can provide an example of how to run the script

PR #2512, opened 2020-05-18
TBD-12: User-friendly 500-error page

wazelin:
ezgif-2-a74e81f80811

[TBD-12](https://oat-sa.atlassian.net/browse/TBD-12) feat: replace big and scary 500-error messages with a bit calmer ones on the error page

To be backported.


views/templates/error/error500.tpl
```diff
@@ -15,37 +16,40 @@
...
+    <div class="error-page">
+        <div class="error-page__form">
+            <img class="error-page__form__logo"
+                 src="<?= ROOT_URL ?>tao/views/media/thunderbolt.svg"
+                 alt="<?= __('Error') ?>"
+            >
+            <div>
+                <h2 class="error-page__form__title"><?= __('Something unexpected happened.') ?></h2>
+                <p class="error-page__form__message">
+                    <?= __(
```
habemuscode, line 30: I don't see any locals in PR, you run translations for these new keys?

wazelin: No need as there's no actual translation available, they keys will be updated during a release, just as usual.

PR #2513, opened 2020-05-19
Fix/uno 199/provide correct mime type

SergiiTao:
Ticket: https://oat-sa.atlassian.net/browse/UNO-199

Check mime type by file content if an extension is not available


helpers/class.Http.php
```diff
@@ -239,7 +239,7 @@ public static function returnFile($filename, $contenttype = true, $svgzSupport =
     {
         if (tao_helpers_File::securityCheck($filename, true)) {
             if (file_exists($filename)) {
-                $mimeType = tao_helpers_File::getMimeType($filename, true);
+                $mimeType = tao_helpers_File::getMimeType($filename);
                 if ($contenttype) {
                     header('Content-Type: ' . $mimeType);
```
wazelin, line 244:
```suggestion
                if ($contenttype) {
                    header('Content-Type: ' . tao_helpers_File::getMimeType($filename));
```

No need to actually read the file's MIME type if it's wasn't requested.

PR #2519, opened 2020-05-26
Create new endpoint to return platform languages

gabrielfs7:
Create new endpoint to return platform languages

https://oat-sa.atlassian.net/browse/NEX-1121

Request example:

```
curl --location --request GET 'https://test-tao-deploy.docker.localhost/tao/Languages/index' \
--header 'Content-Type: application/json' \
--header 'Cookie: tao=4cb8e3ee77544be12662ec05ed4437b4'
```


actions/structures.xml
```diff
@@ -55,6 +55,7 @@
                     <action id="role-move" name="Move role" url="/tao/Roles/moveInstance" context="instance" group="none" binding="moveNode">
                         <icon id="icon-move" src=""/>
                     </action>
+                    <action id="get-languages" name="Get Languages" url="/tao/Languages/index" group="none" context="none" binding="getLanguages" />
```
siwane, line 58: As u r providing an api, u do not need this part

actions/class.Languages.php
```diff
@@ -0,0 +1,40 @@
...
+class tao_actions_Languages extends tao_actions_CommonModule
+{
+    public function index(): void
+    {
+        try {
+            $this->getResponseFormatter()
+                ->withExpiration(time() + 60);
+
+            $this->returnSuccessJsonResponse(
+                tao_helpers_I18n::getAvailableLangsByUsage(
```
siwane, line 32: could u avoid static usage pls, I do not know if there is already an alternative or not

doc/taoApi.yml
```diff
@@ -0,0 +1,59 @@
...
+          type: boolean
+          example: false
+        code:
+          type: integer
+          example: 400
+        message:
+          type: string
+  responses:
+    LanguagesResponse:
+      description: Bad request
```
siwane, line 43: Good request

manifest.php
```diff
@@ -138,54 +138,55 @@
...
+        [AccessRule::GRANT, TaoRoles::TAO_MANAGER,          ['ext' => 'tao','mod' => 'TaskQueue']],
+        [AccessRule::GRANT, TaoRoles::TAO_MANAGER,          ['ext' => 'tao','mod' => 'Users']],
+        [AccessRule::GRANT, TaoRoles::TAO_MANAGER,          ['ext' => 'tao','mod' => 'WebService']],
+        [AccessRule::GRANT, TaoRoles::TAO_MANAGER,          ['ext' => 'tao','mod' => 'Security']],
+        [AccessRule::GRANT, TaoRoles::REST_PUBLISHER,       ['ext' => 'tao','mod' => 'TaskQueue', 'act' => 'get']],
+        [AccessRule::GRANT, TaoRoles::REST_PUBLISHER,       ['ext' => 'tao','mod' => 'TaskQueue', 'act' => 'getStatus']],
+        [AccessRule::GRANT, TaoRoles::SYSTEM_ADMINISTRATOR, ['ext' => 'tao','mod' => 'ExtensionsManager']],
+        [AccessRule::GRANT, TaoRoles::LOCK_MANAGER,     'tao_actions_Lock@forceRelease'],
+        [AccessRule::GRANT, TaoRoles::PROPERTY_MANAGER, 'tao_actions_PropertiesAuthoring'],
         [AccessRule::GRANT, TaoRoles::SYSTEM_ADMINISTRATOR, oat\tao\controller\api\Users::class],
```
siwane, line 190: As u are refromating this part, could simplify fqn pls?

models/classes/http/Controller.php
```diff
@@ -115,4 +118,32 @@ protected function setContentHeader($contentType, $charset = 'UTF-8')
         $this->response = $this->getPsrResponse()->withHeader('content-type', $contentType . ';' . $charset);
         return $this;
     }
+
+    protected function returnSuccessJsonResponse(array $data, int $statusCode = 200): void
```
siwane, line 122: This controller aims to handle pure http, and abstract TAO related stuff. These methods does return anything, could u adapt the name pls

siwane, line 122: Also, I would not put TAO response formatting as default: 'success' may be not kept as TAO http response standard. We should put this stuff outside of httpController, in a trait, or in a TAO controller
wdyt?

scripts/update/Updater.php
```diff
@@ -1344,5 +1344,21 @@ public function update($initialVersion)
         }
 
         $this->skip('42.0.4', '42.10.2');
+
+        if ($this->isVersion('42.10.2')) {
+            AclProxy::applyRule(
```
siwane, line 1349: Is it required to have specific ACL for this controller? I think we can work without, need to double check

siwane:

models/classes/http/Controller.php
gabrielfs7: Agreed. Fixed!

gabrielfs7: Moved!

scripts/update/Updater.php
gabrielfs7: It was the way it worked today as we discussed

manifest.php
gabrielfs7: done

doc/taoApi.yml
gabrielfs7: fixed

actions/class.Languages.php
gabrielfs7: Today it is really tied to static methods, I would have to refactor some places. I agree with you and will try to move it if possible :)

actions/structures.xml
gabrielfs7: Removed, thanks!

models/classes/http/Controller.php
```diff
@@ -21,6 +21,9 @@
 
 namespace oat\tao\model\http;
 
+use oat\tao\model\http\formatter\ResponseFormatter;
```
siwane, line 24: nope

gabrielfs7: removed

actions/class.Languages.php
gabrielfs7: I think this is not feasible to do in the scope of this task since the method uses several static methods inside: 

https://github.com/oat-sa/tao-core/blob/master/helpers/class.I18n.php#L122

It would require a bigger refactoring and many impact points. So I would not add to the scope of this task.

One alternative would be to have a `LanguageRepository::findAll()` delegating a call for this helper, so we deprecate the helper and use the repository, but still we will be using static calls inside the repository.

wdyt? We can create a tech debt (one more) also for this :)

models/classes/http/HttpJsonResponseTrait.php
```diff
@@ -0,0 +1,67 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2019 (original work) Open Assessment Technologies SA;
```
boajer, line 18:
```suggestion
 * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
```

PR #2520, opened 2020-05-27
Fix: SecureResourceServiceInterface::getAllChildren() implementation, not returning data, mapped by resource URIs

SergiiTao:
Keys added to the array


wazelin:
Approved. Next time, though, could you please pay a bit more attention to the PR's title and description?
The title is what we're eventually gonna see in the changelog. `Return array must have keys` says nothing without a context.

wazelin:
I've modified the title, feel free to improve it, as the one I'd come up with is a bit bulky :)

PR #2528, opened 2020-06-02
Use Doctrine Migrations STABLE

bugalot:
**Goal 1: Migrate to beta1/stable and remove dependencies to @internal classes**
The first goal is to replay all validation scenarios on Doctrine Migrations beta1. Most of the effort was to remove the dependency from `oat-sa/tao-core` to `Doctrine\Migrations\MigrationRepository` which is an @internal class, meaning that bacwkard compatibility is not a promise...

And indeed, between alpha1 and beta1, the interface got broken, so we cannot afford to extend this class so we found a little bit of a hack. Icing on the cake, the code is reduced.

**Goal 2: Provide the addReport() method**
As mentioned in https://oat-sa.atlassian.net/wiki/spaces/OAT/pages/380764311/1+PR+Review+Case+Study+-+New+Updater#No-More-Reports we found a regression. No more Reports allowed, as in the previous updaters. We identified this as a regression and implemented a basic support of this feature.


gitromba:
- [] New code is respecting code style rules
- [] New code is respecting best practices
- [] New code is not subject to concurrency issues (if applicable)
- [] Changelog is updated according to changes (if applicable)
- [] Documentation is updated according to changes (if applicable)

gitromba:
I couldn't confirm one acceptance criteria: `A scenario in our Case Study reflects the change`.

scripts/tools/migrations/AbstractMigration.php
```diff
@@ -23,15 +23,47 @@
...
+     *
+     * Will write to output the given $report Report object. This method
+     * enables developers to communicate important information about the execution
+     * of their migrations e.g. to Operations.
+     *
+     * @param Report $report
+     */
+    protected function addReport(Report $report): void
+    {
+        $this->write(ReportHelper::renderToCommandLine($report, false));
```
gitromba, line 67: Using the `logger` instance provided by the local `LoggerAwareTrait ` would result in lower coupling between Doctrine migrations and our implementation. Currently there's absolutely no adapter layer between the 2, but it's always a good idea to maintain low coupling between components.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
```diff
@@ -0,0 +1,66 @@
...
+            {
+                $this->addReport(Report::createFailure('Migration Down!'));
+            }
+        };
+
+        $expectedUpReportMessage = 'Migration Up!' . PHP_EOL;
+        $expectedDownReportMessage = 'Migration Down!' . PHP_EOL;
+
+        $loggerMock->expects($this->exactly(2))
+                   ->method('notice')
```
gitromba, line 57: The test should verify whether the `write` method was called instead of checking on the `notice` method call because `addReport` knows only about `write`.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $connectionMock = $this->createMock(Connection::class);
+        $loggerMock = $this->createMock(LoggerInterface::class);
+        $schemaMock = $this->createMock(Schema::class);
+
+        $migration = new class($connectionMock, $loggerMock) extends AbstractMigration
+        {
+            public function up(Schema $schema): void
+            {
+                $this->addReport(Report::createInfo('Migration Up!'));
```
gitromba, line 44: The `Report` class seems to have ~400 lines and some, first glance, complex method implementations. This dependency can lead to some unexpected surprises. I would recommend to mock it.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            public function down(Schema $schema): void
+            {
+                $this->addReport(Report::createFailure('Migration Down!'));
```
gitromba, line 49: Since the signature of `addReport` demands a valid `\common_report_Report` instance to be passed, I would, either mock ;), or use 1 static creation method only, `createInfo` for instance. This was making me thinking perhaps there's some differences in behaviour, and I've started to dig in to reveal what's going on.

bugalot:
> I couldn't confirm one acceptance criteria: `A scenario in our Case Study reflects the change`.

Please see scenario B taoUpdate.php output. in taoQtiTest Scenario B PR you will find two reports using the feature (use this branch for the tao-core one)

scripts/tools/migrations/AbstractMigration.php
bugalot: Yes but I we use the logger, it will not appear to the output of `taoUpdate.php` which is the one that the Operation Engineer will read after update and communicate to other departments. So it seems mandatory to me to keep this strategy.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
bugalot: OK I'll mock it.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
bugalot: OK I'll mock it.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
bugalot: OK I'll test the write method instead.

bugalot: But actually not because the `ẁrite` method is in the abstract class I want to test (by inheritance) so I cannot mock it ;).

gitromba: Indeed! I've missed this one.

scripts/tools/migrations/AbstractMigration.php
gitromba: Discusses, and as those are 2 distinct logger instances this must be kept as is.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
```diff
@@ -0,0 +1,96 @@
...
+use \common_report_Report as Report;
+
+class AbstractMigrationTest extends TestCase
+{
+    public function testAddReport()
+    {
+        $expectedUpReportMessage = 'Migration Up!';
+        $expectedDownReportMessage = 'Migration Down!';
+
+        $connectionStub = $this->createMock(Connection::class);
```
gitromba, line 39: These are actually mocks, sry if I have been confusing.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
gitromba: 💯  indeed!

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $loggerStub = $this->createMock(LoggerInterface::class);
+        $schemaStub = $this->createMock(Schema::class);
+
+        $upReportStub = $this->createMock(Report::class);
```
gitromba, line 43: With a fluent interface and some rearrangement, you might gain extra readability and maintainability.
`
        $reportMock = $this->createMock(Report::class)
            ->method('getType')
            ->willReturn(Report::TYPE_INFO)
            ->method('getIterator')
            ->willReturn($this->createMock(\Traversable::class));

        $upReportMock = $reportMock->method('getMessage')
            ->willReturn($expectedUpReportMessage);

        $downReportMock = $reportMock->method('getMessage')
            ->willReturn($expectedDownReportMessage);

`

bugalot: Did you try it :smiley: ? It fails. I got the same error when I tried to refactor it, so I'll keep it as it is.

test/unit/scripts/tools/migrations/AbstractMigrationTest.php
bugalot: Fixed.

bugalot:
I did some changes. Let me know if we can put this in the done column.

PR #2531, opened 2020-06-03
TAO-10248 Extend Elastic search query

luisfn:
- Added constant for Assembled Deliveries
- Moved slug creation to helper class


helpers/class.Slug.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+class tao_helpers_Slug
+{
+    public static function create(string $text): string
+    {
+        return strtolower(trim(preg_replace('/[^A-Za-z0-9-]+/', '-', $text)));
+    }
+}
```
uncleempty, line 28: Could you please add a new line?

luisfn: Thanks! Fixed

luisfn:
> Sorry, hurried a bit:
> no manifest/updater script changes. This also _must_ be added before PR is merged
> Everything else is ok

Manifest/Updater will be updated when we have everything ready, on the integration branch

PR #2532, opened 2020-06-05
Feature/tao 10280/resource event handling and indexing

uncleempty:
_NOTE_: this is needed for [this](https://github.com/oat-sa/lib-tao-elasticsearch/pull/18) PR

* Implement the Listener to watch the ResourceCreated event.
* Create a proper task for the task queue which will be responsible for the indexation

~_NOTE2_: no tests yet~


config/default/IndexService.conf.php
```diff
@@ -5,5 +5,8 @@
  */
 
 use oat\tao\model\search\index\IndexService;
+use oat\tao\model\search\index\GenerisDocumentBuilderFactory;
 
-return new IndexService();
+return new IndexService(array(
```
andreluizmachado, line 10: can we use [] instead of array()?

models/classes/search/index/GenerisDocumentBuilderFactory.php
```diff
@@ -0,0 +1,150 @@
...
+                if (!in_array($parent->getUri(), $done)) {
+                    $toDo[] = $parent->getUri();
+                }
+            }
+            $done[] = $class->getUri();
+        }
+
+        return $classes;
+    }
+
```
andreluizmachado, line 149: remove empty line

config/default/IndexService.conf.php
```diff
@@ -5,5 +5,8 @@
...
+    'documentBuilderFactory' => new GenerisDocumentBuilderFactory()
+));
```
andreluizmachado, line 12: add new line character

models/classes/search/index/GenerisDocumentBuilderFactory.php
```diff
@@ -0,0 +1,150 @@
...
+}
```
andreluizmachado, line 150: add new line character

models/classes/search/index/IndexDocumentBuilderInterface.php
```diff
@@ -0,0 +1,42 @@
...
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): ?IndexDocument;
+
+    /**
+     * Creates IndexDocument object from the array data
+     * @param $resource $array
+     * @return IndexDocument
+     * @throws \common_Exception
+     */
+    public function createDocumentFromArray(array $resource): ?IndexDocument;
+}
```
andreluizmachado, line 42: add new line character

models/classes/search/index/IndexService.php
```diff
@@ -23,25 +23,18 @@
 
 namespace oat\tao\model\search\index;
 
-use ArrayIterator;
-use core_kernel_classes_Class;
-use core_kernel_classes_Property;
-use core_kernel_classes_Resource;
-use Iterator;
-use oat\generis\model\OntologyAwareTrait;
+use Elasticsearch\Common\Exceptions\Missing404Exception;
```
andreluizmachado, line 26: I Really Liked the refactory 👍

models/classes/resources/ResourceWatcher.php
```diff
@@ -79,6 +84,10 @@ public function catchUpdatedResourceEvent(ResourceUpdated $event)
             $this->updatedAtCache[$resource->getUri()] = $now;
             $resource->editPropertyValues($property, $now);
         }
+
+        $queueDispatcher = $this->getServiceLocator()->get(QueueDispatcherInterface::SERVICE_ID);
+        $queueDispatcher->setOwner('Index');
```
luisfn, line 89: Same thing

models/classes/resources/ResourceWatcher.php
```diff
@@ -59,6 +60,10 @@ public function catchCreatedResourceEvent(ResourceCreated $event)
         $this->updatedAtCache = [];
         $this->updatedAtCache[$resource->getUri()] = $now;
         $resource->editPropertyValues($property, $now);
+
+        $queueDispatcher = $this->getServiceLocator()->get(QueueDispatcherInterface::SERVICE_ID);
+        $queueDispatcher->setOwner('Index');
+        $queueDispatcher->createTask(new UpdateResourceInIndex(), [$resource->getUri()], __('Adding/Updating search index for created resource ', $resource->getUri()));
```
luisfn, line 66: `Index` seems to be too generic, maybe IndexService? Not sure how this name is used later, so if needs big changes, it can stay as is.

Also, these 3 lines are repeated below, would recommend move it to some method or At leat change the messages, one message exclusive for `Create` and other exclusive for `Update`

models/classes/search/tasks/UpdateResourceInIndex.php
```diff
@@ -0,0 +1,80 @@
...
+ * @author Ilya Yarkavets <ilya@taotesting.com>
+ * @package oat\tao\model\search\tasks
+ */
+class UpdateResourceInIndex implements Action, ServiceLocatorAwareInterface, TaskAwareInterface
+{
+    use ServiceLocatorAwareTrait;
+    use OntologyAwareTrait;
+    use TaskAwareTrait;
+
+    public function __invoke($params): Report
```
luisfn, line 46: `__invoke(array $params)`?

models/classes/search/tasks/UpdateResourceInIndex.php
```diff
@@ -0,0 +1,80 @@
...
+        $searchService = $this->getServiceLocator()->get(Search::SERVICE_ID);
+
+        $numberOfIndexed = $searchService->index([$indexDocument]);
+
+        if ($numberOfIndexed === 1) {
+            $type = Report::TYPE_SUCCESS;
+            $message = "Document in index was succesfully updated.";
+        } else {
+            $type = Report::TYPE_WARNING;
+            $message = "There is a difference between number of documents to be indexed/reindexed and number of indexed/reindexed documents.";
```
luisfn, line 75: Maybe simplify this message? Seems a little confusing

models/classes/search/tasks/UpdateResourceInIndex.php
```diff
@@ -0,0 +1,80 @@
...
+        $documentBuilder = $factory->getDocumentBuilderByResourceType($resourceType);
+
+        $indexDocument = $documentBuilder->createDocumentFromResource($createdResource);
+
+        /** @var Search $searchService */
+        $searchService = $this->getServiceLocator()->get(Search::SERVICE_ID);
+
+        $numberOfIndexed = $searchService->index([$indexDocument]);
+
+        if ($numberOfIndexed === 1) {
```
luisfn, line 70: There is a real change of this being different than 1? if there is a problem when indexing maybe it should be triggering an exception?

models/classes/search/index/GenerisDocumentBuilderFactory.php
```diff
@@ -0,0 +1,150 @@
...
+    {
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): ?IndexDocument
+    {
+        $tokenGenerator = new SearchTokenGenerator();
```
luisfn, line 50: Can this be injected somehow?

config/default/IndexService.conf.php
uncleempty: Of course, thanks

models/classes/resources/ResourceWatcher.php
uncleempty: Will do, thanks!

models/classes/search/tasks/UpdateResourceInIndex.php
uncleempty: it would be incompatible with `Action` class. And I don't think that we should do such a change for the whole project now :)

models/classes/search/tasks/UpdateResourceInIndex.php
uncleempty: Not sure that we should, it's just a message identifying that we indexed either less or more documents than we tried to. This could be a case when document has more than one type (possible case, as I seen in code). IMO, we should log/have a record somewhere that numbers differ.
How do you think?

models/classes/search/tasks/UpdateResourceInIndex.php
uncleempty: As I've answered in previous comment: one document can be pushed in more than one indices, which will result in number different from one, which is a valid case.
About the exception... Hard question. Maybe not an exception, but a log record? Because since it's never caught it would "ruin" "high level" code (like controller/etc).
But I will definitely add a case to log the case when indexed is equal to 0

models/classes/search/index/GenerisDocumentBuilderFactory.php
uncleempty: Only by making a service from it. But should we have such service?

luisfn: Is the same issue I had, I think its fine this way for now

models/classes/search/tasks/UpdateResourceInIndex.php
luisfn: My only concern is with the index/reindex part that makes thing a little ambiguous. If we know what action have triggered would be nice to make it more explicit

Example: `The number or indexed documents is different than the expected total` 

If we have the numbers, it could be even more explicit like: `Indexed 8 of 10 documents'

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,137 @@
...
+
+class GenerisIndexDocumentBuilder implements IndexDocumentBuilderInterface
+{
+    /** @var array */
+    private $map;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): ?IndexDocument
```
luisfn, line 36: From the method can't see why use `?`. It seems it will aways return IndexDocument

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,137 @@
...
+            $indexesProperties
+        );
+
+        return $document;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createDocumentFromArray(array $resource = []): ?IndexDocument
```
luisfn, line 63: Same as before

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
+     * @deprecated should be GenerisDocumentBuilderFactory::createDocumentFromResource instead
      */
-    public function createDocumentFromArray($array = [])
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): ?IndexDocument
     {
-        if (!isset($array['id'])) {
-            throw new MissingOptionException('Missed id property for the index document');
-        }
+        $resourceTypes = $resource->getTypes();
+        $resourceType = current(array_keys($resourceTypes)) ?: "";
```
luisfn, line 84: Single quotes?

models/classes/search/index/AbstractDocumentBuilderFactory.php
```diff
@@ -0,0 +1,36 @@
...
+     * Get the IndexDocument builder based on resource type property
+     * @param string $resourceType
+     * @return IndexDocumentBuilderInterface
+     */
+    public function getDocumentBuilderByResourceType(string $resourceType = ""): IndexDocumentBuilderInterface
+    {
+        return new GenerisIndexDocumentBuilder();
+    }
+
+}
```
andreluizmachado: new empty line missing

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
-     * @throws MissingOptionException
+     * @param \core_kernel_classes_Resource $resource
+     * @return IndexDocument|null
      * @throws \common_Exception
+     * @throws \common_exception_InconsistentData
+     *
+     * @deprecated should be GenerisDocumentBuilderFactory::createDocumentFromResource instead
      */
-    public function createDocumentFromArray($array = [])
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): ?IndexDocument
```
luisfn, line 81: If you agree with previous suggestions, probably will affect here

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
-     * @return mixed
+     * @param array $array
+     * @return IndexDocument
+     * @throws \common_exception_MissingParameter
      * @throws \common_Exception
+     *
+     * @deprecated should be GenerisDocumentBuilderFactory::createDocumentFromArray instead
      */
-    public function getIndexProperties(OntologyIndex $index)
+    public function createDocumentFromArray($array = []): ?IndexDocument
```
luisfn, line 99: And here

models/classes/search/index/GenerisIndexDocumentBuilder.php
uncleempty: Indeed, it was left when switching from one of the previous possible implementations, when null was also possible.
Many thanks, will clean this

models/classes/search/index/GenerisDocumentBuilderFactory.php
```diff
@@ -0,0 +1,27 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\search\index;
+
+class GenerisDocumentBuilderFactory extends AbstractDocumentBuilderFactory
```
Skyld, line 24: What is the purpose of the abstract class if it returns `GenerisIndexDocumentBuilder`? `AbstractDocumentBuilderFactory` should declare the abstract method (or it is even better if you replace the abstract class with an interface), and `GenerisDocumentBuilderFactory` should actually return with `GenerisIndexDocumentBuilder`

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,108 @@
...
+
+        return $document;
+    }
+
+    /**
+     * @param OntologyIndex $index
+     * @return mixed
+     * @throws \common_Exception
+     */
+    public function getIndexProperties(OntologyIndex $index): ?IndexProperty
```
Skyld, line 65: - This shouldn't be a public method
- Method name doesn't match return type (multiple vs single IndexProperty)
- It's a bad practice to use inconsistent return points, it leads to unnecessary type checks by the caller. Let's declare an `IndexPropertyNotFound` exception instead

Skyld: I just realised that it's not even possible to return with `null` here

Skyld, line 65: Shouldn't index properties be actual properties of `OntologyIndex` class? It doesn't seem right that they are completely separated

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,108 @@
...
+
+        return $this->map[$index->getIdentifier()];
+    }
+
+    /**
+     * @param $resource
+     * @return array
+     * @throws \common_exception_Error
+     */
+    public function getTypesForResource(\core_kernel_classes_Resource $resource): array
```
Skyld, line 84: This method shouldn't be public I think

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,108 @@
...
+        while (!empty($toDo)) {
+            $class = new \core_kernel_classes_Class(array_pop($toDo));
+            $classes[] = $class->getUri();
+            foreach ($class->getParentClasses() as $parent) {
+                if (!in_array($parent->getUri(), $done)) {
+                    $toDo[] = $parent->getUri();
+                }
+            }
+            $done[] = $class->getUri();
+        }
```
Skyld, line 104: This is how you can end up with infinite loops. Let's find a less hacky solution for getting all the types.

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,108 @@
...
+
+        return $classes;
```
Skyld, line 106: Return value and method name are not aligned (class vs type)

models/classes/search/index/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,108 @@
...
+            );
+            $this->map[$index->getIdentifier()] = $indexProperty;
+        }
+
+        return $this->map[$index->getIdentifier()];
+    }
+
+    /**
+     * @param $resource
+     * @return array
```
Skyld, line 81: string[]

models/classes/search/index/IndexDocumentBuilderInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\search\index;
+
+interface IndexDocumentBuilderInterface
+{
+    const TYPE_PROPERTY = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type';
```
Skyld, line 26: This constant seems not in use and also scope is missing

models/classes/search/index/IndexDocumentBuilderInterface.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Creates IndexDocument object from the \core_kernel_classes_Resource data
+     * @param \core_kernel_classes_Resource $resource
+     * @return IndexDocument
+     * @throws \common_Exception
+     * @throws \common_exception_InconsistentData
+     */
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): IndexDocument;
```
Skyld, line 35: I think we have a design problem here:

By enforcing `core_kernel_classes_Resource` usage on the interface level you make it pointless to have a `AbstractDocumentBuilderFactory`, because technically you cannot have anything else than a `GenerisDocumentBuilderFactory`, right? So either you should NOT depend on generis on the interface level, or remove the abstract factory and replace it with another interface.

Or am I missing something here?

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
         $indexIterator = new IndexIterator($iterator);
```
Skyld, line 54: Please add return type of the method

models/classes/search/index/IndexService.php
```diff
@@ -51,18 +37,11 @@ class IndexService extends ConfigurableService
...
-        WidgetDefinitions::PROPERTY_TEXTBOX,
-        WidgetDefinitions::PROPERTY_TEXTAREA,
-        WidgetDefinitions::PROPERTY_HTMLAREA,
-        WidgetDefinitions::PROPERTY_CHECKBOX,
-        WidgetDefinitions::PROPERTY_COMBOBOX,
-        WidgetDefinitions::PROPERTY_RADIOBOX,
-    ];
+    public const SERVICE_ID = 'tao/IndexService';
+    public const INDEX_MAP_PROPERTY_DEFAULT = 'default';
+    public const INDEX_MAP_PROPERTY_FUZZY = 'fuzzy';
```
Skyld, line 42: Are these constants in use?

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
-        $body['type'] = $this->getTypesForResource($resource);
-
-        return new IndexDocument(
-            $resource->getUri(),
-            $body,
-            $indexesProperties,
-            $this->getDynamicProperties($body['type'], $resource)
-        );
+        $factory = $this->getOption(self::OPTION_DOCUMENT_BUILDER_FACTORY);
+        return $factory;
```
Skyld, line 70: Please remove temporary variable

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
-    public function createDocumentFromArray($array = [])
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource): IndexDocument
     {
-        if (!isset($array['id'])) {
-            throw new MissingOptionException('Missed id property for the index document');
-        }
+        $resourceTypes = $resource->getTypes();
+        $resourceType = current(array_keys($resourceTypes)) ?: '';
+
+        $document = $this->getDocumentBuilderFactory()->getDocumentBuilderByResourceType($resourceType)->createDocumentFromResource($resource);
```
Skyld, line 86: This doesn't look nice. It would be a better solution if you would "hide" the `getDocumentBuilderByResourceType` part by having a single `IndexDocumentFactory` class which would use the magic with the different builders by resource type INTERNALLY, without exposing it in it's interface. It would look something like this: `$indexDocumentFactory->createFromResourceType($resourceType)`

What do you think?

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
-     * @return mixed
+     * @param array $array
+     * @return IndexDocument
+     * @throws \common_exception_MissingParameter
      * @throws \common_Exception
+     *
+     * @deprecated should be GenerisDocumentBuilderFactory::createDocumentFromArray instead
      */
-    public function getIndexProperties(OntologyIndex $index)
+    public function createDocumentFromArray($array = []): IndexDocument
```
Skyld, line 99: This is not very nice. The only thing you need is the "bodyType" from the array... Could you please try to find a less neglectful alternative?

models/classes/search/index/IndexService.php
```diff
@@ -75,113 +54,60 @@ public function runIndexing()
...
-                if (!in_array($parent->getUri(), $done)) {
-                    $toDo[] = $parent->getUri();
-                }
-            }
-            $done[] = $class->getUri();
-        }
+        $document = $this->getDocumentBuilderFactory()->getDocumentBuilderByResourceType($array['body']['type'])->createDocumentFromArray($array);
 
-        return $classes;
+        return $document;
```
Skyld, line 110: Please get rid of the temporary variable

models/classes/search/tasks/UpdateResourceInIndex.php
```diff
@@ -0,0 +1,83 @@
...
+use Zend\ServiceManager\ServiceLocatorAwareTrait;
+use \common_report_Report as Report;
+
+/**
+ * Class UpdateResourceInIndex
+ *
+ * @author Ilya Yarkavets <ilya@taotesting.com>
+ * @package oat\tao\model\search\tasks
+ */
+class UpdateResourceInIndex implements Action, ServiceLocatorAwareInterface, TaskAwareInterface
```
Skyld, line 40: What you update is the document index, and not resource, no?

models/classes/search/tasks/UpdateResourceInIndex.php
```diff
@@ -0,0 +1,83 @@
...
+        if ($numberOfIndexed === 0) {
+            $type = Report::TYPE_ERROR;
+            $message = "Zero documents were added/updated in index.";
+        } elseif ($numberOfIndexed === 1) {
+            $type = Report::TYPE_SUCCESS;
+            $message = "Document in index was succesfully updated.";
+        } else {
+            $type = Report::TYPE_WARNING;
+            $message = "The number or indexed documents is different than the expected total";
+        }
```
Skyld, line 79: - Please use single quotes instead of double quotes
- There is a typo: `successfully`

models/classes/search/index/GenerisDocumentBuilderFactory.php
uncleempty: Yup, interface would suit better

models/classes/search/index/IndexService.php
uncleempty: agree, but not for now. changing this would break BC, that's why method was marked as deprecated

models/classes/search/index/IndexService.php
uncleempty: From first look no, I have not seen any mentions of them in code, but I'm not sure that they're not used anywhere.

models/classes/search/index/IndexDocumentBuilderInterface.php
uncleempty: reason for that is that IMO, it's better have it here than pasted in 4 places across documentBuilders in library. But, I agree, that's not the best place (or good place at all), so maybe it should be removed and we should have constant in 4 builders.
How do you think?

models/classes/search/index/IndexDocumentBuilderInterface.php
uncleempty: I'm forcing resource on interface level since we, really, shouldn't have anything else passed - it's building document from Resource and Resource === core_kernel_classes_Resource

Does it explain why it is left like that?

models/classes/search/index/GenerisIndexDocumentBuilder.php
uncleempty: not that I'm also fond of it, but would it really make sense to change naming inside function, since we are iterating over classes, which are at same point equal to type property of indexDocument?

models/classes/search/index/GenerisIndexDocumentBuilder.php
uncleempty: houldn't be any infinite loops, IMO. And I'm not sure that there is a better solution for the ResourceClasses (which are stored as rdf triples)

models/classes/search/index/DocumentBuilder/DocumentBuilderFactoryInterface.php
```diff
@@ -0,0 +1,32 @@
...
+namespace oat\tao\model\search\index\DocumentBuilder;
+
+interface DocumentBuilderFactoryInterface
+{
+    /**
+     * Get the IndexDocument builder based on resource type property
+     * @param string $resourceType
+     * @return IndexDocumentBuilderInterface
+     */
+    public function getDocumentBuilderByResourceType(string $resourceType = ""): IndexDocumentBuilderInterface;
```
Skyld, line 31: Why allow empty string? It seems hacky

models/classes/search/index/DocumentBuilder/DocumentBuilderFactoryInterface.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace oat\tao\model\search\index\DocumentBuilder;
+
+interface DocumentBuilderFactoryInterface
+{
+    /**
+     * Get the IndexDocument builder based on resource type property
+     * @param string $resourceType
+     * @return IndexDocumentBuilderInterface
+     */
```
Skyld, line 30: Docblock can be removed

models/classes/search/index/DocumentBuilder/GenerisDocumentBuilderFactory.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace oat\tao\model\search\index\DocumentBuilder;
+
+class GenerisDocumentBuilderFactory implements DocumentBuilderFactoryInterface
+{
+    /**
+     * Get the IndexDocument builder based on resource type property
+     * @param string $resourceType
+     * @return IndexDocumentBuilderInterface
+     */
```
Skyld, line 30: Docblock can be removed

models/classes/search/index/DocumentBuilder/GenerisDocumentBuilderFactory.php
```diff
@@ -0,0 +1,35 @@
...
+    public function getDocumentBuilderByResourceType(string $resourceType = ""): IndexDocumentBuilderInterface
```
Skyld, line 31: Why allow empty string?

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,111 @@
...
+        foreach ($tokenGenerator->generateTokens($resource) as $data) {
+            /** @var OntologyIndex $index */
+            list($index, $strings) = $data;
+            $body[$index->getIdentifier()] = $strings;
+            $indexesProperties[$index->getIdentifier()] = $this->getIndexProperties($index);
+        }
+
+        $body['type'] = $rootResourceType ?: $this->getTypesForResource($resource);
+
+        $document = new IndexDocument(
```
Skyld, line 54: You can return here

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,111 @@
...
+            $body,
+            $indexesProperties
+        );
+
+        return $document;
+    }
+
+    /**
+     * @param OntologyIndex $index
+     * @return IndexProperty
```
Skyld, line 65: Can be removed

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,111 @@
...
+                $index->isDefaultSearchable()
+            );
+            $this->map[$index->getIdentifier()] = $indexProperty;
+        }
+
+        return $this->map[$index->getIdentifier()];
+    }
+
+    /**
+     * @param $resource
```
Skyld, line 83: You can remove

models/classes/search/index/DocumentBuilder/DocumentBuilderFactoryInterface.php
uncleempty: Nope, I've changed the docblocks in implementing classes to have `{@inheritdoc}`

models/classes/search/index/DocumentBuilder/GenerisDocumentBuilderFactory.php
uncleempty: Because we could have resources with empty type, which in order would go into `unclassified` docs index

models/classes/search/index/DocumentBuilder/GenerisDocumentBuilderFactory.php
uncleempty: Changed to be `{@inheritdoc}`

models/classes/search/index/DocumentBuilder/DocumentBuilderFactoryInterface.php
uncleempty: Answered below.

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,106 @@
...
+use oat\tao\model\search\index\IndexDocument;
+use oat\tao\model\search\index\IndexProperty;
+use oat\tao\model\search\index\OntologyIndex;
+use oat\tao\model\search\SearchTokenGenerator;
+use oat\tao\model\TaoOntology;
+
+class GenerisIndexDocumentBuilder extends AbstractIndexDocumentBuilder
+{
+    /** @var array */
+    private $map;
```
Skyld, line 34: It should have a default value i think

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createDocumentFromResource(\core_kernel_classes_Resource $resource, string $rootResourceType = ""): IndexDocument
+    {
+        $tokenGenerator = new SearchTokenGenerator();
```
Skyld, line 41: Dependency injection!

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
uncleempty: An empty array, doesn't really change anything, but I'll add it

PR #2536, opened 2020-06-08
Fix/cached func acl

jbout:
When setting the funcAcl implementation using the service manager the cached version is still used during install, and all access grants will be applied to the previous implementation, meaning no new access roles are written during the same request/command


gitromba:
With a new local TAO CE install I've found no regression, it seems to be fine.

PR #2537, opened 2020-06-09
[BSA-148] Feature: limit nesting level of sub classes

SergiiTao:
In case of large number of classes /taoQtiTest/Items/getItemClasses works very slow.
This PR adds an option to limit the nesting level of a class tree

Breaking changes: 
1. Added return types to methods of `\oat\tao\model\resources\ResourceService` class


models/classes/resources/ResourceService.php
```diff
@@ -50,26 +57,33 @@ class ResourceService extends ConfigurableService
...
     /**
      * Get the class subclasses
+     *
+     * @param core_kernel_classes_Class[] $subClasses
+     * @param int                         $nestingLevel
+     *
      * @return array the classes hierarchy
      */
-    private function getSubClasses($subClasses)
+    private function getSubClasses(array $subClasses, int $nestingLevel = 0): array
```
Alroniks, line 81: Setting nesting level to 0 makes this method backward incompatible. If somebody wants to get all tree he will fail. I guess, need to add the ability to get all children if the method called with only one parameter. The default value may be -1 in this case or zero may mean that it has no limits.

models/classes/resources/ResourceService.php
```diff
@@ -78,26 +92,43 @@ private function getSubClasses($subClasses)
...
      */
-    public function getResources(\core_kernel_classes_Class $rootClass, $format = 'list', $selectedUris = [], $search = '', $offset = 0, $limit = 30)
-    {
+    public function getResources(
+        core_kernel_classes_Class $rootClass,
+        $format = 'list',
+        $selectedUris = [],
+        $search = '',
+        $offset = 0,
+        $limit = 30
```
Alroniks, line 130: Do we have any option for set size of the page? It would be nice to have, but not required.

models/classes/resources/ResourceService.php
SergiiTao: this is a private method :)
`$nestingLevel` doesn't limit the level itself it needs to pass the current level to recursive call to compare it with a limit

models/classes/resources/ResourceService.php
SergiiTao: of what page?

SergiiTao:
> Changes break BC but minor version set. I would leave the old behavior if the second parameter not defined.

do you have issues locally because of these changes?

models/classes/resources/ResourceService.php
Alroniks: Any page. It sets a limit of items for pagination (as described in docblock), so question about configuring the page size for pagination across TAO.

models/classes/resources/ResourceService.php
```diff
@@ -50,26 +57,33 @@ class ResourceService extends ConfigurableService
 
     /**
      * Get the list of  classes from the given root class
+     *
      * @param core_kernel_classes_Class $rootClass the root class
      * @return array the classes hierarchy
      */
-    public function getAllClasses(core_kernel_classes_Class $rootClass)
+    public function getAllClasses(core_kernel_classes_Class $rootClass): array
```
wazelin, line 64: Adding a return type on a public method is technically a breaking change, which makes the class incompatible in case someone had extended it prior to this point.

We should probably reconsider the way we're adding argument and method types, perhaps, doing it in a single batch modification with a major version bump is a right way to go.

models/classes/resources/ResourceService.php
SergiiTao: we do not have pagination on top select class tree

models/classes/resources/ResourceService.php
SergiiTao: yep. make sense. bumping major version

manifest.php
```diff
@@ -57,7 +57,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '42.12.1',
+    'version' => '43.0.0',
```
wazelin, line 60: How do we ususally log the breaking changes, so that other developers/extension users are aware of what they are?

SergiiTao: We don't have CHNGELOGs or something

wazelin: Could you please describe the breaking changes at least in the PR's description?
I believe, those are linked to the release notes in the end.

PR #2539, opened 2020-06-10
TAO-10253 Disabling index modification on all sections when ElasticSearch is enabled

luisfn:
- Removed index setup icon
- Icon is not rendered only when ElasticSearch library is installed and in use
- Changes on backend only

### DIFF
--- a/actions/class.PropertiesAuthoring.php
+++ b/actions/class.PropertiesAuthoring.php
@@ -18,6 +18,8 @@
  * Copyright (c) 2015-2018 Open Assessment Technologies S.A.
  */
 
+declare(strict_types=1);
+
 use oat\oatbox\event\EventManager;
 use oat\tao\model\event\ClassFormUpdatedEvent;
 use oat\generis\model\GenerisRdf;
@@ -472,8 +474,8 @@ class tao_actions_PropertiesAuthoring extends tao_actions_CommonModule
 
     private function isElasticSearchEnabled(): bool
     {
-            $searchService = $this->getServiceLocator()->get(\oat\tao\model\search\Search::SERVICE_ID);
+        $searchService = $this->getServiceLocator()->get(\oat\tao\model\search\Search::SERVICE_ID);
 
-            return get_class($searchService) === \oat\tao\elasticsearch\ElasticSearch::class;
+        return get_class($searchService) === \oat\tao\elasticsearch\ElasticSearch::class;
     }
 }
--- a/actions/class.PropertiesAuthoring.php
+++ b/actions/class.PropertiesAuthoring.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2015-2018 Open Assessment Technologies S.A.
+ * Copyright (c) 2015-2020 Open Assessment Technologies S.A.
  */
 
 declare(strict_types=1);
@@ -30,6 +30,7 @@ use oat\tao\model\search\index\OntologyIndexService;
 use oat\tao\helpers\form\ValidationRuleRegistry;
 use oat\generis\model\OntologyAwareTrait;
 use oat\oatbox\log\LoggerAwareTrait;
+use oat\tao\model\search\Search;
 
 /**
  * Regrouping all actions related to authoring
@@ -303,7 +304,6 @@ class tao_actions_PropertiesAuthoring extends tao_actions_CommonModule
         $formContainer = new tao_actions_form_Clazz($class, $classData, $propertyData, $this->isElasticSearchEnabled());
         $myForm = $formContainer->getForm();
 
-
         if ($myForm->isSubmited()) {
             if ($myForm->isValid()) {
                 //get the data from parameters
@@ -472,7 +472,7 @@ class tao_actions_PropertiesAuthoring extends tao_actions_CommonModule
 
     private function isElasticSearchEnabled(): bool
     {
-        $searchService = $this->getServiceLocator()->get(\oat\tao\model\search\Search::SERVICE_ID);
+        $searchService = $this->getServiceLocator()->get(Search::SERVICE_ID);
 
         return get_class($searchService) === \oat\tao\elasticsearch\ElasticSearch::class;
     }
--- a/actions/class.PropertiesAuthoring.php
+++ b/actions/class.PropertiesAuthoring.php
@@ -54,7 +54,7 @@ class tao_actions_PropertiesAuthoring extends tao_actions_CommonModule
         $this->defaultData();
         $clazz = $this->getClass($this->getRequestParameter('id'));
 
-        $myForm = $this->getClassForm($clazz);
+        $myForm = $this->getClassForm($clazz, true);
         if ($myForm->isSubmited()) {
             if ($myForm->isValid()) {
                 if ($clazz instanceof core_kernel_classes_Resource) {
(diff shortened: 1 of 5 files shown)


actions/class.PropertiesAuthoring.php
```diff
@@ -461,4 +471,11 @@ protected function extractPropertyData($data)
         }
         return $propertyData;
     }
+
+    private function isElasticSearchEnabled(): bool
+    {
+        $searchService = $this->getServiceLocator()->get(\oat\tao\model\search\Search::SERVICE_ID);
+
+        return get_class($searchService) === \oat\tao\elasticsearch\ElasticSearch::class;
```
Skyld, line 479: Wouldn't you get an error if you refer to a class constant when elasticsearch extension is not installed?

actions/class.PropertiesAuthoring.php
```diff
@@ -293,9 +300,12 @@ protected function getCurrentInstance()
     public function getClassForm(core_kernel_classes_Class $clazz)
     {
         $data = $this->getRequestParameters();
-        $formContainer = new tao_actions_form_Clazz($clazz, $this->extractClassData($data), $this->extractPropertyData($data));
+        $classData = $this->extractClassData($data);
+        $propertyData = $this->extractPropertyData($data);
+        $formContainer = new tao_actions_form_Clazz($clazz, $classData, $propertyData, $this->isElasticSearchEnabled());
```
Skyld, line 305: What is clazz ?

actions/class.PropertiesAuthoring.php
```diff
@@ -293,9 +300,12 @@ protected function getCurrentInstance()
     public function getClassForm(core_kernel_classes_Class $clazz)
     {
         $data = $this->getRequestParameters();
-        $formContainer = new tao_actions_form_Clazz($clazz, $this->extractClassData($data), $this->extractPropertyData($data));
+        $classData = $this->extractClassData($data);
+        $propertyData = $this->extractPropertyData($data);
```
Skyld, line 304: Not sure what data is but why not wrap it in a value object? Does it make sense?

actions/form/class.Clazz.php
```diff
@@ -44,15 +44,22 @@ class tao_actions_form_Clazz extends tao_helpers_form_FormContainer
...
+    private $disableIndexChanges;
+
     /**
      * @param core_kernel_classes_Class $clazz
      * @param array $classData
      * @param array $propertyData
+     * @param bool $disableIndexChanges
      */
-    public function __construct(core_kernel_classes_Class $clazz, $classData, $propertyData)
+    public function __construct(core_kernel_classes_Class $clazz, $classData, $propertyData, $disableIndexChanges = false)
```
Skyld, line 58: Flag arguments are not advised to use, it add unnecessary complexity (if /else statements) and also violates the single responsibility principle.

actions/form/class.Clazz.php
```diff
@@ -87,7 +94,12 @@ protected function getTopClazz()
      */
     protected function getPropertyForm($property, $index, $isParentProp, $propData)
```
Skyld, line 95: Where is the static type hinting?

actions/form/class.Clazz.php
Skyld, line 58: Where is the static type hinting?

actions/class.PropertiesAuthoring.php
luisfn: No, I tried it, it work fine

actions/class.PropertiesAuthoring.php
luisfn: I would say it was used before do differentiate the name from class (I saw it in other languages).  If you see, this is not mine, I've just tried to split the line that were too long, but I can rename it I think.

uncleempty: You're correct Luis.

actions/class.PropertiesAuthoring.php
```diff
@@ -293,9 +300,12 @@ protected function getCurrentInstance()
...
+        $formContainer = new tao_actions_form_Clazz($clazz, $classData, $propertyData, $this->isElasticSearchEnabled());
         $myForm = $formContainer->getForm();
 
+
```
uncleempty, line 308: Please remove it :)

actions/class.PropertiesAuthoring.php
uncleempty: Just a strange thought:
maybe we can go in different direction? Just a question - this check should return `true` in case ElasticSearch is enabled? _OR_ in case standard search (the one in tao core) is disabled? Because if yes - what we can do is to just check if `get_class($searchService) === oat\tao\model\search\strategy\GenerisSearch::class` or smth similar.
How does that sound?

actions/class.PropertiesAuthoring.php
luisfn: Started to refactor it, but not a real again unless we go with a full refactor. On the call chain,  the operations are based on these arrays.

actions/form/class.SimpleProperty.php
```diff
@@ -184,7 +183,7 @@ protected function initElements()
         $elementNames[] = $propUriElt;
 
         if (count($elementNames) > 0) {
-            $groupTitle = $this->getGroupTitle($property);
+            $groupTitle = $this->getGroupTitle($property, true);
```
uncleempty, line 186: what does this flag mean? I don't see it's usage in `AbstractProperty` class

actions/class.PropertiesAuthoring.php
```diff
@@ -461,4 +471,11 @@ protected function extractPropertyData($data)
         }
         return $propertyData;
     }
+
+    private function isElasticSearchEnabled(): bool
+    {
+        $searchService = $this->getServiceLocator()->get(\oat\tao\model\search\Search::SERVICE_ID);
```
uncleempty, line 477: why do we need FQCN?

actions/form/class.Clazz.php
uncleempty: Yep, but since it's added to not break BC and passed to a class properties later, IMO - ok for now

actions/class.PropertiesAuthoring.php
luisfn: Some renaming on it

actions/form/class.SimpleProperty.php
luisfn: Should added this on beginning and forgot to remove. Thanks

actions/form/class.Clazz.php
luisfn: Answering @Skyld, Unfortunately I don't see how to achieve it without bigger refactor.

actions/form/class.Clazz.php
luisfn: Fixed

actions/form/class.Clazz.php
luisfn: Fixed

actions/class.PropertiesAuthoring.php
luisfn: Fixed

actions/class.PropertiesAuthoring.php
luisfn: Fixed

actions/class.PropertiesAuthoring.php
```diff
@@ -18,6 +18,8 @@
  * Copyright (c) 2015-2018 Open Assessment Technologies S.A.
```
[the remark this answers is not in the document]
luisfn: Fixed

actions/class.PropertiesAuthoring.php
```diff
@@ -451,7 +459,7 @@ protected function extractClassData($data)
      * @param array $data
      * @return array
      */
-    protected function extractPropertyData($data)
+    protected function extractPropertyData($data): array
```
andreluizmachado, line 462: type hint on $data parameter is missing.

actions/class.RdfController.php
```diff
@@ -229,6 +229,7 @@ protected function editClass(core_kernel_classes_Class $class, core_kernel_class
     protected function getClassForm($class, $resource, $topclass = null)
     {
         $controller = new tao_actions_PropertiesAuthoring();
+        $controller->setServiceLocator($this->getServiceLocator());
```
andreluizmachado, line 232: I did the same thing on my PR :)

actions/class.PropertiesAuthoring.php
luisfn: Fixed

PR #2540, opened 2020-06-10
Fix NEX-1191 don't show error if HEAD request failed because of CORS

bziondik:
Related to: https://oat-sa.atlassian.net/browse/NEX-1191

When user try to add an image by URL function getResourceType called
`getResourceType: function getResourceType(url, callback) {`
https://github.com/oat-sa/tao-core-sdk-fe/blob/d222ef0a28298b76eb1e5718a63a466a2a21356b/src/core/mimetype.js#L41-L61

Sometimes ajax request failed because of CORS policy in this case `request.status === 0`
In `views/js/controller/backoffice.js` we listen to all ajax errors.
No need to show a popup with error in this case because it will be handled in proper Widgets

Steps to reproduce:

1. Go to 'Items' tab.
2. Create a new item and go to Authoring.
3. Put inline bock and insert an image.
4. Put the image's URL into 'File' field in Image Properties.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -57,7 +57,7 @@ return [
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '42.12.2',
+    'version' => '42.12.3',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => [
         'generis' => '>=12.21.0',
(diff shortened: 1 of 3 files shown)


bziondik:
> I'm not able to review it carefully nor to take some time to test it. The general approach looks good but please test the follow cases with HEAD requests ;
> 
> * offline (unplug your network cable, switch off your wifi,  shutdown your local web server)
> * time out (request takes more than 30s for the server to respond)
>   Just to be sure there's no ignored valid error
@krampstudio 
In all these cases + CORS issue, we have the same data in `request` 
`status: 0`
`responseText: ""`
`statusText: "error"`
`readyState: 0`
I can't identify is it CORS error or time out or offline. We use `'Unknown Error'` for all these cases. Or we hide it for all these cases or not.

PR #2543, opened 2020-06-11
Created service to wrap filesystem methods for testing puroses.

SashaTAO:
JIRA ticket: https://oat-sa.atlassian.net/browse/TAO-10137

As part of feature requested in the ticket we need to export QTI test for created delivery (to have a snapshot of test for remote publishing). QTI test export functionality relies on temporary files and local filesystem. We cannot use FileSystemService without much bigger refactoring which may impact big part of TAO functionality.
To be able to cover with tests functionality where native php functions or static methods in helpers are used we need a clacc which we will be able to mock in tests.

At the moment only methods which are used to implement required functionality (export QTI test https://github.com/oat-sa/extension-tao-testqti/blob/develop/helpers/QtiPackageExporter.php) were implemented in helper class.


helpers/FileHelperService.php
```diff
@@ -0,0 +1,82 @@
...
+use tao_helpers_File;
+use oat\oatbox\service\ConfigurableService;
+
+/**
+ * Wrapper on top of tao_helpers_File and some filesystem related functionality.
+ *
+ * Class FileHelperService
+ * @package oat\tao\helpers
+ */
+class FileHelperService extends ConfigurableService
```
Skyld, line 33: I won't prevent the merge but it would be nice so use some abstraction here. This could totally be a custom TaoFilesystemAdapter in League Flysystem package and you could depend on their interface.

PR #2547, opened 2020-06-12
Feature/tao 10283 implement custom properties event property changed

andreluizmachado:
- Introduced new Event PropertiesChangedEvent
- Introduced IndexUpdaterInterface which is responsible to update the properties on the indexes
- Fix getDynamicProperties method on IndexService, it needs to return the properties values according to the type of the property value, not the property type.
- Satisfied the dependency of tao_actions_PropertiesAuthoring on getClassForm method
-  Introduced new Event PropertiesChangedEventListener


actions/class.PropertiesAuthoring.php
```diff
@@ -313,21 +316,49 @@ public function getClassForm(core_kernel_classes_Class $clazz)
 
                 //save all properties values
                 if (isset($data['properties'])) {
+                    $changedProperties = [];
                     foreach ($data['properties'] as $i => $propertyValues) {
```
Skyld, line 320: Please extract loop body into a separate method

actions/class.PropertiesAuthoring.php
```diff
@@ -313,21 +316,49 @@ public function getClassForm(core_kernel_classes_Class $clazz)
...
-                        $this->saveSimpleProperty($propertyValues);
+
+                        $property = $this->getProperty(tao_helpers_Uri::decode($propertyValues['uri']));
+                        $oldPropertyLabel = $property->getLabel();
+                        $oldPropertyType = $property->getOnePropertyValue(
+                            new \core_kernel_classes_Property(WidgetRdf::PROPERTY_WIDGET)
+                        );
+                        $oldProperty = new OldProperty($oldPropertyLabel, $oldPropertyType);
+
+                        $this->saveSimpleProperty($propertyValues, $property);
```
Skyld, line 335: It doesn't look performant to save all properties 1 by 1 (also to read them from DB). Can't we batch the DB calls?

actions/class.PropertiesAuthoring.php
```diff
@@ -338,11 +369,12 @@ public function getClassForm(core_kernel_classes_Class $clazz)
      * Default property handling
      *
      * @param array $propertyValues
+     * @param core_kernel_classes_Resource $property
+     * @throws Exception
      */
-    protected function saveSimpleProperty($propertyValues)
+    protected function saveSimpleProperty($propertyValues, $property)
```
Skyld, line 375: MIssing typehints

models/classes/dto/OldProperty.php
```diff
@@ -0,0 +1,51 @@
...
+
+class OldProperty
+{
+    /** @var string */
+    private $label;
+
+    /** @var core_kernel_classes_Resource|null */
+    private $propertyType;
+
+    public function __construct(string $label, ?core_kernel_classes_Resource $propertyType)
```
Skyld, line 36: Why nullable?

models/classes/search/index/IndexUpdaterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ *
```
luisfn, line 18: Some of your licenses seems to have an extra blank line

models/classes/search/index/IndexService.php
```diff
@@ -251,22 +253,19 @@ private function getDynamicProperties(array $classes, core_kernel_classes_Resour
...
                 }
 
-                $customProperties[$fieldName] = (string)$propertyValue;
-
-                if ($propertyTypeUri === WidgetDefinitions::PROPERTY_COMBOBOX
-                    || $propertyTypeUri === WidgetDefinitions::PROPERTY_RADIOBOX) {
-                    $customProperties[$fieldName] = (string)$propertyValue->getLabel();
+                if ($propertyValue instanceof core_kernel_classes_Literal) {
+                    $customProperties[$fieldName][] = (string)$propertyValue;
+                    $customProperties[$fieldName] = array_unique($customProperties[$fieldName]);
```
luisfn, line 258: Are these operations correct? The second line seems to overwrite the first

models/classes/listener/PropertiesChangedListener.php
```diff
@@ -0,0 +1,67 @@
...
+use oat\tao\model\event\PropertiesChangedEvent;
+use oat\tao\model\search\tasks\RenameIndexProperties;
+use oat\tao\model\taskQueue\QueueDispatcherInterface;
+use RuntimeException;
+
+/**
+ * Class ResourceWatcher
+ * @package oat\tao\model\resources
+ * @author Aleksej Tikhanovich, <aleksej@taotesting.com>
+ */
```
luisfn, line 36: Is this comment needed? Also, your name is Andre :)

models/classes/listener/PropertiesChangedListener.php
```diff
@@ -0,0 +1,67 @@
...
+class PropertiesChangedListener extends ConfigurableService
+{
+    const SERVICE_ID = 'tao/PropertiesChangedListener';
+
+    public function catchPropertiesChangedEvent(PropertiesChangedEvent $event): void
+    {
+        $taskMessage = __('Updating Indexes');
```
luisfn, line 43: Would prefer `updating documents`

models/classes/listener/PropertiesChangedListener.php
```diff
@@ -0,0 +1,67 @@
...
+use oat\tao\model\search\tasks\RenameIndexProperties;
+use oat\tao\model\taskQueue\QueueDispatcherInterface;
+use RuntimeException;
+
+/**
+ * Class ResourceWatcher
+ * @package oat\tao\model\resources
+ * @author Aleksej Tikhanovich, <aleksej@taotesting.com>
+ */
+class PropertiesChangedListener extends ConfigurableService
```
luisfn, line 37: Since properties is a little generic, would call it `ClassPropertiesChangedListener`. Same for the event of course

actions/class.PropertiesAuthoring.php
```diff
@@ -313,21 +316,49 @@ public function getClassForm(core_kernel_classes_Class $clazz)
...
                         $indexes = null;
                         if (isset($propertyValues['indexes'])) {
                             $indexes = $propertyValues['indexes'];
                             unset($propertyValues['indexes']);
                         }
-                        $this->saveSimpleProperty($propertyValues);
+
+                        $property = $this->getProperty(tao_helpers_Uri::decode($propertyValues['uri']));
+                        $oldPropertyLabel = $property->getLabel();
+                        $oldPropertyType = $property->getOnePropertyValue(
```
uncleempty, line 330: Are you sure that all the properties we are using (or which can appear in `$data['properties']` array will have only single value? Because in other case we will be indexing only one possible from all the values.

actions/class.PropertiesAuthoring.php
uncleempty: > Can't we batch the DB calls?
For the moment - it's not possible. Even if we "add here" batch saving, under the hood it will still go one by one.

models/classes/listener/PropertiesChangedListener.php
uncleempty: Even not class, but `ResourcePropertiesChangedListener`

models/classes/search/index/IndexService.php
```diff
@@ -251,22 +253,19 @@ private function getDynamicProperties(array $classes, core_kernel_classes_Resour
...
-                        function (string $propertyValue): string {
-                            return (new core_kernel_classes_Property($propertyValue))->getLabel();
-                        },
-                        $customPropertiesValues
-                    );
-                }
+                $customPropertiesValues = $resource->getPropertyValues($property);
+                $customProperties[$fieldName] = array_map(
+                    function (string $propertyValue): string {
+                        return (new core_kernel_classes_Property($propertyValue))->getLabel();
```
uncleempty, line 265: related not only to this line in this method, but mentioned it first only there:
could you please change all the calls to `new core_kernel_classes_Property(smth)` to `$this->getProperty(smth)`, since `IndexService` is using `OntologyAwareTrait`

scripts/install/RegisterPropertiesChangedEvent.php
```diff
@@ -0,0 +1,42 @@
...
+declare(strict_types=1);
+
+namespace oat\tao\scripts\install;
+
+use oat\oatbox\event\EventManager;
+use oat\oatbox\extension\InstallAction;
+use oat\tao\model\event\PropertiesChangedEvent;
+use oat\tao\model\listener\PropertiesChangedListener;
+
+class RegisterPropertiesChangedEvent extends InstallAction
```
uncleempty, line 31: I know, that all of the changes are gathering into feature branch at first, but...
In order to not forget in future: please add all the install/update scripts into according sections of the manifest file/updater script.
p.s.
For the updater script it may make sense to have them as smth like
```
//if ($this->isVersion(smth)) {
    //do something

   //$this->skip(smth, newSmth);
//}
```
What I mean, since we already need some changes -> we can add them into according scripts, and if for example we don't really want to stick to any versions for now -> just comment them out.

scripts/install/RegisterPropertiesChangedEventListener.php
```diff
@@ -0,0 +1,45 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\scripts\install;
+
+use oat\oatbox\extension\InstallAction;
+use oat\oatbox\service\exception\InvalidServiceManagerException;
+use oat\tao\model\listener\PropertiesChangedListener;
+
+class RegisterPropertiesChangedEventListener extends InstallAction
```
uncleempty, line 30: Same as above

actions/class.PropertiesAuthoring.php
andreluizmachado: I don't think so. @uncleempty Do you know it? @jbout ?

config/default/IndexUpdaterInterface.conf.php
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ *
+ *
+ */
+
+declare(strict_types=1);
+
+use oat\tao\model\search\strategy\GenerisIndexUpdater;
+
+return new GenerisIndexUpdater();
```
uncleempty, line 26: Not sure in file naming, IMO config shouldn't have `...Interface` in name

models/classes/search/index/IndexService.php
```diff
@@ -25,6 +25,7 @@
 
 use ArrayIterator;
 use core_kernel_classes_Class;
+use core_kernel_classes_Literal;
```
uncleempty, line 28: Just a proposal:
maybe use it like `use core_kernel_classes_Literal as Literal;` to "improve" readability inside code. How does that sound?

models/classes/search/index/IndexService.php
```diff
@@ -251,22 +253,19 @@ private function getDynamicProperties(array $classes, core_kernel_classes_Resour
                     continue;
                 }
 
-                $customProperties[$fieldName] = (string)$propertyValue;
-
-                if ($propertyTypeUri === WidgetDefinitions::PROPERTY_COMBOBOX
-                    || $propertyTypeUri === WidgetDefinitions::PROPERTY_RADIOBOX) {
-                    $customProperties[$fieldName] = (string)$propertyValue->getLabel();
+                if ($propertyValue instanceof core_kernel_classes_Literal) {
```
uncleempty, line 256: If previous proposal would be used:
maybe use it like `if ($propertyValue instanceof Literal) {...` to "improve" readability inside code. How does that sound?

actions/class.PropertiesAuthoring.php
uncleempty: My answer is above, actually :) 
I've done a small investigation before answering. But, Joel can now more, let's wait for his answer also.

models/classes/dto/OldProperty.php
andreluizmachado: This can happen when you change a created property that doesn't have a type yet. The application allows you to create properties and change them without to specify the type.

actions/class.PropertiesAuthoring.php
andreluizmachado: Thank you. Refactored.

actions/class.PropertiesAuthoring.php
andreluizmachado: Thank you. Added.

models/classes/search/index/IndexUpdaterInterface.php
andreluizmachado: Thank you. Fixed.

models/classes/search/index/IndexService.php
andreluizmachado: Yes, because properties that belong to parent classes also can also appear on the child classes causing duplicate data. This array_unique is to prevent this situation.

models/classes/listener/PropertiesChangedListener.php
andreluizmachado: Thank you. Removed.

models/classes/listener/PropertiesChangedListener.php
andreluizmachado: Let's put 'Updating search index', as @uncleempty did on his PR?

models/classes/listener/PropertiesChangedListener.php
andreluizmachado: I think it makes more sense Luis Suggestion, given that properties definitions are in the class level, not in the resource level.

actions/class.PropertiesAuthoring.php
```diff
@@ -470,6 +460,58 @@ protected function extractPropertyData(array $data): array
...
+            $indexes = null;
+            if (isset($propertyValues['indexes'])) {
+                $indexes = $propertyValues['indexes'];
+                unset($propertyValues['indexes']);
+            }
+
+            $property = $this->getProperty(tao_helpers_Uri::decode($propertyValues['uri']));
+            $oldPropertyLabel = $property->getLabel();
+            $oldPropertyType = $property->getOnePropertyValue(
+                new \core_kernel_classes_Property(WidgetRdf::PROPERTY_WIDGET)
```
uncleempty, line 482: please, change to the `$this->getProperty` call here also

models/classes/event/PropertiesChangedEvent.php
```diff
@@ -0,0 +1,45 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\model\event;
+
+use oat\oatbox\event\Event;
+
+class PropertiesChangedEvent implements Event
+{
+    /** @var array */
```
uncleempty, line 28: just a proposal:
since `$this->properties` is not a flat array with the  list of properties, could you maybe describe the structure in doc comment?

models/classes/search/index/IndexService.php
```diff
@@ -222,10 +221,11 @@ private function getDynamicProperties(array $classes, core_kernel_classes_Resour
             foreach ($properties as $property) {
```
uncleempty, line 221: just a small note regarding line 218:
maybe it makes sense to change it to `$this->getClass` call also? How does that sound?

models/classes/search/index/IndexUpdater.php
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\search\index;
+
+use Throwable;
+
+interface IndexUpdater
```
uncleempty, line 26: since it's an interface: please consider renaming classname/file to have `...Interface` suffix according to PSR naming convetion

scripts/install/RegisterPropertiesChangedEvent.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function __invoke($params)
```
uncleempty, line 31: please add return type hint

scripts/install/RegisterPropertiesChangedEventListener.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * @param $params
+     *
+     * @return \common_report_Report
+     * @throws \common_Exception
+     * @throws InvalidServiceManagerException
+     */
+    public function __invoke($params)
```
uncleempty, line 37: please add return type hint

scripts/update/Updater.php
```diff
@@ -1344,5 +1349,19 @@ public function update($initialVersion)
...
+//            $registerPropertiesChangedEvent->__invoke([]);
+//
+//            $registerPropertiesChangedEventListener = new RegisterPropertiesChangedEventListener();
+//            $registerPropertiesChangedEventListener->setServiceLocator($this->getServiceManager());
+//            $registerPropertiesChangedEventListener->__invoke([]);
+//
+//            $this->setVersion('42.12.0');
+//        }
+//
+//        $this->skip('42.11.0', '42.12.0');
```
uncleempty, line 1365: :+1: 
with a small addition:
if the full commented `if` block is related to our feature -> please remove `$this->skip` call after the `if`

actions/class.PropertiesAuthoring.php
uncleempty: @andreluizmachado , I've seen your changes, and one question regarding them:
you are 100% sure that
```
$property->getProperty(
                WidgetRdf::PROPERTY_WIDGET
            );
```
is fully equal to
```
$property->getOnePropertyValue(
                new \core_kernel_classes_Property(WidgetRdf::PROPERTY_WIDGET)
);
```
?
p.s.
What I meant here when requesting changes was actually smth like:
```
$property->getOnePropertyValue(
                $this->getProperty(WidgetRdf::PROPERTY_WIDGET)
);
```

models/classes/search/tasks/RenameIndexProperties.php
```diff
@@ -0,0 +1,126 @@
...
+        try {
+            $this->getServiceLocator()->get(IndexUpdaterInterface::SERVICE_ID)->updateProperties($indexProperties);
+        } catch (Throwable $exception) {
+            $message = 'Failed during update search index';
+            $this->getLogger()->error($message, (array)$exception);
+
+            return common_report_Report::createFailure(__($message));
+        }
+
+        $message = 'Search index was successful updated.';
```
uncleempty, line 103: I will not block merging of this anymore ;) just a small note:
> Search index was successful updated.

should be 
> Search index was successfully updated.

scripts/install/RegisterClassPropertiesChangedEvent.php
```diff
@@ -0,0 +1,41 @@
...
+class RegisterClassPropertiesChangedEvent extends InstallAction
+{
+    public function __invoke($params): common_report_Report
+    {
+        /** @var EventManager $eventManager */
+        $eventManager = $this->getServiceManager()->get(EventManager::SERVICE_ID);
+        $eventManager->attach(ClassPropertiesChangedEvent::class, [ClassPropertiesChangedListener::SERVICE_ID, 'catchPropertiesChangedEvent']);
+        $this->getServiceManager()->register(EventManager::SERVICE_ID, $eventManager);
+
+        return new common_report_Report(common_report_Report::TYPE_SUCCESS, 'ClassPropertiesChangedEvent events are registered');
```
uncleempty, line 39: just a note, maybe it'd be better to have:
> ClassPropertiesChangedEvent events are registered

as
> ClassPropertiesChangedEvent listener registered

How does that sound?

models/classes/event/ClassPropertiesChangedEvent.php
```diff
@@ -0,0 +1,57 @@
...
+namespace oat\tao\model\event;
+
+use oat\oatbox\event\Event;
+
+class ClassPropertiesChangedEvent implements Event
+{
+    /**
+     * @var array[{property: \core_kernel_classes_Property, oldProperty: \oat\tao\model\dto\OldProperty}]
+     *
+     * list of properties that it was changed following the structure:
```
uncleempty, line 31: > list of properties that it was changed following the structure

better to have it like
> list of properties that were changed, with the following structure

How does that sound?

models/classes/listener/ClassPropertiesChangedListener.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $taskMessage = __('Updating search index');
+
+        $queueDispatcher = $this->getServiceLocator()->get(QueueDispatcherInterface::SERVICE_ID);
+        $queueDispatcher->createTask(
+            new RenameIndexProperties(),
+            array_map(
+                function (array $property) {
+                    if (!isset($property['oldProperty'], $property['property'])) {
+                        throw new RuntimeException('property data are not set');
```
uncleempty, line 45: Maybe it would be better to rephrase it as
> property data is empty

How does that sound?

PR #2549, opened 2020-06-15
Fetching lines for translation from scripts

Alroniks:
The original task is [BSA-128](https://oat-sa.atlassian.net/browse/BSA-128)

It is required for the proper building lines for translation from scripts, like [there](https://github.com/oat-sa/extension-tao-colorcontrast/pull/10).

### DIFF
--- a/scripts/class.TaoTranslate.php
+++ b/scripts/class.TaoTranslate.php
@@ -97,7 +97,8 @@ class tao_scripts_TaoTranslate extends tao_scripts_Runner
         'views',
         'helper',
         'controller',
-        'model'
+        'model',
+        'scripts',
     ];
 
     protected $verbose = false;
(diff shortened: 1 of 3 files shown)


scripts/class.TaoTranslate.php
```diff
@@ -97,7 +97,8 @@ class tao_scripts_TaoTranslate extends tao_scripts_Runner
         'views',
         'helper',
         'controller',
-        'model'
+        'model',
+        'scripts',
```
wazelin, line 101: Why would you need translation calls in `scripts`?
I fail to understand the use case.

Alroniks: To be able to catch lines that can be defined in scripts, as it works with theme configurations, wherein the config stores theme id and its name. So, this name sets by the script but should be translated.

PR #2569, opened 2020-06-24
upgrade doctrine migrations to 3.0.1

hutnikau:
Change the doctrine/migrations dependency in order to get latest fixes.
Fix notice: `Notice: Trying to access array offset on value of type int in D:\domains\package-tao\vendor\symfony\console\Input\ArrayInput.php on line 135`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "psr/http-message": "~1.0",
         "slim/slim": "^3.0",
         "symfony/dotenv": "~4.3",
-        "doctrine/migrations" : "~3.0.1"
+        "doctrine/migrations" : "3.0.1"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -60,7 +60,7 @@
         "psr/http-message": "~1.0",
         "slim/slim": "^3.0",
         "symfony/dotenv": "~4.3",
-        "doctrine/migrations" : "3.0.0"
+        "doctrine/migrations" : "^3.0.1"
```
bugalot, line 63: Please use fixed versions, this made me retrieve 

```
Package operations: 0 installs, 5 updates, 0 removals
  - Updating doctrine/migrations (3.0.0 => dev-master da238fb)
```

This is why I really recommend using fixed versions. This is not the same commit as the latest release which is a problem

PR #2571, opened 2020-06-24
Fix/indexation fixes

uncleempty:


models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
```diff
@@ -21,88 +21,6 @@
...
 namespace oat\tao\model\search\index\DocumentBuilder;
 
-use oat\generis\model\OntologyRdfs;
-use oat\tao\model\search\index\IndexDocument;
-use oat\tao\model\search\index\IndexProperty;
-use oat\tao\model\search\index\OntologyIndex;
-use oat\tao\model\search\SearchTokenGenerator;
-use oat\tao\model\TaoOntology;
-
 class GenerisIndexDocumentBuilder extends AbstractIndexDocumentBuilder
```
andreluizmachado, line 24: Do we still need this class?

models/classes/search/index/DocumentBuilder/IndexDocumentBuilderInterface.php
```diff
@@ -45,13 +47,5 @@ public function createDocumentFromResource(\core_kernel_classes_Resource $resour
      * @throws \common_Exception
      */
     public function createDocumentFromArray(array $resource = [], string $rootResourceType = ""): IndexDocument;
```
andreluizmachado, line 49: do we still need the rootResourceType?

andreluizmachado (commented):
Just feel comments.

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
uncleempty: Nope, but do you think it'd be better to return abstract builder instead of the "real" one, even if currently real builder (this one) has no code inside?

models/classes/search/index/DocumentBuilder/IndexDocumentBuilderInterface.php
uncleempty: Actually, the same reason as for previous one - it will allow to make the implementation more or less easily extendable. Or we can get rid of and revert to the initial code.
What do you think?

models/classes/search/index/DocumentBuilder/GenerisIndexDocumentBuilder.php
andreluizmachado: IMO yes because it'll decrease the code complexity, and also if we consider the YAGNI principle. So I'd suggest thinking about it just when necessary.

models/classes/search/index/DocumentBuilder/IndexDocumentBuilderInterface.php
andreluizmachado: I think that if we do not need it now. we should remove it. :(

andreluizmachado (approved):

PR #2572, opened 2020-06-24
Fix/tao 10285 error removing property

luisfn:
With this fix we are passing the property real name (like `TextBox_prop`) instead of the property object. This was needed since when the event was captured, the property was not available on database anymore for further operations

### DIFF
--- a/actions/class.PropertiesAuthoring.php
+++ b/actions/class.PropertiesAuthoring.php
@@ -33,6 +33,7 @@ use oat\tao\model\event\ClassFormUpdatedEvent;
 use oat\tao\model\event\ClassPropertiesChangedEvent;
 use oat\tao\model\search\index\OntologyIndex;
 use oat\tao\model\search\index\OntologyIndexService;
+use oat\tao\model\search\tasks\IndexTrait;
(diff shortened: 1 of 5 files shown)


actions/class.PropertiesAuthoring.php
```diff
@@ -128,8 +130,12 @@ public function removeClassProperty(): void
 
         $class = $this->getClass($this->getRequestParameter('classUri'));
         $property = $this->getProperty($this->getRequestParameter('uri'));
+        $propertyType = $this->getPropertyType($property);
 
-        $this->getEventManager()->trigger(new ClassPropertyRemovedEvent($class, $property));
+        if ($propertyType !== null) {
```
andreluizmachado, line 135: suggestion usually I prefer to compare null !== propertyType.
https://en.wikipedia.org/wiki/Yoda_conditions

luisfn: I really don't like this. Would agree with you if was a PSR or in company standards. As the same page says:

`Yoda conditions are widely criticized for compromising readability by increasing the cognitive load of reading the code.`

I see some usage of this on TAO codebase but majority is in the `normal` way, so I would prefer to keep it as is

PR #2574, opened 2020-06-25
PUT under the REST

zagovorichev:
Pull request summary
 
Related to: https://oat-sa.atlassian.net/browse/TCA-678
 
REST API doesn't allow us to use a data from the body, that required with PUT action.
 
 - send PUT with the data that has to be updated in the resource
 - nothing changed
  
#### How to test
 
```
curl --location --request PUT 'http://tao-act.loc/taoLti/RestService/index?uri=http%3A%2F%2Fsample%2Fact.rdf%23i5ef44f33c222b694e6664782566e0038' \
--header 'Accept: application/json' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Authorization: Basic ###' \
--data-raw '{
    "oauth_consumer_key": "valfromput2",
    "oauth-key": "valfromput2_oauth",
    "label": "labelfromput2",
    "oauth-secret": "secretfromput2",
    "oauth-callback-url": "callbackfromput2"
}'
```

**uri** - ID of the LTI consumer
 
or use any other tools like a Storm's REST tool or Postman.

### DIFF
--- a/actions/class.CommonRestModule.php
+++ b/actions/class.CommonRestModule.php
@@ -221,6 +221,31 @@ abstract class tao_actions_CommonRestModule extends tao_actions_RestController
         }
     }
 
+    private function getPostData(): array
+    {
+        if (!is_array($parameters = $this->getPsrRequest()->getParsedBody())) {
+            $parameters = [];
+        }
+
+        return $parameters;
+    }
+
+    private function getBodyData(): array
+    {
+        $data = null;
+        $json = $this->getPsrRequest()->getBody()->getContents();
+        if ($json) {
+            $data = @json_decode($json, true);
+        }
+
+        return is_array($data) ? $data : [];
+    }
+
+    private function getRequestData(): array
+    {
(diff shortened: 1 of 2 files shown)


SashaTAO:
Could you please create PR with correct ticket ID in branch name, so it will be linked to JIRA ticket?

actions/class.CommonRestModule.php
```diff
@@ -221,21 +221,44 @@ protected function put($uri)
...
+
+        return $parameters;
+    }
+
+    private function getBodyData(): array
+    {
+        $data = null;
+        $json = $this->getPsrRequest()->getBody()->getContents();
+        if ($json) {
+            $data = @json_decode($json, true);
```
augustas, line 238: Would it be possible to add some middleware in request creation, which would only decode the json if request has header `application/json`? We have some API's where XML is exchanged in the body

zagovorichev:
@augustas @SashaTAO This PR moved here: https://github.com/oat-sa/tao-core/pull/2579

PR #2579, opened 2020-06-29
Fix/tca 678/rest put request

zagovorichev:
Pull request summary
 
Related to: https://oat-sa.atlassian.net/browse/TCA-678
 
REST API doesn't allow us to use a data from the body, that required with PUT action.
 
 - send PUT with the data that has to be updated in the resource
 - nothing changed
  
#### How to test
 
```
curl --location --request PUT 'http://tao-act.loc/taoLti/RestService/index?uri=http%3A%2F%2Fsample%2Fact.rdf%23i5ef44f33c222b694e6664782566e0038' \
--header 'Accept: application/json' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Authorization: Basic ###' \
--data-raw '{
    "oauth_consumer_key": "valfromput2",
    "oauth-key": "valfromput2_oauth",
    "label": "labelfromput2",
    "oauth-secret": "secretfromput2",
    "oauth-callback-url": "callbackfromput2"
}'
```

**uri** - ID of the LTI consumer
 
or use any other tools like a Storm's REST tool or Postman.

### DIFF
--- a/actions/class.CommonRestModule.php
+++ b/actions/class.CommonRestModule.php
@@ -232,10 +232,9 @@ abstract class tao_actions_CommonRestModule extends tao_actions_RestController
 
     private function getBodyData(): array
     {
-        $data = null;
-        $json = $this->getPsrRequest()->getBody()->getContents();
-        if ($json) {
-            $data = @json_decode($json, true);
+        $data = $this->getPsrRequest()->getBody()->getContents();
+        if ($data && $this->hasHeader('Accept') && $this->getHeader('Accept') === 'application/json') {
+            $data = @json_decode($data, true);
         }
 
         return is_array($data) ? $data : [];
--- a/actions/class.CommonRestModule.php
+++ b/actions/class.CommonRestModule.php
@@ -252,7 +252,7 @@ abstract class tao_actions_CommonRestModule extends tao_actions_RestController
      *
      * @return array
      */
-    protected function getParameters(): array
+    protected function getParameters()
     {
         $effectiveParameters = [];
         $missedAliases = [];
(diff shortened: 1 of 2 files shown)


zagovorichev:
@augustas I've added accepted header checker, but what do you think: do I need to add xml parser here? 
(according to this commentary: https://github.com/oat-sa/tao-core/pull/2574#discussion_r446102487)

actions/class.CommonRestModule.php
```diff
@@ -221,21 +221,43 @@ protected function put($uri)
...
+
     /**
      * Returns all parameters that are URIs or Aliased with values
      *
      * @throws \common_exception_RestApi If a mandatory parameter is not found
      *
      * @return array
      */
-    protected function getParameters()
+    protected function getParameters(): array
```
SashaTAO, line 255: If we add return type declaration here we have to add it in classes which extend this one.

actions/class.CommonRestModule.php
```diff
@@ -221,21 +221,43 @@ protected function put($uri)
...
+            $parameters = [];
+        }
+
+        return $parameters;
+    }
+
+    private function getBodyData(): array
+    {
+        $data = $this->getPsrRequest()->getBody()->getContents();
+        if ($data && $this->hasHeader('Accept') && $this->getHeader('Accept') === 'application/json') {
```
SashaTAO, line 236: `$this->getHeader('Accept')` returns an array and condition is always `false`.

actions/class.CommonRestModule.php
zagovorichev: done

actions/class.CommonRestModule.php
zagovorichev: fixed

PR #2580, opened 2020-06-29
Feature/nex 1243/fix error json response

bartlomiejmarszal:
https://oat-sa.atlassian.net/browse/NEX-1243

In order to handle error response backend need to provide `errorCode`


models/classes/http/response/ErrorJsonResponse.php
```diff
@@ -40,12 +40,14 @@ public function __construct(int $errorCode, string $message, array $data = null)
 
     /**
      * @inheritDoc
+     * @deprecated code
```
siwane, line 43: Is it deprecated?

boajer: If this is deprecated,  please mentioned where is the preferable alternative to use, instead of deprecated  method

bartlomiejmarszal: My point was that a `code` key in the array is deprecated.

habemuscode: Is a custom solution because old FE code, for that the label I suppose

gabrielfs7: This is not deprecated. Please remove the annotation.

- [ ] Remove the annotation.

Btw, this PR should be declined. Why are we duplicating this code?

- [ ] Cancel this PR.

habemuscode:
@gabrielfs7 The standard response in all TAO platform right now is:

- If error, send 200 anyway
- send `errorCode` because is the way used in FE since many years ago

We cant refactor all TAO backoffice platform. Do that fix will introduce bugs in ALL parts of backoffice

habemuscode:
PR refactored, only includes necessary part for FE

habemuscode:
Closed because not need to BE changes. Changes will cover [in this PR](https://github.com/oat-sa/tao-core/pull/2584)

PR #2583, opened 2020-07-02
Feature/uno 336/unit tests

SergiiTao:


models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -311,7 +311,7 @@ private function updateProperties(Value $value): void
      * @throws ValueConflictException
      * @noinspection PhpDocMissingThrowsInspection
      */
-    private function verifyUriUniqueness(Value $value): void
+    protected function verifyUriUniqueness(Value $value): void
```
wazelin, line 314: I prefer keeping `public/protected/private` order in a class definition. If those have to be `protected`, they should be moved higher.

wazelin:
I'll merge and we can review everything in a single PR.

Thanks a lot, @SergiiTao!

PR #2589, opened 2020-07-07
Fix/UNO-383/Case-sensitive search

shpran:
Related to: [UNO-383](https://oat-sa.atlassian.net/browse/UNO-383)

**Changes:**
* Search is now not case-sensitive.

**Hot to test:**
1. Create a new Item or a test property via Manage Schema on a directory;
2. Set the property's type to List – Multiple choice – Search input;
3. Set the property's list relation;
4. Add a new item/test;
5. Try typing a lowercase/uppercase text in the property's field.

**Dependencies:**
* [#82](https://github.com/oat-sa/extension-tao-udir/pull/82)


models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -303,7 +303,10 @@ private function enrichQueryWithSubject(ValueCollectionSearchRequest $searchRequ
 
         $query
             ->andWhere(
-                $this->getPersistence()->getPlatForm()->getQueryBuilder()->expr()->like('element.object', ':subject')
+                $this->getPersistence()->getPlatForm()->getQueryBuilder()->expr()->like(
+                    'lower(element.object)',
+                    'lower(:subject)'
```
wazelin, line 308: I'm pretty sure this is covered by a test case.
Could you please adapt it to your modification?

shpran: You are right, thanks for remark!
Done.

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -303,7 +303,10 @@ private function enrichQueryWithSubject(ValueCollectionSearchRequest $searchRequ
 
         $query
             ->andWhere(
-                $this->getPersistence()->getPlatForm()->getQueryBuilder()->expr()->like('element.object', ':subject')
+                $this->getPersistence()->getPlatForm()->getQueryBuilder()->expr()->like(
```
SergiiTao, line 306: it can have huge impact on performance as LOWER prevents using index. please consider creating a functional index or find another approach.

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -303,7 +303,10 @@ private function enrichQueryWithSubject(ValueCollectionSearchRequest $searchRequ
...
+                    'LOWER(element.object)',
+                    'LOWER(:subject)'
```
SergiiTao, line 308: `:subject` should be lowercased by PHP

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
wazelin: While not an SQL standard, PosgreSQL offers `ILIKE`.

I suggest extending the query builder to support it, depending on the current DBMS driver.

SergiiTao: `ILIKE` is even slower than `LOWER`

wazelin: Too bad I didn't check it before leaving the comment. Thanks!

wazelin: Huh, actually, it seems like the `LIKE` condition didn't weight much. For your task with long lists it may be a different case, though, I admit.

Here's how it goes now, without any modifications:
```
 Nested Loop  (cost=1.50..38.41 rows=1 width=143)
   ->  Nested Loop  (cost=0.96..17.01 rows=4 width=103)
         ->  Index Scan using k_sp on statements property  (cost=0.54..8.56 rows=1 width=36)
               Index Cond: (((subject)::text = 'https://demo.udir.taocloud.org/tao.rdf#i5f056d80b4fbc3463a973756e7b31672'::text) AND ((predicate)::text = 'http://www.w3.org/2000/01/rdf-schema#range'
::text))
         ->  Index Scan using k_po on statements collection  (cost=0.41..8.43 rows=1 width=103)
               Index Cond: (((predicate)::text = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'::text) AND (object = property.object))
   ->  Index Scan using k_sp on statements element  (cost=0.54..5.34 rows=1 width=107)
         Index Cond: (((subject)::text = (collection.subject)::text) AND ((predicate)::text = 'http://www.w3.org/2000/01/rdf-schema#label'::text))
         Filter: (object ~~ 'eng%'::text)
```

This is the `ILIKE` approach:
```
 Nested Loop  (cost=1.50..38.41 rows=1 width=143)
   ->  Nested Loop  (cost=0.96..17.01 rows=4 width=103)
         ->  Index Scan using k_sp on statements property  (cost=0.54..8.56 rows=1 width=36)
               Index Cond: (((subject)::text = 'https://demo.udir.taocloud.org/tao.rdf#i5f056d80b4fbc3463a973756e7b31672'::text) AND ((predicate)::text = 'http://www.w3.org/2000/01/rdf-schema#range'
::text))
         ->  Index Scan using k_po on statements collection  (cost=0.41..8.43 rows=1 width=103)
               Index Cond: (((predicate)::text = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'::text) AND (object = property.object))
   ->  Index Scan using k_sp on statements element  (cost=0.54..5.34 rows=1 width=107)
         Index Cond: (((subject)::text = (collection.subject)::text) AND ((predicate)::text = 'http://www.w3.org/2000/01/rdf-schema#label'::text))
         Filter: (object ~~* 'eng%'::text)
```

And this is the `LOWER` approach:
```
 Nested Loop  (cost=1.50..38.42 rows=1 width=143)
   ->  Nested Loop  (cost=0.96..17.01 rows=4 width=103)
         ->  Index Scan using k_sp on statements property  (cost=0.54..8.56 rows=1 width=36)
               Index Cond: (((subject)::text = 'https://demo.udir.taocloud.org/tao.rdf#i5f056d80b4fbc3463a973756e7b31672'::text) AND ((predicate)::text = 'http://www.w3.org/2000/01/rdf-schema#range'
::text))
         ->  Index Scan using k_po on statements collection  (cost=0.41..8.43 rows=1 width=103)
               Index Cond: (((predicate)::text = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'::text) AND (object = property.object))
   ->  Index Scan using k_sp on statements element  (cost=0.54..5.34 rows=1 width=107)
         Index Cond: (((subject)::text = (collection.subject)::text) AND ((predicate)::text = 'http://www.w3.org/2000/01/rdf-schema#label'::text))
         Filter: (lower(object) ~~ 'eng%'::text)
```

wazelin: So, there's not much of a difference to it. `LOWER` is a tiny bit slower than the `ILIKE`, but in the end, unfortunately, it actually filters them one-by-one, even now.

So, I wouldn't really modify anything yet. Let's solve this performance issue when we get to it.

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
wazelin: It shouldn't really cause any performance issue at all. The query in the end will be executed by the same scenario.

ktsirkou:
How can we test this on mariaDB + linux? I have always comparison case-insesitive by default

shpran:
> How can we test this on mariaDB + linux? I have always comparison case-insesitive by default

@ktsirkou If I'm not mistaken, MariaDB behaves just like MySQL. Therefore, by default, they use case-insensitive search (correct me if I am wrong).
I reproduced this error on PostgreSQL.

ktsirkou:
> > How can we test this on mariaDB + linux? I have always comparison case-insesitive by default
> 
> @ktsirkou If I'm not mistaken, MariaDB behaves just like MySQL. Therefore, by default, they use case-insensitive search (correct me if I am wrong).
> I reproduced this error on PostgreSQL.

Yes, indeed for MariaDB is by default on Linux case-insensitive comparison. 
I am asking if somehow we can disable this on DB server settings or we have to verify it on postgres only.
Does someone knows if and how we can disable case-insensitive comparison on MariaDB + linux? 
I tried to set `lower_case_table_names=1` on my.cnf doesn't work after server restart the variable is always set to 0 :(

wazelin:
@ktsirkou, you have to change the column's or the whole table's collation to a case sensitive one, like `utf_bin` (without `_ci` at the end).

ktsirkou:
> @ktsirkou, you have to change the column's or the whole table's collation to a case sensitive one, like `utf8_bin` (without `_ci` at the end).

Thanks @wazelin, worked like charm

wazelin:
Don't forget to update the `version` to a proper one when resolving the conflicts, @shpran.
GJ!

PR #2591, opened 2020-07-08
Proper search text box validation

Alroniks:
**Require this PR**

https://github.com/oat-sa/tao-core-ui-fe/pull/153

**Related to**

[UNO-401](https://oat-sa.atlassian.net/browse/UNO-401)

**Description**

Now search text box extends `MultipleElement` class and overrides `getRawData`, that returns list to the validator instead of one value. Also, PSR improvements applied.

**Steps to reproduce**

- Create a new Item or a test property via Manage Schema on a directory
- Set the property's type to List – Multiple choice – Search input
- Set the property's list relation
- Set the property Validation Rules to `notEmpty`
- Add a new item/test
- Add a few values to the input
- Try saving the item/test

**Expected**

The item/test is saved without any errors

**Experienced**

This field is required validation error appears next to the field

**Suggested solution**

class.Searchtextbox.php should inherit the class.MultipleElement.php and implement the following method.

```
public function getRawValue()
{
    return $this->values;
}
```


helpers/form/elements/AbstractSearchTextBox.php
```diff
@@ -22,24 +21,26 @@
...
-    private $values = [];
+    protected $values = [];
 
     /**
      * @inheritDoc
      */
     public function feed(): void
     {
-        $this->values = [];
-
```
wazelin, line 42: It's there for a reason as far as I remember.

Try adding a few values to the property, saving them, and then removing one or two ;)

helpers/form/validators/class.NotEmpty.php
```diff
@@ -19,53 +19,39 @@
...
      *
      * @access public
      * @author Joel Bout, <joel.bout@tudor.lu>
-     * @param  values
-     * @return boolean
+     * @param  mixed $values
+     * @return bool
      */
-    public function evaluate($values)
+    public function evaluate($values): bool
```
wazelin, line 41: I hope no one extended this method.
Anyway, adding a return type to a non-private method is technically a breaking change. Kindly consider bumping the major version.

helpers/form/class.FormElement.php
```diff
@@ -21,6 +21,10 @@
  */
 
 use oat\oatbox\validator\ValidatorInterface;
+use oat\tao\helpers\form\elements\xhtml\SearchTextBox;
+
+// Defining aliases for old style class names for backward compatibility
+class_alias(SearchTextBox::class, \tao_helpers_form_elements_xhtml_Searchtextbox::class);
```
wazelin, line 27: Neat trick!

helpers/form/validators/class.NotEmpty.php
Alroniks: You are right, need to check to ensure.

helpers/form/elements/AbstractSearchTextBox.php
Alroniks: Property already defined as an empty list. Is it required to empty it before each feed action?

wazelin: I don't really remember how, but that property is not going to be empty when the `feed`'s being called AFTER you had already assigned some values to a property, and saved it.

helpers/form/validators/class.NotEmpty.php
Alroniks: In our org I didn't find any usage when this class inherited, so it should be pretty safe.

wazelin: Still technically a breaking change, though.

Alroniks: I reverted these changes, as it only a bug. Need to do real refactoring to fix all such cases.

helpers/form/elements/AbstractSearchTextBox.php
Alroniks: Yes, it brakes removing items from the list. I reverted it back.

Alroniks:
@ktsirkou It seems comes from SearchTextBox implementation itself, I saw it before my changes were applied. So no related to the validator. BTW I will check

Alroniks:
So, issue with validation text color is fixed, but we still need to fix the issue with red border around the input.
Eksamens- og prøvesystemet

habemuscode:
Let's merge first npm package, I release and then upgrade version here

habemuscode:
Other PR have final version of the solution :)

wazelin:
Waiting for the linked PR to be merged and released.
Then, within this PR `css` will have to be compiled.

habemuscode:
@Alroniks Release done: `@oat-sa/tao-core-ui@1.5.4`

Update `package.json` and `package.lock`

Alroniks:
@ktsirkou Please, have a look on the latest changes.

PR #2605, opened 2020-07-17
TAO-10365 Changed IndexDocumentBuilder.php to not rely on Extension Manager

luisfn:
On previous implementation we were checking if the `taoSimpleDac` extension was enabled, what isn't really a reliable way to check. Instead, we are now checking if the PermissionProvider implements an interface

### DIFF
--- a/models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
+++ b/models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
@@ -261,7 +261,7 @@ class IndexDocumentBuilder extends InjectionAwareService implements IndexDocumen
             return null;
         }
 
-        $accessRights = $permissionProvider->getResourceAccessData($resource);
+        $accessRights = $permissionProvider->getResourceAccessData($resource->getUri());
         $accessRightsURIs = ['read_access' => array_keys($accessRights)];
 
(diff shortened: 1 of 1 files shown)


models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
```diff
@@ -253,20 +253,17 @@ function (string $propertyValue): string {
         return new ArrayIterator($customProperties);
     }
 
-    private function getAccessProperties(core_kernel_classes_Resource $resource): Iterator
+    private function getAccessProperties(core_kernel_classes_Resource $resource): ?Iterator
     {
-        $accessRights = AdminService::getUsersPermissions($resource->getUri());
+        $permissionProvider = $this->getServiceLocator()->get(PermissionInterface::SERVICE_ID);
 
+        if (!$permissionProvider instanceof ReverseRightLookupInterface) {
```
wazelin, line 260: Don't forget to bump a `generis` dependency version.

luisfn: Thanks for the reminder. We will fix these versions when merging `feature/TAO-10203-advanced-search` to `develop`

PR #2609, opened 2020-07-20
UNO-392 BREAKING CHANGE: extract non-core widgets from generis

wazelin:
- [UNO-392](https://oat-sa.atlassian.net/browse/UNO-392) BREAKING CHANGE: remove duplicate value widget ID constant usages
- [UNO-392](https://oat-sa.atlassian.net/browse/UNO-392) feat: extract non-core widget definitions from `generis`
- [UNO-392](https://oat-sa.atlassian.net/browse/UNO-392) fix: a typo in `widgetdefinitions.rdf` file name
- [UNO-392](https://oat-sa.atlassian.net/browse/UNO-392) BREAKING CHANGE: depend on `generis >= 13.0.0`

⚠️ Depends on `generis` [#823](https://github.com/oat-sa/generis/pull/823).
⚠️ Has to be deployed together with `extension-tao-testqti` [#1874](https://github.com/oat-sa/extension-tao-testqti/pull/1874).

🔗 Required for `extension-tao-resource-workflow` [#24](https://github.com/oat-sa/extension-tao-resource-workflow/pull/24)
🔗 Required for `extension-tao-delivery-rdf` [#357](https://github.com/oat-sa/extension-tao-delivery-rdf/pull/357)
🔗 Required for `extension-tao-publishing` [#90](https://github.com/oat-sa/extension-tao-publishing/pull/90)

### TODO

- [x] Check for `statements` table diff after running the migration

### Diff verification process

1. Install a fresh TAO instance
2. Clone `statements` table into `statements_initial`
3. Perform the migration
4. Verify the diff via the following query
```sql
SELECT 'statements' AS `set`, st.*
FROM statements st
WHERE ROW (st.modelid, st.subject, st.predicate, st.object, st.l_language) NOT IN
      (
          SELECT modelid, subject, predicate, object, l_language
          FROM statements_initial
      )
UNION ALL
SELECT 'statements_initial' AS `set`, initial_st.*
FROM statements_initial initial_st
WHERE ROW (initial_st.modelid, initial_st.subject, initial_st.predicate, initial_st.object, initial_st.l_language) NOT IN
      (
          SELECT modelid, subject, predicate, object, l_language
          FROM statements
      );
```


models/classes/WidgetDefinitions.php
```diff
@@ -21,19 +21,14 @@
...
-    const PROPERTY_PASSWORD = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#Password';
-    const PROPERTY_HIDDENBOX = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HiddenBox';
-    const PROPERTY_RADIOBOX = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#RadioBox';
-    const PROPERTY_COMBOBOX = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#ComboBox';
-    const PROPERTY_CHECKBOX = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#CheckBox';
-    const PROPERTY_FILE = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#AsyncFile';
-    const PROPERTY_VERSIONEDFILE = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#VersionedFile';
-    const PROPERTY_JSONOBJECT = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#JsonObject';
+    public const PROPERTY_CALENDAR          = self::NAMESPACE . '#Calendar';
+    public const PROPERTY_TREEBOX           = self::NAMESPACE . '#TreeBox';
```
Alroniks, line 29: If this constant widely used already? I prefer to have words separated by `_` like `PROPERTY_TREE_BOX` and others. If they are used already, probably deprecation and defining new ones may be an option.

As we already doing breaching change, make sense to do it once. :)

wazelin: Done.
That adds [one more small PR](https://github.com/oat-sa/extension-tao-testqti/pull/1874), though ;)

models/classes/WidgetDefinitions.php
```diff
@@ -21,19 +21,14 @@
...
-    const PROPERTY_CHECKBOX = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#CheckBox';
-    const PROPERTY_FILE = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#AsyncFile';
-    const PROPERTY_VERSIONEDFILE = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#VersionedFile';
-    const PROPERTY_JSONOBJECT = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#JsonObject';
+    public const PROPERTY_CALENDAR          = self::NAMESPACE . '#Calendar';
+    public const PROPERTY_TREE_BOX          = self::NAMESPACE . '#TreeBox';
+    public const PROPERTY_FILE              = self::NAMESPACE . '#AsyncFile';
+    public const PROPERTY_VERSIONED_FILE    = self::NAMESPACE . '#VersionedFile';
+    public const PROPERTY_JSON_OBJECT       = self::NAMESPACE . '#JsonObject';
+    public const PROPERTY_WIDGET_SEARCH_BOX = self::NAMESPACE . '#SearchTextBox';
```
Alroniks, line 33: Seems should be `PROPERTY_WIDGET_SEARCH_TEXT_BOX` (I don't know why suggestions do not work here)

helpers/form/elements/AbstractSearchTextBox.php
```diff
@@ -23,15 +23,14 @@
...
 namespace oat\tao\helpers\form\elements;
 
-use oat\generis\model\WidgetRdf;
 use tao_helpers_form_FormElement;
 use tao_helpers_Uri;
 
 abstract class AbstractSearchTextBox extends tao_helpers_form_FormElement
 {
-    protected const VALUE_DELIMITER = ',';
+    public const WIDGET_ID = 'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#SearchTextBox';
```
SergiiTao, line 31: it looks like all widgets must have `public const WIDGET_ID`.
I am thinking of contracts current and new widgets should fallow and I think that `abstract static public getId(): string` could be a bit safer.
what do think?

wazelin: I was considering that approach at first, but saw a couple of instances when the value is being used as another constant's. It's just easier to get it from another constant in that case.

Besides, it's not safe to introduce such a method, I'm not sure I found all the extensions of a base widget form class.

SergiiTao: but those widgets you probably haven't found still won't work without `WIDGET_ID` right?

wazelin: They will, I didn't remove `protected $widget` property.

PR #2620, opened 2020-07-27
TAO-10286 Normalize type before indexing document

luisfn:
Adjusting document type before indexation

### DIFF
--- a/models/classes/search/index/IndexService.php
+++ b/models/classes/search/index/IndexService.php
@@ -100,11 +100,15 @@ class IndexService extends ConfigurableService
      */
     public function createDocumentFromArray($array = []): IndexDocument
     {
-        if (!isset($array['body'])) {
-            throw new \common_exception_MissingParameter('body');
+        if (!isset($array['body']['type'])) {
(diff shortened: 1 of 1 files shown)


models/classes/search/index/IndexService.php
```diff
@@ -100,13 +100,17 @@ public function createDocumentFromResource(\core_kernel_classes_Resource $resour
...
+            throw new \common_exception_MissingParameter('body[type]');
         }
         if (!isset($array['id'])) {
             throw new \common_exception_MissingParameter('id');
         }
 
+        if (is_string($array['body']['type'])) {
+            $array['body']['type'] = [$array['body']['type']];
+        }
+
```
Skyld, line 113: This looks for me like a dirty workaround. Why can we have string here if it's unexpected?

luisfn: And it, but it's hard to change it since the method is used by other extensions. 
I will Reject/Close this PR and verify it on the `lib-tao-elasticsearch`

PR #2621, opened 2020-07-29
Fix/BSA-202/Add base64 helper

shpran:
Relates to: [BSA-202](https://oat-sa.atlassian.net/browse/BSA-202)

**Changes:**
* Added base64 helper.

**Companion PRs:**
* [#418](https://github.com/oat-sa/extension-tao-item/pull/418)
* [#1499](https://github.com/oat-sa/extension-tao-itemqti/pull/1499)
* [#81](https://github.com/oat-sa/extension-tao-testqti-previewer/pull/81)

### DIFF
--- a/helpers/Base64.php
+++ b/helpers/Base64.php
@@ -14,8 +14,18 @@ class Base64
      *
      * @return bool
      */
+    public function isEncoded($data): bool
+    {
+        return is_string($data) && preg_match('/^data:.*;base64/', $data)
+    }
+
+    /**
+     * @param $data
+     *
+     * @return bool
+     */
     public static function isEncodedImage($data): bool
     {
-        return is_string($data) && preg_match('/^(data:.*;base64)/', $data) && getimagesize($data) !== false;
+        return self::isEncoded($data) && getimagesize($data) !== false;
     }
 }
--- a/helpers/Base64.php
+++ b/helpers/Base64.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace oat\tao\helpers;
+
+/**
+ * Class Base64
+ *
+ * @package oat\tao\helpers
+ */
+class Base64
+{
+    /**
+     * @param $data
+     *
+     * @return bool
+     */
+    public static function isEncodedImage($data): bool
+    {
+        return is_string($data) && getimagesize($data) !== false;
+    }
+}
--- a/helpers/Base64.php
+++ b/helpers/Base64.php
@@ -16,6 +16,6 @@ class Base64
      */
     public static function isEncodedImage($data): bool
     {
-        return is_string($data) && getimagesize($data) !== false;
+        return is_string($data) && preg_match('/^(data:.*;base64)/', $data) && getimagesize($data) !== false;
     }
 }
--- a/helpers/Base64.php
+++ b/helpers/Base64.php
@@ -16,7 +16,7 @@ class Base64
      */
     public static function isEncoded($data): bool
     {
-        return is_string($data) && preg_match('/^data:.*;base64/', $data)
+        return is_string($data) && preg_match('/^data:.*;base64/', $data);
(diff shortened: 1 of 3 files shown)


manifest.php
```diff
@@ -62,7 +62,7 @@
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '45.1.2',
+    'version' => '45.1.3',
```
wazelin, line 65: This looks like a new feature to me, kindly consider bumping a minor version instead.
```suggestion
    'version' => '45.2.0',
```

helpers/Base64.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * @param $data
+     *
+     * @return bool
+     */
+    public static function isEncodedImage($data): bool
+    {
+        return is_string($data) && preg_match('/^(data:.*;base64)/', $data) && getimagesize($data) !== false;
+    }
```
wazelin, line 20: Consider adding a less strict `isEncoded()` method, just to complete the helper.

```suggestion
    public function isEncoded($data): bool
    {
        return is_string($data) && preg_match('/^data:.*;base64/', $data)
    }

    /**
     * @param $data
     *
     * @return bool
     */
    public static function isEncodedImage($data): bool
    {
        return self::isEncoded($data) && getimagesize($data) !== false;
    }
```

You don't need a capturing group in the regex BTW.

wazelin:
Nice touch with extracting the method into a `core` helper!

Would you consider adding a few simple test cases for it?

helpers/Base64.php
```diff
@@ -0,0 +1,31 @@
...
+ * @package oat\tao\helpers
+ */
+class Base64
+{
+    /**
+     * @param $data
+     *
+     * @return bool
+     */
+    public function isEncoded($data): bool
```
wazelin, line 17: My apologies, I forgot to add a `static` keyword in my original suggestion.
```suggestion
    public static function isEncoded($data): bool
```

test/unit/helpers/Base64Test.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @dataProvider dataProvider
+     *
+     * @param $value
+     * @param bool $result
+     */
+    public function testBase64EncodingImage($value, bool $result): void
+    {
+        $this->assertEquals(Base64::isEncodedImage($value), $result);
```
wazelin, line 57: `assertEquals()` accepts an expected value as a first argument.

```suggestion
        $this->assertEquals($result, Base64::isEncoded($value));
    }

    /**
     * @dataProvider dataProvider
     *
     * @param $value
     * @param bool $result
     */
    public function testBase64EncodingImage($value, bool $result): void
    {
        $this->assertEquals($result, Base64::isEncodedImage($value));
```

wazelin:
GJ!
Just a very minor quibble to address, and the PR is good to go.

test/unit/helpers/Base64Test.php
shpran: You are right! Thanks for your comment! :)

test/unit/helpers/Base64Test.php
```diff
@@ -0,0 +1,84 @@
...
+            ],
+            'string' => [
+                'value' => 'string',
+                'result' => false,
+            ],
+            'encodedString' => [
+                'value' => self::ENCODED_IMAGE,
+                'result' => true,
+            ],
+            'encodedImage' => [
```
ktsirkou, line 78: What is the difference between the last 2 cases? 
I would like to add a case where is encoded string but not image, wdyt?

shpran: Thanks for your comment :)
Done.

PR #2627, opened 2020-08-03
Feature/nsup 27/disable validation tokens

tikhanovichA:
NSUP-28 and NSUP-27

Depends on https://github.com/oat-sa/tao-core-sdk-fe/pull/68

For one of the our clients, we need to disable xsrf tokens. By default, we can disable only validation on the server-side, but on FE side we will still try to get a token from the queue.
Because of that we added a new option which allows to reuse tokens from the queue again and send it to the server. 

How to reproduce that issue:
1. `config/taoQtiTest/testRunner.conf.php` set: `'csrf-token' => false`
2. `config/tao/security-xsrf-token.conf.php` set `poolSize' => 5`
3. Run a test with more than 5 items
4.  While in passing the test you will see the next error: `No tokens available. Please refresh the page.`

How to check the fix:
1. Update tao-core with tao-core-sdk-fe.
2. `config/taoQtiTest/testRunner.conf.php` set: `'csrf-token' => false`
3. `config/tao/security-xsrf-token.conf.php` set `poolSize' => 5` and `validateTokens => false`
4. Run a test with more than 5 items
5. You will finish the test without any issues

### DIFF
--- a/config/default/security-xsrf-token.conf.php
+++ b/config/default/security-xsrf-token.conf.php
@@ -2,5 +2,6 @@
 return new oat\tao\model\security\xsrf\TokenService([
     'store' => new oat\tao\model\security\xsrf\TokenStoreSession(),
     'poolSize' => 10,
-    'timeLimit' => 0
+    'timeLimit' => 0,
+    'validateTokens' => true
 ]);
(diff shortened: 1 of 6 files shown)


alexeyvy:
**Review Checklist**
(only the folder 'unit')
- [x] The feature is working correctly on my local machine (if applicable)
- [x] Code should be self-documented

PR #2636, opened 2020-08-07
Fix TAO-10108 apply ACL to search results

wazelin:
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) chore: apply `strict_types` to `class.Search.php`
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) chore: declare return types of `tao_actions_Search` methods
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) chore: inject services into methods of `tao_actions_Search` instead of fetching them directly from a service locator
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) fix: remove a catch of never-thrown `SyntaxException` from `tao_actions_Search::search()`
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) feat: display `Access Denied` instead of a search result element in cases a current user has no `READ` permission on the element
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) feat: add `readonly` property to a `search` response, blocking any action on a search result element
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) fix: remove a non-existent icon from search result `open` action button
- [TAO-10108](https://oat-sa.atlassian.net/browse/TAO-10108) chore(version): bump a fix one

### DIFF
--- a/views/js/layout/search.js
+++ b/views/js/layout/search.js
@@ -48,16 +48,18 @@ function($, _, __, section, actionManager, feedback){
         //create a datatable
         $tableContainer.datatable({
             url: data.url,
-            model : _.values(data.model),
-            actions : {
-                open : function openResource(id){
+            model: _.values(data.model),
+            actions: [{
+                id: 'open',
+                label: 'Open',
+                action: function openResource(id) {
                     actionManager.trigger('refresh', {
-                        uri : id
+                        uri: id
                     });
                 }
-            },
-            params : {
-                params : data.params,
+            }],
+            params: {
+                params: data.params,
                 filters: data.filters,
                 rows: 20
             }
--- a/actions/class.Search.php
+++ b/actions/class.Search.php
@@ -21,11 +21,13 @@
 
 declare(strict_types=1);
 
+use oat\generis\model\data\permission\PermissionHelper;
+use oat\generis\model\data\permission\PermissionInterface;
 use oat\generis\model\OntologyAwareTrait;
 use oat\generis\model\OntologyRdfs;
 use oat\tao\model\search\index\OntologyIndexService;
 use oat\tao\model\search\ResultSet;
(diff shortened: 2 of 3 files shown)


views/js/layout/search.js
```diff
@@ -48,16 +48,18 @@ function($, _, __, section, actionManager, feedback){
         //create a datatable
         $tableContainer.datatable({
             url: data.url,
-            model : _.values(data.model),
-            actions : {
-                open : function openResource(id){
+            model: _.values(data.model),
+            actions: [{
```
bziondik, line 52: Why you changed options to array? 
https://github.com/oat-sa/tao-core-ui-fe/blob/49a9464aea20c389e59d310082fdbdf27bafd8dc/src/datatable.js#L123

bziondik: https://github.com/oat-sa/tao-core-ui-fe/blob/49a9464aea20c389e59d310082fdbdf27bafd8dc/src/datatable.js#L371

wazelin: Because I checked not only the API, but the implementation itself: https://github.com/oat-sa/tao-core-ui-fe/blob/master/src/datatable.js#L374

This allows setting a label without defining an icon DOM element in the template, which doesn't exist ;)

bziondik: I see, can you update JSDoc here https://github.com/oat-sa/tao-core-ui-fe/blob/49a9464aea20c389e59d310082fdbdf27bafd8dc/src/datatable.js#L123 ?

wazelin: This should be done whenever we work on the `core-ui` next time, and not a scope of this PR IMO.

PR #2655, opened 2020-08-25
Feature/tca 764/create new csrf token storage interface

SashaTAO:
JIRA: https://oat-sa.atlassian.net/browse/TCA-764

There is a known issue in current BE token storage implementation. Because of this issue it may happen that two concurrent requests from client will rewrite CSRF token generated by another process. In this case there is a situation where there is `valid` token stored in client tokens pool, but it already doesn't exist in BE token storage.

Solution:
Instead of using pool of tokens on BE create a new interface which will allow to get/add/revoke one token atomically without rewriting all tokens in storage.

Requires:
- [x] https://github.com/oat-sa/generis/pull/827 (for key value storage implementation)


wazelin:
Wait, concurrent requests cannot invalidate the same token, can they?
I mean, client-side dequeues a token prior to sending a request, so when another request is being sent, a new token from the queue will be used.

wazelin (commented):
Wait, concurrent requests cannot invalidate the same token, can they?
I mean, client-side dequeues a token prior to sending a request, so when another request is being sent, a new token from the queue will be used.

SashaTAO:
> Wait, concurrent requests cannot invalidate the same token, can they?
> I mean, client-side dequeues a token prior to sending a request, so when another request is being sent, a new token from the queue will be used.

@wazelin you are right, two concurrent requests will not invalidate the same token. But it may happen that two concurrent requests will read tokens pool, after that request A will generate a new token, store it and return to the client. After that request B will do the same and rewrite token pool without token generated by the first request. When it happens at some moment client will send request signed with token which doesn't exist in the pool on BE.

wazelin:
Right, very good point.

We should definitely consider getting rid of that pool, and using a single token for all the requests within a session.
Having a pool has no security benefits.

models/classes/security/xsrf/TokenService.php
```diff
@@ -109,23 +108,26 @@ public function createToken(): Token
...
+     *
+     * @param string|Token $token The given token to validate
+     *
+     * @return boolean
+     * @throws InvalidService
+     */
+    public function checkFormToken($token): bool
+    {
+        $token = $this->normaliseToken($token);
+        $savedToken = $savedToken = $this->getStore()->getToken(self::FORM_TOKEN_NAMESPACE);
```
wazelin, line 128:
```suggestion
        $savedToken = $this->getStore()->getToken(self::FORM_TOKEN_NAMESPACE);
```

models/classes/security/xsrf/TokenService.php
```diff
@@ -140,30 +142,20 @@ public function validateToken($token): bool
     {
         $isValid = false;
         $expired = false;
-        $pool = $this->getStore()->getTokens();
-
-        if (is_object($token) && $token instanceof Token) {
-            $token = $token->getValue();
+        $token = $this->normaliseToken($token);
+        if (($savedToken = $this->getStore()->getToken($token)) === null) {
```
wazelin, line 146: It's weird to see `getToken` method being called with a `$token` argument.
Perhaps, it's better to rename `$token` into `$tokenId`, what do you think?

models/classes/security/xsrf/TokenStore.php
```diff
@@ -27,22 +27,39 @@
...
+     * @param string $tokenId
+     * @return bool
+     */
+    public function hasToken(string $tokenId): bool;
+
+    /**
+     * @param string $tokenId
+     * @return bool
      */
-    public function getTokens();
```
wazelin, line 36: Technically, that, among others, is a breaking change.

wazelin (changes requested):

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -41,60 +43,108 @@ class TokenStoreKeyValue extends ConfigurableService implements TokenStore
...
-     * @throws \common_exception_Error
+     * {@inheritDoc}
      */
-    protected function getKey()
+    public function clear(): void
+    {
+        $tokens = $this->getPersistence()->hGetAll($this->getKey());
+        foreach ($tokens as $tokenId => $value) {
+            $this->getPersistence()->hDel($this->getKey(), $tokenId);
+        }
```
augustas, line 130: I think you could just do `$this->getPersistence()->del($this->getKey());`, it will remove all hash fields/values, no need for iteration and multiple calls.

models/classes/security/xsrf/TokenService.php
```diff
@@ -253,34 +207,36 @@ public function getTokenName(): string
      *
      * @throws InvalidService
      */
-    protected function invalidate(array $pool): array
+    protected function invalidate(array $tokens): array
```
augustas, line 210: The name of this function is a bit unclear for me. I expected all passed tokens will be invalidated, but it only invalidated the oldest one. Violates POLA principle :t-rex:

augustas (changes requested):
Seems to work fine, but the bug was not reproduced neither by QA nor by devs, so it's hard to tell if it's definitely this problem that caused those errors.
The changes seem to be pretty clear, but security is a tricky topic and I would have liked to see a 3rd party library used for this.

models/classes/security/xsrf/TokenService.php
SashaTAO: Agree, it looks a bit weir. At the moment there is no token ID,  therefore I use token string as key in KV storage.
It's called `$tokenId` in token storage implementation, but here in TokenService it's actually token string value. Do you think it will be better to rename it here to `$tokenId`

models/classes/security/xsrf/TokenStore.php
SashaTAO: Ideally TokenStorage  should be kind of `private` class, used only from TokenService. Therefore I didn't want to increase major version. 
Fixed.

models/classes/security/xsrf/TokenStoreKeyValue.php
SashaTAO: Done

models/classes/security/xsrf/TokenService.php
wazelin: Hmm, you're right. OK then.

wazelin (approved):
- [ ] Pull request title and description are meaningful ⚠️ it would've been better to describe the actual feature's impact there IMO

augustas (approved):

PR #426, opened 2015-07-10
fixed 404, updated tpl on user_error.tpl

dietertaotesting:

PR #2669, opened 2020-09-08
Feature/NEX-1689/Create previewer registry service

shpran:
Relates to: [NEX-1689](https://oat-sa.atlassian.net/browse/NEX-1689)

**Changes:**
* Added abstract previewer registry service (from taoItems);
* Added unit test (from taoItems).

**Companion PRs:**
* [#352](https://github.com/oat-sa/extension-tao-test/pull/352);
* [#426](https://github.com/oat-sa/extension-tao-item/pull/426).


models/classes/preview/PreviewerRegistryService.php
```diff
@@ -0,0 +1,194 @@
...
+use oat\tao\model\modules\DynamicModule;
+use oat\tao\model\ClientLibConfigRegistry;
+use oat\oatbox\service\ConfigurableService;
+
+/**
+ * Class PreviewerRegistryService
+ *
+ * @package oat\tao\model\preview
+ */
+abstract class PreviewerRegistryService extends ConfigurableService
```
SergiiTao, line 35: why it is an abstract if it doesn't contain any abstract methods?

SergiiTao (changes requested):

models/classes/preview/PreviewerRegistryService.php
shpran: It's abstract to avoid using this service directly

SergiiTao: is it possible to create a service that can be used directly?

ktsirkou: I guess what @shpran means here is that if this service is not declared as abstract then someone can initialize an object by that and then use, which will fail as the `registryEntryKey` property will never be set. 

I do agree that abstract class with no abstract methods doesn't really make much sense. 
What about defining an abstract method to get the `registryEntryKey`?

SergiiTao: to define an abstract method just to add a small sense to keep an abstract class? :)
PreviewerRegistryService is self-sufficient.
As I already offered. IMHO it would be better to create a common service an then use directly or initialize it with factory.
in such way we will have less coupling and `composition over inheritance`

models/classes/preview/PreviewerRegistryService.php
```diff
@@ -0,0 +1,208 @@
...
+use oat\tao\model\modules\DynamicModule;
+use oat\tao\model\ClientLibConfigRegistry;
+use oat\oatbox\service\ConfigurableService;
+
+/**
+ * Class PreviewerRegistryService
+ *
+ * @package oat\tao\model\preview
+ */
+class PreviewerRegistryService extends ConfigurableService implements PreviewerRegistryServiceInterface
```
SergiiTao, line 35: is it possible to use this service, not for previewer adapters only, but for any kind of adapters stored in `ClientLibConfigRegistry`?

SergiiTao, line 35: what do you think of using `InjectionAwareService` instead of `ConfigurableService`?

SergiiTao (changes requested):

SergiiTao (approved):

ktsirkou (approved):

PR #2678, opened 2020-09-11
[Fix] UNO 588 - Unexpected scroll bar appears

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/UNO-588

**Description**

When user adds any interaction, goes to response and tries to hover on any tooltip unexpected scroll bar appears.

**Steps to reproduce**

1. Login as admin
2. Go to Items tab
3. Create a new item and go to Authoring
4. Add any interaction to the canvas (f.e. choice interaction)
5. Go to Response and try to hover any tooltip in the response properties

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -62,7 +62,7 @@ return [
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '45.10.0',
+    'version' => '45.10.1',
     'author' => 'Open Assessment Technologies, CRP Henri Tudor',
     'requires' => [
         'generis' => '>=13.4.0',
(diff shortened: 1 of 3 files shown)


bziondik (approved):
Code review only
- [ ]  New code is not subject to concurrency issues (if applicable)
- [ ]  Feature is working correctly on my local machine (if applicable)

zagovorichev (changes requested):
@habemuscode I've found that it works like that without your changes:

not sure that changing of global styles for the all project it is safe operation, we have to check everything with all browsers

So the question is: do you think we still need to have these changes?

habemuscode:
Looks fixed in Playground UDIR and BOSA, I will close

PR #2689, opened 2020-09-24
Move common used method for running scripts to AbstractMigration as a helper

Alroniks:
TAO requires two types of scripts, one for installation and one for updating, which now is called migrations. But during the migration sometimes need to run Action scripts, to initialize service or update config, that works fine in both cases, on install, and on update. 
We used to write our own methods on each migration, which looks like extra work. To avoid such duplication I suggest move this helper method to the main class and use where it needed.

Example of such method and usage [here](https://github.com/oat-sa/extension-tao-test-preview-ui-loader/pull/2/files#diff-c1fecae5f7a542a4b23d3749490321edR81-R88) and [here](https://github.com/oat-sa/extension-tao-test-preview-ui-loader/pull/2/files#diff-c1fecae5f7a542a4b23d3749490321edR30-R32).


wazelin (approved):

hutnikau (approved):

PR #2695, opened 2020-10-01
[UNO-492] Error when selecting root when assets is loading 

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/UNO-492
 
update package.json
  
#### How to test

- navigate to assets tab
- create a new passage
- navigate to authoring for this passage
- try to add an image
- while content not loaded there should be no folders in navigation tree
  
#### Dependencies
 
Requires :
 - [ ] https://github.com/oat-sa/tao-core-ui-fe/pull/188


bziondik (approved):
Code review only

PR #2704, opened 2020-10-13
Fix NEX-2150 no updater needed

SergiiTao:
Since updater scripts are deprecated we shouldn't create new ones
But currently, they are still mandatory to have.

This fix says: if an updater script is not defined let's just set the version mentioned in the manifest.

Depends on oat-sa/generis#850


wazelin (changes requested):
image

SergiiTao:
@wazelin I've seen that. but it is not related to current development. It is in `develop`. Unfortunately, I don't have time to fix that

wazelin:
Tests should be fine now.
The CI build seems to run `composer install`, not updating the pre-existing locked `dev-develop` `generis` dependency, though. And still fails.

Perhaps, an empty commit should refresh the cache.

gitromba:
> Tests should be fine now.
> The CI build seems to run `composer install`, not updating the pre-existing locked `dev-develop` `generis` dependency, though. And still fails.
> 
> Perhaps, an empty commit should refresh the cache.

Yep, with a new commit it should be fine.

Some background details, because this behaviour might seem related to composer install vs update, but in fact, it's due to Docker cache:
as a rule of thumb, each CI build does a subsequent container build where dependencies are being installed (yes, with `composer install`). When one re-runs the build on the same commit, the very same composer.json will be generated. And thanks to Docker cache (see 1st step of `install` stage) no actual composer command will be executed.

shpran (approved):

PR #2706, opened 2020-10-14
Fix/tao 10569 manage schema list value

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-10569
 
Reset list type value after property re-render
  
#### How to test
 
- Login as admin
- Go to Items or Tests
- Select any folder and click Manage Schema
- Add one property, fill all the fields, but don't save
- Add one more property
- Fill all the fields and click save


bziondik (changes requested):
I think it is wrong place to fix issue.
I debugged and see that for every request that add property is run initOntoForms
image
And then it trigger change for all properties
image
And then we go here
image

We need to execute showPropertyList only for new property not for all opened

atsymuk:
> I think it is wrong place to fix issue.
> I debugged and see that for every request that add property is run initOntoForms
> image
> And then it trigger change for all properties
> image
> And then we go here
> image
> 
> We need to execute showPropertyList only for new property not for all opened

i see your point, but unfortunately i designed this way that it listens to AJAX call and re-render the whole form and because of this it is hardly able to recognize what exactly was changed

bziondik (approved):
Fix is working. Can be refactored in future.
- [x]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  New code is not subject to concurrency issues (if applicable)
- [x]  Feature is working correctly on my local machine (if applicable)

lecosson (approved):
- [x]  New code is not subject to concurrency issues (if applicable)
- [x]  Feature is working correctly on my local machine (if applicable)

bziondik (approved):

PR #2709, opened 2020-10-16
Fix/TAO-10583/set proper content type for css request

dresha-dev:
**Related to task:** https://oat-sa.atlassian.net/browse/TAO-10583
**Description:** Set response header content type 'text/css' according to accept header

**The issue technically**
CSS files are stored without the file extension .css, so the mimetype isn't recognized automatically in getFileSystem.php

**Steps to reproduce:** 
1. Login as admin 
2. Go to items and create a new item
3. Go to authoring, add any interaction to the canvas (f.e. associate interaction)
4. Go to Style Editor in item properties and add a style sheet that is attached  
5. Create delivery
Expected behaviour: the custom css file should load with "text/css" value for header content type and apply to the item


models/classes/websource/BaseWebsource.php
```diff
@@ -152,6 +153,11 @@ public function getMimetype($filePath)
             }
         }
 
+        if (strpos($acceptHeader, 'text/css') !== false)
+        {
```
alexeyvy, line 157:
```suggestion
        if (strpos($acceptHeader, 'text/css') !== false) {
```

SergiiTao (changes requested):
dirty hack :)
we should found more clear solution

alexeyvy:
@SergiiTao the other approach may be storing proper extensions for CSS files, but it would require refactoring of how it's saved and a migration for those CSS files already added without an extension which is much heavier. I'd suggest we merge this fix as a compromise and revisit the whole approach of working with files later

AndreyNegrov (approved):

alexeyvy:
Having talked to Joel and Sergii, we decided to close the PR and publish the change only as a hotfix in v45.10.0.1, as it's not the best way of fixing things in one of the central components adding such a hack.

We will further need to revert back to the issue and address it in a more consistent manner.

PR #2718, opened 2020-10-28
fix/TR-43-adaptieve-size-for-checkboxes

meefox:
This is 2 of 4 parts that is needed for https://oat-sa.atlassian.net/browse/TR-43
Update a version for @oat-sa/tao-core-ui for https://oat-sa.atlassian.net/browse/TR-43

### DIFF
--- a/views/package-lock.json
+++ b/views/package-lock.json
@@ -39,9 +39,8 @@
             "integrity": "sha512-S6/2ol3NpT99su9ZlZgoVvZHKW+yRVro562Ab1pizyCwv2YfO1r2FbpEDTS/cgfqI2JvgMtuP1V1SjR3T8Avrg=="
         },
         "@oat-sa/tao-core-ui": {
-            "version": "1.11.2",
-            "resolved": "https://registry.npmjs.org/@oat-sa/tao-core-ui/-/tao-core-ui-1.11.2.tgz",
-            "integrity": "sha512-2YjytI3Na3QEnIvd2/UMQLP8WksKMojyhGRCw8IEeBLxliFQYiGt1j6n4F+gGh3ca0+X67iOe04oLyReNkOAoA=="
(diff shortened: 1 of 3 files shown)


views/package-lock.json
```diff
@@ -39,9 +39,8 @@
             "integrity": "sha512-S6/2ol3NpT99su9ZlZgoVvZHKW+yRVro562Ab1pizyCwv2YfO1r2FbpEDTS/cgfqI2JvgMtuP1V1SjR3T8Avrg=="
         },
         "@oat-sa/tao-core-ui": {
-            "version": "1.11.2",
-            "resolved": "https://registry.npmjs.org/@oat-sa/tao-core-ui/-/tao-core-ui-1.11.2.tgz",
-            "integrity": "sha512-2YjytI3Na3QEnIvd2/UMQLP8WksKMojyhGRCw8IEeBLxliFQYiGt1j6n4F+gGh3ca0+X67iOe04oLyReNkOAoA=="
+            "version": "github:oat-sa/tao-core-ui-fe#65c2696f3c5522b33946107f2c40732f55a1bf6d",
+            "from": "github:oat-sa/tao-core-ui-fe#fix/TR-43-adaptieve-size-for-checkboxes"
```
jsconan, line 43: **issue:** caution, this will introduce a lock to github instead of npm. Once the companion PR will be merged, we'll need to delete this file as well as the node_module, and regenerate them before merge.

jsconan (changes requested):
We need a version bump in the `manifest.php` file as well.
There is also an issue with the package-lock which should not committed while the package file targets a GitHub repository. This should be done only once the related package has been published to NPM and has a proper version. Otherwise, this will introduce a lock issue.

jsconan (changes requested):
Looks good on the code side.
Just one remark: the dependency add changes in the SCSS, this should be compiled and the changes added to the PR as well
    ```
    cd tao/views/build
    npm grunt taosass
    ```

btamas (approved):

PR #2722, opened 2020-11-02
Indexing parent tree string

bartlomiejmarszal:
This change will index string that will describe a tree that the current resource is in

https://oat-sa.atlassian.net/browse/CON-23


models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
```diff
@@ -274,4 +288,22 @@ private function getAccessProperties(core_kernel_classes_Resource $resource): ?I
 
         return new ArrayIterator($accessRightsURIs);
     }
+
+    private function getParentClasses(array $types, string $path = ''): string
+    {
+        foreach ($types as $type) {
+            $path = $type->getUri() . $path;
+            if (!$this->isRootClass($type->getUri())) {
+                $path = '/' . $path;
```
gabrielfs7, line 297: Suggestion to use `|` as separator, since `/` is already part of the class paths.

models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
```diff
@@ -274,4 +288,22 @@ private function getAccessProperties(core_kernel_classes_Resource $resource): ?I
...
+                $path = $this->getParentClasses($type->getParentClasses(), $path);
```
gabrielfs7, line 298: As we discussed before, I would use in the index a more verbose name instead of `parent`, cause it is misleading. We will be using items inside tests too, so might be confuse. 

We could use something like `parent_class_path` or `parent_classes`. Wdyt?

gabrielfs7 (commented):
Small comments! Good job!

models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
mike-ionut-mihai-sandu-tao: not something that should prevent the merge but i agree that 
we could find a better name 
imo this is a tree like path and not a path per say so `parent_tree` would make more sense 
but again not something very important

models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
mike-ionut-mihai-sandu-tao: i would  not create the glue like this but use something that can be reused 
but also i am ok with `|` just to avoid the possible bugs

mike-ionut-mihai-sandu-tao (approved):

models/classes/search/index/DocumentBuilder/IndexDocumentBuilder.php
gabrielfs7: If we put a misleading name it will remain there for a long time and a reindex would be required to fix (not good solution if we move to production). We can prevent this now and it is low cost. 

Either "path" or "tree" are suitable, but I would put "class" as part of the name to avoid confusion. I just think that `tree` seems to be a field that will receive an array inside and here we have a string, so for me `path` makes more sense.

gabrielfs7 (approved):

gabrielfs7 (approved):

PR #2730, opened 2020-11-06
fix error when property has not have a widget

bartlomiejmarszal:
When widget is not present in property this error will be thrown:

```200: Fatal error: Uncaught Error: Call to a member function getUri() on null in /tao/code/tao/models/classes/Lists/Busin ess/Service/ClassMetadataService.php:187```

### DIFF
--- a/models/classes/Lists/Business/Service/ClassMetadataService.php
+++ b/models/classes/Lists/Business/Service/ClassMetadataService.php
@@ -184,6 +184,10 @@ class ClassMetadataService extends InjectionAwareService
 
     private function isTextWidget(core_kernel_classes_Property $property): bool
     {
+        if ($property->getWidget()) {
+            $this->logWarning($property->getLabel() . 'is not including a widget property');
+            return false;
+        }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: models/classes/Lists/Business/Service/ClassMetadataService.php [hunk 1]
```diff
@@ -184,6 +184,10 @@ private function getPropertyValues(core_kernel_classes_Property $property): ?arr
 
     private function isTextWidget(core_kernel_classes_Property $property): bool
     {
+        if ($property->getWidget()) {
+            $this->logWarning($property->getLabel() . 'is not including a widget property');
+            return false;
+        }
```
gabrielfs7, line 190: Why we log it? Is this an error? 

For me, this is just a business rule to avoid include widgets. We will flood out logs with warnings unnecessary.

I would not log it at all.

gabrielfs7 (changes requested):
Unnecessary log

habemuscode (approved):
Code review only

PR #2733, opened 2020-11-13
Remove obsolete warning message

Alroniks:
Hotfix for [CGF-69](https://oat-sa.atlassian.net/browse/CGF-69)

The problem, that in templates used helper `Template::inc` that requires blocks by the path, but theming used, it passes this path to a special method in Theme implementation, but such as these classes deprecated, it works unproperly. It tries to check template id and switch construction always fails with the warning message in the log when templating works well.

As a hotfix, and unblock customers I just removed this warning, but total refactoring will be done under [CGF-134](https://oat-sa.atlassian.net/browse/CGF-134)

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -63,7 +63,7 @@ return [
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
-    'version' => '46.7.2',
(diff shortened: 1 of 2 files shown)


Alroniks:
Closing as a wrong solution.

PR #2734, opened 2020-11-16
Bump npm bridge version to make it compatible with composer 2.0

Alroniks:
Work under [CGF-22](https://oat-sa.atlassian.net/browse/CGF-22)


AndreyNegrov (approved):
Looks good

PR #2736, opened 2020-11-16
Revert "Bump npm bridge version to make it compatible with composer 2.0"

SashaTAO:
Reverts oat-sa/tao-core#2734 to unblock tao update/install


gitromba (approved):
LGTM.

PR #2743, opened 2020-11-26
Feature/CON-76/advance search in result page

bziondik:
Related to: https://oat-sa.atlassian.net/browse/CON-76
https://github.com/oat-sa/extension-tao-outcomeui/pull/357
https://github.com/oat-sa/lib-tao-elasticsearch/pull/40
- [ ] https://github.com/oat-sa/tao-core-ui-fe/pull/222

FE part: changes in `views/js/layout/search.js`:

1. search sends to `searchModal`  hideResourceSelector: true id it is Results page, and placeholder
2. after selecting search result store` delivery result id` in `searchComponent.container`

How to test:
- go to any page and check that placeholder for search input `Search Item`, `Search Result` used also in searchModal
- go to Results page
- run search (use delivery name or test taker name)
- in searchModal should be hidden `add criteria` button and resource selector
- check table with search results and press `Go To Item`
- delivery result should be loaded


views/js/layout/search.js
```diff
@@ -96,11 +97,26 @@ define(['jquery', 'layout/actions', 'ui/searchModal', 'core/store', 'context', '
...
 
         searchModalInstance.on('store-updated', manageSearchStoreUpdate);
-        searchModalInstance.on('refresh', uri => {
+        searchModalInstance.on('refresh', (uri, data) => {
             actionManager.trigger('refresh', { uri });
+            if (isResultPage && data) {
+                // TODO data.result is delivery result id
+                searchComponent.container.data(
+                    'show-result',
+                    data.result ||
```
bziondik, line 116: @bartlomiejmarszal I assumed that I will have in response `result` filed with delivery result id

habemuscode (approved):
Code review only

config/default/ResultSetMapper.conf.php
```diff
@@ -0,0 +1,35 @@
...
+        ResultSetMapper::OPTION_STRUCTURE_MAP =>
+            [
+                'default' => [
+                    'label' => [
+                        'id' => 'label',
+                        'label' => __('Label'),
+                        'sortable' => false
+                    ]
+                ],
+                'results' => [
```
bziondik, line 16: @bartlomiejmarszal please, update `label` for all `'label' => __('Delivery'),`, add `Delivery URI` as we discussed, and `Start Time`
```'delivery' => array(
                'id' => 'delivery',
                'label' => 'Delivery Uri',
                'sortable' => false,
                'visible' => false)
```
and change order according mockup - first will be `Test Taker Name`

migrations/Version202011301939462234_tao.php
```diff
@@ -0,0 +1,61 @@
...
+    public function getDescription(): string
+    {
+        return 'Register ResultSetMapper';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->getServiceManager()->register(ResultSetMapper::SERVICE_ID, new ResultSetMapper(
+            [
+                ResultSetMapper::OPTION_STRUCTURE_MAP =>
```
bziondik, line 26: Can we use here `config/default/ResultSetMapper.conf.php`?

adriloma (approved):
FE review only. Notice my comment on `package.json`

- [ ] bumping versions are applied (if applicable)

habemuscode (approved):
- [x] bumping versions are applied (if applicable)

config/default/ResultSetMapper.conf.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+use oat\tao\model\search\ResultSetMapper;
+
+return new ResultSetMapper(
+    [
+        ResultSetMapper::OPTION_STRUCTURE_MAP =>
```
shpran, line 7: I think it's better to not move the open parenthesis to the next line.
So, instead of:
```
ResultSetMapper::OPTION_STRUCTURE_MAP =>
    [
        ... code ...
    ]
```
use this one:
```
ResultSetMapper::OPTION_STRUCTURE_MAP => [
    ... code ...
]

File: migrations/Version202011301939462234_tao.php (hunk 3 above)
Comment (shpran) [inline migrations/Version202011301939462234_tao.php:26]: Same

File: models/classes/search/ResultSetMapper.php [hunk 5]
```diff
@@ -0,0 +1,31 @@
...
+    public const OPTION_STRUCTURE_MAP = 'optionStructureMap';
+
+    public function getPromiseModel(string $structure)
+    {
+        $map = $this->getOption(self::OPTION_STRUCTURE_MAP);
+        if (isset($map[$structure])) {
+            return $map[$structure];
+        }
+
+        return $map['default'];
```
shpran, line 19:
```suggestion
        return $map[$structure] ?? $map['default'];
```

models/classes/search/ResultSetResponseNormalizer.php
```diff
@@ -48,29 +46,30 @@ public function normalize(SearchQuery $searchQuery, ResultSet $resultSet): array
...
                 }
 
-                $instanceProperties = [
-                    'id' => $instance->getUri(),
-                    OntologyRdfs::RDFS_LABEL => $instance->getLabel(),
-                ];
+                $instanceProperties[] = $this->getResultSetMapper()->getResultSetModel($content, $structure);
 
-                $response['data'][] = $instanceProperties;
+                $response['data'] = $instanceProperties;
```
shpran, line 66: This block looks a bit strange for me.
WDYT about this?
```
$response['data'][] = $this->getResultSetMapper()->getResultSetModel($content, $structure);
```

shpran (commented):
Looks good. 👍🏻 
Minor cs comments :)

config/default/ResultSetMapper.conf.php
gabrielfs7: This is also PSR recommendation. I agree with @shpran

config/default/ResultSetMapper.conf.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+use oat\tao\model\search\ResultSetMapper;
+
+return new ResultSetMapper(
```
gabrielfs7, line 5: I am wondering why this needs to be configurable? If I change this, most probably a custom extesion / layout would have to be applied. I would not open this to be configurable outside right now.

models/classes/search/ResultSetMapper.php
```diff
@@ -0,0 +1,31 @@
...
+namespace oat\tao\model\search;
+
+use oat\oatbox\service\ConfigurableService;
+
+class ResultSetMapper extends ConfigurableService
+{
+    public const SERVICE_ID = 'tao/ResultSetMapper';
+    public const OPTION_STRUCTURE_MAP = 'optionStructureMap';
+
+    public function getPromiseModel(string $structure)
```
gabrielfs7, line 12: Do you think a `mapper`, that should only `map` should have methods like `getPromiseModel` or `getResultSetModel`? 

Should it not have the method `map` instead?

gabrielfs7 (changes requested):
SRP violated on `ResultSetMapper`

config/default/ResultSetMapper.conf.php
bartlomiejmarszal: I can move configuration to internal const variables. 
Yes. 

I think base on multiple extensions we may want to introduce additional indexation also over non-resource content 
(I mean other then Groups, Tests, Items etc). 

At this moment we are indexing this way `delivery-results` only but I expect that we will extend this in the future. 

If we keep our result set configurable we will have control over what we want to display in `searchModal` result data table. 

Having configurable service allow us to quickly extend our search result set to our needs.

models/classes/search/ResultSetMapper.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $map = $this->getOption(self::OPTION_STRUCTURE_MAP);
+
+        return $map[$structure] ?? $map['default'];
+    }
+
+    public function getResultSetModel(array $content, string $structure)
+    {
+        $map = $this->getOption(self::OPTION_STRUCTURE_MAP);
+        $allowedKeys = $map[$structure] ?? $map['default'];
```
shpran, line 22: To avoid code duplication we can replace this with `getPromiseModel` method call:
```
$allowedKeys = $this->getPromiseModel($structure);
```
Wdyt?

migrations/Version202011301939462234_tao.php
bartlomiejmarszal: I don't think I can use this file here. @boajer @mike-ionut-mihai-sandu-tao ?

config/default/ResultSetMapper.conf.php
bartlomiejmarszal: I think I fix all this changes. Can you have a look?

config/default/ResultSetMapper.conf.php
gabrielfs7: I agree, but it seems this configuration is not enough. We have to do several changes in the FE to add new indexes and search criteria. I am more concerned about this https://github.com/oat-sa/tao-core/pull/2743#discussion_r534939705 to be honest.

models/classes/search/ResultSetMapper.php
bartlomiejmarszal: `ResultSetModelBuilder`?

migrations/Version202011301939462234_tao.php
mike-ionut-mihai-sandu-tao: if you want to have this inside a specific place you could technically doing but i don't see the benefits of that

models/classes/Lists/Business/Service/ClassMetadataService.php
```diff
@@ -130,6 +131,9 @@ private function getClassMetadata(core_kernel_classes_Class $class): MetadataCol
         $collection = new MetadataCollection();
 
         foreach ($class->getProperties(true) as $property) {
+            if (!$this->isWidget($property)) {
```
mike-ionut-mihai-sandu-tao, line 134: are you sure we need this extra check ? 
we are checking 
```
if (!$this->isTextWidget($property) && !$this->isListWidget($property)) {
```

models/classes/Lists/Business/Service/ClassMetadataService.php
```diff
@@ -203,4 +207,9 @@ private function getListItemsUri(core_kernel_classes_Property $property, ?array
 
         return sprintf(self::BASE_LIST_ITEMS_URI, urlencode($property->getUri()));
     }
+
+    private function isWidget(core_kernel_classes_Property $property): bool
```
mike-ionut-mihai-sandu-tao, line 211: i am not sure that we need this and that this is not available anyware

models/classes/search/ResultSetFilter.php
```diff
@@ -0,0 +1,42 @@
...
+use oat\oatbox\service\ConfigurableService;
+
+class ResultSetFilter extends ConfigurableService
+{
+    public function filter(array $content, string $structure): array
+    {
+        $allowedKeys = $this->getResultSetMapper()->map($structure);
+
+        return array_filter($content, function ($key) use ($allowedKeys) {
+            return $key === 'id' || in_array($key, array_keys($allowedKeys));
```
mike-ionut-mihai-sandu-tao, line 34: maybe add 3rd element to 
```
in_array
```

models/classes/Lists/Business/Service/ClassMetadataService.php
bartlomiejmarszal: This was causing an error when Item property did not have `getWidget` set. I fix this error in our PR.

models/classes/Lists/Business/Service/ClassMetadataService.php
bartlomiejmarszal: This is solution of an error when Item property did not have getWidget set. I fix this error in our PR.

models/classes/search/ResultSetMapper.php
```diff
@@ -0,0 +1,27 @@
...
+    public const SERVICE_ID = 'tao/ResultSetMapper';
+    public const OPTION_STRUCTURE_MAP = 'optionStructureMap';
+
+    public function map(string $structure)
+    {
+        $map = $this->getOption(self::OPTION_STRUCTURE_MAP);
+
+        if (!$this->getAdvancedSearchChecker()->isEnabled()) {
+            return $map['default'];
+        }
```
gabrielfs7, line 19: Do you think we should return a map for an undesired structure because advanced search is not active? I saw the default one contains label, but what if we add another field or the result you are mapping does not have label?

Should not we have a `default` map per each type of structure instead? This is not extensible and will cause issues if fields change. I recommend change the mapping to something like:


```
                'results' => 
                    'default' => [
                        'label' => [
                            'id' => 'label',
                            'label' => __('Label'),
                            'sortable' => false
                        ],
                    ],
                    'advanced' => [
                        'label' => [
                            'id' => 'label',
                            'label' => __('Label'),
                            'sortable' => false
                        ],
                        'test_taker_name' => [
                            'id' => 'test_taker_name',
                            'label' => __('Test Taker'),
                            'sortable' => false
                        ],
                        'test_taker' => [
                            'id' => 'test_taker',
                            'label' => __('Test Taker'),
                            'sortable' => false
                        ],
                    ]
```

And then you can do something like this to get the results:

```
if (!$this->getAdvancedSearchChecker()->isEnabled()) {
    return $map[$structure]['default'] ?? [];
}

return $map[$structure]['advanced'] ?? [];
```

migrations/Version202011301939462234_tao.php
```diff
@@ -0,0 +1,60 @@
...
+    public function getDescription(): string
+    {
+        return 'Register ResultSetMapper';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->getServiceManager()->register(ResultSetMapper::SERVICE_ID, new ResultSetMapper(
+            [
+                ResultSetMapper::OPTION_STRUCTURE_MAP => [
```
gabrielfs7, line 26: I left a comment here. The result mapping default should be done by subject to avoid coupling among all different result types we will have:

https://github.com/oat-sa/tao-core/pull/2743/files#r537280131

models/classes/search/ResultSetResponseNormalizer.php
```diff
@@ -48,29 +46,28 @@ public function normalize(SearchQuery $searchQuery, ResultSet $resultSet): array
...
-            foreach ($resultsRaw as $uri) {
-                $instance = $this->getResource($uri);
-                $isAccessible = isset($accessibleResultsMap[$uri]);
+            foreach ($resultsRaw as $content) {
+                $isAccessible = isset($accessibleResultsMap[$content['id']]);
 
                 if (!$isAccessible) {
-                    $instance->label = __('Access Denied');
+                    $content['label'][] = __('Access Denied');
+                    continue;
```
gabrielfs7, line 61: Now the results with `Access Denied` will never be added for the result and it will cause broken lists. Does business agree with that? It is changing the behavior and interface.

If they agree, we also do not need then to have this line:

```
$content['label'][] = __('Access Denied');
```

...cause is never used. But I am afraid we should have same amount of results per page.

gabrielfs7 (changes requested):
Inconsistence in the results as it was before and also coupled result mapping should change

models/classes/search/ResultSetResponseNormalizer.php
```diff
@@ -22,18 +22,16 @@
...
 use oat\generis\model\OntologyAwareTrait;
-use oat\generis\model\OntologyRdfs;
 use oat\oatbox\service\ConfigurableService;
 
 class ResultSetResponseNormalizer extends ConfigurableService
 {
     use OntologyAwareTrait;
 
-    public function normalize(SearchQuery $searchQuery, ResultSet $resultSet): array
+    public function normalize(SearchQuery $searchQuery, ResultSet $resultSet, string $structure): array
```
gabrielfs7, line 34: $structure should be identified by the `ResultSet` object IMO. They are interdependent therefore should contain this info.

test/unit/models/classes/search/ResultSetFilterTest.php
```diff
@@ -0,0 +1,128 @@
...
+                    'label' => [
+                        'id' => 'label',
+                        'label' => __('Label'),
+                        'sortable' => false
+                    ]
+                ]
+            ],
+        ];
+    }
+}
```
gabrielfs7, line 128:
```suggestion
}

```

models/classes/search/ResultSetFilter.php
```diff
@@ -0,0 +1,42 @@
...
+
+class ResultSetFilter extends ConfigurableService
+{
+    public function filter(array $content, string $structure): array
+    {
+        $allowedKeys = $this->getResultSetMapper()->map($structure);
+
+        return array_filter($content, function ($key) use ($allowedKeys) {
+            return $key === 'id' || in_array($key, array_keys($allowedKeys), true);
+        }, ARRAY_FILTER_USE_KEY);
```
gabrielfs7, line 35:
```suggestion
        return array_filter(
            $content, 
            function ($key) use ($allowedKeys) {
                return $key === 'id' || in_array($key, array_keys($allowedKeys), true);
            }, 
            ARRAY_FILTER_USE_KEY
        );
```

Please, apply PSR-12.

When breaking lines to pass parameters to a function, every parameter should be in a new line.

shpran: This part is OK in my opinion. :)
You can find the same "breaking lines" in [PSR-12](https://www.php-fig.org/psr/psr-12/#47-method-and-function-calls):
```
somefunction($foo, $bar, [
  // ...
], $baz);

$app->get('/hello/{name}', function ($name) use ($app) {
    return 'Hello ' . $app->escape($name);
});
```

models/classes/search/ResultSetMapper.php
bartlomiejmarszal: When elastic search is not enabled we are using GenerisSearch to search over the labels. This means that we always have labels and we can assume that it is default model when elastic search is not enabled.

models/classes/search/ResultSetFilter.php
gabrielfs7: It is not ok, there are 2 arguments here. read it: 

```
Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line
```

models/classes/search/ResultSetMapper.php
gabrielfs7: You can assume, but if you search over generis URI or other fields we have issues. It is for the sake of extensibility and maintainability.

models/classes/search/ResultSetFilter.php
shpran: Take a look on this example:
image

models/classes/search/ResultSetResponseNormalizer.php
bartlomiejmarszal: We can set  structure string as a type of `ResultSet`

models/classes/search/ResultSetFilter.php
gabrielfs7: @shpran Yes, this is "optional" for anonymous functions and arrays, but honestly do you think it makes sense to have such different approaches?  This contradicts the first sentence.

```
When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line
```

For me is worse to read like this. It takes time to understand what is parameter for the anonymous function and what is not.

models/classes/search/ResultSetMapper.php
bartlomiejmarszal: https://github.com/oat-sa/tao-core/pull/2743/commits/4e60e857c029e27543e379224371c4279964b058
Is this solving your comment?

migrations/Version202011301939462234_tao.php
bartlomiejmarszal: https://github.com/oat-sa/tao-core/pull/2743/commits/4e60e857c029e27543e379224371c4279964b058

actions/class.Search.php
```diff
@@ -44,22 +44,36 @@ class tao_actions_Search extends tao_actions_CommonModule
...
+        $this->setSuccessJsonResponse([
             'url' => _url('search'),
             'params' => [
-                'query' => $rawQuery,
-                'rootNode' => $this->getRequestParameter('rootNode'),
-                'parentNode' => $this->getRequestParameter('parentNode'),
+                'query' => $parsedBody['query'],
+                'rootNode' => $queryParams['rootNode'],
+                'parentNode' => $parsedBody['parentNode'],
+                'structure' => $parsedBody['structure'],
```
gabrielfs7, line 73: Why structure of result response is part of the query provided by enduser?

migrations/Version202011301939462234_tao.php
```diff
@@ -0,0 +1,60 @@
...
+                    'default' => [
+                        'label' => [
+                            'id' => 'label',
+                            'label' => __('Label'),
+                            'sortable' => false
+                        ]
+                    ],
+                    'results' => [
```
gabrielfs7, line 34: This needs to be fixed as here: https://github.com/oat-sa/tao-core/pull/2743/files#diff-5de033e84d58407715598c1bd76c5d4051a24e962bbf891dcdb47785110cc2bbR15

models/classes/search/SearchQueryFactory.php
```diff
@@ -42,7 +42,7 @@ public function create(ServerRequestInterface $request): SearchQuery
 
         return new SearchQuery(
             $params['params']['query'],
-            $params['params']['rootNode'],
+            $params['params']['structure'],
```
gabrielfs7, line 45: Will it work as it was before? You used the root node for generis query, no?

test/unit/models/classes/search/ResultSetFilterTest.php
```diff
@@ -0,0 +1,128 @@
...
+                    ResultSetMapper::SERVICE_ID => $this->resultSetMapperMock
+                ]
+            )
+        );
+    }
+
+    /**
+     * @dataProvider getResultSetData
+     */
+    public function testGetResultSetModel(array $resultDataSet, string $structure, array $filteredResults, array $map): void
```
gabrielfs7, line 53: Inconsistent test name

test/unit/models/classes/search/ResultSetFilterTest.php
```diff
@@ -0,0 +1,128 @@
...
+        $this->resultSetMapperMock
+            ->method('map')
+            ->willReturn($map);
+
+        $result = $this->subject->filter($resultDataSet, $structure);
+
+        $this->assertEquals($filteredResults, $result);
+    }
+
+    public function getResultSetData()
```
gabrielfs7, line 64:
```suggestion
    public function getResultSetData(): array
```

test/unit/models/classes/search/ResultSetFilterTest.php
```diff
@@ -0,0 +1,128 @@
...
+            'results data set' => [
+                [
+                    'test_taker_name' => [
+                        'id' => 'someId'
+                    ],
+                    'some_other_key' => [
+                        'id' => 'anotherId'
+                    ]
+                ],
+                'results',
```
gabrielfs7, line 76: SHould not this be compatible with new format (with `default` and `advanced` indexes)?

bartlomiejmarszal: In this test its is not relevant. It's relevant only in `ResultSetMapperTest`

models/classes/search/SearchQueryFactory.php
bartlomiejmarszal: this attribute is relevant only for elastic search query. Gerneris search is taking parent node as a type:
`\oat\tao\model\search\GenerisSearchBridge::search` root node is not involved here.

actions/class.Search.php
bartlomiejmarszal: This is not provided by enduser. This string is generated by FE. It's referring to tab that we want to perform search query.

models/classes/search/ResultSetFilter.php
bartlomiejmarszal: Can I resolve this comment or you suggest to apply some changes here?

actions/class.Search.php
gabrielfs7: FE in this case is the user, cause enduser uses FE to reach us. But, is a `structure` of mapping representing the `tab` a good reference? 

Do not you think a more generic/agnostic name i.e `searchType` would fit better? When you read structure what do you think this is about?

models/classes/search/SearchQueryFactory.php
gabrielfs7: So how will this query still work for elastic search? Are you providing this in another way? The `SearchQuery` should be agnostic of search engine. Could you help me understand how it is working now?

gabrielfs7 (approved):
I think there are other things that could be done, but only my opinion. 

Approved! Good job! :)

models/classes/search/ResultSetFilter.php
gabrielfs7: I think nobody has issues with this format of coding except me, so I am closing :)

PR #2760, opened 2020-12-21
Fix/CON-144/Booklet preview fails due to timeout

shpran:
Relates to: [CON-144](https://oat-sa.atlassian.net/browse/CON-144)

**Changes:** default KV storage will be used to store DOM validation result for items. This will allow us to avoid unnecessary validations.

**How to test:**
1. Create or import a large test (15+ items);
2. Generate a booklet;
3. Open this booklet and preview it.

**Previous result:** _the preview takes a long time (1 - 1.5 seconds per item)._
**Current result:** _the preview takes a short time (1-3 seconds for the whole test)._

_**Note: if a booklet was generated before this fix was applied - you need to preview it or regenerate to store validation results in storage.**_

**Require PR:** [#1603](https://github.com/oat-sa/extension-tao-itemqti/pull/1603)


models/classes/class.Parser.php
```diff
@@ -227,136 +190,101 @@ public function validateMultiple($xsds = [])
...
+
                         $this->content = tao_helpers_Request::load($this->source, true);
                         break;
                     case self::SOURCE_STRING:
                         $this->content = $this->source;
                         break;
                     case self::SOURCE_FLYFILE:
-                        if (! $this->source->exists()) {
-                            throw new Exception('Source file does not exists ("' . $this->source->getBasename() . '").');
+                        if (!$this->source->exists()) {
```
gabrielfs7, line 275: The PSR-2 exception is for closures and arrays, in this case is one parameter per line. Can you please format/ident it?

models/classes/class.Parser.php
```diff
@@ -227,136 +190,101 @@ public function validateMultiple($xsds = [])
...
+                            throw new Exception(sprintf(
+                                'Source file does not exists ("%s").',
+                                $this->source->getBasename()
+                            ));
                         }
-                        if (! $this->content = $this->source->read()) {
-                            throw new Exception('Unable to read file ("' . $this->source->getBasename() . '").');
+                        if (!$this->content = $this->source->read()) {
+                            throw new Exception(sprintf(
```
gabrielfs7, line 282: The PSR-2 exception is for closures and arrays, in this case is one parameter per line. Can you please format/ident it?

models/classes/class.Parser.php
```diff
@@ -366,49 +294,116 @@ public function getContent($refresh = false)
...
     {
         $this->errors = [];
     }
 
-    /**
-     * Creates a report without title of the parsing result
-     * @return common_report_Report
-     */
-    public function getReport()
+    protected function getKvStorage(): KeyValuePersistence
```
gabrielfs7, line 360: - This should be private. Least privilege principle.
- You are forcing the `default_kv` always now, this means that it wont take by configuration. Are you sure this is a good idea?

models/classes/class.Parser.php
```diff
@@ -163,55 +119,62 @@ public function getSource()
...
                 $dom = new DomDocument();
                 $dom->formatOutput = true;
                 $dom->preserveWhiteSpace = false;
-                
+
                 $this->valid = $dom->loadXML($content);
 
-                if ($this->valid && !empty($schema)) {
+                $validationKey = $this->getValidationKey($dom, $schema);
+                $hashedContent = md5(preg_replace('/\s+<responseProcessing\/>/', '', $content));
```
gabrielfs7, line 147: Can you explain why this is needed? If this is important, I recommend to extract the new methods added + this part to a new class with proper name and self-explanatory methods.

gabrielfs7 (changes requested):
Good job! I will need come explanation on how this solution fixed the issue. :) 

Please, consider moving all new code to a proper place where this new requirements understandable for next developers. Also always provide storage by option.

models/classes/class.Parser.php
shpran: Moved to a new class with explanation 😄

models/classes/ItemDomValidation.php
```diff
@@ -0,0 +1,117 @@
...
+    /** @var bool */
+    private $containsAssessmentItem;
+
+    /** @var string */
+    private $validationKey;
+
+    /** @var string */
+    private $hashedContent;
+
+    public function __construct(
```
gabrielfs7, line 48: As much as I prefer dependencies through constructor, if we want to extend this with new options provided by the user, this might be done only using configurable service inheritance.
- Also if seems you will have ton instantiate this service directly on the client class, so not good for testing (cause we do not have the DI).

I wold consider make this a ConfigurableService.

models/classes/class.Parser.php
```diff
@@ -20,114 +20,61 @@
...
 class tao_models_classes_Parser
 {
+    public const SOURCE_FILE = 1;
+
+    public const SOURCE_URL = 2;
+
+    public const SOURCE_STRING = 3;
+
+    // Current file is \oat\oatbox\filesystem\File object
+    public const SOURCE_FLYFILE = 4;
```
gabrielfs7, line 46:
```suggestion
    public const SOURCE_FILE = 1;
    public const SOURCE_URL = 2;
    public const SOURCE_STRING = 3;
    
    /**
     * Current file is \oat\oatbox\filesystem\File object
     */
    public const SOURCE_FLYFILE = 4;
```

Thanks for adding constant visibility. No line break is required unless there is a comment

models/classes/class.Parser.php
```diff
@@ -163,55 +115,57 @@ public function getSource()
...
             try {
                 libxml_use_internal_errors(true);
 
                 $dom = new DomDocument();
                 $dom->formatOutput = true;
                 $dom->preserveWhiteSpace = false;
-                
+
                 $this->valid = $dom->loadXML($content);
+                $itemDomValidation = new ItemDomValidation($dom, $schema, $content);
```
gabrielfs7, line 141: I sent to you a message explaining details about this, but you can replace from line 134 to line 154 to rely only on the validator to do everything. The outcome will be something like this:

```
$errors = $this->getItemDomValidator()->validateSchema($schema, $content);
$this->valid = count($errors) === 0;

if (!$this->valid) {
    $this->addErrors($errors);
}
```

This way the validation is entirely done outside of this class and the other classes do not need to load DOMElement by themselves just to use it. Also you return a collection of errors instead of true/false, which will help us understand the exact error.

models/classes/DomValidator.php
```diff
@@ -0,0 +1,210 @@
...
+            $this->keyValuePersistence = $this->getOption(self::OPTION_KV_PERSISTENCE);
+
+            if ($this->keyValuePersistence === null || !$this->keyValuePersistence instanceof KeyValuePersistence) {
+                $this->keyValuePersistence = $this->getPersistenceManager()->getPersistenceById(
+                    'default_kv'
+                );
+            }
+        }
+
+        return $this->keyValuePersistence;
```
gabrielfs7, line 165: It is safer to inject the persistence id instead of the object, cause on seed files it will not work :/ 

```suggestion
        if (!isset($this->keyValuePersistence)) {
            $this->keyValuePersistence = $this->getPersistenceManager()
                ->getPersistenceById($this->getOption(self::OPTION_KV_PERSISTENCE) ?? 'default_kv');
        }

        return $this->keyValuePersistence;
```

models/classes/DomValidator.php
```diff
@@ -0,0 +1,210 @@
...
+        return $this->getServiceLocator()->get(PersistenceManager::SERVICE_ID);
+    }
+
+    private function setValidationResult(bool $isValid): void
+    {
+        if ($this->assessmentItem !== null) {
+            $ttl = $this->getKeyValuePersistenceTTL();
+
+            $this->getKeyValuePersistence()->set($this->validationKey . '.content', $this->hashedContent, $ttl);
+            $this->getKeyValuePersistence()->set($this->validationKey . '.result', $isValid, $ttl);
```
gabrielfs7, line 179: Does it work for SQL or just for Redis?

gabrielfs7 (commented):
Good job @shpran! This is a tough topic.

I think it is better to have a chat about this, cause on the first test preview we would still have 504 cause test could be too big. Even using cache, I might have a huge booklet and timeout will arrive. 

Suggestion: Maybe we should allow paginated preview for booklets? So we reduce the amount of computation required? Maybe combine the 2 solutions?

I am not sure the TTL will work as expected to others KV storage and if we use Redis, what would be the impact by adding all item content and validation inside of it?

PR #2773, opened 2021-01-07
fix: TR-348 update cognia messages

lecosson:
Related to: https://oat-sa.atlassian.net/browse/TR-348

Requires: related PR's
- https://github.com/oat-sa/extension-tao-mp/pull/513
- https://github.com/oat-sa/extension-tao-mpart/pull/846

To check:
- AC5
- AC7
- AC8


btamas (changes requested):
As we discussed this won't be necessary.

PR #2776, opened 2021-01-11
Fix/updater alarm norification error

hutnikau:
fix of the following fatal at time of update:
```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to oat\oatbox\reporting\Report::filterChildrenByTypes() must be of the type array, null given, called in /Users/aleh/domains/package-tao/tao/models/classes/notifications/AlarmNotificationService.php on line 43 and defined in /Users/aleh/domains/package-tao/generis/common/oatbox/reporting/Report.php:316
```

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -64,7 +64,7 @@ return [
     'label' => 'TAO Base',
     'description' => 'TAO meta-extension',
     'license' => 'GPL-2.0',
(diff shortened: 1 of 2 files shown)


AndreyNegrov (approved):
looks good

PR #2802, opened 2021-02-25
fix: define ROOT_URL constant at the beginning of the installation

shpran:
**Related PR:** [oat-sa/generis#874](https://github.com/oat-sa/generis/pull/874)

**Changes:** define the ROOT_URL constant with the right value at the beginning of the installation, because it can be used in manifest files.

### DIFF
--- a/install/class.Installator.php
+++ b/install/class.Installator.php
@@ -103,12 +103,17 @@ class tao_install_Installator
              */
             $this->log('i', "Checking install data");
             self::checkInstallData($installData);
-            
+
             $this->log('i', "Starting TAO install");
-            
+
             // Sanitize $installData if needed.
             if (!preg_match("/\/$/", $installData['module_url'])) {
                 $installData['module_url'] .= '/';
+            }
+
+            // Define the ROOT_URL constant if not defined (can be used in manifest files)
+            if (!defined('ROOT_URL')) {
+                define('ROOT_URL', $installData['module_url']);
             }
 
             if (isset($installData['extensions'])) {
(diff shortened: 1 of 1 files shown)


install/class.Installator.php
```diff
@@ -103,14 +103,19 @@ public function install(array $installData, callable $callback = null)
...
-            
+
             // Sanitize $installData if needed.
             if (!preg_match("/\/$/", $installData['module_url'])) {
                 $installData['module_url'] .= '/';
             }
 
+            // Define the ROOT_URL constant if not defined (can be used in manifest files)
+            if (!defined('ROOT_URL')) {
+                define('ROOT_URL', $installData['module_url']);
```
gabrielfs7, line 116: So at this point the `config/generis.conf.php` was not created? I am wondering why the other constants there are not an issue too.

Maybe you have a corrupted environment?

mike-ionut-mihai-sandu-tao: this is happening for multiple developers,  there was a big thread inside software-engineers  channel

mike-ionut-mihai-sandu-tao (approved):

install/class.Installator.php
gabrielfs7: Just to let you know that **any constant** inside the manifest.php that was not declared before **will have the same issue** now, so I think we are maybe fixing the wrong place.

Before we read the php file as a simple simple file (which I am not saying is good), but now we use require. 

https://github.com/oat-sa/generis/commit/85812fbbd6733c31ec31a736ea79cf61109fbc83#diff-5848b98eb9a5d8c8f1f36cacfa3329f1619a1f6c8a2598c3e48aeef98b2f3592R321

So now any constant can become an issue. I am just warning that this can lead for other problems in the future, so might be that the fix is to have all the constants declared by generis, before reading any other manifest files :)

hutnikau (approved):

PR #2803, opened 2021-02-25
fix: define constant variables at the beginning of the installation script

shpran:
Create `config/generis.conf.php` file and define all constants at the beginning of the installation script.
In this case we will be sure that all necessary constants are defined and can be used in all parts of the installation script.


install/class.Installator.php
```diff
@@ -122,19 +120,49 @@ public function install(array $installData, callable $callback = null)
...
+                'ROOT_PATH' => $this->options['root_path'],
+                'FILES_PATH' => $installData['file_path'],
+                'ROOT_URL' => $installData['module_url'],
+                'DEFAULT_LANG' => $installData['module_lang'],
+                'DEBUG_MODE' => $installData['module_mode'] === 'debug',
+                'TIME_ZONE' => $installData['timezone'],
+                'DEFAULT_ANONYMOUS_INTERFACE_LANG' => $installData['anonymous_lang'] ?? $installData['module_lang'],
+            ];
+
+            $configWriter = new ConfigWriter(
```
gabrielfs7, line 141: 2 small comments:

- Why not use `tao_install_utils_ConfigWriter` anymore? Cause before it was enough.
- Since you moved all this code, should we put this either on a private method or other class? Wdyt?

install/utils/ConfigWriter.php
```diff
@@ -0,0 +1,126 @@
...
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\install\utils;
+
+use tao_install_utils_Exception as InstallUtilsException;
+
+class ConfigWriter
```
gabrielfs7, line 27: Why `tao_install_utils_ConfigWriter` is not used anymore?

install/utils/class.ConfigWriter.php
```diff
@@ -1,116 +0,0 @@
...
- * The ConfigWriter class enables you to create config file from samples
- * and to write the constants inside.
- *
- * @author Bertrand CHEVRIER <bertrand.chevrier@tudor.lu>
- * @access public
- * @package tao
-
- *
- */
-class tao_install_utils_ConfigWriter
```
gabrielfs7, line 33: If you delete this class it is a breaking change. Could you do a delegate + deprecate here instead?

gabrielfs7 (commented):
Good job, left some comments: :) 

install/utils/class.ConfigWriter.php
shpran: I don't think that we should mark it as deprecated. This class used only in one place, so we can safely replace this with a new one.

install/utils/ConfigWriter.php
shpran: New class base on `tao_install_utils_ConfigWriter`.

I want to normalize file name, class name and namespace + I refactored it a bit:

**Previous:**
1. Copy sample to `config/generis.conf.php`;
2. Get content of `config/generis.conf.php`;
3. Replace content with the necessary data;
4. Save this content.

**New:**
1. Get sample content;
2. Replace content with the necessary data;
3. Define constants during replacement (if not defined);
3. Save as `config/generis.conf.php`.

install/class.Installator.php
shpran: 1. https://github.com/oat-sa/tao-core/pull/2803#discussion_r583303578
2. Do you mean something like:
 `private function generateGenerisConfig(array $constants = []): void {}`
 or
 `private function generateGenerisConfig(array $installationData = []): void {}`
 ?

install/utils/class.ConfigWriter.php
gabrielfs7: If you are 100% sure is used only here, but we should always go for deprecation/delegation, never remove files till we release a major / breaking change. 

Do not forget that we have custom extensions and other versions of tao around.

It is a good practice to deprecate/delegate specially when it is an open/source project and not an internal private SaaS application.

install/class.Installator.php
gabrielfs7: Probably only the second one works for what I suggested :)

install/utils/ConfigWriter.php
gabrielfs7: got it after, thanks! :)

PR #2808, opened 2021-03-02
fix:missing user and password in command to register webhoook

bartlomiejmarszal:
Fixed missing user and password for basic authentication

https://oat-sa.atlassian.net/browse/AUT-16

### DIFF
--- a/scripts/tools/RegisterEventWebhook.php
+++ b/scripts/tools/RegisterEventWebhook.php
@@ -37,6 +37,7 @@ use oat\tao\model\webhooks\WebhookRegistryManagerInterface;
 use oat\tao\model\webhooks\WebhookSerializableEventInterface;
 use oat\taoPublishing\model\publishing\event\RemoteDeliveryCreatedEvent;
 use ReflectionClass;
+use function Webmozart\Assert\Tests\StaticAnalysis\implementsInterface;
 
 class RegisterEventWebhook extends ScriptAction
 {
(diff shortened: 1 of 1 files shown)


scripts/tools/RegisterEventWebhook.php
```diff
@@ -37,6 +37,7 @@
 use oat\tao\model\webhooks\WebhookSerializableEventInterface;
 use oat\taoPublishing\model\publishing\event\RemoteDeliveryCreatedEvent;
 use ReflectionClass;
+use function Webmozart\Assert\Tests\StaticAnalysis\implementsInterface;
 
 class RegisterEventWebhook extends ScriptAction
```
ktsirkou, line 42: Based on the name I would expect this to be generic EventWebhook registration. 
Actually it's not, as it registers only `RemoteDeliveryCreatedEvent`  ([L104]( https://github.com/oat-sa/tao-core/pull/2808/files#diff-a9266bbd193492cbc15fc5540ea50546c73fb0181500254f61895998262805a3L104 )
and the webhook config associates only the id `remoteDeliveryWebHook` ([L178](https://github.com/oat-sa/tao-core/pull/2808/files#diff-a9266bbd193492cbc15fc5540ea50546c73fb0181500254f61895998262805a3R178))

Either we have to rename the class to reflect that is only registering RemoteDeliverCreatedEvent or make it parametric

PR #2809, opened 2021-03-02
fix: remove unknown template warning

hutnikau:


models/classes/theme/ConfigurableTheme.php
```diff
@@ -158,7 +158,6 @@ public function getTemplate($id, $context = Theme::CONTEXT_BACKOFFICE)
                 $template = Template::getTemplate('blocks/login.tpl', 'tao');
                 break;
             default:
-                \common_Logger::w('Unknown template ' . $id);
```
alroniks, line 161: I would suggest replacing it by debug, instead of just removing it. Because it may be useful when creating custom theming for clients.

hutnikau: done

alroniks (approved):

tikhanovichA (approved):

PR #2813, opened 2021-03-11
feat: show 404 page when unable to resolve url

hutnikau:
Tao logs are full of `Unable to resolve` errors which actually not errors but expected behaviour when user (or bot/parser) make request to non-existent page. 404 error page should be shown in that case.
Screenshot 2021-03-11 at 10 56 33

Log level needs to be changed to debug and instead of redirect to login page 404 error page should be shown

How to test:

Open any inexistent Tao page: `<tao_url>/foobar`

### DIFF
--- a/models/classes/controllerMap/ActionNotFoundException.php
+++ b/models/classes/controllerMap/ActionNotFoundException.php
@@ -29,5 +29,8 @@ namespace oat\tao\model\controllerMap;
  */
 class ActionNotFoundException extends \common_Exception
 {
-    
+    public function getSeverity()
+    {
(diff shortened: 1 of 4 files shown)


models/classes/controllerMap/ActionNotFoundException.php
```diff
@@ -29,5 +29,8 @@
  */
 class ActionNotFoundException extends \common_Exception
 {
-    
+    public function getSeverity()
+    {
+        return \common_Logger::DEBUG_LEVEL;
```
alroniks, line 34: Why do not use import instead of root namespace?

alroniks (approved):

PR #2827, opened 2021-03-22
ci: add release and conventional-commit actions

hutnikau:


krampstudio (approved):
It looks good to me.
But since this may have an important impact on developers, we should decide when we want to roll this out. To prevent any inconvenience and test it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+          mkdir -p tao/views/locales/en-US/
+          echo '{"serial":"9","date":1615820392,"version":"3.3.0-9","translations":{}}' > tao/views/locales/en-US/messages.json
+          touch index.php
+          mkdir -p config/
+
+       # Install node
+      - name: Use Node.js 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
```
Cicatrice, line 59: you may move such version number to env variable in beginning of file for maintaining purpose

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      # Install PHP, composer v2
+      - name: Setup PHP,
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.2'
```
Cicatrice, line 22: you may move such version number to env variable in beginning of file for maintaining purpose

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          COMPOSER=composer-release.json composer install --no-dev --no-interaction --prefer-source
+
+      # Prepare to release
+      - name: Prepare to release
+        run: |
+          mkdir -p taoQtiItem/views/js/mathjax/
+          touch taoQtiItem/views/js/mathjax/MathJax.js
+          mkdir -p tao/views/locales/en-US/
+          echo '{"serial":"9","date":1615820392,"version":"3.3.0-9","translations":{}}' > tao/views/locales/en-US/messages.json
```
Cicatrice, line 51: not sure how those data are generated, but maybe they may belong in an env var too

Cicatrice (commented):
I see no major pitfall for the moment to use such workflow. Maybe @gitromba  has more comment on it but he is off for the week.

We didnt work yet on a support for CG application for our CICD chain, and we may have to change later how this release process is working, still to be discussed.

PR #2831, opened 2021-03-23
ci: remove pr-merge check

hutnikau:


krampstudio (approved):
It looks correct even if I can't test it

PR #2832, opened 2021-03-24
ci: discard changes in repository before release

hutnikau:
Automated release of the extensions fails because of the changes in repository:
```
➡ Checking extension status
The extension tao has local changes, please clean or stash them before releasing
```
The changes are:
```
Changes not staged for commit:
	modified:   views/package-lock.json
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
         run: |
           php -r '$composerArray = json_decode(file_get_contents("./composer.json"), true);
             $composerArray["require"][$composerArray["name"]] = "dev-develop";
-            $composerArray["repositories"] = is_array($composerArray["repositories"]) ? $composerArray["repositories"] : [];
+            $composerArray["repositories"] = array_key_exists("repositories", $composerArray) ? $composerArray["repositories"] : [];
             $composerArray["repositories"][] = [
               "type" => "vcs",
               "url" => "https://github.com/".$composerArray["name"].".git"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -71,5 +71,7 @@ jobs:
           git config --global user.name github-actions
           git config --global user.email github-actions@github.com
           git remote set-url --push origin https://${GITHUB_TOKEN}@github.com/$REPO_NAME.git
+          git add .
+          git reset --hard
```
jsconan, line 75: Is it not better to do `git checkout .` instead?

hutnikau: this will not discard already indexed changes (changes to be committed)

krampstudio:
This should be solved by https://github.com/oat-sa/composer-npm-bridge/pull/17

.github/workflows/release.yml
jsconan: Indexed changes can be reverted by a `git reset`, then the `git checkout .` will rollback any changes to the last commit.
I wonder if the `--hard` mode is the best fit, especially since it may rewrite history.
But anyway, it does the job as well.

.gitignore
```diff
@@ -11,6 +11,7 @@ views/js/test/.htaccess
 /views/css/tao-3.css.map.orig
 /views/css/tao-3.css.orig
 /views/index.html
+/views/package-lock.json
```
jsconan, line 14: Not sure it should be discarded. This will have impact on deploys, especially when https://github.com/oat-sa/composer-npm-bridge/pull/17 will be merged.

krampstudio: yes, it's necessary the problem comes from the bridge that shouldn't update it

krampstudio (approved):

jsconan (approved):
Looks good

PR #2865, opened 2021-04-07
[Fix - ADF 193] UDIR Pre-Prod: Console errors after removing all 'Grant' permissions from the object in Access Control panel.

habemuscode:
**Issue**

https://oat-sa.atlassian.net/browse/ADF-193

**Description**

When user with Global Manager role (f. e. Administrator) removes all the 'Grant' permissions from any object in Back Office, there are 2 errors in Browser console appear right after removing the last 'Grant' permission. Also there's an uninformative error message randomly appears after the saving attempt (please see attached images).

**Steps to reproduce**

- Log in as admin and go to Items tab.
- Choose any class or item and click 'Access Control' button.
- Remove checkboxes from 'Grant' permissions and try to click 'Save'.

**Actual result**

2 errors appear in Browser console right after removing the last 'Grant' permission. Also, there's an uninformative error message pops up after clicking 'Save' button (occurs randomly).

**Expected result**

No errors appear in browser console, clicking 'Save' button causes the report to appear in Background tasks with the explanation that saving demands at least one role or user to have 'Grant' permission.


habemuscode:
Deployed temporarily here: http://test-bosa.playground.kitchen.it.taocloud.org:49151 Ping me if not solved, maybe overwrited deployment

views/js/lib/popper/tooltip.js
```diff
@@ -443,19 +445,21 @@ var Tooltip = function () {
...
+                return;
+              }
+              var popper = _this3.popperInstance.popper;
+              if (reference.contains(e.target) || popper.contains(e.target)) {
+                return;
+              }
+              func(e);
+            }, true);
+          }
         }
```
andrey-ponamarev, line 463: I can be mistaken, but it doesn't look like a library issue, I feel the issue could be on our side, and it smells like a memory leak.
Please try to avoid modifications of external libraries sources instead check if the issue exist and fixed there, then update the whole [library source](https://github.com/popperjs/popper-core/blob/v1.16.1/packages/tooltip/src/index.js#L351).

andrey-ponamarev (changes requested):

views/js/lib/popper/tooltip.js
habemuscode: Thanks!

Change all popper library will require a more extensive QA test that we don't have planned. Also, we need to check if all methods continue to exist or not or have modifications in params.

I can create another issue for this and go forward with this small workaround for now

https://oat-sa.atlassian.net/browse/ADF-262

404passionFruit: Maybe leave a comment with a hint to the issue and the way this should be handled when we have more time?

habemuscode: I already comment the new issue link: https://oat-sa.atlassian.net/browse/ADF-262

404passionFruit (approved):
Approved as temp fix

bziondik (changes requested):
The root of issue in our code
taoDacSimple/views/js/controller/admin/index.js
we don't have `$submitter` , maybe markup was changed
image
Please, fix `$submitter`

habemuscode:
To avoid modifying the external library I will follow @bziondik suggestion in this other PR: https://github.com/oat-sa/extension-tao-dac-simple/pull/164

PR #2883, opened 2021-04-19
Feature/ADF-154/Add ActionAccessControl service and tool to add/remove action permissions

shpran:
**Relates to:** _[ADF-154](https://oat-sa.atlassian.net/browse/ADF-154)_

**Changes:**
* `oat\tao\model\accessControl\ActionAccessControl` class added. Allow adding `READ`, `WRITE`, `GRANT` permissions to roles for specific controller actions;
* `oat\tao\scripts\tools\accessControl\SetRolesAccess` tool added. Allow adding/removing rules and/or permissions for `oat\tao\model\accessControl\ActionAccessControl` config;
* `oat\tao\test\unit\accessControl\ActionAccessControlTest` test added.


models/classes/accessControl/ActionAccessControl.php
```diff
@@ -0,0 +1,176 @@
...
+                unset($permissions[$controller]);
+            }
+        }
+
+        $this->setOption(self::OPTION_PERMISSIONS, $permissions);
+    }
+
+    public function hasReadAccess(string $controller, string $action, ?User $user = null): bool
+    {
+        return $this->hasAccess([self::READ, self::WRITE, self::GRANT], $controller, $action, $user);
```
gabrielfs7, line 120: Should not be like this?

```suggestion
        return $this->hasAccess([self::READ], $controller, $action, $user);
```

models/classes/accessControl/ActionAccessControl.php
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    public function hasWriteAccess(string $controller, string $action, ?User $user = null): bool
+    {
+        return $this->hasAccess([self::WRITE, self::GRANT], $controller, $action, $user);
```
gabrielfs7, line 125: Should not be like this?

```suggestion
        return $this->hasAccess([self::WRITE], $controller, $action, $user);
```

models/classes/accessControl/ActionAccessControl.php
shpran: Nope.
A user with WRITE access should be able to READ - he needs to view data which he wants to change.
Same with GRANT.

models/classes/accessControl/ActionAccessControl.php
shpran: https://github.com/oat-sa/tao-core/pull/2883#discussion_r616088763

scripts/tools/accessControl/ApplyRules.php
```diff
@@ -0,0 +1,111 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\scripts\tools\accessControl;
+
+use oat\oatbox\extension\script\ScriptAction;
+use oat\tao\model\accessControl\func\AclProxy;
+use oat\tao\model\accessControl\func\AccessRule;
+
+class ApplyRules extends ScriptAction
```
gabrielfs7, line 29:
```suggestion
class ApplyActionRules extends ScriptAction
```

Do we plan to have the ACL for other purposes than Action here?

scripts/tools/accessControl/SetActionAccessPermissions.php
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\scripts\tools\accessControl;
+
+use oat\oatbox\extension\script\ScriptAction;
+use oat\tao\model\accessControl\ActionAccessControl;
+
+class SetActionAccessPermissions extends ScriptAction
```
gabrielfs7, line 28: Do not you think those 2 Script are kind of doing the same but with different input? Maybe they are the same?

https://github.com/oat-sa/tao-core/pull/2883/files#diff-250d1b7911b70c3d82945c816b77a615f3d05616ce4fdcfda0e05b98b4fe3736R29

scripts/tools/accessControl/ApplyRules.php
shpran: This tool is able to apply any mask: whole extension, specified controller or specified action.
I don't think we should rename it.

models/classes/accessControl/ActionAccessControl.php
gabrielfs7: But the method is `hasReadAccess`

shpran: It's correct. If the user has WRITE access - he is also able to READ, so he has READ access

scripts/tools/accessControl/SetActionAccessPermissions.php
shpran: Nope.
First script is just a wrapper for `AplyRule/RevokeRule`, which allow to apply/revoke multiple rules (helps us to avoid code duplication in migrations).

This one: to add/remove (configure) our "custom" permissions: READ, WRITE, GRANT

scripts/tools/accessControl/ApplyRules.php
gabrielfs7: Ok understood :) 

Maybe we could also write an example with this case in the php docs then? Wdyt?

test/unit/accessControl/ActionAccessControlTest.php
```diff
@@ -0,0 +1,201 @@
...
+    {
+        return $this->actionAccessControl->hasGrantAccess(
+            self::TEST_CONTROLLER,
+            self::TEST_ACTION,
+            $this->user
+        );
+    }
+}
+
+class User extends TestUser
```
gabrielfs7, line 188: It is not good practice to have 2 classes in the same file. You can create either:

- An anonymous class
- Create a mock.

scripts/tools/accessControl/SetActionAccessPermissions.php
```diff
@@ -0,0 +1,93 @@
...
+    public const OPTION_ACTION = 'action';
+    public const OPTION_PERMISSIONS = 'permissions';
+
+    public const ACTION_ADD = 'add';
+    public const ACTION_REMOVE = 'remove';
+
+    /**
+     * @example [
+     *     '--action', 'add',
+     *     '--permissions', [
```
gabrielfs7, line 39: So in the command line we should pass a string of a php array as parameter? What do you think of providing a .json file with the permissions map?

models/classes/accessControl/ActionAccessControl.php
gabrielfs7: Lets check together. I think there is some confusion here :)

models/classes/accessControl/ActionAccessControl.php
gabrielfs7: So you are saying that if I want to check is I have READ access, I should have `GRANT`, 'READ' or `WRITE`?

I think I got it now :), For the scope we have now it is OK, but we will have different permission for NMC like `UPDATE`, 'CREATE' and 'DELETE' so we will have to change this logic here. Cause the permission will need to be exact the same.

models/classes/accessControl/ActionAccessControl.php
gabrielfs7: https://github.com/oat-sa/tao-core/pull/2883/files#r616375379

models/classes/accessControl/ActionAccessControl.php
shpran: If it will be required - it can easily fixed 🙂 
Just change a bit of method declaration (first parameter `array` -> `string`)
and change comparison: `in_array($permission, $allowedPermissions, true)` -> `$permissions === allowedPermission`

scripts/tools/accessControl/SetActionAccessPermissions.php
shpran: It's a good idea, will add this 🙂 
Thanks

scripts/tools/accessControl/SetActionAccessPermissions.php
gabrielfs7: I understood :) 

But now, imagine that I want to start adding/removing roles for users in the platform. As an API user I will have to do it in 2 different ways and in two places.

What I was suggesting is to have a format (contract - input) where we can apply both. One "idead" would be something like this:

```
[
    'role1' => [
        [
             'ext' => 'extension', 
             'mod' => 'controller', (optional)
             'act' => 'action',  (optional)
             'permissions' => (optional) [
                  'GRANT',
                  'READ',
                  'WRITE',
              ]
    ],
```

Also, I think we should write a doc for this :)

README.md
```diff
@@ -96,4 +96,44 @@ $permissionChecker->hasReadAccess('resourceId');
...
+
+### Config example
+```json
+{
+    "rules": {
+        "role": [
+            {
+                "ext": "ext",
+                "mod": "mod",
+                "act": "act"
```
gabrielfs7, line 126:
```suggestion
                "ext": "extensionIdentifier",
                "mod": "ActionControllerName",
                "act": "actionMethodName"
```

Just to be clear what should be the parameters:

@shpran

models/classes/accessControl/ActionAccessControl.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public function hasGrantAccess(string $controller, string $action, ?User $user = null): bool
+    {
+        return $this->hasAccess([self::GRANT], $controller, $action, $user);
+    }
+
+    private function hasAccess(array $allowedPermissions, string $controller, string $action, ?User $user = null): bool
```
gabrielfs7, line 134: We could cache this verification, so no needed to run over this loop every time. Wdyt?

scripts/tools/accessControl/SetRolesAccess.php
```diff
@@ -0,0 +1,184 @@
...
+     * @example config
+     * [
+     *     '--config', [
+     *         'rules' => [
+     *             'role1' => [
+     *                 ['ext' => 'extension', 'mod' => 'controller', 'act' => 'action'],
+     *             ],
+     *             'role2' => [
+     *                 ['ext' => 'extension', 'mod' => 'controller', 'act' => 'action'],
+     *             ],
```
gabrielfs7, line 70: Suggestion to improve readability.

```suggestion
     *             'role1' => [
     *                 [
     *                     'ext' => 'extension', 
     *                     'mod' => 'controller', 
     *                      'act' => 'action'
     *                 ],
     *             ],
     *             'role2' => [
     *                 [
     *                     'ext' => 'extension', 
     *                     'mod' => 'controller', 
     *                     'act' => 'action'
     *                 ],
     *             ],
```

gabrielfs7 (approved):
LGTM, good job!

I just missed a cache layer on the `ActionAccessControl` to avoid future performance issues.

models/classes/accessControl/ActionAccessControl.php
boajer: I believe we already have approach to cache ACL on the actions level `\oat\funcAcl\helpers\CacheHelper::getControllerAccess`

boajer (approved):

PR #2901, opened 2021-04-27
fix/AUT-396/update @oat-sa/tao-core-ui

bziondik:
Related to: [AUT-396](https://oat-sa.atlassian.net/browse/AUT-396)
Related to: https://github.com/oat-sa/tao-core-ui-fe/pull/279

**Description**
Added deleted node to `deleted.deleter` event. It is important because this event fires on the parent node, and we should have a possibility to detect which node was removed.


AnastasiaArcadia (approved):

AntonNeviarouski:
QA test passed on http://test-hanna.playground.kitchen.it.taocloud.org:40141/

habemuscode (approved):

PR #2926, opened 2021-05-25
Advanced search: Add criteria option is unclickable after opening Modal view.

habemuscode:
**Ticket**

https://oat-sa.atlassian.net/browse/ADF-343

**Description**

The issue is randomly reproduced (about 1 of 10 times) - when the user opens the Modal view by clicking the Search icon, the "add criteria" button is unclickable and the "+" icon on the left of it is replaced with endlessly turning spinner. After clicking the 'Clear' button the user is able to click on the button and to see a dropdown list but the spinner is still there.

**Steps to reproduce**

- Open any tab.
- Click the Search icon to open Modal view.

**Actual result**

About 1 of 10 times the 'add criteria' button becomes unclickable and has a constantly turning spinner icon on its left.

**Expected result**

There's no case when 'add criteria' becomes unclickable right after opening Modal view.


boajer (approved):
LGFM

PR #2981, opened 2021-06-28
fix: removing namespace check

mike-ionut-mihai-sandu-tao:
removing namespace check


julien-sebire (dismissed):

julien-sebire (changes requested):
Don't do this at all. No change needed in the code. Please see https://oatsa.slack.com/archives/C02739D5RUG/p1624906798067000.

PR #2984, opened 2021-07-01
feat: allow to disable actions defined in env variables

bartlomiejmarszal:
This will allow to define actions that has to be disable for tao instance

https://oat-sa.atlassian.net/browse/AUT-553

### DIFF
--- a/models/classes/featureFlag/ActionDisabledFeatureFlag.php
+++ b/models/classes/featureFlag/ActionDisabledFeatureFlag.php
@@ -1,0 +1,39 @@
+<?php
+
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
+ */
+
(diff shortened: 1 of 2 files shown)


models/classes/featureFlag/ActionDisabledFeatureFlag.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace oat\tao\model\featureFlag;
+
+use oat\oatbox\service\ConfigurableService;
+
+class ActionDisabledFeatureFlag extends ConfigurableService
+{
+    private const DISABLED_ACTIONS = 'DISABLED_ACTIONS';
+
+    public function isActionDisabled(string $feature): bool
```
mike-ionut-mihai-sandu-tao, line 31: as i remember last time we spoke about this with @gabrielfs7 everything is disabled unless we set it as active 
so creating a disabled list makes sense in a way that we want everything on unless we say so , 
so we don't have to create a list with all existing actions
let's have a call that will include Gabriel and Misha and white board this decision

mike-ionut-mihai-sandu-tao (approved):

mike-ionut-mihai-sandu-tao (commented):
i think we need some tests or at least a technical dept ticket

models/classes/featureFlag/ActionDisabledFeatureFlag.php
gabrielfs7: So as I understood you want to disable actions from the Menu based on another env var.

Just a remind that they might still be available from mother parts. Example, disabling groups from menu do not hide them from delivery edit page. Is this ok? Might complicate even more the management...

Why no use the already existing https://github.com/oat-sa/tao-core/blob/master/models/classes/featureFlag/FeatureFlagChecker.php#L29
?

You could have feature flags called `FEATURE_FLAG_DISABLE_ACION_{ACTION_NAME}`

So you can reuse in another places.

But to be honest, I would like to understand why not manage this with ACL? If there is a new section not accessible by default user, then we would not map it on ACL. 

Wdyt?

mike-ionut-mihai-sandu-tao: we were able to split the ticket related to this feature and we will have a proper implementation in the next sprit that will cover a bit more cases

mike-ionut-mihai-sandu-tao:
do we want to still keep this ? as i see we should go for a more general solution

PR #2986, opened 2021-07-02
fix: update manifest with new sprint number

bartlomiejmarszal:
This commit will increase sprint version in manifest file


wazelin (approved):

mike-ionut-mihai-sandu-tao (approved):

jsconan (approved):
The change look ok
But usually, we also update the translations.

PR #3003, opened 2021-07-15
fix: options cast to array

bartlomiejmarszal:
fix: Introduced service when no option is configured causing fatal error

### DIFF
--- a/models/classes/action/ActionBlackList.php
+++ b/models/classes/action/ActionBlackList.php
@@ -36,9 +36,9 @@ class ActionBlackList extends ConfigurableService
 
     public function isDisabled(string $action): bool
     {
-        $disabledActionsMap = $this->getOption(self::OPTION_DISABLED_ACTIONS_FLAG_MAP);
+        $disabledActionsMap = $this->getOption(self::OPTION_DISABLED_ACTIONS_FLAG_MAP, []);
 
-        return array_search($action, (array) $this->getOption(self::OPTION_DISABLED_ACTIONS)) !== false ||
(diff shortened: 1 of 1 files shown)


models/classes/action/ActionBlackList.php
```diff
@@ -38,7 +38,7 @@ public function isDisabled(string $action): bool
     {
         $disabledActionsMap = $this->getOption(self::OPTION_DISABLED_ACTIONS_FLAG_MAP);
 
-        return array_search($action, $this->getOption(self::OPTION_DISABLED_ACTIONS)) !== false ||
+        return array_search($action, (array) $this->getOption(self::OPTION_DISABLED_ACTIONS)) !== false ||
```
gabrielfs7, line 41:
```suggestion
        return array_search($action, (array) $this->getOption(self::OPTION_DISABLED_ACTIONS, [])) !== false ||
```

models/classes/action/ActionBlackList.php
```diff
@@ -38,7 +38,7 @@ public function isDisabled(string $action): bool
     {
         $disabledActionsMap = $this->getOption(self::OPTION_DISABLED_ACTIONS_FLAG_MAP);
```
gabrielfs7, line 39:
```suggestion
        $disabledActionsMap = $this->getOption(self::OPTION_DISABLED_ACTIONS_FLAG_MAP, []);
```

gabrielfs7 (changes requested):
Use default value too, please

gabrielfs7 (approved):

PR #3023, opened 2021-07-23
feat: add method to receive User UI language code

shpran:
**Relates to:** _[ADF-167](https://oat-sa.atlassian.net/browse/ADF-167)_

**Changes:** added new helper method to get user language interface code.

---

**Companion PR:** https://github.com/oat-sa/extension-tao-udir/pull/129

### DIFF
--- a/helpers/UserHelper.php
+++ b/helpers/UserHelper.php
@@ -26,8 +26,6 @@ use oat\generis\model\GenerisRdf;
 use oat\generis\model\OntologyRdfs;
 use oat\oatbox\service\ServiceManager;
 use oat\oatbox\user\User;
-use core_kernel_classes_Resource;
-use core_kernel_users_GenerisUser;
 use Jig\Utils\StringUtils;
 use oat\oatbox\user\UserService;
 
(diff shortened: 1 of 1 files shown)


helpers/UserHelper.php
```diff
@@ -153,13 +151,27 @@ public static function getUserName(User $user, $defaultToLabel = false)
...
             $userName = self::getUserLabel($user);
         }
 
         return $userName;
     }
+
+    /**
+     * @param User|string $user
+     *
+     * @return string
```
boajer, line 167:
```suggestion
```

helpers/UserHelper.php
```diff
@@ -153,13 +151,27 @@ public static function getUserName(User $user, $defaultToLabel = false)
...
+     */
+    public static function getUserUILanguageCode($user): string
+    {
+        if (is_string($user)) {
+            $user = self::getUser($user);
+        }
+
```
gabrielfs7, line 174: Do you think we could have strict types as type hint? Mixed type generally opens a precedent to increase coupling and difficult to maintain / refactor the method after.

```suggestion
    public static function getUserUILanguageCode(string $userId): string
    {
        $user = self::getUser($userId);

```

gabrielfs7: Do you think we should have this type of logic here instead? 

https://github.com/oat-sa/generis/blob/master/common/oatbox/user/UserLanguageService.php

gabrielfs7 (commented):

PR #3045, opened 2021-08-07
Feature/ADF-421/Extend remote list with extra optional parameter to define dependency field

shpran:
**Relates to:** _[ADF-421](https://oat-sa.atlassian.net/browse/ADF-421)_

- [x] Extend remote list with extra optional parameter to define dependency field
- [x] Add and configure feature flag for this feature

---

**Companion PRs:**
* https://github.com/oat-sa/extension-tao-backoffice/pull/107
* https://github.com/oat-sa/extension-tao-item/pull/501


models/classes/Context/AbstractContext.php
```diff
@@ -0,0 +1,74 @@
...
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\Context;
+
+use InvalidArgumentException;
+
+abstract class AbstractContext
```
gabrielfs7, line 27:
```suggestion
abstract class AbstractContext implements ContextInterface
```

models/classes/Context/AbstractContext.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function getParameter(string $parameter, $default = null)
+    {
+        $this->checkParameterSupport($parameter);
+
+        return $this->parameters[$parameter] ?? $default;
+    }
+
+    public function setParameter(string $parameter, $parameterValue): void
```
gabrielfs7, line 48: Shouldn't it also be part of the interface?

models/classes/Lists/Business/Domain/RemoteSourceContext.php
```diff
@@ -0,0 +1,71 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\model\Lists\Business\Domain;
+
+use InvalidArgumentException;
+use oat\tao\model\Context\AbstractContext;
+use oat\tao\model\Context\ContextInterface;
+
+class RemoteSourceContext extends AbstractContext implements ContextInterface
```
gabrielfs7, line 29: Make the abstraction implement it, please :) 

```suggestion
class RemoteSourceContext extends AbstractContext
```

models/classes/Lists/Business/Domain/RemoteSourceContext.php
```diff
@@ -0,0 +1,71 @@
...
+            self::PARAM_SOURCE_URL,
+            self::PARAM_URI_PATH,
+            self::PARAM_LABEL_PATH,
+            self::PARAM_DEPENDENCY_URI_PATH,
+            self::PARAM_PARSER,
+            self::PARAM_JSON,
+        ];
+    }
+
+    protected function validateParameter(string $parameter, $parameterValue): void
```
gabrielfs7, line 50: It is missing validate the other parameters, like: PARAM_URI_PATH, PARAM_PARSER, etc...

models/classes/Lists/Business/Domain/RemoteSourceContext.php
```diff
@@ -0,0 +1,71 @@
...
+    public const PARAM_JSON = 'json';
+
+    protected function getSupportedParameters(): array
+    {
+        return [
+            self::PARAM_SOURCE_URL,
+            self::PARAM_URI_PATH,
+            self::PARAM_LABEL_PATH,
+            self::PARAM_DEPENDENCY_URI_PATH,
+            self::PARAM_PARSER,
```
gabrielfs7, line 45: Will it be a parser service as reference? Should it be here?

models/classes/Lists/Business/Domain/Value.php
```diff
@@ -44,19 +44,28 @@ class Value implements JsonSerializable
...
     public function __construct(?int $id, string $uri, string $label)
     {
-        $this->id          = $id;
-        $this->uri         = $uri;
-        $this->label       = $label;
-        $this->originalUri = null === $id ? null : $uri;
+        $this->id = $id;
+        $this->uri = $uri;
+        $this->label = $label;
+
```
gabrielfs7, line 55: Why separate line?

```suggestion
```

models/classes/Lists/Business/Domain/Value.php
```diff
@@ -84,14 +93,25 @@ public function setLabel(string $label): self
...
 
-    public function hasChanges(): bool
+    public function getDependencyUri(): ?string
     {
-        return $this->hasChanges;
+        return $this->dependencyUri;
     }
 
-    public function getLabel(): string
+    public function setDependencyUri(?string $dependencyUri): self
```
gabrielfs7, line 101: Why allow to set it to null back again? Does it make sense to happen? This is a VO, so in theory is should not even change the state.

models/classes/Lists/Business/Service/RemoteSource.php
```diff
@@ -45,13 +46,32 @@ public function __construct(array $parsers, Client $client = null)
...
+            RemoteSourceContext::PARAM_URI_PATH => $uriPath,
+            RemoteSourceContext::PARAM_LABEL_PATH => $labelPath,
+            RemoteSourceContext::PARAM_PARSER => $parser,
+        ]);
+
+        yield from $this->fetchContext($context);
+    }
 
-        $body = json_decode((string)$response->getBody(), true);
+    public function fetchContext(RemoteSourceContext $context): Traversable
```
gabrielfs7, line 64: 👏 👏 👏  🚀

models/classes/Lists/Business/Service/RemoteSourceParserInterface.php
```diff
@@ -14,27 +15,21 @@
...
-     * @param string $labelRule
-     *
-     * @return iterable|Value[]
-     *
-     * @throws RuntimeException
+     * @deprecated Use $this->iterateContext()
      */
     public function iterate(array $json, string $uriRule, string $labelRule): iterable;
+
+    public function iterateContext(ContextInterface $context): iterable;
```
gabrielfs7, line 34: You are not actually iterating the context, but based on it. Wdyt?

```suggestion
    public function iterateByContext(ContextInterface $context): iterable;
```

models/classes/Lists/Business/Service/RemoteSource.php
gabrielfs7: Cause you are fetching the resources by context, not fetching the context. Wdyt?

```suggestion
    public function fetchByContext(RemoteSourceContext $context): Traversable
```

models/classes/Lists/Business/Service/RemoteSourceJsonPathParser.php
```diff
@@ -14,48 +15,76 @@
...
+        yield from $this->iterateContext(
+            new RemoteSourceContext([
+                RemoteSourceContext::PARAM_JSON => $json,
+                RemoteSourceContext::PARAM_URI_PATH => $uriRule,
+                RemoteSourceContext::PARAM_LABEL_PATH => $labelRule,
+            ])
+        );
+    }
+
+    public function iterateContext(ContextInterface $context): iterable
```
gabrielfs7, line 49: https://github.com/oat-sa/tao-core/pull/3045/files#r684744084

```suggestion
    public function iterateByContext(ContextInterface $context): iterable
```

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -83,11 +84,14 @@ public function findAll(ValueCollectionSearchRequest $searchRequest): ValueColle
...
         $values = [];
         foreach ($query->execute()->fetchAll() as $rawValue) {
-            $values[] = new Value(
-                (int)$rawValue[self::FIELD_ITEM_ID],
+            $value = new Value(
+                (int) $rawValue[self::FIELD_ITEM_ID],
                 $rawValue[self::FIELD_ITEM_URI],
                 $rawValue[self::FIELD_ITEM_LABEL]
             );
+            $value->setDependencyUri($rawValue[self::FIELD_DEPENDENCY_ITEM_URI]);
```
gabrielfs7, line 92: I would add this as a construct parameter, seems we will alway set it for now on. Wdyt?

models/classes/accessControl/AbstractContext.php
```diff
@@ -6,6 +6,9 @@
 
 use InvalidArgumentException;
 
+/**
+ * @deprecated Use \oat\tao\model\Context\AbstractContext
+ */
 abstract class AbstractContext
```
gabrielfs7, line 12: You can erase all the content and extend the proper abstraction, so you do not keep the same logic in 2 places.

```suggestion
abstract class AbstractContext extends \oat\tao\model\Context\AbstractContext
```

models/classes/accessControl/ContextInterface.php
```diff
@@ -22,8 +22,9 @@
 
 namespace oat\tao\model\accessControl;
 
-use oat\oatbox\user\User;
-
+/**
+ * @deprecated Use \oat\tao\model\Context\ContextInterface
+ */
 interface ContextInterface
```
gabrielfs7, line 28: You can make this interface extends the proper one and to not keep the same methods in 2 places.

```suggestion
interface ContextInterface extends \oat\tao\model\Context\ContextInterface
```

gabrielfs7, line 28: You can extends the proper interface and remove the methods from here, so it is easier/safer to migrate to eliminate deprecated implementations.

```suggestion
interface ContextInterface extends \oat\tao\model\Context\ContextInterface
```

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -167,20 +171,18 @@ protected function insert(ValueCollection $valueCollection, Value $value): void
...
-                    'uri'     => $value->getUri(),
-                    'label'   => $value->getLabel(),
-                    'listUri' => $valueCollection->getUri()
-                ]
-            )
+            ->values([
+                self::FIELD_ITEM_LABEL => ':label',
+                self::FIELD_ITEM_URI => ':uri',
+                self::FIELD_ITEM_LIST_URI => ':listUri',
+                self::FIELD_DEPENDENCY_ITEM_URI => ':dependencyUri',
```
gabrielfs7, line 178: Did not we agree to use a separate table to keep the dependencies?

scripts/install/CreateRdsListStore.php
```diff
@@ -62,21 +62,43 @@ public function create(Schema $schema): void
...
+            'string',
+            ['length' => 255]
+        );
+        $listItemsTable->addColumn(
+            RdsValueCollectionRepository::FIELD_ITEM_LIST_URI,
+            'string',
+            ['length' => 255]
+        );
+        $listItemsTable->addColumn(
+            RdsValueCollectionRepository::FIELD_DEPENDENCY_ITEM_URI,
```
gabrielfs7, line 86: Did not we agree to use a separate table to keep the dependencies?

gabrielfs7 (changes requested):
Good job @shpran! 🥇 

I think should use a separate table to store the dependencies relation as we agreed before, so we can be prepared to more complex relations in the future.

There are some other comments about deprecation and OO, etc

models/classes/Lists/Business/Domain/RemoteSourceContext.php
shpran: They are all covered by this validation:
```php
if ($parameter !== self::PARAM_JSON && is_string($parameterValue)) {
    return;
}
```

The only check we need - just to check that they are string-values

models/classes/Lists/Business/Domain/RemoteSourceContext.php
shpran: Nope, it's just a parser key (string value)

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
shpran: Yes, we need to discuss this one more time 🙂

migrations/Version202007151446482234_tao.php
```diff
@@ -1,32 +1,120 @@
...
-
     public function getDescription(): string
     {
         return 'Create an RDS store for lists';
     }
 
     public function up(Schema $schema): void
     {
-        (new CreateRdsListStore())->create($schema);
+        $fromSchema = clone $schema;
```
boajer, line 41: Why are you modifying existing migrations?

models/classes/Lists/Business/Domain/Value.php
```diff
@@ -107,8 +116,9 @@ public function hasModifiedUri(): bool
     public function jsonSerialize(): array
     {
         return [
-            'uri'   => tao_helpers_Uri::encode($this->uri),
+            'uri' => tao_helpers_Uri::encode($this->uri),
             'label' => $this->label,
+            'dependencyUri' => $this->dependencyUri,
```
boajer, line 121: Why while we have two different types of urls in serialized object one is encoded, but another - not?

models/classes/featureFlag/FeatureFlagCheckerInterface.php
```diff
@@ -24,5 +24,9 @@
 
 interface FeatureFlagCheckerInterface
 {
+    public const FEATURE_FLAG_ADVANCED_SEARCH_DISABLED = 'FEATURE_FLAG_ADVANCED_SEARCH_DISABLED';
```
boajer, line 27: What is the purpose of introducing such constants?  IMO despite it may be helpful to have all feature flags in the centric location it may introduce even more code coupling

actions/form/class.RemoteList.php
```diff
@@ -55,20 +57,28 @@ public function initElements()
         $this->createTextBoxElement(self::FIELD_SOURCE_URL, __('Data source URI'));
         $this->createTextBoxElement(self::FIELD_ITEM_LABEL_PATH, __('Label Path'));
         $this->createTextBoxElement(self::FIELD_ITEM_URI_PATH, __('URI Path'));
+
+        if (($this->options[self::IS_LISTS_DEPENDENCY_ENABLED] ?? false)) {
+            $this->createTextBoxElement(
+                self::FIELD_DEPENDENCY_ITEM_URI_PATH,
+                __('Dependency URI Path'),
```
gabrielfs7, line 64: Will it be required to update translations files for UDIR with this?

migrations/Version202007151446482234_tao.php
gabrielfs7: Yes, indeed, we need a new migration for the new table

migrations/Version202007151446482234_tao.php
```diff
@@ -1,32 +1,120 @@
...
     }
 
     public function down(Schema $schema): void
     {
         $schema->dropTable(RdsValueCollectionRepository::TABLE_LIST_ITEMS);
+
+        $this->addReport(
+            Report::createSuccess(
+                sprintf(
+                    'Table "%s" successfully dropped',
```
gabrielfs7, line 62: This is already existing table, we should not drop it.

migrations/Version202007151446482234_tao.php
```diff
@@ -1,32 +1,120 @@
...
+                    RdsValueCollectionRepository::TABLE_LIST_ITEMS
+                )
+            )
+        );
+    }
+
+    private function createTable(Schema $schema): void
+    {
+        $listItemsTable = $schema->createTable(RdsValueCollectionRepository::TABLE_LIST_ITEMS);
```
gabrielfs7, line 71: Probably it should be the new table

migrations/Version202108101839442234_tao.php
```diff
@@ -0,0 +1,120 @@
...
+        );
+
+        $listItemsDependenciesTable->addIndex([RdsValueCollectionRepository::FIELD_LIST_ITEM_ID]);
+        $listItemsDependenciesTable->addIndex([RdsValueCollectionRepository::FIELD_LIST_ITEM_FIELD]);
+        $listItemsDependenciesTable->addIndex([RdsValueCollectionRepository::FIELD_LIST_ITEM_VALUE]);
+
+        $listItemsDependenciesTable->addForeignKeyConstraint(
+            RdsValueCollectionRepository::TABLE_LIST_ITEMS,
+            [RdsValueCollectionRepository::FIELD_LIST_ITEM_ID],
+            [RdsValueCollectionRepository::FIELD_ITEM_ID]
```
gabrielfs7, line 97: Is `FIELD_ITEM_ID` a key on `TABLE_LIST_ITEMS`?

models/classes/Lists/Business/Service/RemoteSource.php
```diff
@@ -45,13 +48,38 @@ public function __construct(array $parsers, Client $client = null)
...
+            RemoteSourceContext::PARAM_URI_PATH => $uriPath,
+            RemoteSourceContext::PARAM_LABEL_PATH => $labelPath,
+            RemoteSourceContext::PARAM_PARSER => $parser,
+        ]);
 
-        $body = json_decode((string)$response->getBody(), true);
+        yield from $this->fetchByContext($context);
+    }
+
+    public function fetchByContext(RemoteSourceContext $context): Traversable
```
gabrielfs7, line 66: Should we use ContextInterface instead?

models/classes/Lists/Business/Service/RemoteSource.php
```diff
@@ -45,13 +48,38 @@ public function __construct(array $parsers, Client $client = null)
...
+    {
+        $response = $this->getClient()->get($context->getParameter(RemoteSourceContext::PARAM_SOURCE_URL));
+        $context->setParameter(
+            RemoteSourceContext::PARAM_JSON,
+            json_decode((string) $response->getBody(), true)
+        );
+        $context->setParameter(
+            RemoteSourceContext::PARAM_IS_LISTS_DEPENDENCY_ENABLED,
+            $this->getFeatureFlagChecker()->isEnabled(
```
gabrielfs7, line 75: Why parser does not check itself the feature flag, but instead we provide it here, so if another place call the parses needs to remember to do the same?

models/classes/Lists/Business/Service/RemoteSourceJsonPathParser.php
```diff
@@ -14,48 +15,82 @@
...
             return;
         }
 
-        if ($count !== $labels->count()) {
+        if ($urisCount !== $labels->count()) {
             throw new RuntimeException('Count of URIs and labels should be equal');
         }
 
+        $dependencyUri = null;
+        $isListsDependencyEnabled = $context->getParameter(
```
gabrielfs7, line 71: I suggest to use feature flag checker directly here to avoid clients of this parser to have to do it everytime. 

https://github.com/oat-sa/tao-core/pull/3045/files#r686543834

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -165,23 +209,48 @@ protected function verifyUriUniqueness(Value $value): void
...
+    {
+        if ($this->isListsDependencyEnabled() && $value->getDependencyUri() !== null) {
+            $qb->insert(self::TABLE_LIST_ITEMS_DEPENDENCIES)
+                ->values([
+                    self::FIELD_LIST_ITEM_ID => ':list_item_id',
+                    self::FIELD_LIST_ITEM_FIELD => ':field',
+                    self::FIELD_LIST_ITEM_VALUE => ':value',
+                ])
+                ->setParameters([
+                    'list_item_id' => $qb->getConnection()->lastInsertId('list_items_id_seq'),
```
gabrielfs7, line 248: We need to check if it will work with postgres and myqsl/mariadb

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -138,12 +150,44 @@ public function delete(string $valueCollectionUri): void
...
 
+        $this->deleteListItemsDependencies($query, $valueCollectionUri);
+
         $query->delete(self::TABLE_LIST_ITEMS)
             ->where($query->expr()->eq(self::FIELD_ITEM_LIST_URI, ':list_uri'))
             ->setParameter('list_uri', $valueCollectionUri)
             ->execute();
     }
 
+    private function deleteListItemsDependencies(QueryBuilder $query, string $valueCollectionUri): void
```
gabrielfs7, line 161: Our CR guidelines reinforce private methods should be always after public, protected.

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -165,23 +209,48 @@ protected function verifyUriUniqueness(Value $value): void
...
+                    'uri' => $value->getUri(),
+                    'label' => $value->getLabel(),
+                    'listUri' => $valueCollection->getUri(),
+                ])
+                ->execute();
+
+            $this->insertListItemsDependency($qb, $value);
+
+            $platform->commit();
+        } catch (\Throwable $e) {
```
gabrielfs7, line 233: FQN

```suggestion
        } catch (Throwable $e) {
```

models/classes/featureFlag/FeatureFlagCheckerInterface.php
gabrielfs7: Where would we document map all the FF we have? I do agree it is coupling, but at the same time we do not have a place to track what can be "activated" in the platform.

boajer: What if we can have such FF mapping not in the code itself, but in documentation stored within the codebase?  
If not plain format, but annotations to the classes which can be parsed and aggregated, might be even with help of GitHub actions to be a bit more automatic

gabrielfs7: It could also be, but it wont avoid us to create constants and forget to update this new doc.

migrations/Version202007151446482234_tao.php
shpran: @gabrielfs7 we talked about this yesterday - I'm extracting logic from Installation Script into this migration + I created new migration for new table

gabrielfs7: So this change here will be reverted?

actions/form/class.RemoteList.php
shpran: Translation files can be updated, but this is not necessary - `Dependency URI path` will be used by default if there are no translations.

migrations/Version202007151446482234_tao.php
shpran: I do not fully understand you. I extracted this table creation from Installation script.
Previously, we called installation script directly, and it was not so good.

migrations/Version202108101839442234_tao.php
shpran: Yes

models/classes/Lists/Business/Service/RemoteSource.php
shpran: We cannot use `serviceLocator` inside the parser (I tried 😅).
If we want this - we need to propagate it before. So we will need to remember to propagate it in all other places.
In my case, if we forgot to provide feature flag in the context - it will be `false` by default

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
shpran: Thanks - I forgot about this

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
shpran: With postgres - yes (locally I worked with it).
I will configure playground to use `mysql`

models/classes/featureFlag/FeatureFlagCheckerInterface.php
shpran: So, what should I do?

I will update the README.md file. Something else?

migrations/Version202007151446482234_tao.php
shpran: Nope.
Previously, we executed the installation script here.
In this case, if I will add something to the installation scrip - this may affect on this migration (or on the new ones). Because this migration will contain changes from installation scrip.
That's why I extracted old logic (list_items table creation) from installation script into this migration + I created a new migration for the new table.
And now I can safely update Installation script with my new table and I can be sure, that this will not affect on something except fresh installation.

models/classes/Lists/Business/Service/RemoteSource.php
gabrielfs7: Is it possible to make parser extend ConfigurableService?

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
shpran: Checked with `postgres` and `mysql` - both work

models/classes/Lists/Business/Service/RemoteSource.php
shpran: This may confuse you, but this parser already extend this ConfigurableService 😅

migrations/Version202007151446482234_tao.php
gabrielfs7: Yes, I see your point and it might already have been happened. If you tested the migration it is ok then

migrations/Version202007151446482234_tao.php
gabrielfs7: got it from previous comments, thanks!

migrations/Version202007151446482234_tao.php
gabrielfs7: Yes, not it is clear why you did this. :)

models/classes/Lists/Business/Service/RemoteSource.php
shpran: Done. Will propagate parser during `getParser` method execution 🙂

models/classes/Lists/Business/Domain/Value.php
```diff
@@ -24,8 +24,8 @@
 
 namespace oat\tao\model\Lists\Business\Domain;
 
-use JsonSerializable;
 use tao_helpers_Uri;
+use JsonSerializable;
```
gabrielfs7, line 28: Why change the alphabetical order?

models/classes/featureFlag/FeatureFlagCheckerInterface.php
gabrielfs7: It is ok for me to keep it here, since core is used in all extensions and this is only about a feature flag mapping, we do not have the features logic here

scripts/install/CreateRdsListStore.php
```diff
@@ -24,48 +24,64 @@
...
-            $persistence->exec($query);
-        }
-
-        return Report::createSuccess(
-            sprintf('Table "%s" successfully created', RdsValueCollectionRepository::TABLE_LIST_ITEMS)
-        );
+        return [$fromSchema, $schema];
     }
 
-    public function create(Schema $schema): void
```
gabrielfs7, line 61: Is it a breaking change?

shpran: Nope - this method used only in this installation script.
Plus, as we know, installation scrip should be executed by `__invoke` method, that it's a bit strange that this method was public 😅 I will roll back to the previous name (to be sure that we will not break anything)

gabrielfs7 (approved):

migrations/Version202007151446482234_tao.php
```diff
@@ -1,32 +1,120 @@
...
+
+        $listItemsTable->addColumn(
+            RdsValueCollectionRepository::FIELD_ITEM_ID,
+            'integer',
+            ['autoincrement' => true]
```
boajer, line 76: Any change to avoid auto increments by design? That will not work with NewRDS

models/classes/Context/AbstractContext.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->checkParameterSupport($parameter);
+        $this->validateParameter($parameter, $parameterValue);
+
+        $this->parameters[$parameter] = $parameterValue;
+    }
+
+    abstract protected function getSupportedParameters(): array;
+
+    /**
+     * @param $parameterValue
```
boajer, line 59:
```suggestion
     * @param mixed $parameterValue
```

boajer, line 59:
```suggestion
     * @param mixed $parameterValue
```

models/classes/featureFlag/FeatureFlagCheckerInterface.php
boajer: Yes, lets keep readme updated, I've realized that plenty of people in the company not aware yet we have feature flags at all, not even speaking about which exactly we have. So Readme would be the first place to read about it

scripts/install/CreateRdsListStore.php
```diff
@@ -24,48 +24,64 @@
...
         $listItemsTable = $schema->createTable(RdsValueCollectionRepository::TABLE_LIST_ITEMS);
 
-        $listItemsTable->addColumn(RdsValueCollectionRepository::FIELD_ITEM_ID, 'integer', ['autoincrement' => true]);
-        $listItemsTable->addColumn(RdsValueCollectionRepository::FIELD_ITEM_LABEL, 'string', ['length' => 255]);
-        $listItemsTable->addColumn(RdsValueCollectionRepository::FIELD_ITEM_URI, 'string', ['length' => 255]);
-        $listItemsTable->addColumn(RdsValueCollectionRepository::FIELD_ITEM_LIST_URI, 'string', ['length' => 255]);
+        $listItemsTable->addColumn(
+            RdsValueCollectionRepository::FIELD_ITEM_ID,
+            'integer',
+            ['autoincrement' => true]
```
boajer, line 68: same as https://github.com/oat-sa/tao-core/pull/3045/files#r686760292

boajer (approved):
I would strongly encourage  to deal with https://github.com/oat-sa/tao-core/pull/3045/files#r686760292 before merging, meanwhile it's not a blocker 

migrations/Version202007151446482234_tao.php
gabrielfs7: @boajer this was previous existing logic that only supports RDS by default, so Andrei cannot remove it here, but I do agree with you :) 

We could add a techdebt for this (which was not created by our squad)

models/classes/Lists/Business/Domain/Value.php
shpran: `JsonSerializable` (16 characters) is longer word, than `tao_helpers_Uri` (15 characters) 🙂
First I order by length, then - by alphabet

gabrielfs7: PHPStorm orders by alphabet always, I am not sure we should create another type of ordering, but it is ok :)

PR #3101, opened 2021-09-06
Counter service

alroniks:
Counter service allows counting amount of calls of the event configured in the service.

For example, it used to count every delivery executions creation as well as delivery executions state changes to provide information for another system, for analyzing daily activity.

https://oat-sa.atlassian.net/browse/CGF-445

### DIFF
--- a/migrations/Version202108081928492234_tao.php
+++ b/migrations/Version202108081928492234_tao.php
@@ -1,38 +1,28 @@
 <?php
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 namespace oat\tao\migrations;
 
-use common_Exception;
 use Doctrine\DBAL\Schema\Schema;
 use oat\oatbox\service\exception\InvalidServiceManagerException;
-use oat\oatbox\service\ServiceNotFoundException;
 use oat\tao\model\counter\CounterService;
 use oat\tao\scripts\install\RegisterCounterService;
 use oat\tao\scripts\tools\migrations\AbstractMigration;
 
-/**
- * Auto-generated Migration: Please modify to your needs!
- */
 final class Version202108081928492234_tao extends AbstractMigration
 {
-
     public function getDescription(): string
     {
         return 'Register ' . CounterService::class;
     }
 
-    /**
-     * @param Schema $schema
-     */
     public function up(Schema $schema): void
     {
         $this->propagate(new RegisterCounterService)();
     }
 
     /**
-     * @param Schema $schema
      * @throws InvalidServiceManagerException
      */
     public function down(Schema $schema): void
--- a/migrations/Version202108081928492234_tao.php
+++ b/migrations/Version202108081928492234_tao.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
+namespace oat\tao\migrations;
+
+use common_Exception;
+use Doctrine\DBAL\Schema\Schema;
+use oat\oatbox\service\exception\InvalidServiceManagerException;
+use oat\oatbox\service\ServiceNotFoundException;
+use oat\tao\model\counter\CounterService;
+use oat\tao\scripts\install\RegisterCounterService;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version202108081928492234_tao extends AbstractMigration
+{
+
+    public function getDescription(): string
+    {
+        return 'Register ' . CounterService::class;
+    }
+
+    /**
+     * @param Schema $schema
+     * @throws InvalidServiceManagerException
+     * @throws common_Exception
+     * @throws ServiceNotFoundException
+     */
+    public function up(Schema $schema): void
+    {
+        $registerCounterService = new RegisterCounterService();
+        $this->propagate($registerCounterService);
+        $registerCounterService([]);
+    }
+
+    /**
+     * @param Schema $schema
+     * @throws InvalidServiceManagerException
+     */
+    public function down(Schema $schema): void
+    {
+        $this->getServiceManager()->unregister(CounterService::SERVICE_ID);
+    }
+}
--- a/migrations/Version202108081928492234_tao.php
+++ b/migrations/Version202108081928492234_tao.php
@@ -25,15 +25,10 @@ final class Version202108081928492234_tao extends AbstractMigration
 
     /**
      * @param Schema $schema
-     * @throws InvalidServiceManagerException
-     * @throws common_Exception
-     * @throws ServiceNotFoundException
      */
     public function up(Schema $schema): void
     {
-        $registerCounterService = new RegisterCounterService();
-        $this->propagate($registerCounterService);
-        $registerCounterService([]);
+        $this->propagate(new RegisterCounterService)();
     }
 
     /**
--- a/models/classes/counter/CounterService.php
+++ b/models/classes/counter/CounterService.php
@@ -15,46 +15,45 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
  * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
- *
  */
 
-declare(strict_types=1);
+declare(strict_types = 1);
 
 namespace oat\tao\model\counter;
 
 use common_Exception;
-use common_persistence_KeyValuePersistence;
+use common_persistence_KeyValuePersistence as KeyValuePersistence;
 use oat\oatbox\event\Event;
 use oat\oatbox\event\EventManager;
 use oat\oatbox\service\ConfigurableService;
 use oat\oatbox\service\exception\InvalidServiceManagerException;
 use oat\oatbox\service\ServiceNotFoundException;
 
-/**
- *
- */
 class CounterService extends ConfigurableService
 {
     public const SERVICE_ID = 'tao/CounterService';
+
     public const OPTION_PERSISTENCE = 'persistence';
     public const OPTION_COUNTER_KEY_PREFIX = 'counterKeyPrefix';
     public const OPTION_EVENTS = 'events';
+
     public const DEFAULT_PREFIX = 'tao:counter:';
+
     protected const OPTION_CALLBACK = 'keyCallbackMethod';
     protected const OPTION_SHORT_NAME = 'shortName';
 
     /**
-     * @return common_persistence_KeyValuePersistence
+     * @return KeyValuePersistence
      * @throws CounterServiceException
      */
-    protected function getPersistence(): common_persistence_KeyValuePersistence
+    protected function getPersistence(): KeyValuePersistence
     {
         $persistenceId = $this->getOption(self::OPTION_PERSISTENCE);
-        $persistence = common_persistence_KeyValuePersistence::getPersistence($persistenceId);
+        $persistence = KeyValuePersistence::getPersistence($persistenceId);
 
-        if (!$persistence instanceof common_persistence_KeyValuePersistence) {
+        if (!$persistence instanceof KeyValuePersistence) {
             $msg = "Persistence '${persistenceId}' must be an instance of '";
-            $msg .= common_persistence_KeyValuePersistence::class . "', ";
+            $msg .= KeyValuePersistence::class . "', ";
             $msg .= get_class($persistence) . ' persistence given.';
             throw new CounterServiceException($msg, CounterServiceException::CODE_INVALID_PERSISTENCE);
         }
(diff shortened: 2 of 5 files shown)


ks16 (approved):

migrations/Version202108081928492234_tao.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return 'Register ' . CounterService::class;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->propagate(new RegisterCounterService)();
```
siwane, line 31: Should you register the service here?

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+    public const DEFAULT_PREFIX = 'tao:counter:';
+
+    protected const OPTION_CALLBACK = 'keyCallbackMethod';
+    protected const OPTION_SHORT_NAME = 'shortName';
+
+    /**
+     * @return KeyValuePersistence
+     * @throws CounterServiceException
+     */
+    protected function getPersistence(): KeyValuePersistence
```
siwane, line 49: wdyt about putting visibility to private? it eases the maintenance and there is no value to inherit this method (as it is configurable)

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+     * @param string      $eventFqcn
+     * @param string      $shortName
+     * @param string|null $keyCallbackMethod
+     *
+     * @throws CounterServiceException
+     * @throws InvalidServiceManagerException
+     * @throws ServiceNotFoundException
+     * @throws common_Exception
+     */
+    public function attach(string $eventFqcn, string $shortName, ?string $keyCallbackMethod = null): void
```
siwane, line 74: Attach and detach method should be part of the install and not part of the runtime. I think it is better to create another service to alter config folder

siwane (commented):
lgtm, some comments

siwane:
Im wondering about the purpose of this PR.

scripts/install/RegisterCounterService.php
```diff
@@ -0,0 +1,94 @@
...
+        $logMsg = "Counter Service registered with persistence '${persistence}' and key prefix '" . CounterService::DEFAULT_PREFIX . "'.";
+        $this->logInfo($logMsg);
+    }
+
+    /**
+     * @return string|null
+     * @throws ServiceNotFoundException
+     * @throws InvalidServiceManagerException
+     */
+    protected function discoverPersistenceId(): ?string
```
gabrielfs7, line 70: As @siwane pointed, lets focus on least privilege principle and try to avoid protected/public unless they are required and are used.

It makes eayser to maintain code. Imagine that if you have a protected method/attribute/constant and have to change the signature it is already a **breaking change** and hard to track where is it used, since this is an open-source code.

scripts/install/RegisterCounterService.php
```diff
@@ -0,0 +1,94 @@
...
+
+        $logMsg = "Counter Service registered with persistence '${persistence}' and key prefix '" . CounterService::DEFAULT_PREFIX . "'.";
+        $this->logInfo($logMsg);
+    }
+
+    /**
+     * @return string|null
+     * @throws ServiceNotFoundException
+     * @throws InvalidServiceManagerException
+     */
```
gabrielfs7, line 69: No need for redundant annotations.

```suggestion
    /**
     * @throws ServiceNotFoundException
     * @throws InvalidServiceManagerException
     */
```

scripts/install/RegisterCounterService.php
```diff
@@ -0,0 +1,94 @@
...
+        $persistenceManager = $this->getServiceManager()->get(PersistenceManager::SERVICE_ID);
+
+        /**
+         * Search for a suitable persistence. In case of an installation by Seed, or an Update
+         * of an existing infrastructure on the OAT ecosystem, we might find the best fit.
+         *
+         * 1. Most popular Key Value persistence ID in OAT ecosystem is 'redis'. This is the best fit.
+         * 2. As a fail-over, 'default_kv' persistence is implemented in all TAO Setups (See generis/manifest.php's
+         * Registered Installation Actions).
+         */
```
gabrielfs7, line 84: Why do you think this comment is necessary?

scripts/install/RegisterCounterService.php
```diff
@@ -0,0 +1,94 @@
...
+     * @throws common_Exception
+     */
+    public function __invoke($params = [])
+    {
+        $persistence = $this->discoverPersistenceId();
+
+        /**
+         * The default CounterService is registered for client code usage. We expect the client code
+         * to register counters for specific custom needs.
+         */
```
gabrielfs7, line 48: Why do you think this comment is necessary? It is implicit in TAO that a configurable service can be override by other extensions. We do not need to advertise it in everyplace we use :)

scripts/install/RegisterCounterService.php
```diff
@@ -0,0 +1,94 @@
...
+
+        ]);
+
+        $this->registerService(
+            CounterService::SERVICE_ID,
+            $counterService
+        );
+
+        $logMsg = "Counter Service registered with persistence '${persistence}' and key prefix '" . CounterService::DEFAULT_PREFIX . "'.";
+        $this->logInfo($logMsg);
```
gabrielfs7, line 62: Unnecessary variable:

```suggestion
        $this->logInfo(
            sprintf(
                "Counter Service registered with persistence '${persistence}' and key prefix '%s'",
                CounterService::DEFAULT_PREFIX
            )
        );
```

scripts/install/RegisterCounterService.php
alroniks: Not my, it was Jerome's explanation, but I fully agree that it has to be there with an explanation of how it works. Code should be self-explanatory, but the logic behind not always clear. I prefer to leave it here.

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+    public const SERVICE_ID = 'tao/CounterService';
+
+    public const OPTION_PERSISTENCE = 'persistence';
+    public const OPTION_COUNTER_KEY_PREFIX = 'counterKeyPrefix';
+    public const OPTION_EVENTS = 'events';
+
+    public const DEFAULT_PREFIX = 'tao:counter:';
+
+    protected const OPTION_CALLBACK = 'keyCallbackMethod';
+    protected const OPTION_SHORT_NAME = 'shortName';
```
gabrielfs7, line 43: https://github.com/oat-sa/tao-core/pull/3101/files#r704079172

```suggestion
    private const OPTION_CALLBACK = 'keyCallbackMethod';
    private const OPTION_SHORT_NAME = 'shortName';
```

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+    {
+        return (int)$this->getPersistence()->get($this->buildKey($eventFqcn, $keyCallbackValue));
+    }
+
+    /**
+     * @param string $eventFqcn
+     * @param string|null $keySuffix
+     * @return string
+     */
+    protected function buildKey(string $eventFqcn, ?string $keySuffix = null): string
```
gabrielfs7, line 178: Keep least privilege in mind. It is very important to avoid extra maintenance costs.

https://github.com/oat-sa/tao-core/pull/3101/files#r704079172

Unnecessary annotation, we already have typehint and return types

```suggestion
    private function buildKey(string $eventFqcn, ?string $keySuffix = null): string
```

scripts/install/RegisterCounterService.php
alroniks: It is tao-core, root extension. I assume it is the best place to advertise such things. :) Even if we all know that it is possible, but newbies don't know it.

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+
+use common_Exception;
+use common_persistence_KeyValuePersistence as KeyValuePersistence;
+use oat\oatbox\event\Event;
+use oat\oatbox\event\EventManager;
+use oat\oatbox\service\ConfigurableService;
+use oat\oatbox\service\exception\InvalidServiceManagerException;
+use oat\oatbox\service\ServiceNotFoundException;
+
+class CounterService extends ConfigurableService
```
gabrielfs7, line 32: Unit tests?

models/classes/counter/CounterService.php
gabrielfs7: https://github.com/oat-sa/tao-core/pull/3101/files#r704079172

gabrielfs7: Also, we can avoid redundant annotation when we have type hint/return type. Less code to read /maintain is always better :)

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+    protected const OPTION_SHORT_NAME = 'shortName';
+
+    /**
+     * @return KeyValuePersistence
+     * @throws CounterServiceException
+     */
+    protected function getPersistence(): KeyValuePersistence
+    {
+        $persistenceId = $this->getOption(self::OPTION_PERSISTENCE);
+        $persistence = KeyValuePersistence::getPersistence($persistenceId);
```
gabrielfs7, line 52: Is a static call the best way?

models/classes/counter/CounterService.php
gabrielfs7: Suggestion: I would create an object (or array) to pass as parameter (This is a Command from CQRS). This list of params can grow/shrink and then we will have troubles extending it.

models/classes/counter/CounterService.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * @param string $eventFqcn
+     *
+     * @throws InvalidServiceManagerException|CounterServiceException
+     * @throws \oat\oatbox\service\ServiceNotFoundException
+     * @throws \common_Exception
+     */
+    public function detach(string $eventFqcn): void
```
gabrielfs7, line 103: Suggestion: I would avoid redundant annotations and also import FGN

```suggestion
    /**
     * @throws InvalidServiceManagerException|CounterServiceException
     * @throws ServiceNotFoundException
     * @throws common_Exception
     */
    public function detach(string $eventFqcn): void
```

gabrielfs7:
Code sniffing is breaking: http://qa.kitchen.it.taocloud.org/blue/organizations/jenkins/ci-pipeline-cg-extension-pr%2Ftao-core/detail/PR-3101/4/pipeline

gabrielfs7 (changes requested):
Good job! 

Some required changes required:

- Least privilege
- Code sniffing is breaking http://qa.kitchen.it.taocloud.org/blue/organizations/jenkins/ci-pipeline-cg-extension-pr%2Ftao-core/detail/PR-3101/4/pipeline
- Missing unit tests

migrations/Version202108081928492234_tao.php
```diff
@@ -0,0 +1,42 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use oat\oatbox\service\exception\InvalidServiceManagerException;
+use oat\oatbox\service\ServiceNotFoundException;
+use oat\tao\model\counter\CounterService;
+use oat\tao\scripts\install\RegisterCounterService;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
```
gabrielfs7, line 17: Not needed

```suggestion
```

models/classes/counter/CounterService.php
gabrielfs7: Suggestion: I would create an interface for this service, since it seems it can be implemented in the feature with different approach

migrations/Version202108081928492234_tao.php
alroniks: But it registers the service here. It uses a script, which is used for both cases, installation and updating. `RegisterCounterService` is an invokable script.

models/classes/counter/CounterService.php
alroniks: To change it need to refactor too deep things https://github.com/oat-sa/generis/blob/master/common/persistence/class.Persistence.php#L46, so I would say yes, no way.

scripts/install/RegisterCounterService.php
gabrielfs7: I think it pollutes the codebase. I got many times outdated code, cause implementation changed and I never rely on comments, specially in TAO. 

Also this is describing how ConfigurableService works. The best place to advertise is in the README.md or user guide, not in the places that use it :) 

I think less is more, so less to read is always better

scripts/install/RegisterCounterService.php
gabrielfs7: The code iterates an array and when find the key, returns it. Not sure we need an explanation for that. If KV is recommended, this might be the case now, but for other purposes, other infra, it could be different. This is a decision taken outside of the application, not to be documented here IMO, but this is my suggestion :)

models/classes/counter/CounterService.php
gabrielfs7: You can use PersistenceManager

```
/** @var PersistenceManager $persistenceManager */
$persistenceManager = $this->getServiceManager()->get(PersistenceManager::SERVICE_ID);

$persistenceManager->getPersistenceById($persistenceId);
```

models/classes/counter/CounterService.php
SashaTAO: This service is quite simple and does exactly what it has to do - count registered events. Maybe it would be better to make it final and avoid customization? In this case we will have less issues, WDYT?

gabrielfs7: It is reasonable to say that in the future we might need to attach event timestamp, payload, userId or other data to provide to more granular counting metrics.

We cannot predict the future, but having a list of parameters like this (including one optional parameter) already gives a hint that this can change.

I would work with a context parameter https://github.com/oat-sa/tao-core/blob/master/models/classes/Context/ContextInterface.php

scripts/install/RegisterCounterService.php
alroniks: I moved this description from the method body to the method description, it should help with code readability itself, but I still want to save somewhere explanation. Sometimes it helps a lot especially you want to get a quick answer during support instead of digging the code.

PR #3172, opened 2021-10-11
feat: Block field edition based on usages

gabrielfs7:
Block field edition based on usages

https://oat-sa.atlassian.net/browse/ADF-710

- [x] Handle form display on BE
- [x] Handle form display on FE
- [x] Add validators in the BE
- [x] Get new ElementAware interface from @shpran PRs when it is ready
- [x] Limit BE validators scope due complexity and time constraints (Pending business confirmation)
- [x] Cover with unit tests

**Related PR**
https://github.com/oat-sa/tao-core-ui-fe/pull/373


AntonNeviarouski:
QA verified

PR #3183, opened 2021-10-21
Fix TR-1911 - fresh installation fails

wazelin:
fix: remove infinite recursive calls of `SearchProxy::supportCustomIndex()`

Fresh installation of the latest version is impossible ATM, let's fix it ASAP.


Businesspunk:
image

PR #3188, opened 2021-10-21
[Feat] ADF-495: GREP Cascaded deletion

clagosarias:
**Ticket**
https://oat-sa.atlassian.net/browse/ADF-495

**Description**
Implement cascaded deletion of values for secondary properties when primary property value is altered.
This also contributes to the proper functioning of filtering secondary properties values.

**How to test**
- Create a primary property, with an eligible remote list, for every available type (single dropdown, single search, multiple search)
- Create a secondary property with an eligible remote list, for every available type (single dropdown, single search, multiple search) for every primary property
- Test that the addition/removal/filtering of the secondary property values are aligned and in sync with the changes to the primary property value


AntonNeviarouski:
QA verified

PR #3198, opened 2021-10-27
Feature/ADF-711/Handling changes on dependency definition

shpran:
# [ADF-711](https://oat-sa.atlassian.net/browse/ADF-711)

## Changes
* Added pre-validation for dependent values;
* If `Depends on property` value set - the `Depends on property` field allows selecting ` --- none --- ` to revoke the dependency.

## Require PRs
* https://github.com/oat-sa/generis/pull/934

## Tasks
- [x] Fix unit tests
- [x] Add new unit tests
- [ ] Update `oat-sa/generis` version in `composer.json`


models/classes/Lists/ServiceProvider/ListsServiceProvider.php
```diff
@@ -124,5 +126,15 @@ public function __invoke(ContainerConfigurator $configurator): void
                     service(DependentPropertySpecification::class),
                 ]
             );
+
+        $services
+            ->set(DependsOnPropertyFormFieldFactory::class, DependsOnPropertyFormFieldFactory::class)
```
gabrielfs7, line 131: It the place using this using the PsrContainer?

models/classes/Lists/Presentation/Web/Factory/DependsOnPropertyFormFieldFactory.php
```diff
@@ -46,66 +62,81 @@ public function withElement(tao_helpers_form_FormElement $element): self
...
+        return $this->featureFlagChecker->isEnabled(
+            FeatureFlagCheckerInterface::FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED
+        );
+    }
 
+    private function initElement(int $index): tao_helpers_form_FormElement
+    {
         $element = $this->element ?? tao_helpers_form_FormFactory::getElement(
-            "{$index}_depends-on-property",
+            $index . '_depends-on-property',
```
gabrielfs7, line 85: sprintf?

models/classes/Lists/Presentation/Web/Factory/DependsOnPropertyFormFieldFactory.php
```diff
@@ -46,66 +62,81 @@ public function withElement(tao_helpers_form_FormElement $element): self
     public function create(array $options): ?tao_helpers_form_FormElement
     {
         // @TODO Remove feature flag after we can rely on repository output
-        if (
-            !$this->getFeatureFlagChecker()->isEnabled(
-                FeatureFlagCheckerInterface::FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED
-            )
-        ) {
+        if (!$this->isListsDependencyEnabled()) {
```
gabrielfs7, line 65: Do we still need this @todo?

gabrielfs7 (approved):

models/classes/Lists/Presentation/Web/Factory/DependsOnPropertyFormFieldFactory.php
shpran: I don't think that there's a bit necessary for `sprintf` here.
I think we should use concatenation for such cases as:
* concatenation at the beginning: `$i . ' string'`;
* concatenation as the end: `'string ' . $i`;
* etc.

And `sprintf` for strings that might be difficult to read: `'Some text ' . $i . ' to read'` ---> `sprintf('Some text %s to read', $i)`

But it's just my opinion 🙂

models/classes/Lists/ServiceProvider/ListsServiceProvider.php
shpran: https://github.com/oat-sa/tao-core/blob/master/actions/form/class.SimpleProperty.php#L279

AntonNeviarouski:
QA verified

PR #3209, opened 2021-10-28
Fix/AUT-787/select moved resource

bziondik:
Related to: https://oat-sa.atlassian.net/browse/AUT-787
**Description**
Whenever user moves the class with it's content to another place within Items, Tests or Deliveries tab, the 'Move to' menu (where user choose destination class) becomes inactive and can be removed by page refresh or clicking another class. However, the class is moved so this seems like a visual issue which is reproduced randomly.

**Steps to reproduce:**
Go to Deliveries tab.
Click on class with content.
Click 'Move to' and choose destination class in opened menu.
Click 'Move'.
**Expected result:** Class moved and highlighted, it's properties are displayed.
**Actual result:** 'Move to' menu is not closed and becomes 'dimmed' with loading indicator spinning on the 'Move' button.

**Was made fix:**
Issue can be reproduce In case destination class is not opened in tree on the left panel.
In this case tree refresh `$(tree).trigger('refresh.taotree', [destinationClassUri]);` doesn't work, because node isn't present in tree(tree parts are loaded by demand)
For this case in parallel with loading classes on destination selector should be loaded classes in tree
But for empty folder should be also loaded data by clicking on it, in case class is filled it will be refreshed without clicking

### DIFF
--- a/views/js/layout/actions/common.js
+++ b/views/js/layout/actions/common.js
@@ -32,19 +32,35 @@ define([
     'ui/destination/selector',
     'uri',
     'ui/feedback',
-    'ui/dialog/confirm',
-], function($, __, _, Promise, request, section, binder, permissionsManager, resourceProviderFactory, destinationSelectorFactory, uri, feedback, confirmDialog) {
+    'ui/dialog/confirm'
+], function(
+    $,
+    __,
+    _,
+    Promise,
+    request,
+    section,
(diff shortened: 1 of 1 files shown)


bziondik:
Fix will be done in https://github.com/oat-sa/tao-core/pull/3216

PR #3233, opened 2021-11-16
Fix/ADF-309/Do not remove classes due to restricted resources while deletion

shpran:
# [ADF-309](https://oat-sa.atlassian.net/browse/ADF-309)

## Companion PRs:
* https://github.com/oat-sa/extension-tao-test/pull/409
* https://github.com/oat-sa/extension-tao-item/pull/548
* https://github.com/oat-sa/extension-tao-mediamanager/pull/349

## How to test (Example on `Items` section)
1. Create new user with `Item Author` role;
2. Open `Items` section;
3. Make the following structure
    <details>
    <summary>Expand</summary>

    ```
    Root Class
    |
    |____Class 1
    |      |
    |      |____Class 1_1
    |      |      |
    |      |      |____Class 1 - Item 1    <================
    |      |      |____Class 1 - Item 2
    |      |
    |      |____Class 1_2
    |      |      |
    |      |      |____Class 2 - Item 1
    |      |      |____Class 2 - Item 2
    |      |
    |      |____Class 1_3    <================
    |             |
    |             |____Class 3 - Item 1
    |             |____Class 3 - Item 2
    |
    |____Root Class - Item 1    <================
    |____Root Class - Item 2
    ```
    </details>
4. Open an access control for the marked by arrows elements and replace `Back Office` role with `Global Manager` (Grant);
5. Log in as recently created user;
6. Select `Class 1` and try to delete it.

**Old behavior:** _`Class 1` with all subclasses and instances (items) was deleted._
**New behavior:** _Only accessible subclasses and instances (items) were deleted. `Class 1` still present and contains all protected elements._

## Tasks:
- [x] Add unit tests


actions/class.RdfController.php
```diff
@@ -1046,19 +1045,29 @@ public function deleteClass()
...
+
+            $classDeleter = $this->getPsrContainer()->get(ClassDeleter::class);
+            $classDeleter->delete($class, $this->getRootClass());
+
+            $success = true;
+            $deleted = $classDeleter->isDeleted($class);
+
+            $msg = $success
+                ? __('%s has been deleted', $label)
+                : __('Unable to delete %s', $label);
```
gabrielfs7, line 1064: $success is always true.

```suggestion
            $msg = __('%s has been deleted', $label);
```

shpran: I forgot to use `$deleted` instead 😅
Thanks!

models/classes/resources/Service/ClassDeleter.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        $this->permissionChecker = $permissionChecker;
+        $this->ontology = $ontology;
+    }
+
+    public function delete(core_kernel_classes_Class $class, core_kernel_classes_Class $rootClass): void
+    {
+        if (!$class->equals($rootClass)) {
+            $this->deleteClassRecursively($class);
+        }
```
gabrielfs7, line 52: Maybe we should trow an exception if they are the same?

models/classes/resources/Service/ClassDeleterInterface.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\model\resources\Service;
+
+use core_kernel_classes_Class;
+
+interface ClassDeleterInterface
+{
+    public function delete(core_kernel_classes_Class $class, core_kernel_classes_Class $rootClass): void;
```
gabrielfs7, line 29: Do we really need the root class as parameter?

models/classes/resources/Service/ClassDeleter.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        return !$class->exists();
+    }
+
+    private function deleteClassRecursively(core_kernel_classes_Class $class): bool
+    {
+        $isClassDeletable = true;
+
+        foreach ($class->getSubClasses() as $subClass) {
+            $isClassDeletable = $this->deleteClassRecursively($subClass) && $isClassDeletable;
```
gabrielfs7, line 65: hmmm. If the last time you called `$this->deleteClassRecursively` it returns true, it will ignore the previous `false` that you got.

gabrielfs7: Forget it. Got now the second && validation

models/classes/resources/Service/ClassDeleter.php
```diff
@@ -0,0 +1,127 @@
...
+
+        return $this->deleteClass($class, $isClassDeletable);
+    }
+
+    private function deleteClass(core_kernel_classes_Class $class, bool $isClassDeletable): bool
+    {
+        return $this->deleteInstances($class->getInstances())
+            && $isClassDeletable
+            && $this->deleteProperties($class->getProperties())
+            && $class->delete();
```
gabrielfs7, line 76: We must check if the user has permission to delete the class too. It needs to be checked before any deletion is performed.

shpran: Yeah, this will be my second step 🙂 
First I wanted to check for the resource accessibility

models/classes/resources/Service/ClassDeleter.php
```diff
@@ -0,0 +1,127 @@
...
+        $status = true;
+
+        foreach ($instances as $instance) {
+            if (!$this->permissionChecker->hasWriteAccess($instance->getUri())) {
+                $status = false;
+
+                continue;
+            }
+
+            $instance->delete();
```
hectoras, line 93: nitpick: Maybe using a standard `if`/`else` branching condition here makes it more readable instead jumping with `continue;` ?

```suggestion
            if (!$this->permissionChecker->hasWriteAccess($instance->getUri())) {
                $status = false;
            } else {
                $instance->delete();
            }
```

hectoras: We can even invert the condition and put the "happy path" first:

```suggestion
            if ($this->permissionChecker->hasWriteAccess($instance->getUri())) {
                $instance->delete();
            } else {
                $status = false;
            }
```

hectoras (approved):

actions/class.RdfController.php
```diff
@@ -1046,19 +1046,34 @@ public function deleteClass()
...
             $label = $class->getLabel();
-            $success = $this->getClassService()->deleteClass($class);
-            $msg = $success ? __('%s has been deleted', $label) : __('Unable to delete %s', $label);
+
+            $classDeleter = $this->getClassDeleter();
+
+            try {
+                $classDeleter->delete($class);
+                $success = true;
+            } catch (InvalidArgumentException $exception) {
```
gabrielfs7, line 1062: Maybe catch a Throwable instead? I am not sure this is the only exception that can be thrown. Internally our ontology can throw another. Are you wrapping them into a `InvalidArgumentException`? If yes, does it make sense to be this exception?

actions/class.RdfController.php
```diff
@@ -1046,19 +1046,34 @@ public function deleteClass()
...
+            $msg = $deleted
+                ? __('%s has been deleted', $label)
+                : __('Unable to delete %s', $label);
         }
 
         $this->returnJson([
             'success' => $success,
             'message' => $msg,
-            'deleted' => $success
+            'deleted' => $deleted,
```
gabrielfs7, line 1076: Will it make sense to have success true and deleted false?

models/classes/ClassServiceTrait.php
```diff
@@ -50,38 +50,22 @@ public function deleteResource(\core_kernel_classes_Resource $resource)
...
+
+        try {
+            $classDeleter->delete($class);
+        } catch (InvalidArgumentException $exception) {
+            return false;
         }
-        
-        return (bool) $returnValue;
+
+        return $classDeleter->isDeleted($class);
```
gabrielfs7, line 68: You should move this return inside of the `try` block.

models/classes/resources/Service/ClassDeleter.php
gabrielfs7: We do not use else as recommendation of code calisthenics. https://williamdurand.fr/2013/06/03/object-calisthenics/#2-dont-use-the-else-keyword

You never need else.

models/classes/resources/Service/RootClassesListService.php
```diff
@@ -0,0 +1,57 @@
...
+    public function __construct(Ontology $ontology)
+    {
+        $this->ontology = $ontology;
+    }
+
+    public function list(): array
+    {
+        $rootClasses = [];
+
+        foreach (MenuService::getAllPerspectives() as $perspective) {
```
gabrielfs7, line 43: Please inject this `MenuService::getAllPerspectives()` in order to perform the unit testing.

models/classes/resources/Service/RootClassesListService.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /** @var Ontology */
+    private $ontology;
+
+    public function __construct(Ontology $ontology)
+    {
+        $this->ontology = $ontology;
+    }
+
+    public function list(): array
```
gabrielfs7, line 39: Maybe you should also have a method `listUris(): array` and `isRoorClass(string $uri): bool` and use it instead of this one, cause we do not need to load all the root classes objects just to check if it is a root class.

models/classes/resources/Service/ClassDeleter.php
hectoras: Then I'd use something like your last example (however `$status` should be kept as false if at least one call to `hasWriteAccess()` returns `false`, since it is used as the return value of the method as a whole).

```suggestion
            if ($this->permissionChecker->hasWriteAccess($instance->getUri())) {
                $instance->delete();
                continue;
            }
            
            $status = false;         
```

actions/class.RdfController.php
shpran: I think it makes sense.
`success = false && deleted = false` - operation was not finished due to unexpected/expected issue.
`success = true && deleted = false` - operation finished, partially class was deleted (instances, subclasses and their properties), but required class was not deleted due to some "protected" elements.
`success = true && deleted = true` - operation finished, everything was deleted.

models/classes/ClassServiceTrait.php
shpran: Are you sure?
I don't remember that `$class->exists()` can throw any exception. (At least, I don't know what should happen to break it 😅 ).
But if you are sure - I will do this.

models/classes/resources/Service/RootClassesListService.php
shpran: I agree to have `listUris` method.
But for root class check I created a specification - are you sure, that we want to have this `is` method inside list service?

models/classes/ClassServiceTrait.php
gabrielfs7: I mean, it makes sense to be inside the try logically. You do not want to execute it after the catch block. Got it?

models/classes/resources/Service/RootClassesListService.php
gabrielfs7: It is ok if you have an specification :)

models/classes/resources/Service/ClassDeleter.php
gabrielfs7: I updated my example. The `deletion` status should also be considered IMO. 

Ping @shpran

shpran: Ping @hectoras @gabrielfs7 
```php
private function deleteInstances(array $instances): bool
{
    $status = true;

    foreach ($instances as $instance) {
        if (!$this->permissionChecker->hasWriteAccess($instance->getUri()) || !$instance->delete()) {
            $status = false;
        }
    }

    return $status;
}
```

actions/class.RdfController.php
gabrielfs7: Is it respecting the previous behavior?

gabrielfs7 (commented):
Amazing job! Just missing tests :) 

models/classes/resources/Service/ClassDeleter.php
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * @param bool $isClassDeletable Class is not deletable if it contains at least one protected sub class,
+     *                               instance or property
+     */
+    private function deleteClass(core_kernel_classes_Class $class, bool $isClassDeletable): bool
+    {
+        $classUri = $class->getUri();
+
+        return $this->permissionChecker->hasReadAccess($classUri)
```
gabrielfs7, line 96: The order of the conditions **cannot be changed at all**. I am just concerned if it is clear enough for the other developers how critical can be if we change the order.  Do you think is there another way to expose this code making sure it is understandable?

test/unit/models/classes/resources/Service/ClassDeleterTest.php
```diff
@@ -0,0 +1,204 @@
...
+use core_kernel_classes_Property;
+use core_kernel_classes_Resource;
+use oat\generis\model\data\Ontology;
+use PHPUnit\Framework\MockObject\MockObject;
+use oat\tao\model\resources\Service\ClassDeleter;
+use oat\tao\model\accessControl\PermissionCheckerInterface;
+use oat\tao\model\Specification\ClassSpecificationInterface;
+
+/**
+ * @TODO Improve tests
```
gabrielfs7, line 36: Still improving?

models/classes/resources/Service/ClassDeleter.php
shpran: At least, I can add comments-block here 😅 
I will take a look on this

test/unit/models/classes/resources/Service/ClassDeleterTest.php
shpran: I want to add some tests to cover other cases:
* Root class provided
* User doesn't have read access to the class
* User doesn't have write access to the resource
* User doesn't have write access to the class
* Resource/property/class was not deleted

+ I want to beautify this test, because we have a lot of duplications.

But I think we should not block this PR by these improvements.
I will do everything anyway, but maybe in a different PR 🙂

models/classes/resources/Service/ClassDeleter.php
hectoras: Some thoughts: Maybe split into multiple `if` clauses (check/deletion/check/deletion)? It's longer, but explicit

```
if (!$this->permissionChecker->hasReadAccess($classUri)) {
    return false;
}

if (!$this->deleteInstances($class->getInstances())) {
    return false;
}

if (!$isClassDeletable || !$this->permissionChecker->hasWriteAccess($classUri)) {
  return false;
}

if (!$this->deleteProperties($class->getProperties()) {
  return false;
}

return $class->delete();
```

or

```
if (!$this->permissionChecker->hasReadAccess($classUri)) {
    return false;
}

if (!$this->deleteInstances($class->getInstances())) {
    return false;
}

if ($isClassDeletable) {
  if ($this->permissionChecker->hasWriteAccess($classUri)
  && $this->deleteProperties($class->getProperties()) {
      return $class->delete();
    }
  }
}

return false;
```

We may also prevent propagating the `$isClassDeletable` flag here and move it into `deleteClassRecursively`? So we move the first 2 checks to the caller and avoid calling methods with side effects in conditionals as much as we can

```
private function deleteClassRecursively(core_kernel_classes_Class $class): bool
{
    ...
    if ($this->permissionChecker->hasReadAccess($classUri) && $isClassDeletable) {
      return $this->deleteClass($class)
    }
    return false;
}

private function deleteClass(core_kernel_classes_Class $class): bool
{
        $classUri = $class->getUri();

        if ($this->permissionChecker->hasWriteAccess($classUri)) {
                 return $this->deleteProperties($class->getProperties())
                     && $class->delete();
        }

       return false;
}

```

views/js/layout/actions/common.js
```diff
@@ -280,6 +280,11 @@ define([
                             });
 
                         } else {
+                            if (response.success && !response.deleted) {
+                                $(actionContext.tree).trigger('refresh.taotree');
+                                reject(response.msg || __('Unable to delete the selected resource because you do not have the required rights to delete part of its content.'));
```
gabrielfs7, line 285: As you can see in the line 270 we are using `message` instead of msg. Why did you change this in the BE instead of the FE?

https://github.com/oat-sa/lib-tao-elasticsearch/pull/54/files

hectoras:
Verified locally on a fresh install the fix works as expected when switching `composer.json` to the feature branches for this fix

```
  "oat-sa/tao-core": "dev-fix/adf-309/check-instances-access-during-deletion as v48.58.1",
  "oat-sa/extension-tao-item": "dev-fix/adf-309/check-instances-access-during-deletion as v11.19.1",
  "oat-sa/extension-tao-test": "dev-fix/adf-309/check-instances-access-during-deletion as v15.10.1",
  "oat-sa/extension-tao-mediamanager": "dev-fix/adf-309/check-instances-access-during-deletion as v12.18.0",
```

image

gabrielfs7 (approved):

AntonNeviarouski:
QA verified

PR #3236, opened 2021-11-17
fix: filter values of depends on property select

clagosarias:
### [ADF-797](https://oat-sa.atlassian.net/browse/ADF-797)

**Description**
Prevent user from being able to create multiple level dependencias.

I have achieved this by adding an `.on('change')` event to the `depends on` property. Whenever a value is selected we will iterate over all other `depends on` properties that have a possible option targeting the recently changed one and mark it as disabled. 

This way, the user will not be able to create a dependency with a property that already has a value selected in the `depends on` property.


habemuscode (approved):
LGTM

views/js/form/depends-on-property.js
```diff
@@ -71,9 +71,14 @@
         ];
     }
 
+    function hasAnyDependency($container) {
```
adriloma, line 74: missing jsdoc

views/js/uiForm.js
```diff
@@ -775,21 +776,37 @@
...
+                            dependsOn.toggle($dependsOnSelect, $dependsOnSelect.parent(), $this.closest('.property-edit-container'));
                         } else {
-                            dependsOnSelect.parent().hide();
+                            $dependsOnSelect.parent().hide();
                         }
                     }
                 });
             }
 
+            function filterDependsOnProperty() {
```
adriloma, line 797: misssed jsdoc.

adriloma (approved):
Code review only, I suggest document every new function.

- [x] Commits are atomic

AntonNeviarouski:
QA verified

PR #3269, opened 2021-12-07
feat/ADF-842: Use default UI language for users with no UI language set

hectoras:
Associated Jira task: [ADF-842](https://oat-sa.atlassian.net/browse/ADF-842)

These changes extracts business logic out of the User Settings controller, moving it into the model instead, and sets the interface language to a default in the user settings page if the user does not have one set.

Related pull requests:
- https://github.com/oat-sa/extension-tao-udir/pull/142
- https://github.com/oat-sa/generis/pull/950


models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+ * Copyright (c) 2018-2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+namespace oat\tao\model\user\implementation;
+
+use oat\tao\model\user\UserSettings;
+use DomainException;
+
+class UserSettingsBuilder
```
gabrielfs7, line 27: Do we really need a builder for this? Seems we are just forwarding constructor parameters.

models/classes/user/UserSettings.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+
+namespace oat\tao\model\user;
+
+/**
+ * Settings for the current user.
+ *
+ * Right now this interface is used to return settings related with user language
+ */
```
gabrielfs7, line 28: It is self-explanatory

```suggestion
```

models/classes/user/UserSettingsFormFieldsPresenter.php
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * This method returns a Resource array with the following keys:
+     *
+     * - ui_lang:   Language selected for the Graphical User Interface.
+     * - data_lang: Language selected to access the data in persistent memory.
+     * - timezone:  Timezone selected to display times and dates.
+     *
+     * @return string[] The URIs of the languages.
+     */
+    public function getFormFields(): array
```
gabrielfs7, line 47: Do you think this class could be a form factory instead? And return `tao_actions_form_UserSettings`. The only need for this method is to be used as form constructor arguments.

models/classes/user/implementation/UserSettingsImpl.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Copyright (c) 2018-2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+namespace oat\tao\model\user\implementation;
+
+use oat\tao\model\user\UserSettings;
+
+class UserSettingsImpl implements UserSettings
```
gabrielfs7, line 26: We generally name the interface with the "Interface" suffix to avoid having to add "Impl" on every class (Which is a Java standard, but we generally not use in PHP). 

```suggestion
class UserSettings implements UserSettingsInterface
```

models/classes/user/implementation/UserSettingsImpl.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    /** @var string|null */
+    protected $dataLanguageCode = null;
+
+    /** @var string|null */
+    protected $uiLanguageCode = null;
+
+    /** @var string */
+    protected $timezone;
```
gabrielfs7, line 35: Least privilege principle. We should never/ever open the scope of methods and attributes unless we need it and it is really justifiable.

```suggestion
    /** @var string|null */
    private $dataLanguageCode;

    /** @var string|null */
    private $uiLanguageCode;

    /** @var string */
    private $timezone;
```

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2018-2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
gabrielfs7, line 18: Please check the other blocks

```suggestion
 * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ */
+
```
gabrielfs7, line 21: We are missing to declare strict types in the new classes.

models/classes/user/UserSettingsFormFieldsPresenter.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $fields = [
+            'timezone' => $this->userSettings->getTimezone(),
+        ];
+
+        if (!empty($this->userSettings->getUILanguageCode())) {
+            $fields['ui_lang'] = $this->userSettings->getUILanguageCode();
+        }
+
+        if (!empty($props[GenerisRdf::PROPERTY_USER_DEFLG])) {
```
shpran, line 57: Did you forget to add `$props` variable?

models/classes/user/UserSettingsFormFieldsPresenter.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2018-2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+namespace oat\tao\model\user;
```
shpran, line 22:
```suggestion
 * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
 */

declare(strict_types=1);

namespace oat\tao\model\user;
```

models/classes/user/UserSettingsService.php
```diff
@@ -0,0 +1,28 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ */
+
+namespace oat\tao\model\user;
```
shpran, line 21:
```suggestion
 * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
 */

declare(strict_types=1);

namespace oat\tao\model\user;
```

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+class UserSettingsBuilder
+{
+    /** @var string|null */
+    protected $dataLanguageCode = null;
+
+    /** @var string|null */
+    protected $uiLanguageCode = null;
+
+    /** @var string */
+    protected $timezone;
```
shpran, line 36: Do we need them to be `protected` ?

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+
+    public function __construct(string $defaultTimezone)
+    {
+        if (empty($defaultTimezone)) {
+            throw new DomainException("Default Timezone is required");
```
shpran, line 41: Please, use single quotes.

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+
+        return $this;
+    }
+
+    public function withTimezone(string $timezone = null): self
+    {
+        $this->timezone = trim($timezone) ?? null;
+
+        return $this;
+    }
```
shpran, line 66: It's not critical, but as for me, `setters` are more habitual 🙂

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+    public function __construct(string $defaultTimezone)
+    {
+        if (empty($defaultTimezone)) {
+            throw new DomainException("Default Timezone is required");
+        }
+
+        $this->timezone = $defaultTimezone;
+    }
+
+    public function withDataLanguage(string $dataLanguage = null): self
```
shpran, line 47: Do we need this default `null` ?

models/classes/user/implementation/UserSettingsBuilder.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $this->dataLanguageCode = trim($dataLanguage) ?? null;
```
shpran, line 49:
```php
trim(' ') ?? null; // result is ''
trim(null) ?? null; // result is ''

trim(' ') ?: null; // result is null
trim(null) ?: null; // result is null
```

models/classes/user/implementation/UserSettingsImpl.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2018-2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+namespace oat\tao\model\user\implementation;
```
shpran, line 22:
```suggestion
 * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
 */

declare(strict_types=1);

namespace oat\tao\model\user\implementation;
```

models/classes/user/implementation/UserSettingsImpl.php
```diff
@@ -0,0 +1,57 @@
...
+    protected $dataLanguageCode = null;
+
+    /** @var string|null */
+    protected $uiLanguageCode = null;
+
+    /** @var string */
+    protected $timezone;
+
+    public function __construct(string $timezone, string $uiLanguageCode = null, string $dataLanguageCode = null) {
+        $this->timezone = $timezone; ;
```
shpran, line 38:
```suggestion
        $this->timezone = $timezone;
```

models/classes/user/implementation/UserSettingsServiceImpl.php
```diff
@@ -0,0 +1,70 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ */
+
+namespace oat\tao\model\user\implementation;
```
shpran, line 21:
```suggestion
 * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
 */

declare(strict_types=1);

namespace oat\tao\model\user\implementation;
```

models/classes/user/implementation/UserSettingsServiceImpl.php
```diff
@@ -0,0 +1,70 @@
...
+use Laminas\ServiceManager\ServiceLocatorAwareTrait;
+use oat\generis\model\OntologyAwareTrait;
+
+class UserSettingsServiceImpl implements UserSettingsService
+{
+    use ServiceLocatorAwareTrait;
+    use OntologyAwareTrait;
+
+    /** @var string */
+    protected $defaultTimeZone;
```
shpran, line 37: https://github.com/oat-sa/tao-core/pull/3269/files#r764337945

models/classes/user/implementation/UserSettingsServiceImpl.php
```diff
@@ -0,0 +1,70 @@
...
+use oat\tao\model\user\UserSettingsService;
+use core_kernel_classes_Resource;
+
+use Laminas\ServiceManager\ServiceLocatorAwareTrait;
+use oat\generis\model\OntologyAwareTrait;
+
+class UserSettingsServiceImpl implements UserSettingsService
+{
+    use ServiceLocatorAwareTrait;
+    use OntologyAwareTrait;
```
shpran, line 34: Please, don't use them.
As I see, you don't need ServiceLocatorAwareTrait;

If you want to use Ontology - use `Ontology` interface as a constructor argument instead (via DI container)

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: I think using it makes `UserSettingsServiceImpl` a little bit easier to read, but no strong opinion either way. This may help in case we add more fields to `UserSettings` in the future, as in this way we should not need to change callers for the builder but just the builder itself.

But, indeed, this may be overkill and we may just use local variables in the caller instead (`UserSettingsServiceImpl`) and the code will still be easily readable. 

```
        if (!empty($props[GenerisRdf::PROPERTY_USER_TIMEZONE])) {
            $timezone = current($props[GenerisRdf::PROPERTY_USER_TIMEZONE]);
        }

        if (!empty($props[GenerisRdf::PROPERTY_USER_UILG])) {
            $uiLanguageCode = current($props[GenerisRdf::PROPERTY_USER_UILG])->getUri();
        }

        if (!empty($props[GenerisRdf::PROPERTY_USER_DEFLG])) {
            $dataLanguageCode = current($props[GenerisRdf::PROPERTY_USER_DEFLG])->getUri();
        }

       return new UserSettingsImpl($timezone ?? $this->defaultTimeZone, $uiLanguageCode ?? null, $dataLanguageCode ?? null);
```

Does something like that look better?

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15 :+1:

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15 :+1:

models/classes/user/UserSettings.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ *
+ */
+
+namespace oat\tao\model\user;
```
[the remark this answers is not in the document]
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15 :+1:

models/classes/user/UserSettingsFormFieldsPresenter.php
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15 :+1:

models/classes/user/UserSettingsService.php
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15 :+1:

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: You are right, we don't. Changed in e4d6f5d4a86192bf63a705631870d0c93b64766c :+1:  Thanks!

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: Fixed :point_right: 4c59e4a8c2283baf44a7a0abfb08eb1d9a90f786. Thanks!

models/classes/user/implementation/UserSettingsImpl.php
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15, thanks!

models/classes/user/implementation/UserSettingsImpl.php
hectoras: :facepalm: :point_right: 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15 Thanks :+1:

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: :facepalm: Fixed in 69d63c03dc7fa7f590335ee8b85e73633c8bf3ea, thanks :)

models/classes/user/implementation/UserSettingsServiceImpl.php
hectoras: Fixed in 1cb1381f10f97bd33e3cad97db2a07c5c24b2e15, thanks :slightly_smiling_face:

models/classes/user/implementation/UserSettingsImpl.php
hectoras: I prefer the other way around (interfaces (aka contracts) with no suffix, implementation with it), but no strong opinion either way.

However, if we do "swap things" here (suffix the interface and not the implementation), then we'll end up with this one and the service interface as `UserSettingsInterface` and `UserSettingsServiceInterface` while also having other interfaces (those that were already here) named as `LockoutStorage` and `UserLocks` (with no `-Interface` suffix) on the same tao-core namespace.

That seems inconsistent for me (and also was a reason to not use the suffix myself in the first place), but I'm OK with that if it is for you both, just wanted to raise my concerns first so we can make an informed decision.

OTOH adding `getUserSettings()` into the existing `GenerisUserService` may be an option worth considering as well.

(Also, I think we should not consider renaming also existing interfaces from tao-core to use the `-Interface` suffix since that would probably be a breaking change).

@gabrielfs7 @shpran Thoughts? Shall we change it, keep as it is now, merge into `GenerisUserService`?

models/classes/user/UserSettingsFormFieldsPresenter.php
hectoras: I like the idea, I'll rename it as `UserSettingsFormFactory` and move instantiation of `tao_actions_form_UserSettings` into it. Thanks :ok_hand:

hectoras: Renamed to `UserSettingsFormFactory` in 23cb88e96c326dec5f7f70468c9a0218956822aa.

The method wraps the interaction with `tao_actions_form_UserSettings`, and returns a `tao_helpers_form_Form` instance.

models/classes/user/implementation/UserSettingsImpl.php
hectoras: Done ea0aaae7eea969e9baccbd1c1671e55ebe298876 :ok_hand:

models/classes/user/UserSettingsFormFieldsPresenter.php
hectoras: I missed that one, thank you! It is fixed in f6c800d3f7cb57e442ca6163a2bdcdaeca23c64a.

(The former code for this got its info from an array, the new one calls a getter from a UserSettings instance. Data used to initialize the new class can be found in [UserSettingsServiceImpl](https://github.com/oat-sa/tao-core/pull/3269/files#diff-9ace29faa8dbcc1d5769d4a48242ee8de1d24357846ac6e51aad73f451c56632R45)).

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: Technically no, since right now the parameters for these are generated by `getUri()` calls form properties, and those always return a string.

Removed `null` defaults in 2b895e97147bb4008c1d11b360e191288d0be5cd :+1:

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: Note for myself

- [ ] TODO Rename builder methods from `withField()` to `setField()`

hectoras: Done: 795aa2986daa0603790e674d4fb7e222e153a36b :+1:

models/classes/user/implementation/UserSettingsServiceImpl.php
hectoras: :white_check_mark:  Done in 0c9a340beb20b4fd53c6d25035b824d095910381, thanks :ok_hand:

models/classes/user/implementation/UserSettingsBuilder.php
gabrielfs7: The Builder makes sense if it is helping you to build the object with extra logic, extra support services, etc, but in your case it is just the constructor parameters. IMO it is overkill.

models/classes/user/implementation/UserSettingsServiceImpl.php
hectoras: ... and removed `ServiceLocatorAwareTrait` as well in 314929641cf8338a3b64dceb3020610f5bd29f09 :)

models/classes/user/UserSettingsService.php
```diff
@@ -0,0 +1,28 @@
...
+
+use core_kernel_classes_Resource;
+
+interface UserSettingsService
+{
+    public function getUserSettings(core_kernel_classes_Resource $user): UserSettings;
```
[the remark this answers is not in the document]
gabrielfs7: Humm, We should not repeat that class name in methods because of this. It is redundant and not necessary. 

Imagine a `UserRepository::find()` method. It is a generic name and it makes sense in the context of this repository.

Btw, if you have a class implementing `UserSettingsService` that has multiple behaviours with `get` you are probably trying to implement it in the wrong place. 

This interface should be used through composition in client services and in implementation classes, those should be SRP and not have more responsibilities to `get` other things. 

Do you find a good idea to have a class implementing UserSettingsService having other responsibilities attached?

models/classes/user/implementation/UserSettingsBuilder.php
hectoras: Removed :point_right: 1ba59a3b5a75d40ab4d61460a985cdf2f9ea5a8e :heavy_check_mark:

models/classes/user/UserSettingsService.php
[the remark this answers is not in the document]
hectoras: From that point of view it makes sense, I'll rename the method.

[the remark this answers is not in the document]
hectoras: Renamed in ef71906d8efffa98cd5de21232b627561a19e80e

models/classes/user/UserSettingsFormFieldsPresenter.php
hectoras: Resolving conversation since it has been already renamed as `UserSettingsFormFactory`.

models/classes/user/implementation/UserSettingsImpl.php
hectoras: Swapped suffixes (interfaces with `-Interface` suffix, implementations with no suffix) in e03bf0aa8a187ae6e4651ba30d0c82b5ab73c616.

models/classes/user/implementation/UserSettingsServiceImpl.php
```diff
@@ -0,0 +1,70 @@
...
+
+use oat\generis\model\GenerisRdf;
+use oat\tao\model\user\UserSettings;
+use oat\tao\model\user\UserSettingsService;
+use core_kernel_classes_Resource;
+
+use Laminas\ServiceManager\ServiceLocatorAwareTrait;
+use oat\generis\model\OntologyAwareTrait;
+
+class UserSettingsServiceImpl implements UserSettingsService
```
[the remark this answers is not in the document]
hectoras: Swapped suffixes (interfaces with `-Interface` suffix, implementations with no suffix) in e03bf0aa8a187ae6e4651ba30d0c82b5ab73c616.

actions/class.UserSettings.php
```diff
@@ -69,87 +74,82 @@ public function password()
...
+    {
+        return $this->getPsrContainer()->get(UserSettingsServiceInterface::class);
+    }
+
+    protected function getUserLanguageService(): UserLanguageServiceInterface
+    {
+        return $this->getPsrContainer()->get(UserLanguageServiceInterface::class);
+    }
+
+    protected function getLanguageService(): tao_models_classes_LanguageService
```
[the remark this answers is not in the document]
hectoras: Done :point_right: 5dfff29e078f8563af79373f2f491c51c225b5ed

actions/class.UserSettings.php
```diff
@@ -47,18 +48,22 @@ class tao_actions_UserSettings extends tao_actions_CommonModule
...
+            $passwordFormContainer = new tao_actions_form_UserPassword([], [FormContainer::CSRF_PROTECTION_OPTION => true]);
+            $passwordForm = $passwordFormContainer->getForm();
+
+            if ($passwordForm->isSubmited() && $passwordForm->isValid()) {
                 $user = $this->getUserService()->getCurrentUser();
-                $this->getServiceLocator()->get(tao_models_classes_UserService::SERVICE_ID)
-                    ->setPassword($user, $myForm->getValue('newpassword'));
+                $newPassword = $passwordForm->getValue('newpassword');
+
+                $this->getUserService()->setPassword($user, $newPassword);
```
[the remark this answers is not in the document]
hectoras: Done :point_right:  5dfff29e078f8563af79373f2f491c51c225b5ed

actions/class.UserSettings.php
```diff
@@ -69,87 +74,82 @@ public function password()
...
+            'formTitle',
+            __('My settings (%s)', DisplayHelper::htmlEscape($currentUser->getLabel()))
+        );
+
+        $this->setData('myForm', $settingsForm->render());
 
         $this->setView('form/settings_user.tpl');
     }
 
+    protected function getExtensionManager(): common_ext_ExtensionsManager
```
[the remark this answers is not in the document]
hectoras: It is new code, so switching them to private in 5dfff29e078f8563af79373f2f491c51c225b5ed. Thanks!

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $fields = [
+            'timezone' => $this->userSettings->getTimezone(),
+            'ui_lang' => $this->defaultLanguage,
+        ];
+
+        if (!empty($this->userSettings->getUILanguageCode())) {
+            $fields['ui_lang'] = $this->userSettings->getUILanguageCode();
+        }
+        if (!empty($this->userSettings->getDataLanguageCode())) {
```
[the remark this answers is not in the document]
hectoras: Fixed in 5dfff29e078f8563af79373f2f491c51c225b5ed

models/classes/user/implementation/UserSettingsService.php
```diff
@@ -0,0 +1,74 @@
...
+namespace oat\tao\model\user\implementation;
+
+use oat\generis\model\data\Ontology;
+use oat\generis\model\GenerisRdf;
+use oat\oatbox\user\UserTimezoneServiceInterface;
+use oat\tao\model\user\UserSettingsInterface;
+use oat\tao\model\user\UserSettingsServiceInterface;
+use core_kernel_classes_Resource;
+
+class UserSettingsService implements UserSettingsServiceInterface
```
gabrielfs7, line 32: - [ ] Lets add a test when possible, please :)

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,67 @@
...
+
+declare(strict_types=1);
+
+namespace oat\tao\model\user;
+
+use tao_helpers_form_FormContainer;
+use tao_actions_form_UserSettings;
+use tao_helpers_form_Form;
+
+class UserSettingsFormFactory
```
gabrielfs7, line 29: - [ ] Lets add a test when possible, please :)

actions/class.UserSettings.php
```diff
@@ -69,87 +74,82 @@ public function password()
...
+        $userSettings = $this->getUserSettingsService()->get($currentUser);
+
+        $languageService = $this->getLanguageService();
+        $userLangService = $this->getUserLanguageService();
+
+        $defaultUserLangCode = $userLangService->getDefaultLanguage();
+        $defaultLanguage = $languageService->getLanguageByCode($defaultUserLangCode);
 
-            $currentUser = $this->getUserService()->getCurrentUser();
+        $presenter = new UserSettingsFormFactory($userSettings, $defaultLanguage);
```
gabrielfs7, line 86: - [ ] Please, do not forget to get it from container :)

models/classes/user/UserSettingsFormFactory.php
gabrielfs7: - [ ] Lets add this to the container

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,67 @@
...
+    /** @var string */
+    private $defaultLanguage;
+
+    public function __construct(UserSettingsInterface $userSettings, string $defaultLanguage)
+    {
+        $this->userSettings = $userSettings;
+        $this->defaultLanguage = trim($defaultLanguage);
+    }
+
+    public function getForm(bool $csrf = true): tao_helpers_form_Form
```
gabrielfs7, line 43: Some suggestions:

- This is a factory, so it `creates`. Maybe `create()` better represents the intent of this?
- Use extensible parameters for public interface when possible. I suggest at least an array `array $params` 
- Having bool parameter for public/protected contract is a code smell and we should avoid. More details here https://martinfowler.com/bliki/FlagArgument.html

```suggestion
    public function create(array $params): tao_helpers_form_Form
```

And inside you can get `$params['useCSRFProtection']`, for example.

models/classes/user/UserSettingsInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA.
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\user;
+
+interface UserSettingsInterface
```
gabrielfs7, line 25: 👌  🚀

models/classes/user/UserSettingsServiceProvider.php
```diff
@@ -0,0 +1,49 @@
...
+use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
+
+class UserSettingsServiceProvider implements ContainerServiceProviderInterface
+{
+    public function __invoke(ContainerConfigurator $configurator): void
+    {
+        $services = $configurator->services();
+
+        $services
+            ->set(UserSettingsServiceInterface::class, UserSettingsService::class)
```
gabrielfs7, line 40: I recommend not use the interface as service Identifier, cause when we have multiple implementations the interface will be already hardcoded to this one. The interface should be just type hinted in the client class to follow open/closed. Wdyt?

```suggestion
            ->set(UserSettingsService::class, UserSettingsService::class)
```

hectoras:
Rebased with last changes from `develop`

models/classes/user/UserSettingsServiceProvider.php
hectoras: Then we'll need to bind `tao_actions_UserSettings` to `UserSettingsService` instead of `UserSettingsServiceInterface`.

That is, we'll be getting the service by asking the PSR container for the particular implementation instead of the interface: `$this->getPsrContainer()->get(UserSettingsService::class);` instead of `$this->getPsrContainer()->get(UserSettingsServiceInterface::class);`.

That's ok?

models/classes/user/UserSettingsFormFactory.php
hectoras: Changed: b41d8e620299f0c5d980c37bbbf1b92a9c791da8

models/classes/user/UserSettingsServiceProvider.php
gabrielfs7: It is ok :). It is your call, not a problem for me any solution in this case.

hectoras: Changed: 616291e818a1d222696f0fd1940a7dcd3cd53772

models/classes/user/UserSettingsFormFactory.php
hectoras: Changed to get the FormFactory from the container: a1210c0312c31f2d10849af2c535fa47d4c75afd

hectoras: Tests added here: 859f4bb412c8b7025fa86bbc9c094bc4a0857cb2

test/unit/user/UserSettingsFormFactoryTest.php
```diff
@@ -20,59 +20,209 @@
...
-                'timezone' => '	Europe/Luxembourg'
-            ]
+            'Having all values set' => [
+                'expectedTimezone' => 'Europe/Luxembourg',
+                'expectedUiLanguage' => 'en-US',
+                'expectedDataLanguage' => 'en-US',
+
+                'defaultUiLanguage' => 'en-US',
+
+                'userTimezone' => '	Europe/Luxembourg',
```
hectoras, line 141: It seems this language is in fact being provided as a string containing extra leading whitespaces: I'm keeping those here for the test to match the real behavior.

models/classes/user/UserSettingsFormFactory.php
hectoras: Updated tests here: 5e5f3915eb05117fa77e3c88869100a1a8076f87

actions/class.UserSettings.php
```diff
@@ -69,87 +75,85 @@ public function password()
...
+        );
+
+        $this->setData('myForm', $settingsForm->render());
 
         $this->setView('form/settings_user.tpl');
     }
 
+    private function getExtensionManager(): common_ext_ExtensionsManager
+    {
+        return $this->getServiceLocator()->get(common_ext_ExtensionsManager::SERVICE_ID);
```
shpran, line 127:
```suggestion
        return $this->getPsrrContainer()->get(common_ext_ExtensionsManager::SERVICE_ID);
```

actions/form/class.UserSettings.php
```diff
@@ -25,15 +25,39 @@
...
+            $this->setLanguageService($options['LanguageService']);
+        }
+
+        if (isset($options['ServiceManager'])) {
+            $this->setServiceManager($options['ServiceManager']);
+        }
+
+        if (isset($options['UserTimezoneService'])) {
+            $this->setUserTimezoneService($options['UserTimezoneService']);
+        }
```
shpran, line 57: Maybe we can store this option names as constants?
E.g.
```php
public const OPTION_LANGUAGE_SERVICE = 'LanguageService';

// ...

if (isset($options[self::OPTION_LANGUAGE_SERVICE])) {
    $this->setLanguageService($options[self::OPTION_LANGUAGE_SERVICE]);
}
```

Also I'm thinking about provided values. If someone provides wrong value (string or something else) - setter will fail with wrong type exception.
```php
public const OPTION_LANGUAGE_SERVICE = 'LanguageService';

// ...

if (($options[self::OPTION_LANGUAGE_SERVICE] ?? null) instanceof tao_models_classes_LanguageService) {
    $this->setLanguageService($options[self::OPTION_LANGUAGE_SERVICE]);
}
```

hectoras: Adding the option names as constants sounds like a good idea for me, changed here: e755569b7e0f7fb0c0c38e4baa8e7c95baf2b5b4 Thanks!

For the second part... I'm not sure if we want to mask an error caused by providing wrong data types.

I would prefer to _let it crash_ in this case, as that would be the case (throwing a `TypeError`, as you already pointed out) if we had proper constructor parameters for them and they were directly passed instead of having them as array elements. Silent errors are tricky to diagnose later :worried: 

However, an intermediate approach would be to let wrong types here but log an error in that case, with the tradeoff of coupling this to the logger.

What do you think is best here? Crash, ignore, or ignore+logError?

Note this is a legacy class, but it seems right now it is only called from our `UserSettingsFormFactory` class.

cc/ @gabrielfs7

actions/form/class.UserSettings.php
```diff
@@ -114,4 +132,48 @@ private function addTimezoneEl($form): void
...
+    }
+
+    private function getServiceManager(): ServiceManager
+    {
+        if (!$this->serviceManager) {
+            $this->serviceManager = ServiceManager::getServiceManager();
+        }
+
+        return $this->serviceManager;
+    }
```
gabrielfs7, line 173: Lets try to use the DI container instead: 

```suggestion
    private function getContainer(): ContainerInterface
    {
        if (!$this->container) {
            $this->container = ServiceManager::getServiceManager()->getContainer();
        }

        return $this->container;
    }
```

models/classes/user/UserSettingsFormFactory.php
gabrielfs7: Thanks!

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * @throws common_Exception
+     */
+    public function create(
+        UserSettingsInterface $userSettings,
+        string $uiLanguage = null,
+        array $params = []
+    ): tao_helpers_form_Form {
```
gabrielfs7, line 48: Why not having just `array $params` as signature?

gabrielfs7 (approved):

actions/form/class.UserSettings.php
hectoras: Changed in f0d12cbfbb341c933c603a82e884395f14a84f73, thanks!

models/classes/user/UserSettingsFormFactory.php
hectoras: - For `UserSettingsInterface`: This way I can type hint it on the method signature instead of manually testing it inside with something like `if (!isset($params['userSettings'])  || !($params['userSettings'] instanceof UserSettingsInterface)) { throw new RuntimeException('...'); }`. Also, in this case the form is always created from a `UserSettings` instance, so it is always mandatory.

- For `$uiLanguage`: That way the caller does not need to know the name of the "param" that it needs to add to the `$param` array, so the signature is self-explanatory.
  - But, in fact, maybe we should either pass all as parameters as parameters on its own through the method signature (i.e. `useCSRFProtection` too, which is then passed to `tao_actions_form_UserSettings` constructor), or pass all inside `$params` and add validations inside.

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,74 @@
...
+use tao_helpers_form_Form;
+use tao_helpers_form_FormContainer;
+use tao_models_classes_LanguageService;
+
+class UserSettingsFormFactory
+{
+    /** @var tao_models_classes_LanguageService */
+    private $languageService;
+
+    public function __construct(tao_models_classes_LanguageService $languageService = null)
```
shpran, line 36: Why this one is `null` by default? 🙂
And the second question - why we are not providing this value in Service Provider?
Just to know 😅

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,74 @@
...
+
+        if (!empty($userSettings->getDataLanguageCode())) {
+            $fields['data_lang'] = $userSettings->getDataLanguageCode();
+        }
+
+        $options = [
+            tao_helpers_form_FormContainer::CSRF_PROTECTION_OPTION => (bool) ($params['useCSRFProtection'] ?? true),
+        ];
+
+        if ($this->languageService != null) {
```
shpran, line 66: Please, use strict comparison every time 🙂 
```suggestion
        if ($this->languageService !== null) {
```

models/classes/user/UserSettingsFormFactory.php
```diff
@@ -0,0 +1,74 @@
...
+        if (!empty($userSettings->getUILanguageCode())) {
+            $fields['ui_lang'] = $userSettings->getUILanguageCode();
+        }
+
+        if (!empty($userSettings->getDataLanguageCode())) {
+            $fields['data_lang'] = $userSettings->getDataLanguageCode();
+        }
+
+        $options = [
+            tao_helpers_form_FormContainer::CSRF_PROTECTION_OPTION => (bool) ($params['useCSRFProtection'] ?? true),
```
shpran, line 63: I advise using constants for parameters 🙂
This will be more visible for users and much easier to use them in all places instead of simple text

test/unit/user/UserSettingsFormFactoryTest.php
```diff
@@ -0,0 +1,224 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA.
+ */
+
+namespace oat\tao\test\unit\models\classes\service;
```
shpran, line 21:
```suggestion
 */

declare(strict_types=1);

namespace oat\tao\test\unit\models\classes\service;
```

test/unit/user/UserSettingsFormFactoryTest.php
```diff
@@ -0,0 +1,224 @@
...
+use oat\tao\model\user\UserSettingsFormFactory;
+use tao_models_classes_LanguageService;
+use core_kernel_classes_Resource;
+use core_kernel_classes_Class;
+use core_kernel_persistence_smoothsql_SmoothModel;
+use ReflectionProperty;
+
+/**
+ * @runClassInSeparateProcess since this test changes global state
+ */
```
shpran, line 40: Maybe I'm wrong, but as I know, Unit tests should not change global state

test/unit/user/UserSettingsFormFactoryTest.php
```diff
@@ -0,0 +1,224 @@
...
+        );
+
+        $lang->setModel($this->ontologyMock);
+        $lang->setLabel($name);
+        $lang->setPropertyValue(
+            $this->ontologyMock->getProperty(OntologyRdf::RDF_VALUE),
+            $code
+        );
+
+        return $lang;
```
shpran, line 222: I'm not sure about this part.
We should not create any instances and set their values.
If you need them - just mock and mock their necessary methods to get something (like label, model and other things).
If you want - we may discuss this one 🙂

models/classes/user/UserSettingsFormFactory.php
hectoras: I was allowing null because it was not obvious for me how to provide a LanguageService instance through from the service provider, but at the same time I need to be able to inject it so I can mock it in tests.

However, after spending some time with this I figured out a way to make it work: We need to register the language service in a service provider, then we can use it on the service provider for the UserSettings service. I'll commit a fix for this shortly.

hectoras: Here it is :point_right: e1df61255ade18212d9845ce90d1c1de1fda8dcf :slightly_smiling_face:

models/classes/user/UserSettingsFormFactory.php
hectoras: This comparison has been removed altogether, marking this thread as solved.

models/classes/user/UserSettingsFormFactory.php
hectoras: You are right, I'm adding a public constant for this here: 908eee95cfd8ec9ad193fac35823fed7b158ddaf Thanks!

test/unit/user/UserSettingsFormFactoryTest.php
hectoras: You are absolutely right. I found some tricky issues related with the existing functionality, so I needed to be _creative_ to test this thing. 

First of all: I needed to change the singleton instance for the `ServiceManager` (which is a static variable of the `ServiceManager` class itself) to define a couple of services the form factory that are then fetched from the User Settings form by calling ServiceManager::getServiceManager(). That call to the ServiceManager from my code returns the singleton instance I'm mocking, so that's part of the global state I'm changing.

Apart from that, I'm also doing a nasty trick with reflection to manually initialize the `WidgetRegistry` cache (if I don't, the cache is `null` for some reason, and calls done from classes we are extending ourselves from make it crash. Under normal circumstances it should have been initialized to an empty array or some sort of iterable object at some point, but it seems for tests it is not.

Introducing changes in the `WidgetRegistry` code seems scary :sweat_smile: (and a bit out of scope for this task) so here I'm using reflection to initialize its cache to a value that  makes it work (i.e. an empty array). And, as I'm doing something I should not (modifying global state), I'm keeping these tests on its own process to prevent side effects on other tests.

So, this keeps other tests isolated from changes to `WidgetRegistry` and the `ServiceManager` this test makes.

test/unit/user/UserSettingsFormFactoryTest.php
hectoras: You're right, changed to use a mock in 3feb1f984cf2f6662f46b7a7da2f750867281d9e

test/unit/user/UserSettingsFormFactoryTest.php
gabrielfs7: There are some parts in tao you cannot test without making static content to be injectable in the target class. Sometimes we cannot test due a compromise. 

If this test is changing global state we should not have it.

hectoras: I'll try to make the actual form class returned by `UserSettingsFormFactory` configurable in order to then be able to use a mock class on the tests that does not call any `tao_helpers_form_FormContainer` legacy methods (so it does not change global state).

models/classes/user/UserSettingsFormFactory.php
gabrielfs7: It is not forbidden to pass objects inside the array. I am closing cause this is not mandatory :) 

Thanks!

test/unit/user/UserSettingsFormFactoryTest.php
hectoras: Okay, we've changed our approach. After a conversation with @gabrielfs7 we've decided to split this method in two (one to get the form fields, and another one to get the form options), and test the values returned by those that would be used to build the form itself :point_right: 5b1fc9f8131a26e65a22901979b32e0ecfe528c9

This way we test everything but except the actual call to the form builder (that is, the logic itself from this class is tested (_conditional branching code_)). No more global state changes from this test :tada:

models/classes/user/implementation/UserSettingsService.php
hectoras: Added: a56da1fc289200b6db60cab09b63f3479867b29a :+1:

test/unit/user/UserSettingsFormFactoryTest.php
hectoras: I'm marking this thread as solved. Please, reopen it if you think there are more things to address for this specific issue. Thanks!

shpran (approved):

actions/class.UserSettings.php
hectoras: This was already handled, marking this thread as solved

actions/form/class.UserSettings.php
hectoras: `instanceof` checks added in 2ee916525e1472405ff3076cbbda66d53f2afe59

hectoras:
Rebased with the `develop` branch

hectoras:
This pull request has 2 approvals and QA verified these changes. Therefore, I'm merging this pull request.

hectoras:
Released as [v48.74.1](https://github.com/oat-sa/tao-core/releases/tag/v48.74.1).

PR #3270, opened 2021-12-07
Feature/ADF-759/Efficient loading of lists with huge amount of elements

shpran:
# [ADF-759](https://oat-sa.atlassian.net/browse/ADF-759)

## Companion PRs
* https://github.com/oat-sa/extension-tao-backoffice/pull/116

## Changes
* Added `setTotalCount`/`getTotalCount` methods to `ValueCollection`;
* Added `hasOffset`/`setOffset`/`getOffset` methods to `ValueCollectionsSearchRequest`;
* Added new specification to check if class is editable or not;
* Added new specification to check if class is list or not;
* Added possibility to filter result (tested on local/remote list elements) by offset;
* Changed request to get list elements:
  * Now method is `GET` instead of `POST`;
  * Now `...` element is adding on FE side by checking for the list type (remote or not);
  * Now request is using new response.

## Tasks
- [ ] Add/update unit tests

### DIFF
--- a/models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
+++ b/models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
@@ -38,6 +38,9 @@ class ValueCollectionSearchRequest
     /** @var string[] */
     private $excluded = [];
 
+    /** @var int */
+    private $offset = 0;
+
     /** @var int|null */
     private $limit;
 
@@ -143,6 +146,23 @@ class ValueCollectionSearchRequest
     public function addExcluded(string $excluded): self
     {
         $this->excluded[] = $excluded;
+
+        return $this;
+    }
+
+    public function hasOffset(): bool
+    {
+        return $this->offset > 0;
+    }
+
+    public function getOffset(): int
+    {
+        return $this->offset;
+    }
+
+    public function setOffset(int $offset): self
+    {
+        $this->offset = $offset;
 
         return $this;
     }
--- a/models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
+++ b/models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
@@ -38,8 +38,8 @@ class ValueCollectionSearchRequest
     /** @var string[] */
     private $excluded = [];
 
-    /** @var int */
-    private $offset = 0;
+    /** @var int|null */
+    private $offset;
 
     /** @var int|null */
     private $limit;
@@ -152,10 +152,10 @@ class ValueCollectionSearchRequest
 
     public function hasOffset(): bool
     {
-        return $this->offset > 0;
+        return $this->offset !== null && $this->offset >= 0;
     }
 
-    public function getOffset(): int
+    public function getOffset(): ?int
     {
         return $this->offset;
     }
--- a/models/classes/Lists/Business/Specification/LocalListClassSpecification.php
+++ b/models/classes/Lists/Business/Specification/LocalListClassSpecification.php
@@ -25,24 +25,35 @@ namespace oat\tao\model\Lists\Business\Specification;
 use core_kernel_classes_Class;
 use oat\generis\model\GenerisRdf;
 use oat\tao\model\Specification\ClassSpecificationInterface;
+use oat\tao\model\Lists\Business\Service\RemoteSourcedListOntology;
 
 class LocalListClassSpecification implements ClassSpecificationInterface
 {
     /** @var ClassSpecificationInterface */
-    private $remoteListClassSpecification;
+    private $listClassSpecification;
 
-    public function __construct(ClassSpecificationInterface $remoteListClassSpecification)
+    public function __construct(ClassSpecificationInterface $listClassSpecification)
     {
-        $this->remoteListClassSpecification = $remoteListClassSpecification;
+        $this->listClassSpecification = $listClassSpecification;
     }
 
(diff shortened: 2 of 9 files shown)


views/js/uiForm.js
```diff
@@ -728,28 +728,34 @@
...
                 if (classUri && classUri.trim()) {
-                    $this.parent("div").children("div.form-error").remove();
+                    $this.parent('div').children('div.form-error').remove();
+
                     $.ajax({
                         url: context.root_url + 'taoBackOffice/Lists/getListElements',
-                        type: "POST",
-                        data: {listUri: classUri},
-                        dataType: 'json',
+                        type: 'GET',
```
hectoras, line 743: _For reference: This matches [this change in the controller](https://github.com/oat-sa/extension-tao-backoffice/blob/8df36234c7e2ee6013bd834e206b27f535b745b1/controller/Lists.php#L234) to explicitly expect this to be passed as a GET parameter._

models/classes/Lists/Business/Specification/LocalListClassSpecification.php
```diff
@@ -25,26 +25,24 @@
...
-            || $this->isBoolean($class)
-        ) && $this->isNotRemoteList($class);
+        $this->remoteListClassSpecification = $remoteListClassSpecification;
     }
 
-    private function isBoolean(core_kernel_classes_Class $class): bool
+    public function isSatisfiedBy(core_kernel_classes_Class $class): bool
     {
-        return $class->getUri() === GenerisRdf::GENERIS_BOOLEAN;
+        return $this->isBoolean($class) || !$this->remoteListClassSpecification->isSatisfiedBy($class);
```
gabrielfs7, line 41: - [ ] We need to revert this, this will bring a false positive in case of any class that is not a remote list (including Item class, Test class, etc)

gabrielfs7 (changes requested):

models/classes/Lists/Business/Specification/LocalListClassSpecification.php
shpran: Are you sure?
We don't have remote Boolean lists. In this case you f class is Boolean - it's local list.
Or I'm wrong?

gabrielfs7: Try to call this specification passing an object of class `http://www.tao.lu/Ontologies/TAOItem.rdf#Item` and you will see it returns true when should not, cause anything will be different of `$this->remoteListClassSpecification->isSatisfiedBy($class)`

hectoras (approved):

views/js/uiForm.js
```diff
@@ -728,28 +728,38 @@
...
-                            let html = "<ul class='form-elt-list'>",
+                            let html = '<ul class="form-elt-list">',
                                 property;
-                            for (property in response) {
-                                if(!response.hasOwnProperty(property)) {
+
+                            for (property in response.data.elements) {
+                                if (!response.data.elements.hasOwnProperty(property)) {
                                     continue;
                                 }
```
hectoras, line 754: Suggestion: Maybe it would be slightly better to use `Object.prototype.hasOwnProperty.call()` instead of calling the method directly on the object instance (ESLint would probbaly complain about this). See https://dev.to/aman_singh/what-s-the-deal-with-object-prototype-hasownproperty-call-4mbj,  https://stackoverflow.com/questions/12017693 & https://eslint.org/docs/rules/no-prototype-builtins

```suggestion
                                if (!Object.prototype.hasOwnProperty.call(response.data.elements, property)) {
                                    continue;
                                }
```

models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
```diff
@@ -147,6 +150,23 @@ public function addExcluded(string $excluded): self
         return $this;
     }
 
+    public function hasOffset(): bool
+    {
+        return $this->offset > 0;
```
gabrielfs7, line 155: 0 is a valid offset, I did not get your validation. Suggestion: 

```suggestion
        return $this->offset >= 0;
```

An you can make the default null

models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
```diff
@@ -38,6 +38,9 @@ class ValueCollectionSearchRequest
     /** @var string[] */
     private $excluded = [];
 
+    /** @var int */
+    private $offset = 0;
```
gabrielfs7, line 42: According to this comment https://github.com/oat-sa/tao-core/pull/3270/files#r771116688

```suggestion
    private $offset = null;
```

gabrielfs7 (approved):

views/js/uiForm.js
shpran: Thanks for your suggestion 🙂
@bziondik @adriloma , can you please check this?

models/classes/Lists/Business/Domain/ValueCollectionSearchRequest.php
shpran: Thanks, added. A bit modified your solution by additional check for null:
```php
return $this->offset !== null && $this->offset >= 0;
```
Because `null >= 0` is `true` - just to avoid this 🙂

views/js/uiForm.js
bziondik: yes, it is correct

AntonNeviarouski:
QA Verified

PR #3274, opened 2021-12-09
feat: migrate controller to DI container

gabrielfs7:
https://oat-sa.atlassian.net/browse/ADF-719

Tutorial: 

https://github.com/oat-sa/generis/blob/feature/ADF-719%2Fcontainer-for-actions/core/DependencyInjection/README.md#actionscontrollers-as-part-of-di-container

related:

- [ ] https://github.com/oat-sa/generis/pull/949


models/classes/routing/Service/ActionFinder.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        $params = [];
+
+        foreach ($constructorParameters as $parameter) {
+            $paramClass = $parameter->getClass();
+
+            if (!$paramClass) {
+                continue;
+            }
```
hectoras, line 64: I think, in case this happens, it does not add the required value to `$params`, so then `newInstanceArgs` will call the constructor with a wrong number of args (which would probably fail): Would it will be better to just `return null` if this happens (and maybe log an error)?

models/classes/routing/Service/ActionFinder.php
```diff
@@ -0,0 +1,94 @@
...
+        $constructorParameters = $reflectionClass->getConstructor()->getParameters();
+
+        if (count($constructorParameters) < 1) {
+            return null;
+        }
+
+        $params = [];
+
+        foreach ($constructorParameters as $parameter) {
+            $paramClass = $parameter->getClass();
```
hectoras, line 60: It seems getClass() [is deprecated](https://www.php.net/manual/en/reflectionparameter.getclass.php). 

Probably this can be replaced by usages of [getType](https://www.php.net/manual/en/reflectionparameter.gettype.php) (which unfortunately does not seems as straightforward to use as `getClass()`). Maybe something like this would work (please note I've not tested it):

```
if(PHP_MAJOR_VERSION >= 7) {
  $paramType = $parameter->getType();
  if($paramType == null) { // Maybe if($paramType instanceof ReflectionNaedType) would be better?
    return null; // Cannot call constructor with missing parameters
  } 

  $paramClassName = $paramType->getName();
} else {
  // Former code
}
```

models/classes/routing/Service/ActionFinder.php
gabrielfs7: I agree. I logged as a `debug`, cause although for DI we do not support, we might support for legacy algorithm.

models/classes/routing/Service/ActionFinder.php
```diff
@@ -0,0 +1,115 @@
...
+            ? $className::SERVICE_ID
+            : $className;
+    }
+
+    private function isAutoWiredController(string $className, string $serviceId): bool
+    {
+        return $className === $serviceId
+            && class_exists($className)
+            && method_exists($className, '__construct')
+            && !in_array(Controller::class, class_implements($className), true);
```
hectoras, line 113: Suggestion: Maybe `is_subclass_of($className, Controller::class)` would be a bit more readable.

hectoras (approved):

test/unit/model/routing/Service/ActionFinderTest.php
```diff
@@ -0,0 +1,118 @@
...
+    public function __construct(DummyDependencyClass $DummyDependencyClass)
+    {
+        $this->dummyDependencyClass = $DummyDependencyClass;
+    }
+
+    public function getDummyDependencyClass(): DummyDependencyClass
+    {
+        return $this->dummyDependencyClass;
+    }
+}
```
shpran, line 49: Is it possible to use anonymous classes?

test/unit/model/routing/Service/ActionFinderTest.php
```diff
@@ -0,0 +1,118 @@
...
+class ActionFinderTest extends TestCase
+{
+    /** @var ActionFinder */
+    private $sut;
+
+    /** @var MockObject|ContainerInterface */
+    private $container;
+
+    /** @var MockObject|LoggerInterface */
+    private $logger;
```
shpran, line 60: As I see, we are not using this class variable - so maybe we can replace this one with a local one?

test/unit/model/routing/Service/ActionFinderTest.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public function testFindNotLegacyControllerWillReturnObject(): void
+    {
+        $dummyDependencyClass = new DummyDependencyClass();
+
+        $this->container
+            ->method('has')
+            ->willReturnCallback(
+                function (string $class) use ($dummyDependencyClass) {
```
shpran, line 102:
```suggestion
                function (string $class) use ($dummyDependencyClass): bool {
```

shpran (approved):

test/unit/model/routing/Service/ActionFinderTest.php
gabrielfs7: I tried, in this case I cant, cause it needs to be autoloaded.

test/unit/model/routing/Service/ActionFinderTest.php
gabrielfs7: done

test/unit/model/routing/Service/ActionFinderTest.php
gabrielfs7: Done

models/classes/routing/Service/ActionFinder.php
gabrielfs7: Done, thanks!

PR #3293, opened 2021-12-21
feat: set max list title length, add validations and logging

shpran:
# [ADF-885](https://oat-sa.atlassian.net/browse/ADF-885)

## Companion PRs
* https://github.com/oat-sa/extension-tao-backoffice/pull/122


models/classes/Lists/Business/Domain/ValueCollection.php
```diff
@@ -39,6 +39,9 @@ class ValueCollection implements IteratorAggregate, JsonSerializable, Countable
 
     private $totalCount = 0;
 
+    /** @var bool */
+    private $isReloading = false;
```
gabrielfs7, line 43: It seems that this attribute does not belong to this VO, but to some client that needs to know in which context it was generated. I would not keep it here.

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -122,15 +122,15 @@ public function findAll(ValueCollectionSearchRequest $searchRequest): ValueColle
     public function persist(ValueCollection $valueCollection): bool
     {
         if ($valueCollection->hasDuplicates()) {
-            throw new ValueConflictException("Value Collection {$valueCollection->getUri()} has duplicate values.");
+            throw new ValueConflictException(
+                sprintf('List "%s" has duplicate values.', $valueCollection->getUri()),
+                __('List "%s" has duplicate values.', $valueCollection->getUri())
```
gabrielfs7, line 127: Is there a way to point what are the duplicated values? At least the first one? So the user can take action :) Wdyt?

gabrielfs7, line 127:
```suggestion
                sprintf('List "%s" has duplicated values.', $valueCollection->getUri()),
                __('List "%s" has duplicated values.', $valueCollection->getUri())
```

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -144,22 +144,35 @@ public function persist(ValueCollection $valueCollection): bool
 
             return true;
         } catch (ValueConflictException $exception) {
-            throw $exception;
+            throw new ValueConflictException($exception->getMessage());
```
gabrielfs7, line 147:
```suggestion
            throw $exception;
```

models/classes/Lists/Business/Domain/ValueCollection.php
shpran: Done. Previously, I used this to check if we are reloading a list or creating a new one (to avoid `uniqueness` check).
But now I'm using existing URI list to check this uniqueness.

Fixed.

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
shpran: At the moment - nope.
But we can refactor a bit `hadDuplicates` method.
So instead of checking for the first duplications - we can collect all the duplications and check if this `array` is empty or not.
And then we can use this array of duplications 🙂 

Wdyt?

models/classes/Lists/Business/Domain/ValueCollection.php
```diff
@@ -56,32 +56,45 @@ public function getUri(): ?string
     public function addValue(Value $value): self
     {
         $value = $this->ensureValueProperties($value);
-
-        if ($value->getUri() === '') {
-            $this->values[] = $value;
-        } else {
-            $this->values[$value->getUri()] = $value;
-        }
+        $this->values[] = $value;
```
hectoras, line 59: Would it be easier to just have a 2D array, that is, have it indexed by URI as before but have each array element as an array of values?  That is, here you would do something like:

```
if (!isset($this->values[$value->getUri()]) {
  $this->values[$value->getUri()] =  [];
}

$this->values[$value->getUri()][] = $value
```

Then `extractValueByUri()` would do something like

```
if (isset($this->values[$value->getUri()]) {
  return current($this->values[$value->getUri()]);
}
```

`getUris()` would just do (assuming you don't need to return the duplicate URIs more than once): `return array_keys($this->values);`

And `getDuplicatedValues()` would do something like this (avoiding the `O(n²)` implementation (that is, calling `extractValueByUri()`, which is `O(n)`, `n` times):

```
public function getDuplicatedValues(int $limit = null): ValueCollection
    {
        $duplicates = new ValueCollection();
        $counter = 0;

        foreach ($this->values as $valuesForUri) {
           // This only is executed with URIs with actual duplicates (we have an smaller n here)
           if (count($valuesForUri) > 1) { // If this URI has duplicates...
                // I think the first one should not be added, should it?
                foreach(array_slice($valuesForUri, 1) as $value) { // .... iterate its values except the first one
                        $duplicates->addValue($value); // .... and add them to the set of duplicates

                        if ($limit !== null && ++$counter >= $limit) { // If we reached the max number of items...
                             break 2; // .. exit this loop and the parent one
                        }
                }
            }
        }

        return $duplicates;
    }
```

Could it be done that way? _(Keep in mind maybe I'm missing something, I'm not too familiar with this component; and maybe more parts of this class would need to be changed too)_

gabrielfs7: I agree with you, but since Andrei is not here this week and this code is working, I will not change it right now. Lets see if we have time, we can apply this as an improvement :) 

Thanks!

gabrielfs7 (approved):

hectoras (approved):
There are some minor improvements that could be made to make the code simpler, but LGTM overall.

PR #3295, opened 2021-12-22
fix: TR-3163. Restore focusing functionality

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-3163

tao-test-runner-qti-fe repository has broken tests. Root cause of this - styles changes which do not allow focus chcheckboxes programmatically. Current fix - updating styles based on changes in this PR: https://github.com/oat-sa/tao-core-ui-fe/pull/406

How to test: do it in terms of pointed PR


jsconan (commented):
What is actually changed here?
The CSS should be compiled with all PR having a change either in a npm dependency or in the SCSS. Then it will be also checked and compiled by the auto-release process.

I don't see any update neither in the npm dependencies nor in the SCSS... Is it because a previous PR/release did not update properly the CSS? If so, it will be overwritten by the next one.

Sorry, but for me this PR does not make sense, unless you forgot to commit the root change.

PR #3303, opened 2021-12-28
feat: Optional test for max number of items in ValueCollectionService::persist()

hectoras:
**Associated Jira issue:** [ADF-788]( https://oat-sa.atlassian.net/browse/ADF-788)

Feature needed by https://github.com/oat-sa/extension-tao-backoffice/pull/121

### DIFF
--- a/models/classes/Lists/Business/Service/ValueCollectionService.php
+++ b/models/classes/Lists/Business/Service/ValueCollectionService.php
@@ -84,7 +84,6 @@ class ValueCollectionService extends InjectionAwareService
 
     /**
      * @param ValueCollection $valueCollection
-     * @param int $maxItems
      *
      * @return bool
      *
--- a/models/classes/Lists/Business/Service/ValueCollectionService.php
+++ b/models/classes/Lists/Business/Service/ValueCollectionService.php
@@ -42,6 +42,9 @@ class ValueCollectionService extends InjectionAwareService
     /** @var ValueCollectionRepositoryInterface */
     private $repositories;
 
+    /** @var int */
+    private $maxItems = 0;
+
     public function __construct(ValueCollectionRepositoryInterface ...$repositories)
     {
         parent::__construct();
@@ -88,9 +91,9 @@ class ValueCollectionService extends InjectionAwareService
      * @throws ValueConflictException
      * @throws OverflowException
      */
-    public function persist(ValueCollection $valueCollection, int $maxItems = 0): bool
+    public function persist(ValueCollection $valueCollection): bool
     {
-        if ($maxItems > 0 && $valueCollection->count() > $maxItems) {
+        if ($this->maxItems > 0 && $valueCollection->count() > $this->maxItems) {
             throw new OverflowException("Collection exceeds the allowed number of items");
         }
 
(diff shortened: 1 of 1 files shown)


models/classes/Lists/Business/Service/ValueCollectionService.php
```diff
@@ -80,13 +81,19 @@ public function delete(ValueCollectionDeleteInput $input): void
...
      * @param ValueCollection $valueCollection
+     * @param int $maxItems
      *
      * @return bool
      *
      * @throws ValueConflictException
+     * @throws OverflowException
      */
-    public function persist(ValueCollection $valueCollection): bool
+    public function persist(ValueCollection $valueCollection, int $maxItems = 0): bool
```
gabrielfs7, line 91: What about centralize the control of max items? 

Example: Use `$this->getListService()->getMaxItems()` inside this class instead of a new parameter.

hectoras: I think this is used for remote lists as well.

[Lists::saveLists()](https://github.com/oat-sa/extension-tao-backoffice/blob/feature/ADF-788_maximum-elements/controller/Lists.php#L268) calls `ValueCollectionService::persist()` for local lists (with `maxItems` limit), but
 [Lists::sync()](https://github.com/oat-sa/extension-tao-backoffice/blob/feature/ADF-788_maximum-elements/controller/Lists.php#L190) also calls it for remote lists (with a `ValueCollection` obtained from a `RemoteSource`). I think in the later case we don't want to enforce the limit (as they are coming from a 3rd party), do we?
 
Maybe we can do that if we record the fact it is a remote list inside the `ValueColllection` somehow:
- `protected bool ValueColllection::$isRemote` maybe? But I don't think adding info about the source being local or remote to this domain entity would be a good idea
- Maybe having a subclass of `ValueCollection` for remote lists? But again it would add info about the data source to the domain layer

Another option would be having a separate operation for `persist()` in the service to deal with remote lists. We'd be adding info about the source being local or remote to the domain layer, but not to the collection itself. That may work, and would remove the max number of items from the controller on the persist operation (this is the option I like the most).

BTW keep in mind the controller still needs to know the limit (i.e. call `$this->getListService()->getMaxItems()`) in order to set up the max items limit for the frontend Javascript logic).

gabrielfs7: Maybe add to the `RdfValueCollectionRepository::persist()` method then?

gabrielfs7: Keeping this second parameter is like partially delegating the validation rules to the client of the class. It is like having a method `UserRepository::save(User $user, int $maxLoginLength)`.

hectoras: Sorry for the delay in the response, yesterday afternoon I was testing locally how to do this. Right now:

- I have removed the parameter from `ValueCollectionService::persist()`, as well as made additional changes to inject `SessionService` and `UserLanguageServiceInterface` on its constructor to prevent extending from the deprecated `InjectionAwareService` class
-  I'm passing `int $maxItems = 0` to `RdfValueCollectionRepository` constructor. To do so, I'm registering both `RdfValueCollectionRepository` and `ValueCollectionService` in `ListsServiceProvider`.

I'll try to commit those changes soon.

gabrielfs7: It is too complicated and List management is done in tao-core and tao-backoffice in a way that we cannot refactor right now.

We are closing the PR

gabrielfs7 (approved):

hectoras:
This was automatically released as [v48.74.0](https://github.com/oat-sa/tao-core/releases/tag/v48.74.0).

PR #3312, opened 2022-01-04
feat: Do not ksort list items for types that are already sorted

hectoras:
**Associated Jira task:** [ADF-870](https://oat-sa.atlassian.net/browse/ADF-870)

`ElementMapFactory` calls `Property::getRange()` to get the allowed values (list items) for list-type properties.

This pull request makes `ElementMapFactory` to skip the sorting step (that is, the call to `ksort()`) for lists items if they belong to a list with a type among a pre-defined set of types (controlled by a new `PresortedListSpecification` class).

## Tasks
- [ ] [Add unit tests](https://oat-sa.atlassian.net/browse/ADF-922)


helpers/form/ElementMapFactory.php
```diff
@@ -23,6 +23,8 @@
 namespace oat\tao\helpers\form;
 
 use common_Logger;
+use oat\tao\model\Lists\Business\Specification\PresortedListSpecification;
```
mike-ionut-mihai-sandu-tao, line 26: can you optimize the imports please ,  as i assume the addon was just done on top

helpers/form/ElementMapFactory.php
```diff
@@ -98,6 +100,7 @@ public function create(core_kernel_classes_Property $property): ?tao_helpers_for
             FeatureFlagChecker::FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED
         );
 
+        $parentProperty = null;
```
mike-ionut-mihai-sandu-tao, line 103: why ?

helpers/form/ElementMapFactory.php
```diff
@@ -163,11 +166,16 @@ public function create(core_kernel_classes_Property $property): ?tao_helpers_for
                             }
                         }
                     }
-                    ksort($options);
+
+                    if (!$this->getPresortedListSpecification()->isSatisfiedBy($property)) {
+                        ksort($options);
```
mike-ionut-mihai-sandu-tao, line 171: why do we need the ksort at all ?

helpers/form/ElementMapFactory.php
andreluizmachado, line 103: I believe that you do not need to initiate this variable here as you already do in line 105. Could you check it, please?

models/classes/Lists/Business/Specification/PresortedListSpecification.php
```diff
@@ -0,0 +1,84 @@
...
+use tao_helpers_form_elements_Radiobox;
+use tao_helpers_form_elements_Combobox;
+use tao_helpers_form_elements_Checkbox;
+use core_kernel_classes_Class;
+
+/**
+ * Presorted lists are lists that are considered "already sorted", so UI
+ * components can skip sorting the options for them.
+ */
+class PresortedListSpecification implements PropertySpecificationInterface
```
mike-ionut-mihai-sandu-tao, line 35: so instead of a ksort we are creating a new service that will setup a new core_kernel_classes_Class inside the constructor
and that does a very expensive get `getRange()` and some other logic 

i think this is a bit too much and will cost and decrease the performance

mike-ionut-mihai-sandu-tao (changes requested):
it looks like a nice coded solution but we need to think about performance also , 
so please reconsider and do not replace ksort with a functionality that has very expensive calls

models/classes/Lists/Business/Specification/PresortedListSpecification.php
```diff
@@ -0,0 +1,84 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA.
+ */
+
```
andreluizmachado, line 20: is it possible to add strict_types declaration?

test/unit/models/classes/Lists/Business/Specification/PresortedListSpecificationTest.php
```diff
@@ -0,0 +1,192 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA;
```
andreluizmachado, line 18: 2022

test/Asset/CustomRootClassFixture.php
```diff
@@ -0,0 +1,36 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA.
+ */
+
+namespace oat\tao\test\Asset;
+
+use core_kernel_classes_Class;
+
+class CustomRootClassFixture extends core_kernel_classes_Class
```
andreluizmachado, line 25: Do we really need to create this stub class? Can't it be done by using mockBuilder? maybe an anonymous class?

helpers/form/ElementMapFactory.php
shpran: I think he added this to avoid a possible issue at [line 144](https://github.com/oat-sa/tao-core/blob/d61675218da1ff0a9cfb58562817bbf86d791caa/helpers/form/ElementMapFactory.php#L144)

As you can see there, we use this `$parentProperty`, but previously we initialized this one inside if-statement, so sometimes our property was not exists and might fail at this line.

models/classes/Lists/Business/Specification/PresortedListSpecification.php
shpran: https://github.com/oat-sa/tao-core/pull/3312#discussion_r783749039

helpers/form/ElementMapFactory.php
shpran: https://github.com/oat-sa/tao-core/pull/3312#discussion_r783749039

helpers/form/ElementMapFactory.php
shpran: I'm not sure, that we need this one, but I decided to keep this one to avoid any unexpected issues.

At the moment, I removed specification and moved `ksort` inside the if statement.
So, we will not use `ksort` for lists (local or remote), but we will keep this one for other properties, like `Item Model`, `Roles`, etc.
(In accordance with the task ACs)

Pint @mike-ionut-mihai-sandu-tao

helpers/form/ElementMapFactory.php
```diff
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2020-2021 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2020-2022 (original work) Open Assessment Technologies SA;
```
andreluizmachado, line 18: Do we already have automated tests covering these changes?

helpers/form/ElementMapFactory.php
```diff
@@ -162,12 +165,16 @@ public function create(core_kernel_classes_Property $property): ?tao_helpers_for
...
+
+                        ksort($options);
                     }
-                    ksort($options);
+
                     $sortedOptions = [];
-                    foreach ($options as $id => $values) {
+
+                    foreach ($options as $values) {
                         $sortedOptions[$values[0]] = $values[1];
```
andreluizmachado, line 175: What is value[0] and value[1]? shouldn't we validate if they are set or not?

helpers/form/ElementMapFactory.php
shpran: Nope, we didn't have any tests covering sorting

helpers/form/ElementMapFactory.php
shpran: Updated - now these variables are named.
About validation - there's no need, because we are filling this `$options` a bit upper 🙂

andreluizmachado (approved):

bartlomiejmarszal (changes requested):
Why there is no unit tests?

shpran:
> Why there is no unit tests?

Will be covered by unit tests in a separate task to avoid blocking the task, as this one is a sprint goal.
https://oat-sa.atlassian.net/browse/ADF-922

bartlomiejmarszal (approved):
LGTM

helpers/form/ElementMapFactory.php
hectoras: `ksort` is the former behavior.

Here I was keeping the call to `ksort()` for widget types that are not listed in the acceptance criteria from the ticket: For what I remember, the change to keep the former order should be applied only to:

-  _Single choice lists with Radio buttons_ ("list", `tao_helpers_form_elements_Radiobox::WIDGET_ID`)
-  _Single choice lists with Drop downs_ ("longlist", `tao_helpers_form_elements_Combobox::WIDGET_ID`) 
- `Multiple choice lists with Check box` ("multilist", `tao_helpers_form_elements_Checkbox::WIDGET_ID`)

My idea was to use `PresortedListSpecification` to check if the list was for one of those types and, only in that case, skip the `ksort` call.

models/classes/Lists/Business/Specification/PresortedListSpecification.php
hectoras: Please note `getRange()` call was already being called by the former code, here I was just moving that call along with the logic for what should and what should not be sorted to a separate class.

AntonNeviarouski:
QA verified

helpers/form/ElementMapFactory.php
hectoras: My former idea was to move the logic that checks if a list has to be sorted to a specification class (`PresortedListSpecification`) and test that part independently.

PR #3315, opened 2022-01-07
Feature/tr 3180/review panel item button list component

oatymart:
https://oat-sa.atlassian.net/browse/TR-3180

This PR is just to apply the branch from:
- https://github.com/oat-sa/tao-core-ui-fe/pull/409

TODO:
- [x] replace branch ref with published package version before merge


olga-kulish (approved):
Required dependency is updated. Will still need to use released version.

Github action should update js bundle, but you can also add bundle in PR.

PR #3339, opened 2022-01-17
Draft: feature/ADF-869: Delete previous statements for a property when edited with an empty value

hectoras:
**Associated Jira issue:** [ADF-869](https://oat-sa.atlassian.net/browse/ADF-869)

The fix itself is mostly contained in 7c43473e4eb7d976e4861f02f2ada6d54415dd14.

Additionally, this pull request adds unit tests for the whole data binder class, covering all class methods and branches except the static call to `ServiceManager::getServiceManager()`, that is now wrapped inside a private method to allow injecting the manager from tests.

Note empty values for arrays are still saved ([here](https://github.com/oat-sa/tao-core/pull/3339/files#diff-4d74295bd463ae29b09b9177d43f4d72316391c82320b9ab196ed4c8c9cee65dR145) and [here](https://github.com/oat-sa/tao-core/pull/3339/files#diff-4d74295bd463ae29b09b9177d43f4d72316391c82320b9ab196ed4c8c9cee65dR162)); that is, no filtering is done for arrays but only for string properties.

**Superseded by #3361**

---
Self-checks before moving to code review:

- [x] Pull request's target is not master


models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +64,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+                $instance->removePropertyValues($property);
+            } else {
+                $instance->editPropertyValues($property, $propertyValue);
+            }
+        }
+    }
+
+    private static function isEmptyValue(string $value): bool
+    {
+        return (('' == $value) || (' ' == $value) || strlen(trim($value)) == 0);
```
shpran, line 176: 1. Please, use strict comparison
2. There's no need to wrap these comparisons with brackets
3. Maybe we can use `empty(trim($value))` instead?
```php
return empty(trim($value));
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+
+        $this->emptyCollectionMock
+            ->method('count')
+            ->willReturn(0);
+
+        $this->sut =
+            new tao_models_classes_dataBinding_GenerisInstanceDataBinder(
+                $this->resource,
+                $this->eventManagerMock
+            );
```
shpran, line 91:
```suggestion
        $this->sut = new tao_models_classes_dataBinding_GenerisInstanceDataBinder(
            $this->resource,
            $this->eventManagerMock
        );
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    public function testBindScalarWithPreviousValue(): void
+    {
+        $this->eventManagerMock
+            ->expects($this->once())
+            ->method('trigger')
+            ->with($this->callback(function (MetadataModified $e): bool {
```
shpran, line 99: Please use a better variable name.

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+                return (
+                    $e->getResource()->getLabel() == $this->resource->getUri()
+                    && ($e->getMetadataUri() == self::URI_PROPERTY_1)
+                    && ($e->getMetadataValue() == 'Value 1'));
```
shpran, line 103: 1. Please, use strict comparison
2. There's no need to wrap these comparisons with brackets

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+            }));
+
+        $this->resource
+            ->expects($this->once())
+            ->method('setType')
+            ->with($this->callback(function (core_kernel_classes_Class $c) {
```
shpran, line 109: Please use a better name for the property.

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+                return ($c->getUri() == self::URI_TYPE_1);
```
shpran, line 110: 1. Please, use strict comparison
2. There's no need to wrap these comparisons with brackets

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+            }))
+            ->willReturn(true);
+
+        $this->resource
+            ->method('getTypes')
+            ->willReturn([
+                new core_kernel_classes_Class(self::URI_TYPE_1)
```
shpran, line 117: Please, return configured mock object

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+            ]);
+
+        // There is a previous value for prop1 and its new value is a scalar:
+        // The data binder should call editPropertyValues() on the resource.
+        //
+        $this->resource
+            ->method('getPropertyValuesCollection')
+            ->will($this->returnCallback(
+                function (core_kernel_classes_Property $property) {
+                    if ($property->getUri() == self::URI_PROPERTY_1) {
```
shpran, line 127: Please, use strict comparison

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+                        return $this->nonEmptyCollectionMock;
+                    }
+
+                    $this->fail("Unexpected property: {$property->getUri()}");
```
shpran, line 131:
```suggestion
                    $this->fail('Unexpected property: ' . $property->getUri());
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+                }
+            ));
+
+        $this->resource
+            ->expects($this->once())
+            ->method('editPropertyValues')
+            ->with($this->callback(
+                function (core_kernel_classes_Property $property, $_ = null) {
```
shpran, line 139: Please use a better variable name (`$_`).

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,572 @@
...
+                    return ($property->getUri() == self::URI_PROPERTY_1);
```
shpran, line 140: Please, use strict comparison.

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -18,48 +18,41 @@
...
      * @author Jerome Bogaerts, <jerome@taotesting.com>
-     * @param  Resource targetInstance The
-     * @return mixed
      */
-    public function __construct(core_kernel_classes_Resource $targetInstance)
-    {
-        
+    public function __construct(
+        core_kernel_classes_Resource $targetInstance,
+        EventManager $eventManager = null
```
shpran, line 52: I'm not sure about providing here a second parameter, as this class is initializable and fully relates to work with `instance`. So, this one will not be automatically configured by DI container.

As I understood, you added this one for testing.
Maybe using a method `withEventManager` would be better?

It's just a suggestion 🙂 
WDYT?

shpran (changes requested):
Great job!
Left some comments and suggestions 🙂

I didn't comment the whole test (only first test method), as it contain the same issues - please, fix them in accordance with the comments 👨🏻‍💻

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +64,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
-                            $prop,
-                            $propertyValue
-                        );
-                        if (strlen(trim($propertyValue)) == 0) {
-                            //if the property value is an empty space(the default value in a select input field), delete the corresponding triplet (and not all property values)
-                            $instance->removePropertyValues($prop, ['pattern' => '']);
-                        }
-                    }
+            foreach ($data as $propUri => $value) {
+                if (($propUri == OntologyRdf::RDF_TYPE) && (null != $value)) {
```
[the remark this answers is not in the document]
hectoras: Whoops, missed that one. Changed here :point_right:  3cc111e1933bf4656807826bdae50570174a949d, thanks! :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +64,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
-        } catch (common_Exception $e) {
-            $msg = "An error occured while binding property values to instance '': " . $e->getMessage();
-            $instanceUri = $instance->getUri();
-            throw new tao_models_classes_dataBinding_GenerisInstanceDataBindingException($msg);
+
+            return $instance;
+        } catch (Throwable $error) {
+            throw new tao_models_classes_dataBinding_GenerisInstanceDataBindingException(
+                sprintf(
+                    "Error binding property values to instance '%s': %s",
```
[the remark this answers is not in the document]
hectoras: You're right , it is better that way. Changed here :point_right: 3cc111e1933bf4656807826bdae50570174a949d, thanks! :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Yes, this was just for testing, since right now this class is used by instantiating it directly instead of using the DI container nor `ServiceManager`.

Changed in 55b97cd74ac968e3e79a0e98879fc3072fc92637 to inject the `EventManager` in tests with a new setter method (used only from tests).

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Changed to strict comparison (since a type coercion to string has already been done because of the type hint of `$value`), as well as unwrapped the expressions (removed the parenthesis) in 6c1d423b2096038d8351f1ada89d421523cfce60 :+1: Thanks!

Regarding the third point, I think we cannot use `empty()` in this method, since that would handle a literal zero (`"0"`) as an empty value, and we only want to handle as such strings containing only whitespaces. I've just checked just in case:

```
hector@hector-oat:~/snippets$ cat empty-with-zero-str.php 
<?php
$value = '0';

if (empty(trim($value))) {
	echo "Oh no :(" . PHP_EOL;
}
hector@hector-oat:~/snippets$ php empty-with-zero-str.php 
Oh no :(
hector@hector-oat:~/snippets$ 
```

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +64,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+            $instance->setType(new core_kernel_classes_Class($type));
+        }
+    }
+
+    private function bindProperty(
+        core_kernel_classes_Resource $instance,
+        string $propertyUri,
+        $newValue
+    ): void {
+        $prop = new core_kernel_classes_Property($propertyUri);
```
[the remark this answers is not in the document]
hectoras: Done! 5acd1b9a189ea65bcef831f5e6f898f55069cceb :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +64,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+        core_kernel_classes_Resource &$instance,
+        $propertyValue
+    ): void {
+        foreach ($instance->getTypes() as $type) {
+            $instance->removeType($type);
+        }
+
+        $types = !is_array($propertyValue) ? [$propertyValue] : $propertyValue;
+        foreach ($types as $type) {
+            $instance->setType(new core_kernel_classes_Class($type));
```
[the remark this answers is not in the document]
hectoras: Done! 5acd1b9a189ea65bcef831f5e6f898f55069cceb :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 28de603eecea5195605434e6dfc091abb57fc57b :ok_hand:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 28de603eecea5195605434e6dfc091abb57fc57b, but I'm not completely sure about this one. [PSR-12](https://www.php-fig.org/psr/psr-12/) states that "Lines SHOULD NOT be longer than 80 characters; lines longer than that SHOULD be split into multiple subsequent lines of no more than 80 characters each". 

The class name alone is already 56 chars long, so I think putting it in the next line is the only way to keep the lines short. By the way, putting it as a single line will make it 83 chars, so I'll do it that way as you suggested as it is still not too long.

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 28de603eecea5195605434e6dfc091abb57fc57b :ok_hand:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 28de603eecea5195605434e6dfc091abb57fc57b :ok_hand:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 28de603eecea5195605434e6dfc091abb57fc57b :ok_hand:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: TODO
- [ ] Add a unit test for checking "0" is not handled as an empty value

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 5b06cbb3c967251292b1811c866280f51a2c8306 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 5b06cbb3c967251292b1811c866280f51a2c8306 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed in 5b06cbb3c967251292b1811c866280f51a2c8306 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed: c8385d39e8a9cd846f26f0898e07f7ad93d9c77b :ok_hand:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Missed that one, changed in 55bef0e7ea387769500e5a420b626e2a178e7a8f. Thanks!

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Added a unit test in 0e59c66a3cd9f8d7cd447867fc32d8c7663c63c9 that checks if setting a property value to zero works as expected (i.e. calls `$target->setPropertyValue($property, '0')`).

As expected, changing the implementation for `isEmptyValue()` to use `empty(trim($value))` makes the test fail.

hectoras: I think things raised in this thread regarding coding standards and using strict comparison have been already addressed, and we've checked if we can use `empty()` instead of `strlen()` here (as well as added a unit test to ensure a literal zero is not handled as an empty value).

Therefore, I'm marking this thread as solved, but please feel free to reopen it if you think I'm missing something :smile:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+            $instance->removePropertyValues($property);
+
+            foreach ($propertyValue as $aPropertyValue) {
+                $instance->setPropertyValue($property, $aPropertyValue);
+            }
+        } elseif (is_string($propertyValue)) {
+            if (self::isEmptyValue($propertyValue)) {
+                // Setting an empty value for a scalar property deletes
+                // the statement
+                $instance->removePropertyValues($property);
```
hectoras, line 169: Note this is the primary change introduced by this pull request (initially committed in 7c43473e4eb7d976e4861f02f2ada6d54415dd14).

If the value is considered empty (an empty string, or a string containing only whitespaces) and we are editing a scalar field, we unconditionally remove the values for this property under this instance (previously we were doing so with `$instance->removePropertyValues($property, ['pattern' => ''])`, which doesn't work in Postgres as expected).

Other things should work as they did, this pull request just refactor them to be easier to follow.

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                    $this->fail('Unexpected property: ' . $property->getUri());
+                }
+            ));
+
+        $this->target
+            ->expects($this->once())
+            ->method('removePropertyValues')
+            ->with($this->callback(function (core_kernel_classes_Property $p) {
+                return $p->getUri() === self::URI_PROPERTY_1;
+            }));
```
shpran, line 321:
```suggestion
            ->with($this->callback(function (core_kernel_classes_Property $property) {
                return $property->getUri() === self::URI_PROPERTY_1;
            }));
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+    {
+        $this->eventManagerMock
+            ->expects($this->once())
+            ->method('trigger')
+            ->with($this->callback(function (MetadataModified $event): bool {
+                return (
+                    $event->getResource()->getUri() == $this->target->getUri()
+                    && $event->getMetadataUri() === self::URI_PROPERTY_1
+                    && $event->getMetadataValue() == 'Value 1');
+            }));
```
shpran, line 144: I saw, that you repeat this one multiple times. Maybe you can extract this into a separate method and use it instead?
WDYT?

1. There's no need to wrap such condition with brackets
2. Use strict comparison
```suggestion
            ->with(
                $this->callback(
                    function (MetadataModified $event): bool {
                        return $event->getResource()->getUri() === $this->target->getUri()
                            && $event->getMetadataUri() === self::URI_PROPERTY_1
                            && $event->getMetadataValue() === 'Value 1';
                    }
                )
            );
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+        $this->target
+            ->method('getPropertyValuesCollection')
+            ->will($this->returnCallback(
+                function (core_kernel_classes_Property $property) {
+                    if ($property->getUri() === self::URI_PROPERTY_1) {
+                        return $this->nonEmptyCollectionMock;
+                    }
+                    if ($property->getUri() === self::URI_PROPERTY_2) {
+                        return $this->nonEmptyCollectionMock;
+                    }
```
shpran, line 463:
```suggestion
                    if (in_array($property->getUri(), [self::URI_PROPERTY_1, self::URI_PROPERTY_2], true)) {
                        return $this->nonEmptyCollectionMock;
                    }
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                }
+            );
+
+        $this->target
+            ->expects($this->exactly(1))
+            ->method('removePropertyValues')
+            ->willReturnCallback(
+                function (core_kernel_classes_Property $p, $opts = []) {
+                    $this->assertEquals(self::URI_PROPERTY_1, $p->getUri());
+                }
```
shpran, line 485:
```suggestion
                function (core_kernel_classes_Property $property, $opts = []) {
                    $this->assertEquals(self::URI_PROPERTY_1, $property->getUri());
                }
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+            self::URI_PROPERTY_2 => 'Value 2',
+        ]);
+
+        $this->assertIsTargetInstance($resource);
+    }
+
+    public function testBindNewTypesToExistingInstance(): void
+    {
+        $this->eventManagerMock
+            ->expects(self::never())
```
shpran, line 502:
```suggestion
            ->expects($this->never())
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+            ->method('trigger');
+
+        $this->target
+            ->method('getTypes')
+            ->willReturn([
+                new core_kernel_classes_Class(self::URI_TYPE_1),
+            ]);
+
+        $this->target
+            ->expects(self::exactly(1))
```
shpran, line 512:
```suggestion
            ->expects($this->exactly(1))
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+            ->method('removeType')
+            ->with($this->callback(function (core_kernel_classes_Class $class) {
+                return $class->getUri() === self::URI_TYPE_1;
+            }))
+            ->willReturn(true);
+
+        $this->target
+            ->expects(self::exactly(1))
```
shpran, line 520:
```suggestion
            ->expects($this->exactly(1))
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+        $this->target
+            ->method('getProperty')
+            ->willReturnMap([
+                [self::URI_PROPERTY_1, $this->property1],
+                [self::URI_PROPERTY_2, $this->property2],
+            ]);
+
+        // There are no properties other than types for this test
+        $this->target
+            ->expects(self::never())
```
shpran, line 536:
```suggestion
            ->expects($this->never())
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+        $this->target
+            ->method('getPropertyValuesCollection')
+            ->will($this->returnCallback(
+                function (core_kernel_classes_Property $property) {
+                    if ($property->getUri() === self::URI_PROPERTY_1) {
+                        return $this->emptyCollectionMock;
+                    }
+                    if ($property->getUri() === self::URI_PROPERTY_2) {
+                        return $this->emptyCollectionMock;
+                    }
```
shpran, line 583:
```suggestion
                    if (in_array($property->getUri(), [self::URI_PROPERTY_1, self::URI_PROPERTY_2], true)) {
                        return $this->emptyCollectionMock;
                    }
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                }
+            ));
+
+        $this->target
+            ->expects($this->once())
+            ->method('setPropertyValue')
+            ->with(
+                $this->callback(
+                    function (core_kernel_classes_Property $property, $_value = null) {
+                        return ($property->getUri() === self::URI_PROPERTY_1);
```
shpran, line 658:
```suggestion
                        return $property->getUri() === self::URI_PROPERTY_1;
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                    $this->fail('Unexpected property: ' . $property->getUri());
+                }
+            ));
+
+        $this->target
+            ->expects($this->once())
+            ->method('setPropertyValue')
+            ->with(
+                $this->callback(
+                    function (core_kernel_classes_Property $property, $_value = null) {
```
shpran, line 657:
```suggestion
                    function (core_kernel_classes_Property $property, $value = null) {
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+            ]);
+
+        $this->target
+            ->method('getProperty')
+            ->with(self::URI_PROPERTY_1)
+            ->willReturn($this->property1);
+
+        $this->target
+            ->method('getPropertyValuesCollection')
+            ->willThrowException(new Exception("error", 123));
```
shpran, line 694:
```suggestion
            ->willThrowException(new Exception('error', 123));
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+
+        $this->expectException(
+            tao_models_classes_dataBinding_GenerisInstanceDataBindingException::class
+        );
+
+        $this->expectExceptionMessage(
+            "Error binding property values to instance"
+        );
```
shpran, line 702:
```suggestion
        $this->expectExceptionMessage('Error binding property values to instance');
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+
+        $this->assertIsTargetInstance($resource);
+    }
+
+    private function assertIsTargetInstance($resource): void
+    {
+        // Used to check that the binder returns the former instance with
+        // the changes applied
+        $this->assertSame($this->target, $resource);
+    }
```
shpran, line 716: Do we really need this method?
Not critical, you can keep this one if you want 🙂

shpran (changes requested):

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
-                        $instance->editPropertyValues(
-                            $prop,
-                            $propertyValue
-                        );
-                        if (strlen(trim($propertyValue)) == 0) {
-                            //if the property value is an empty space(the default value in a select input field), delete the corresponding triplet (and not all property values)
-                            $instance->removePropertyValues($prop, ['pattern' => '']);
-                        }
-                    }
+            foreach ($data as $propUri => $value) {
```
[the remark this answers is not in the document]
hectoras: I agree. I think this was shortened to keep the lines short (as we had more than 3 levels of indentation), but it is now unneeded since most of the logic like that has been moved to separate methods.

Changed in 0b71577f5446ddb457cc71cd8d70c3b003065872 :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+                    $error->getMessage()
+                ),
+                0,
+                $error
+            );
+        }
+    }
+
+    private function bindTypes(
+        core_kernel_classes_Resource &$instance,
```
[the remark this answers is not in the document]
hectoras: Removed pass-by-reference declaration in 0b71577f5446ddb457cc71cd8d70c3b003065872 :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+        $propertyValue
+    ): void {
+        foreach ($instance->getTypes() as $type) {
+            $instance->removeType($type);
+        }
+
+        $types = !is_array($propertyValue) ? [$propertyValue] : $propertyValue;
+        foreach ($types as $type) {
```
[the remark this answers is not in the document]
hectoras: Changed in 0b71577f5446ddb457cc71cd8d70c3b003065872 :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+                // Setting an empty value for a scalar property deletes
+                // the statement
+                $instance->removePropertyValues($property);
+            } else {
+                $instance->editPropertyValues($property, $propertyValue);
+            }
+        }
+    }
+
+    private static function isEmptyValue(string $value): bool
```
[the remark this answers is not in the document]
hectoras: Changed in 0b71577f5446ddb457cc71cd8d70c3b003065872 :+1:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,124 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
...
+    {
+        return '' === $value || ' ' === $value || strlen(trim($value)) == 0;
+    }
+
+    private function getEventManager(): EventManager
+    {
+        if ($this->eventManager == null) {
```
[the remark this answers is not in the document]
hectoras: Changed in 0b71577f5446ddb457cc71cd8d70c3b003065872 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                    $this->fail('Unexpected property: ' . $property->getUri());
+                }
+            ));
+
+        $this->target
+            ->expects($this->once())
+            ->method('setPropertyValue')
+            ->with(
+                $this->callback(
+                    function (core_kernel_classes_Property $property, $_value = null) {
```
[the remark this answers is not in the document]
hectoras: Changed: 4ce6e76e6ee334de1160502f796c2e59187e35c5 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed: 4ce6e76e6ee334de1160502f796c2e59187e35c5 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: This was put this way to conform to [PSR-12](https://www.php-fig.org/psr/psr-12/) soft limit on line length.

BTW I'm changing it in 47e54e9a5c0d33c8198b547bfad98cf9807cbb11.

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Missed that one, changed on d9d2097b62f2e37fadea731930defb60dd728246.

BTW I generally prefer being explicit when I'm using the result of a condition as a (boolean) value, or when combining multiple boolean expressions on conditional ("`if`") statements), for the reasons stated on this SO comment: https://softwareengineering.stackexchange.com/a/201187

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                        return ($property->getUri() === self::URI_PROPERTY_1);
```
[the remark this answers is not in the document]
hectoras: Changed in ac23cd5cb705abff473deca0cc3a43b5ac8a3868 (see https://github.com/oat-sa/tao-core/pull/3339#discussion_r790629100 for the rationale for wrapping it with parenthesis).

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                    $event->getResource()->getUri() === $this->target->getUri()
+                    && $event->getMetadataUri() === self::URI_PROPERTY_1
+                    && $event->getMetadataValue() === 'Value 1');
+            }));
+
+        $this->target
+            ->expects($this->once())
+            ->method('setType')
+            ->with($this->callback(function (core_kernel_classes_Class $class) {
+                return ($class->getUri() === self::URI_TYPE_1);
```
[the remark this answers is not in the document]
hectoras: Changed in ac23cd5cb705abff473deca0cc3a43b5ac8a3868 (see https://github.com/oat-sa/tao-core/pull/3339#discussion_r790629100 for the rationale for wrapping it with parenthesis).

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,717 @@
...
+                    $this->fail('Unexpected property: ' . $property->getUri());
+                }
+            ));
+
+        $this->target
+            ->expects($this->once())
+            ->method('editPropertyValues')
+            ->with($this->callback(
+                function (core_kernel_classes_Property $property, $value = null) {
+                    return ($property->getUri() === self::URI_PROPERTY_1);
```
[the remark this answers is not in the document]
hectoras: Changed in ac23cd5cb705abff473deca0cc3a43b5ac8a3868 (see https://github.com/oat-sa/tao-core/pull/3339#discussion_r790629100 for the rationale for wrapping it with parenthesis).

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed here: 754c338c79f0b4299aa11c33ce0b42cf8f03482f

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed here: 754c338c79f0b4299aa11c33ce0b42cf8f03482f

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed here: 754c338c79f0b4299aa11c33ce0b42cf8f03482f

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: I thought having it this way made a more clear the intent of that check.

BTW I'm inlining the check and removing the additional method and comment in 376dfcb02dd7c2b2ed5592e768429fa3d52db128 since it is clear that way too.

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Changed: 24948dcdf4ff8ec460c2606746e78fc9fc97dbb2 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Putting it as a single line (`if (in_array($property->getUri(), [self::URI_PROPERTY_1, self::URI_PROPERTY_2], true)) {`) would make the line 109 chars width (exceeding [PSR-12](https://www.php-fig.org/psr/psr-12/) max line lenght), so if we want to still use `in_array` for this one and conform to the linter rules, this should be put this way instead to make the linter happy :) :

```
                    if (
                        in_array(
                            $property->getUri(),
                            [self::URI_PROPERTY_1, self::URI_PROPERTY_2],
                            true
                        )
                    ) {
                        return $this->nonEmptyCollectionMock;
                    }
```
 (Because of PSR-12 §4.7 rules _The first expression of a multi-line control structure must be on the line after the opening parenthesis_ and _Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line._)
 
I personally find it more readable the former, explicit way with two `if`s (and will avoid the reader to know (or check in the documentation, as I did :sweat_smile: ) that the 3rd parameter for `in_array` is to do strict checking  (i.e. to make `in_array` behave as `===` on individual items)).

We may also use this one, but still feels a bit weird ([I put it like that initially](https://github.com/oat-sa/tao-core/blob/46462ce747455a4f3fca73898f2d40726700728f/test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php#L383)):

```
                    if (
                        $property->getUri() === self::URI_PROPERTY_1
                        || $property->getUri() === self::URI_PROPERTY_2
                    ) {
                        return $this->nonEmptyCollectionMock;
                    }
```

I'd prefer to keep it as it is right now unless you have a strong preference for merging the two conditionals using either one of the previous formulas (`in_array` or the multi-line `if`) :)

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: See https://github.com/oat-sa/tao-core/pull/3339#discussion_r790538135

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: Good idea! I'm moving the setup for method expectations to a helper method in dd3faa491d1984a9fb7d729d62148922bd362d2f. Thank you :slightly_smiling_face:

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -71,99 +66,125 @@ public function __construct(core_kernel_classes_Resource $targetInstance)
      */
     protected function getTargetInstance()
```
andreluizmachado, line 67: is it possible to add return type hint here?

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -18,50 +18,45 @@
...
 /**
- * A data binder focusing on binding a source of data to a generis instance
+ * A data binder focusing on binding a source of data to a Generis instance.
  *
  * @access public
  * @author Jerome Bogaerts, <jerome@taotesting.com>
  * @package tao
-
  */
 class tao_models_classes_dataBinding_GenerisInstanceDataBinder extends tao_models_classes_dataBinding_AbstractDataBinder
```
andreluizmachado, line 36: How hard would be to change to namespace instead of FQN on the class name for this class? is it possible?

andreluizmachado (changes requested):

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Yes, it is :) Added such return type hint in a4f5bd618f34029bf2bc82e282620e0aa8fc1faa Thanks!

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: I'm afraid it would be a little risky, since that would be a breaking change and it seems this class was created in 2013.

But maybe we can leave this one as:

```
/**
 * @deprecated Use oat\tao\model\dataBinding\GenerisInstanceDataBinder instead.
 */
class tao_models_classes_dataBinding_GenerisInstanceDataBinder extends GenerisInstanceDataBinder
{
    // void
}
```

and then have the real class as `GenerisInstanceDataBinder` under `namespace oat\tao\model\dataBinding;` and with the proper file name (and rename the usages under tao-core to use the new name). That would ensure prior usages for this class still work while allowing using the namespaced name for new usages (btw I'd rename usages in tao-core itself (such as those from Lists), and other classes under the "dataBinder" directory to keep them consistent with this change).

Would that work for you? :)

hectoras: (Hmm.. by the way I think creating a new file while keeping the former one for each class would mean we'd loose the git history for the class the moment we remove the file with the legacy name)

andreluizmachado: I like your proposal to keep backward compatibility and at the same time, we can start to use namespace for such kinds of classes. Regarding the history. did you try to only rename the file first, commit it and then rename the class and recreate the tao_models_classes_dataBinding_GenerisInstanceDataBinder class? I think like this way will keep the history, no?

andreluizmachado (changes requested):

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
andreluizmachado: or maybe just a comment on the commit explaining that the content of the class was moved would work.

hectoras: That worked like a charm! Renamed file 8a7e3c56fcd64f554e37739e126927e3ec4221c6 and then pushed a new one with the former name 9f495ee25a3698180b4f9dd45c1c84a4e1d1754e et voila! The former git log is correctly tracked under the new name DataBindingException.php. I'll repeat the procedure with the other files.

models/classes/dataBinding/GenerisFormDataBindingException.php
```diff
@@ -0,0 +1,53 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2002-2008 (original work) Public Research Centre Henri Tudor & University of Luxembourg (under the project TAO & TAO2);
+ *               2008-2010 (update and modification) Deutsche Institut für Internationale Pädagogische Forschung (under the project TAO-TRANSFER);
+ *               2009-2012 (update and modification) Public Research Centre Henri Tudor (under the project TAO-SUSTAIN & TAO-DEV);
+  *              2022 (update and modification) Open Assessment Technologies SA;
```
gabrielfs7, line 21: No need to all dates

models/classes/dataBinding/GenerisInstanceDataBindingException.php
```diff
@@ -0,0 +1,55 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2002-2008 (original work) Public Research Centre Henri Tudor & University of Luxembourg (under the project TAO & TAO2);
+ *               2008-2010 (update and modification) Deutsche Institut für Internationale Pädagogische Forschung (under the project TAO-TRANSFER);
+ *               2009-2012 (update and modification) Public Research Centre Henri Tudor (under the project TAO-SUSTAIN & TAO-DEV);
```
gabrielfs7, line 20: No need to all dates

models/classes/dataBinding/GenerisInstanceDataBindingException.php
```diff
@@ -0,0 +1,55 @@
...
+use Throwable;
+
+/**
+ * Exception to be thrown when an error occurs at Binding time between the data
+ * be bound and the Generis Persistent memory.
+ *
+ * @access public
+ * @author Jerome Bogaerts, <jerome@taotesting.com>
+ * @package tao
+ */
```
gabrielfs7, line 36: Not required 

```suggestion
```

models/classes/dataBinding/GenerisFormDataBindingException.php
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Exception to be thrown when an error occurs when binding data coming from a
+ * Instance Form to the Generis persistent memory.
+ *
+ * @access public
+ * @author Jerome Bogaerts, <jerome@taotesting.com>
+ * @package tao
+
+ */
```
gabrielfs7, line 34: Not required

```suggestion
```

models/classes/dataBinding/DataBindingException.php
```diff
@@ -0,0 +1,39 @@
...
+
+use common_Exception;
+
+/**
+ * Exception to be thrown when an errors occurs at Data Binding time.
+ *
+ * @access public
+ * @author Jerome Bogaerts, <jerome@taotesting.com>
+ * @package tao
+ */
```
gabrielfs7, line 35: Not required
```suggestion
```

models/classes/dataBinding/DataBindingException.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2002-2008 (original work) Public Research Centre Henri Tudor & University of Luxembourg (under the project TAO & TAO2);
+ *               2008-2010 (update and modification) Deutsche Institut für Internationale Pädagogische Forschung (under the project TAO-TRANSFER);
+ *               2009-2012 (update and modification) Public Research Centre Henri Tudor (under the project TAO-SUSTAIN & TAO-DEV);
+ *               2022 Open Assessment Technologies SA;
+ *
```
gabrielfs7, line 22: Please, use latest annotation on all your new classes and provide only the creation year, not all dates

models/classes/dataBinding/GenerisFormDataBindingException.php
```diff
@@ -0,0 +1,53 @@
...
+class GenerisFormDataBindingException extends DataBindingException
```
gabrielfs7, line 35: I safe way to deprecate the previous one is make this one extend `tao_models_classes_dataBinding_GenerisFormDataBindingException`

I am afraid this code can cause a breaking change and unpredictable errors.

models/classes/dataBinding/AbstractDataBinder.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
gabrielfs7, line 1: Please, review annotations as other classes

actions/class.Users.php
```diff
@@ -233,7 +233,7 @@ public function delete()
      * @return void
      * @throws Exception
      * @throws \oat\generis\model\user\PasswordConstraintsException
-     * @throws tao_models_classes_dataBinding_GenerisFormDataBindingException
+     * @throws GenerisFormDataBindingException
```
gabrielfs7, line 236: This is a breaking change now. More details in the other comments I left.

actions/class.Users.php
```diff
@@ -328,7 +328,7 @@ public function checkLogin()
      * @throws Exception
      * @throws \oat\generis\model\user\PasswordConstraintsException
      * @throws common_exception_Error
-     * @throws tao_models_classes_dataBinding_GenerisFormDataBindingException
+     * @throws GenerisFormDataBindingException
```
gabrielfs7, line 331: This is a breaking change now. More details in the other comments I left.

models/classes/GenerisServiceTrait.php
```diff
@@ -150,7 +151,7 @@ public function createSubClass(core_kernel_classes_Class $parentClazz, $label =
      * @param  core_kernel_classes_Resource $instance
      * @param  array properties
      * @return core_kernel_classes_Resource
-     * @throws \tao_models_classes_dataBinding_GenerisInstanceDataBindingException
+     * @throws GenerisInstanceDataBindingException
```
gabrielfs7, line 154: This is a breaking change now. More details in the other comments I left.

models/classes/class.UserService.php
```diff
@@ -590,7 +590,7 @@ function ($role) {
      * @param string|null $hashForKey
      *
      * @return boolean
-     * @throws tao_models_classes_dataBinding_GenerisFormDataBindingException
+     * @throws GenerisFormDataBindingException
```
gabrielfs7, line 593: This is a breaking change now. More details in the other comments I left.

hectoras:
This pull request has been superseded by #3361 (which commits fewer changes).

AntonNeviarouski:
QA verified

hectoras:
A newer pull request that fix the same issue as this one has been merged tot he `develop` branch: https://github.com/oat-sa/tao-core/pull/3361

PR #3340, opened 2022-01-18
Fix/rel 615/load translation for cli action

siwane:
Fix Japanese translations for error on deferred task

Require https://github.com/oat-sa/generis/pull/962

How to test
Create a test and add any item to it. Save the test and delete the item, which already in the test. Then try to publish the test
1.Go to "Deliveries" tab
2.Go to "New delivery" icon on the left bottom of the screen and click on it
3.Enter three or more characters to select the test you want to publish ( Publishing faild if test contain empty items)


alroniks:
@siwane I don't know why but seems pull request now contains commits from previous PRs. Probably need to rebase it on top of develop.

alroniks (approved):

PR #3354, opened 2022-01-24
Fix/rel 615/move password constraints to tao

siwane:
**Description**
Create PasswordContraintService in tao extension

**Ticket**
https://oat-sa.atlassian.net/browse/REL-615
https://oat-sa.atlassian.net/browse/REL-582


migrations/Version202201200829462234_tao.php
```diff
@@ -0,0 +1,74 @@
...
+
+        if ($removeExistingPasswordConfig) {
+            unset($passwordConfig['constrains']);
+            $generisExtension->setConfig('passwords', $passwordConfig);
+        }
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->getServiceManager()->unregister(PasswordConstraintsServiceInterface::SERVICE_ID);
```
alroniks, line 72: I hope it will be a rare case, but I guess we need return back proper configuration to generis in case rollback.

alroniks (approved):
Works well, but one small comment about migration.

models/classes/password/PasswordConstraintsService.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace oat\tao\model\password;
+
+use oat\oatbox\service\ConfigurableService;
+use tao_helpers_form_Validator;
+
+class PasswordConstraintsService extends ConfigurableService implements PasswordConstraintsServiceInterface
+{
+    /** @var tao_helpers_form_Validator[] */
+    private array $validators = [];
```
alroniks, line 32:
```suggestion
    private $validators = [];
```
Because it works only with PHP 7.4 when we still support 7.2

siwane:
Translations for password constraints will be done in another PR

migrations/Version202201200829462234_tao.php
siwane: PR updated pls check with ```php index.php '\oat\tao\scripts\tools\Migrations' -c rollback -v oat\\tao\\migrations\\Version202201200829462234_tao```

alroniks (approved):

models/classes/password/PasswordConstraintsService.php
augustas: Right, it breaks `tao/scripts/taoUpdate.php` on my php 7.2 instance

hectoras: Same here :point_up:

PR #3356, opened 2022-01-24
Feature/ADF-905/Validate statistical metadata during CSV import

shpran:
# [ADF-905](https://oat-sa.atlassian.net/browse/ADF-905)

## Require PRs
* https://github.com/oat-sa/generis/pull/968

## Changes
* League/CSV extracted to a separate service
* Validation extracted to a separate services
* Added metadata availability validation
* Added metadata uniqueness validation
* Added metadata type validation
* Reports improved

## Tasks
- [ ] ❗❗❗ Update `oat-sa/generis` dependency in `composer.json`
- [x] Add unit tests
- [ ] Add README.md


models/classes/Csv/Service/Reader.php
```diff
@@ -0,0 +1,83 @@
...
+
+namespace oat\tao\model\Csv\Service;
+
+use Iterator;
+use League\Csv\Exception;
+use InvalidArgumentException;
+use League\Csv\InvalidArgument;
+use League\Csv\Reader as LeagueCsvReader;
+
+class Reader
```
gabrielfs7, line 31: I am not sure that this is a service. Is seems to me more like a collection/value object/resource.

models/classes/Csv/Service/Reader.php
```diff
@@ -0,0 +1,83 @@
...
+{
+    /** @var LeagueCsvReader */
+    private $reader;
+
+    public function __construct(LeagueCsvReader $reader)
+    {
+        $this->reader = $reader;
+    }
+
+    public function setDelimiter(string $delimiter): self
```
gabrielfs7, line 41: Why is this method necessary here? We are already providing the `reader` in the constructor. Don't you think we are doing an unnecessary delegate?

models/classes/Csv/Service/Reader.php
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            $this->reader->setDelimiter($delimiter);
+        } catch (Exception | InvalidArgument $exception) {
+            throw new InvalidArgumentException($exception->getMessage());
+        }
+
+        return $this;
+    }
+
+    public function setHeaderOffset(?int $offset): self
```
gabrielfs7, line 52: Why is this method necessary here? We are already providing the `reader` in the constructor. Don't you think we are doing an unnecessary delegate?

models/classes/Csv/Service/Reader.php
gabrielfs7, line 31: If this class is a wrapper for the `Reader`, should not we use an interface?

models/classes/StatisticalMetadata/Contract/Header.php
```diff
@@ -0,0 +1,30 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA.
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\StatisticalMetadata\Contract;
+
+interface Header
```
gabrielfs7, line 25: Why an interface without behavior?

models/classes/Csv/Service/Reader.php
hectoras: I would try to move the logic from `setDelimiter()` & `setHeaderOffset()` to `ReaderFactory` itself, as well as changing the signature of those methods to return instances of League's [TabularDataReader](https://github.com/thephpleague/csv/blob/master/src/TabularDataReader.php) interface instead of the actual `LeagueCsvReader` implementation.

models/classes/StatisticalMetadata/Import/Mapper/StatisticalMetadataMapper.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private function map(array $metadataProperties): array
+    {
+        $map = [];
+
+        foreach ($metadataProperties as $metadataProperty) {
+            $map[Header::METADATA_PREFIX . $metadataProperty->alias][] = [
+                self::KEY_URI => $metadataProperty->uri,
+                self::KEY_DOMAIN => $this->ontology->getClass($metadataProperty->domain),
+                self::KEY_WIDGET => $metadataProperty->widget,
```
hectoras, line 79: Would it be possible for this to return an actual object instead of an array? That way the caller would depend on that object instead of depending on the whole mapper itself.

Alternatively, maybe we can add an interface for the mapper and move the constants there.

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
```diff
@@ -0,0 +1,231 @@
...
+                $template .= '. %d warning(s)';
+                $templateData[] = $warningsCount;
+            }
+
+            if ($errorsCount) {
+                $template .= ', %d error(s)';
+                $templateData[] = $errorsCount;
+            }
+
+            $template .= '.';
```
hectoras, line 212: nitpick: Minor issue, if there are no warnings but only errors, the first part of the message will be separated with a comma instead of a full stop.

```suggestion
            $msg = [];

            [$warningCount, $errorCount] = $this->countReportTypes();

            if ($warningCount > 0) {
                $msg[] = '%d warning(s)';
                $templateData[] = $warningCount;
            }

            if ($errorCount > 0) {
                $msg[] = '%d error(s)';
                $templateData[] = $errorCount;
            }

            $template = 'Imported %d/%d row(s).' . (!empty($msg) ? ' ' . implode(', ', $msg) .'.' : '');
```

models/classes/StatisticalMetadata/Import/Extractor/MetadataAliasesExtractor.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /** @var MetadataHeadersExtractor */
+    private $metadataHeadersExtractor;
+
+    public function __construct(MetadataHeadersExtractor $metadataHeadersExtractor)
+    {
+        $this->metadataHeadersExtractor = $metadataHeadersExtractor;
+    }
+
+    public function extract(array $header): array
```
gabrielfs7, line 37: It is actually returning all the headers, but removing the prefix from the `metatada_`. Do you think this is a proper behavior for an alias extractor?

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
```diff
@@ -0,0 +1,231 @@
...
+
+    private function countReportTypes(): array
+    {
+        $warningsCount = 0;
+        $errorsCount = 0;
+
+        foreach ($this->reports as $report) {
+            $report->getType() === Report::TYPE_WARNING
+                ? ++$warningsCount
+                : ++$errorsCount;
```
hectoras, line 226: Suggestion: I'd use a standard `if`/`else` here instead of using a ternary operator with side effects.

```suggestion
            if ($report->getType() === Report::TYPE_WARNING) {
                ++$warningsCount;            
            } else {
                ++$errorsCount;
            }
```

Another approach would be:
```
$warningsCount = array_reduce($this->reports, function ($report, $carry) {
    return $carry + ($report->getType() === Report::TYPE_WARNING ? 1 : 0);
});
$errorCount = count($this->reports) - $warningsCount;
```

(but I think the `if`/`else` is more clear :D)

models/classes/StatisticalMetadata/Import/Validator/HeaderValidator.php
```diff
@@ -0,0 +1,142 @@
...
+        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HTMLArea',
+    ];
+
+    public function __construct(MetadataHeadersExtractor $metadataHeadersExtractor)
+    {
+        $this->metadataHeadersExtractor = $metadataHeadersExtractor;
+    }
+
+    /** @var MetadataHeadersExtractor */
+    private $metadataHeadersExtractor;
```
hectoras, line 45: Please put the properties before the first method (the constructor) :)

```suggestion
    /** @var MetadataHeadersExtractor */
    private $metadataHeadersExtractor;
    
    public function __construct(MetadataHeadersExtractor $metadataHeadersExtractor)
    {
        $this->metadataHeadersExtractor = $metadataHeadersExtractor;
    }
```

models/classes/StatisticalMetadata/Import/Validator/HeaderValidator.php
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        if (empty($this->metadataHeadersExtractor->extract($header))) {
+            $missedColumns[] = Header::METADATA_PREFIX . '{property_alias}';
+        }
+
+        if (!empty($missedColumns)) {
+            throw new RuntimeException(
+                sprintf(
+                    'required %s missing ("%s")',
```
hectoras, line 64:
```suggestion
                    'Required %s missing ("%s")',
```

models/classes/StatisticalMetadata/Import/Validator/HeaderValidator.php
```diff
@@ -0,0 +1,142 @@
...
+            $this->metadataHeadersExtractor->extract($header),
+            array_keys($metadataMap)
+        );
+
+        if (!empty($invalid)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    '%s not found ("%s")',
+                    $this->getCounteredString($invalid, 'properties', 'property'),
+                    implode('", "',  $invalid)
```
hectoras, line 84:
```suggestion
                    '%s not found ("%s")',
                    $this->getCounteredString($invalid, 'Properties', 'Property'),
                    implode('", "',  $invalid)
```

models/classes/StatisticalMetadata/Import/Validator/HeaderValidator.php
```diff
@@ -0,0 +1,142 @@
...
+                $invalid[] = $metadataName;
+            }
+        }
+
+        if (!empty($invalid)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    '%s used on more than one property ("%s")',
+                    $this->getCounteredString($invalid, 'aliases are', 'alias is'),
+                    implode('", "', $invalid)
```
hectoras, line 105:
```suggestion
                sprintf(
                    '%s used on more than one property ("%s")',
                    $this->getCounteredString($invalid, 'Aliases are', 'Alias is'),
                    implode('", "', $invalid)
```

models/classes/StatisticalMetadata/Import/Validator/HeaderValidator.php
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+
+        if (!empty($invalid)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    '%s of the wrong input type - only TEXT is allowed ("%s")',
+                    $this->getCounteredString($invalid, 'properties are', 'property is'),
+                    implode('", "', $invalid)
+                )
```
hectoras, line 133:
```suggestion
                sprintf(
                    '%s of the wrong input type - only TEXT is allowed ("%s")',
                    $this->getCounteredString($invalid, 'Properties are', 'Property is'),
                    implode('", "', $invalid)
                )
```

models/classes/StatisticalMetadata/Import/Validator/MetadataToBindValidator.php
```diff
@@ -0,0 +1,51 @@
...
+    public function validateRelationToResource(
+        core_kernel_classes_Resource $resource,
+        core_kernel_classes_Class $metadataDomain
+    ): void {
+        if (!$resource->isInstanceOf($metadataDomain)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'provided metadata does not belong to the resource "%s".',
+                    $resource->getUri()
+                )
```
hectoras, line 47:
```suggestion
                sprintf(
                    'Provided metadata does not belong to the resource "%s".',
                    $resource->getUri()
                )
```

models/classes/StatisticalMetadata/Import/Validator/MetadataToBindValidator.php
```diff
@@ -0,0 +1,51 @@
...
+use InvalidArgumentException;
+use core_kernel_classes_Class;
+use core_kernel_classes_Resource;
+
+class MetadataToBindValidator
+{
+    public function validateMetadataValue($value): void
+    {
+        if (!is_string($value)) {
+            throw new InvalidArgumentException('metadata value must be a string.');
```
hectoras, line 34:
```suggestion
            throw new InvalidArgumentException('Metadata value must be a string.');
```

models/classes/StatisticalMetadata/Import/Validator/HeaderValidator.php
shpran: Nice, seems like IDE generated property with an issue 😮 
Thanks!

models/classes/StatisticalMetadata/Import/Validator/RecordResourceValidator.php
```diff
@@ -0,0 +1,72 @@
...
+use oat\tao\model\StatisticalMetadata\Contract\Header;
+use oat\tao\model\StatisticalMetadata\Import\Exception\ErrorValidationException;
+
+class RecordResourceValidator
+{
+    public function validateResourceId(array $record): void
+    {
+        if (empty($record[Header::ITEM_ID]) && empty($record[Header::TEST_ID])) {
+            throw new ErrorValidationException(
+            'At least one of %s or %s is required.',
```
shpran, line 38:
```suggestion
                'At least one of %s or %s is required.',
```

models/classes/Csv/Factory/ReaderFactory.php
```diff
@@ -0,0 +1,58 @@
...
+
+        $reader = LeagueReader::createFromStream($stream);
+
+        try {
+            $reader->setHeaderOffset(0);
+
+            if (!empty($options[self::DELIMITER]) && is_string($options[self::DELIMITER])) {
+                $reader->setDelimiter($options[self::DELIMITER]);
+            }
+        } catch (Exception | InvalidArgument $exception) {
```
gabrielfs7, line 52: Isn't the same in this case? 

```suggestion
        } catch (Throwable $exception) {
```

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
```diff
@@ -0,0 +1,247 @@
...
+
+            foreach ($warningReports as $warningReport) {
+                $lineMainReport->add($warningReport);
+            }
+
+            foreach ($errorReports as $errorReport) {
+                $lineMainReport->add($errorReport);
+            }
+
+            if (empty($errorReports)) {
```
gabrielfs7, line 92: You also need to check that you actually have warning reports too, no?

```suggestion
            if (empty($errorReports) && !empty($warningReports)) {
```

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
```diff
@@ -0,0 +1,247 @@
...
+                $warningAndErrorReports[$line] = $lineMainReport;
+            }
+        }
+
+        if (!empty($headerExceptionReports)) {
+            $this->createAndAddSubReport(
+                $report,
+                $headerExceptionReports,
+                Report::TYPE_ERROR,
+                'Header contain an error(s)'
```
gabrielfs7, line 106: I am not sure `an` will match with both plural an singular

```suggestion
                'Header contain error(s)'
```

gabrielfs7: Same for other messages

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
```diff
@@ -0,0 +1,247 @@
...
+    {
+        if ($reporter->getTotalHeaderErrors()) {
+            return Report::create(
+                $type,
+                'CSV import failed: header is not valid (%d error(s))',
+                [$reporter->getTotalHeaderErrors()]
+            );
+        }
+
+        if ($reporter->getTotalImportedRecords() === 0  || $reporter->getTotalScannedRecords() === 0) {
```
gabrielfs7, line 174:
```suggestion
        if ($reporter->getTotalImportedRecords() === 0 || $reporter->getTotalScannedRecords() === 0) {
```

models/classes/StatisticalMetadata/Import/Extractor/ResourceExtractor.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        return $resource;
+    }
+
+    private function extractResourceHeader(array $record): string
+    {
+        return !empty($record[Header::ITEM_ID])
+            ? Header::ITEM_ID
+            : Header::TEST_ID;
```
gabrielfs7, line 76: Avoiding negation is better to read

```suggestion
        return empty($record[Header::ITEM_ID])
            ? Header::TEST_ID
            : Header::ITEM_ID;
```

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->headerValidator = $headerValidator;
+        $this->metadataPropertiesExtractor = $metadataPropertiesExtractor;
+        $this->resourceExtractor = $resourceExtractor;
+        $this->metadataValuesExtractor = $metadataValuesExtractor;
+        $this->reportBuilder = $reportBuilder;
+    }
+
+    public function process(File $file): Report
+    {
+        $reporter = new ImportReporter();
```
gabrielfs7, line 80: `ImportReporter` seems to be a `Service` name, therefore, I would not instantiate it here. This is a actually a ValueObject. Maybe rename it to `ImportReport` or `ImportResult` ?

models/classes/StatisticalMetadata/Import/Reporter/ImportReporter.php
```diff
@@ -0,0 +1,184 @@
...
+namespace oat\tao\model\StatisticalMetadata\Import\Reporter;
+
+use Throwable;
+use oat\tao\model\StatisticalMetadata\Import\Exception\ErrorValidationException;
+use oat\tao\model\StatisticalMetadata\Import\Exception\HeaderValidationException;
+use oat\tao\model\StatisticalMetadata\Import\Exception\WarningValidationException;
+use oat\tao\model\StatisticalMetadata\Import\Exception\AbstractValidationException;
+use oat\tao\model\StatisticalMetadata\Import\Exception\AggregatedValidationException;
+
+class ImportReporter
```
gabrielfs7, line 32: `ImportReporter` seems to be a `Service` name, therefore, I would not instantiate it here. This is a actually a ValueObject. Maybe rename it to `ImportReport` or `ImportResult` ?

https://github.com/oat-sa/tao-core/pull/3356/files#r794264830

models/classes/StatisticalMetadata/Import/Validator/ResourceMetadataRelationValidator.php
```diff
@@ -0,0 +1,49 @@
...
+use oat\tao\model\StatisticalMetadata\Import\Exception\ErrorValidationException;
+
+class ResourceMetadataRelationValidator
+{
+    public function validate(
+        core_kernel_classes_Resource $resource,
+        core_kernel_classes_Property $metadataProperty
+    ): void {
+        /** @var core_kernel_classes_Class $metadataPropertyDomain */
+        $metadataPropertyDomain = $metadataProperty->getDomain()->get(0);
```
gabrielfs7, line 37: Can this be null. Will it be a problem?

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
shpran: Nope. This if is to add values inside `$onlyWarningReports` array.

models/classes/StatisticalMetadata/Import/Validator/ResourceMetadataRelationValidator.php
shpran: I'm not sure, that this will be `null` in our case, as we are using only properties from `items` and `tests`, which have domaing.
In other side, this might be covered by `try-catch` block.
Also, I can add a check for this and throw an exception. But it still be covered by `try-catch` block 😅

models/classes/StatisticalMetadata/Import/Exception/AggregatedValidationException.php
```diff
@@ -0,0 +1,68 @@
...
+
+    /** @var WarningValidationException[] */
+    private $warnings;
+
+    public function __construct(array $errors, array $warnings)
+    {
+        $this->errors = $errors;
+        $this->warnings = $warnings;
+
+        parent::__construct();
```
hectoras, line 40: nitpick: I'd put the call to the constructor first

```suggestion
        parent::__construct();
        
        $this->errors = $errors;
        $this->warnings = $warnings;
```

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
```diff
@@ -0,0 +1,247 @@
...
+                if ($exception instanceof WarningValidationException) {
+                    $warningReports[] = $this->createReportByException($exception);
+
+                    continue;
+                }
+
+                if ($exception instanceof ErrorValidationException) {
+                    $errorReports[] = $this->createReportByException($exception);
+                }
+            }
```
hectoras, line 69: Suggestion: This is a simple branching code, I'd use a plain-old `if-else` structure here instead of jumping out of the current loop with `continue`. It makes this nested loop more concise as well. (BTW this is also a personal preference, feel free to keep it as it is :D)

```suggestion
            foreach ($exceptions as $exception) {
                if ($exception instanceof HeaderValidationException) {
                    $headerExceptionReports[] = $this->createReportByException($exception);
                } elseif ($exception instanceof WarningValidationException) {
                    $warningReports[] = $this->createReportByException($exception);
                } elseif ($exception instanceof ErrorValidationException) {
                    $errorReports[] = $this->createReportByException($exception);
                }
            }
```

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
hectoras: Maybe a ternary like `count($headerExceptionReports) > 1 ? 'Header contains an error' : 'Header contains errors'` would work in these cases.

models/classes/StatisticalMetadata/Import/Builder/ReportBuilder.php
shpran: Thanks for your suggestion 🙂
I think, that right now I will keep this one as it is, because this `if-elseif` block make a bit of mess 😅 (I mean, it becomes difficult to read). But I agree with you, that in this case this one is better, than using `continue`.

In the future, I want to refactor and optimize this class 🙂
I will keep your suggestion in my mind to use it during this refactoring

models/classes/StatisticalMetadata/Import/Extractor/MetadataAliasesExtractor.php
```diff
@@ -29,18 +29,40 @@ class MetadataAliasesExtractor
...
     {
-        return array_map(
-            static function (string $metadataHeader): string {
-                return str_replace(Header::METADATA_PREFIX, '', $metadataHeader);
-            },
-            $this->metadataHeadersExtractor->extract($header)
-        );
+        $hash = $this->getHash($header);
+
+        if (!array_key_exists($hash, $this->cache)) {
```
hectoras, line 44: `extractAliases` would at least return an empty array, so if a key exists it would at least be set and be an empty array. Therefore, I think we can use `isset` instead of `array_key_exists` here, as it returns true if the key exists even if its value is considered empty

```suggestion
        if (!isset($this->cache[$hash])) {
```

(Just checked :D)
```
~$ php
<?php $a=['k'=>[]]; var_dump(isset($a['k']));
Standard input code:1:
bool(true)
```

models/classes/StatisticalMetadata/Import/Extractor/MetadataHeadersExtractor.php
```diff
@@ -26,8 +26,27 @@
...
 {
+    /** @var array */
+    private $cache = [];
+
     public function extract(array $header): array
     {
-        return preg_grep(sprintf('/^%s/', Header::METADATA_PREFIX), $header);
+        $hash = $this->getHash($header);
+
+        if (!array_key_exists($hash, $this->cache)) {
```
hectoras, line 36: See [#r794312964](https://github.com/oat-sa/tao-core/pull/3356/files/74d62b110d8278ddc548d6c1f3196655ff552f54..a420f5f86b20d6006103c4c8e7c3a880c10df26c#r794312964)

```suggestion
        if (!isset($this->cache[$hash])) {
```

models/classes/StatisticalMetadata/Import/Extractor/MetadataValuesExtractor.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->resourceMetadataRelationValidator = $resourceMetadataRelationValidator;
+    }
+
+    /**
+     * @param core_kernel_classes_Property[] $metadataProperties
+     */
+    public function extract(array $record, core_kernel_classes_Resource $resource, array $metadataProperties): array
+    {
+        $values = [];
+        $warnings = [];
```
gabrielfs7, line 53: Not sure those are only warnings, maybe rename.

models/classes/StatisticalMetadata/Import/Result/ImportResult.php
```diff
@@ -0,0 +1,184 @@
...
+        foreach ($exception->getWarnings() as $warningException) {
+            $this->addInternalException($line, $warningException);
+        }
+
+        foreach ($exception->getErrors() as $errorException) {
+            $this->addInternalException($line, $errorException);
+        }
+
+
+        return $this;
```
gabrielfs7, line 153:
```suggestion
        }

        return $this;
```

gabrielfs7 (approved):
Good job, we are just missing tests :) 

hectoras (approved):

AntonNeviarouski:
QA verified

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
```diff
@@ -0,0 +1,138 @@
...
+            } catch (Throwable $exception) {
+                return $this->reportBuilder->buildByException($exception);
+            }
+        }
+
+        return $this->reportBuilder->buildByResult($result);
+    }
+
+    /**
+     * @TODO Improve the repository to allow persistence of resource properties and use it instead of a data binder.
```
gabrielfs7, line 129: Do we plan to do it? If not, I would remove the comment

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
```diff
@@ -0,0 +1,138 @@
...
+use oat\tao\model\StatisticalMetadata\Import\Builder\ReportBuilder;
+use oat\tao\model\StatisticalMetadata\Import\Validator\HeaderValidator;
+use oat\tao\model\StatisticalMetadata\Import\Extractor\ResourceExtractor;
+use oat\tao\model\StatisticalMetadata\Import\Extractor\MetadataValuesExtractor;
+use oat\tao\model\StatisticalMetadata\Import\Exception\AbstractValidationException;
+use oat\tao\model\StatisticalMetadata\Import\Extractor\MetadataPropertiesExtractor;
+use oat\tao\model\StatisticalMetadata\Import\Exception\AggregatedValidationException;
+use tao_models_classes_dataBinding_GenerisInstanceDataBindingException as DataBindingException;
+
+// @TODO Improve tests - add invalid cases
```
gabrielfs7, line 42: - [ ] Test was added, we can remove it :)

test/unit/models/classes/StatisticalMetadata/Import/Extractor/MetadataAliasesExtractorTest.php
```diff
@@ -0,0 +1,125 @@
...
+                            'metadata_alias_6',
+                        ];
+                    }
+
+                    return [];
+                }
+
+            );
+
+        // Check that $this->metadataHeadersExtractor->expects() is called 2 times instead of 4
```
gabrielfs7, line 119: Is this comment required?

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
shpran: Yes, I want to do this 🙂

models/classes/StatisticalMetadata/Import/Processor/ImportProcessor.php
shpran: We have only valid case. I want to add invalid cases.

test/unit/models/classes/StatisticalMetadata/Import/Extractor/MetadataAliasesExtractorTest.php
shpran: Just an explanation why these methods called multiple times 🙂

PR #3361, opened 2022-01-26
feature/ADF-869: Delete previous statements for a property when edited with an empty value

hectoras:
**Associated Jira issue:** [ADF-869](https://oat-sa.atlassian.net/browse/ADF-869)

This pull request fixes the call to `$instance->editPropertyValues()` (near line 140) to just call `removePropertyValues()` or `editPropertyValues()` depending on the value being considered as empty or not.

Previously,
1. The values provided were inserted as-is by calling `editPropertyValues()`, which removed all the values and then inserted the new ones with no conversion, and then
2. `removePropertyValues()` was called for values under the instance being edited **but** using an empty string (`''`) as the matching pattern for the values to be deleted.

This later step may fail in case the RDBMS doesn't _automagically_ turn strings containing only whitespaces into empty strings in step 1 (i.e. if it does not `trim()` the values on insertion behind the scenes). Instead of that, this pull request calls `removePropertyValues()` with no pattern at all, which has the expected behavior of removing all values for the instance being edited before persisting the new ones. 

Additionally, this pull request adds unit tests for the whole data binder class, covering all class methods and branches except the static call to `ServiceManager::getServiceManager()`, that is now wrapped inside a private method to allow injecting the manager from tests.

Note empty values for arrays are still saved ([here](https://github.com/oat-sa/tao-core/blob/7934e85497517172dc10faf1da7935bb3eed5679/models/classes/dataBinding/class.GenerisInstanceDataBinder.php#L118) and [here](https://github.com/oat-sa/tao-core/blob/7934e85497517172dc10faf1da7935bb3eed5679/models/classes/dataBinding/class.GenerisInstanceDataBinder.php#L137)); that is, no filtering is done for arrays but only for string properties.

This pull request supersedes #3339 by providing the intended fix with no additional changes.

---
Self-checks before moving to code review:

- [x] Pull request's target is not master

### DIFF
--- a/models/classes/dataBinding/class.GenerisInstanceDataBinder.php
+++ b/models/classes/dataBinding/class.GenerisInstanceDataBinder.php
@@ -102,7 +102,7 @@ class tao_models_classes_dataBinding_GenerisInstanceDataBinder extends tao_model
                         $instance->removeType($type);
                     }
 
-                    $types = !is_array($propertyValue) ? [$propertyValue] : $propertyValue;
+                    $types = is_array($propertyValue) ? $propertyValue : [$propertyValue];
 
                     foreach ($types as $type) {
                         $instance->setType(new core_kernel_classes_Class($type));
--- a/models/classes/dataBinding/class.GenerisInstanceDataBinder.php
+++ b/models/classes/dataBinding/class.GenerisInstanceDataBinder.php
@@ -122,7 +122,7 @@ class tao_models_classes_dataBinding_GenerisInstanceDataBinder extends tao_model
                             );
                         }
                     } elseif (is_string($propertyValue)) {
-                        if($this->isEmptyValue($propertyValue)) {
+                        if ($this->isEmptyValue($propertyValue)) {
                             $instance->removePropertyValues($prop);
                         } else {
                             $instance->editPropertyValues(
@@ -151,7 +151,7 @@ class tao_models_classes_dataBinding_GenerisInstanceDataBinder extends tao_model
                     new MetadataModified($instance, $propertyUri, $propertyValue)
                 );
             }
-            
+
             return $instance;
         } catch (common_Exception $e) {
             $msg = "An error occured while binding property values to instance '': " . $e->getMessage();
@@ -162,7 +162,7 @@ class tao_models_classes_dataBinding_GenerisInstanceDataBinder extends tao_model
 
     private function isEmptyValue(string $value): bool
     {
-        return '' === $value || ' ' === $value || strlen(trim($value)) == 0;
+        return strlen(trim($value)) === 0;
     }
 
     private function getEventManager(): EventManager
--- a/models/classes/dataBinding/class.GenerisInstanceDataBinder.php
+++ b/models/classes/dataBinding/class.GenerisInstanceDataBinder.php
@@ -18,13 +18,18 @@
  * Copyright (c) 2002-2008 (original work) Public Research Centre Henri Tudor & University of Luxembourg (under the project TAO & TAO2);
  *               2008-2010 (update and modification) Deutsche Institut für Internationale Pädagogische Forschung (under the project TAO-TRANSFER);
  *               2009-2012 (update and modification) Public Research Centre Henri Tudor (under the project TAO-SUSTAIN & TAO-DEV);
- *
+ *               2022 (update and modification) Open Assessment Technologies SA;
  */
 
 use oat\generis\model\OntologyRdf;
+use oat\tao\model\dataBinding\AbstractDataBinder;
+use oat\tao\model\dataBinding\GenerisInstanceDataBindingException;
+use oat\tao\model\event\MetadataModified;
+use oat\oatbox\event\EventManager;
+use oat\oatbox\service\ServiceManager;
 
 /**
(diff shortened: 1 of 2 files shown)


models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -146,24 +139,38 @@ public function bind($data)
...
             $msg = "An error occured while binding property values to instance '': " . $e->getMessage();
             $instanceUri = $instance->getUri();
             throw new tao_models_classes_dataBinding_GenerisInstanceDataBindingException($msg);
         }
-        
+    }
+
+    private function isEmptyValue(string $value): bool
+    {
+        return '' === $value || ' ' === $value || strlen(trim($value)) == 0;
```
gabrielfs7, line 165: Why do you need this extra comparison? 

```
php > var_dump(trim('              ')); // many spaces
string(0) ""

```

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -94,23 +88,22 @@ protected function getTargetInstance()
...
             foreach ($data as $propertyUri => $propertyValue) {
                 if ($propertyUri == OntologyRdf::RDF_TYPE) {
                     foreach ($instance->getTypes() as $type) {
                         $instance->removeType($type);
                     }
-                    if (!is_array($propertyValue)) {
-                        $types = [$propertyValue] ;
-                    }
+
+                    $types = !is_array($propertyValue) ? [$propertyValue] : $propertyValue;
```
gabrielfs7, line 105: I would avoid negation always as possible, since it is less readable.

```suggestion
                    $types = is_array($propertyValue) ? $propertyValue : [$propertyValue];
```

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
```diff
@@ -0,0 +1,670 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA.
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\test\unit\model\action;
```
gabrielfs7, line 23: Are you sure about the namespace?

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -129,13 +122,13 @@ public function bind($data)
...
-                        if (strlen(trim($propertyValue)) == 0) {
-                            //if the property value is an empty space(the default value in a select input field), delete the corresponding triplet (and not all property values)
-                            $instance->removePropertyValues($prop, ['pattern' => '']);
+                        if($this->isEmptyValue($propertyValue)) {
+                            $instance->removePropertyValues($prop);
+                        } else {
+                            $instance->editPropertyValues(
+                                $prop,
+                                $propertyValue
+                            );
```
gabrielfs7, line 131: Missing small space. 

```suggestion
                        if ($this->isEmptyValue($propertyValue)) {
                            $instance->removePropertyValues($prop);
                        } else {
                            $instance->editPropertyValues(
                                $prop,
                                $propertyValue
                            );
```

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
shpran: I think `strlen(trim($value)) === 0` will be enough.

hectoras: It was a sort-of micro-optimization to avoid the function calls in the most common cases. I'll change it to keep just the function calls.

hectoras: :point_right: fdeb3539eb4014036702bc8089116519706a6a43

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Changed :point_right: 83663b774db2ba839cbd60fef93d53856fde82b7 :+1:

test/unit/models/classes/dataBinding/GenerisInstanceDataBinderTest.php
hectoras: :facepalm:

hectoras: Changed: 1b454e8c72d631b6a81e24b9c3b8581c6d29f9af

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Fixed in fdeb3539eb4014036702bc8089116519706a6a43

shpran (approved):

gabrielfs7 (approved):

AntonNeviarouski:
QA verified

PR #3370, opened 2022-01-31
fix: use string fr exception + remove type hint

siwane:
Fix
- remove type hint to ensure php7.2 compatibility
- use extension id as string for exception message


shpran (approved):
LGTM 👍🏻

PR #3378, opened 2022-02-01
feat: add new property is statistical

gabrielfs7:
https://oat-sa.atlassian.net/browse/ADF-876

New property "is statistical" is now available for schema management

## PRs

- [ ] https://github.com/oat-sa/generis/pull/971

### DIFF
--- a/actions/form/class.SimpleProperty.php
+++ b/actions/form/class.SimpleProperty.php
@@ -62,16 +62,7 @@ class tao_actions_form_SimpleProperty extends tao_actions_form_AbstractProperty
 
         $index = $this->getIndex();
 
-        $propertyProperties = array_merge(
-            tao_helpers_form_GenerisFormFactory::getDefaultProperties(),
-            [
-                $this->getProperty(GenerisRdf::PROPERTY_ALIAS),
-                $this->getProperty(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
-                $this->getProperty(GenerisRdf::PROPERTY_IS_STATISTICAL),
-                $this->getProperty(TaoOntology::PROPERTY_GUI_ORDER),
-                $this->getProperty(ValidationRuleRegistry::PROPERTY_VALIDATION_RULE)
-            ]
-        );
+        $propertyProperties = $this->getSchemaProperties();
         $values = $property->getPropertiesValues($propertyProperties);
 
         $elementNames = [];
@@ -286,6 +277,24 @@ class tao_actions_form_SimpleProperty extends tao_actions_form_AbstractProperty
         }
 
         $element->setDisabledValues(array_unique($disabledValues));
+    }
+
+    private function getSchemaProperties(): array
+    {
+        return array_merge(
+            tao_helpers_form_GenerisFormFactory::getDefaultProperties(),
+            array_filter(
+                [
+                    $this->getProperty(GenerisRdf::PROPERTY_ALIAS),
+                    $this->getProperty(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
+                    $this->getFeatureFlagChecker()->isEnabled('FEATURE_FLAG_STATISTIC_METADATA_IMPORT')
+                        ? $this->getProperty(GenerisRdf::PROPERTY_IS_STATISTICAL)
+                        : null,
+                    $this->getProperty(TaoOntology::PROPERTY_GUI_ORDER),
+                    $this->getProperty(ValidationRuleRegistry::PROPERTY_VALIDATION_RULE)
+                ]
+            )
+        );
     }
 
     private function getDependsOnPropertyFormFieldFactory(): DependsOnPropertyFormFieldFactory
--- a/actions/form/class.SimpleProperty.php
+++ b/actions/form/class.SimpleProperty.php
@@ -65,10 +65,10 @@ class tao_actions_form_SimpleProperty extends tao_actions_form_AbstractProperty
         $propertyProperties = array_merge(
             tao_helpers_form_GenerisFormFactory::getDefaultProperties(),
             [
-                new core_kernel_classes_Property(GenerisRdf::PROPERTY_ALIAS),
-                new core_kernel_classes_Property(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
-                new core_kernel_classes_Property(GenerisRdf::PROPERTY_IS_STATISTICAL),
-                new core_kernel_classes_Property(TaoOntology::PROPERTY_GUI_ORDER),
+                $this->getProperty(GenerisRdf::PROPERTY_ALIAS),
+                $this->getProperty(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
+                $this->getProperty(GenerisRdf::PROPERTY_IS_STATISTICAL),
+                $this->getProperty(TaoOntology::PROPERTY_GUI_ORDER),
                 $this->getProperty(ValidationRuleRegistry::PROPERTY_VALIDATION_RULE)
             ]
         );
@@ -108,7 +108,7 @@ class tao_actions_form_SimpleProperty extends tao_actions_form_AbstractProperty
                     $element->addValidator(tao_helpers_form_FormFactory::getValidator('NotEmpty'));
                 }
                 if (
-                    $propertyProperty->getUri() == GenerisRdf::PROPERTY_IS_STATISTICAL &&
+                    $propertyProperty->getUri() === GenerisRdf::PROPERTY_IS_STATISTICAL &&
                     empty($values[$propertyProperty->getUri()])
                 ) {
                     $element->setValue(tao_helpers_Uri::encode(GenerisRdf::GENERIS_FALSE));
--- a/actions/form/class.SimpleProperty.php
+++ b/actions/form/class.SimpleProperty.php
@@ -67,6 +67,7 @@ class tao_actions_form_SimpleProperty extends tao_actions_form_AbstractProperty
             [
                 new core_kernel_classes_Property(GenerisRdf::PROPERTY_ALIAS),
                 new core_kernel_classes_Property(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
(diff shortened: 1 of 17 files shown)


actions/form/class.SimpleProperty.php
```diff
@@ -67,6 +67,7 @@ protected function initElements()
             [
                 new core_kernel_classes_Property(GenerisRdf::PROPERTY_ALIAS),
                 new core_kernel_classes_Property(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
+                new core_kernel_classes_Property(GenerisRdf::PROPERTY_IS_STATISTICAL),
```
shpran, line 70: Can we use $this->getProperty() instead?

actions/form/class.SimpleProperty.php
```diff
@@ -106,6 +107,13 @@ protected function initElements()
                 if ($propertyProperty->getUri() == OntologyRdfs::RDFS_LABEL) {
                     $element->addValidator(tao_helpers_form_FormFactory::getValidator('NotEmpty'));
                 }
+                if (
+                    $propertyProperty->getUri() == GenerisRdf::PROPERTY_IS_STATISTICAL &&
```
shpran, line 111: Please, use strict comparison

hectoras (approved):

helpers/form/elements/xhtml/class.Calendar.php
```diff
@@ -45,6 +45,23 @@ public function render()
...
+            name="%s"
+            id="%s"
+            data-testid="%s"
+            value="%s"
+            %s
+            >',
+                $this->name,
+                $elementId,
+                $this->getDescription(),
+                empty($this->value) ? '' : $this->getDateOutput(),
```
shpran, line 60: I think, there's no need to check for `empty($this->value)`, as you are doing the same here: https://github.com/oat-sa/tao-core/pull/3378/files#diff-8e6a81364a06744f4c3f0574dee637ca5df18be82b3b6ad258f2c7a7bc37b9f9R100

models/classes/Csv/Resource/Reader.php
```diff
@@ -40,7 +40,14 @@ public function __construct(LeagueCsvReader $reader)
...
     {
-        return $this->reader->getHeader();
+        $headers = $this->reader->getHeader();
+
+        return array_map(
+            function ($header) {
+                return trim($header);
+            },
+            $headers
+        );
```
shpran, line 50:
```suggestion
        return array_map('trim', $headers);
```

models/classes/StatisticalMetadata/Import/Validator/MetadataPropertiesValidator.php
```diff
@@ -26,13 +26,16 @@
...
 
 class MetadataPropertiesValidator
 {
     private const ALLOWED_WIDGETS = [
-        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#TextBox',
-        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#TextArea',
-        'http://www.tao.lu/datatypes/WidgetDefinitions.rdf#HTMLArea',
+        tao_helpers_form_elements_Textbox::WIDGET_ID,
+        tao_helpers_form_elements_Textarea::WIDGET_ID,
+        tao_helpers_form_elements_Htmlarea::WIDGET_ID
```
shpran, line 38: 👍🏻

models/classes/StatisticalMetadata/Import/Validator/MetadataPropertiesValidator.php
```diff
@@ -85,7 +88,10 @@ public function validateMetadataUniqueness(array $metadataProperties): void
             }
 
             if (!array_key_exists($alias, $exceptions)) {
-                $exceptions[$alias] = $this->buildHeaderException($alias, 'Property referenced by "%s" is not unique');
+                $exceptions[$alias] = $this->buildHeaderException(
+                    $metadataProperty->getAlias() ?? $metadataProperty->getLabel(),
```
shpran, line 92: `$alias` contains `$metadataProperty->getAlias()` value, so it can be reused to avoid unnecessary call 🙂 
But it's not critical, so can be skipped

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
```diff
@@ -168,9 +204,23 @@ private function isEmptyValue(string $value): bool
...
     }
+
+    private function getFeatureFlagChecker(): FeatureFlagCheckerInterface
+    {
+        return $this->getServiceManager()->get(FeatureFlagChecker::class);
+    }
+
+    private function getServiceManager(): ServiceManager
+    {
+        if ($this->serviceManager === null) {
```
shpran, line 220: Please, use `isset` instead (optional here, but will be good to use it for typed properties 🙂 ).
If in future we will use typed class properties without a default value - this code will fail.

```php
// PHP <7.4
class A
{
    private $foo;

    public function getFoo(): string
    {
        if ($this->foo === null) {
            $this->foo = 'bar';
        }

        return $this->foo;
    }
}

(new A())->getFoo(); // OK
```

```php
// PHP >=7.4
class A
{
    private $foo;

    public function getFoo(): string
    {
        if ($this->foo === null) {
            $this->foo = 'bar';
        }

        return $this->foo;
    }
}

(new A())->getFoo(); // OK
```

```php
// PHP >=7.4
class A
{
    private ?string $foo = null;

    public function getFoo(): string
    {
        if ($this->foo === null) {
            $this->foo = 'bar';
        }

        return $this->foo;
    }
}

(new A())->getFoo(); // OK
```

```php
// PHP >=7.4
class A
{
    private ?string $foo;

    public function getFoo(): string
    {
        if ($this->foo === null) {
            $this->foo = 'bar';
        }

        return $this->foo;
    }
}

(new A())->getFoo(); // Uncaught Error: Typed property A::$foo must not be accessed before initialization
```

hectoras: Does `isset($this->foo)` throw an exception in PHP >= 7.4 if checking with it an uninitialized property? (I've not tested that locally).

We may as well set the default value for `$serviceManager` to `null` and keep testing here for `null`, right? If we use typed properties here, I think we should use a nullable type in the `$serviceManager` declaration to prevent problems unless it is initialized in the constructor.

 `private ServiceManager ?$serviceManager = null;`

actions/form/class.SimpleProperty.php
gabrielfs7: true, done

helpers/form/elements/xhtml/class.Calendar.php
gabrielfs7: true, done

models/classes/Csv/Resource/Reader.php
gabrielfs7: done

models/classes/dataBinding/class.GenerisInstanceDataBinder.php
hectoras: Just checked locally, `isset` does not throw an exception when used with an uninitialized property, so 4e37e2a4286cf59cffb166a475c4763a79d5af45 should be safe when we add typed properties :)

```
<?php class Foo {
    private ?string $bar;
    public function testIsset() { // Does not crash
         print_r(isset($this->bar));
    }
    public function testNullCmp() { // Crashes
         print_r(null === $this->bar);
    }
}

echo phpversion() . PHP_EOL; // 7.4.27
$foo = new Foo();
$foo->testIsset();
// $foo->testNullCmp();
```

shpran (approved):
Great job!

actions/form/class.SimpleProperty.php
```diff
@@ -280,6 +279,24 @@ private function disableValues(core_kernel_classes_Property $property, Validator
...
+                    $this->getProperty(GenerisRdf::PROPERTY_ALIAS),
+                    $this->getProperty(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),
+                    $this->getFeatureFlagChecker()->isEnabled('FEATURE_FLAG_STATISTIC_METADATA_IMPORT')
+                        ? $this->getProperty(GenerisRdf::PROPERTY_IS_STATISTICAL)
+                        : null,
+                    $this->getProperty(TaoOntology::PROPERTY_GUI_ORDER),
+                    $this->getProperty(ValidationRuleRegistry::PROPERTY_VALIDATION_RULE)
+                ]
+            )
+        );
```
hectoras, line 297: Suggestion/idea: As we are already doing a function call here to `array_merge`, we may put the conditional outside the main array and save the additional call to `array_filter`.

```suggestion
        return array_merge(
            tao_helpers_form_GenerisFormFactory::getDefaultProperties(),
            [
                $this->getProperty(GenerisRdf::PROPERTY_ALIAS),
                $this->getProperty(GenerisRdf::PROPERTY_IS_LG_DEPENDENT),                    
                $this->getProperty(TaoOntology::PROPERTY_GUI_ORDER),
                $this->getProperty(ValidationRuleRegistry::PROPERTY_VALIDATION_RULE)
            ],
            $this->getFeatureFlagChecker()->isEnabled('FEATURE_FLAG_STATISTIC_METADATA_IMPORT')
                        ? $this->getProperty(GenerisRdf::PROPERTY_IS_STATISTICAL)
                        : [],
        );
```

gabrielfs7: I cant, cause I need the new field to be just after `PROPERTY_IS_LG_DEPENDENT`  and I do not want to overcomplicate and create index with order for the fields, etc.

gabrielfs7: I also not want to split in 3 arrays to avoid using array_filter. I think the cost of having array_filter is not hight compared to visibility/readability.

hectoras: Ok then, I did not thought about the field order :facepalm:

gabrielfs7: Thank you for the suggestion. I was also thinking about your proposal, but turned out array_filter was the clear way I found. If you have other idea, please share :)

AntonNeviarouski:
QA verified

PR #3384, opened 2022-02-02
feat/ADF-922: Add tests for ElementMapFactory

hectoras:
**Associated Jira issue:** [ADF-922](https://oat-sa.atlassian.net/browse/ADF-922)

This PR adds tests for the main/common branches from `ElementMapFactory`:

- Adds `withSomething()` methods to allow injecting mocks for services used from tests.
  -  `withElement()` allows mocking both having an element (`element!=null && hasElement=true`) or not having it (`element==null && hasElement=true`)
- Adds a `isStandaloneMode()` whose return value can be mocked from tests as well
- Adds the tests itself, covering the branches detailed in the attached report

<details>
  <summary>PHPUnit coverage report</summary>

image

</details>

### DIFF
--- a/helpers/form/ElementMapFactory.php
+++ b/helpers/form/ElementMapFactory.php
@@ -152,7 +152,7 @@ class ElementMapFactory extends ConfigurableService
         // Horrible hack to fix file widget
         if ($widgetUri === AsyncFile::WIDGET_ID) {
             $expectedWidget = GenerisAsyncFile::WIDGET_ID;
-            $widgetResource = new core_kernel_classes_Resource($expectedWidget);
+            $widgetResource = $property->getResource($expectedWidget);
         }
 
         if ($this->hasElement) {
--- a/helpers/form/ElementMapFactory.php
+++ b/helpers/form/ElementMapFactory.php
@@ -140,8 +140,8 @@ class ElementMapFactory extends ConfigurableService
             return null;
         }
 
-        $widgetUri      = $widgetResource->getUri();
-        $propertyUri    = $property->getUri();
+        $widgetUri = $widgetResource->getUri();
+        $propertyUri = $property->getUri();
         $expectedWidget = $widgetUri;
 
         // Authoring widget is not used in standalone mode
--- a/helpers/form/ElementMapFactory.php
+++ b/helpers/form/ElementMapFactory.php
@@ -140,8 +140,9 @@ class ElementMapFactory extends ConfigurableService
             return null;
         }
 
-        $widgetUri   = $widgetResource->getUri();
-        $propertyUri = $property->getUri();
+        $widgetUri      = $widgetResource->getUri();
+        $propertyUri    = $property->getUri();
+        $expectedWidget = $widgetUri;
 
         // Authoring widget is not used in standalone mode
         if ($widgetUri === Authoring::WIDGET_ID && $this->isStandaloneMode()) {
@@ -150,7 +151,8 @@ class ElementMapFactory extends ConfigurableService
 
         // Horrible hack to fix file widget
         if ($widgetUri === AsyncFile::WIDGET_ID) {
-            $widgetResource = new core_kernel_classes_Resource(GenerisAsyncFile::WIDGET_ID);
+            $expectedWidget = GenerisAsyncFile::WIDGET_ID;
+            $widgetResource = new core_kernel_classes_Resource($expectedWidget);
         }
 
         if ($this->hasElement) {
@@ -166,7 +168,7 @@ class ElementMapFactory extends ConfigurableService
             return null;
         }
 
-        if ($element->getWidget() !== $widgetUri) {
+        if ($element->getWidget() !== $expectedWidget) {
             common_Logger::w(sprintf(
                 'Widget definition differs from implementation: %s != %s',
                 $element->getWidget(),
--- a/helpers/form/ElementMapFactory.php
+++ b/helpers/form/ElementMapFactory.php
@@ -56,9 +56,38 @@ class ElementMapFactory extends ConfigurableService
     /** @var core_kernel_classes_Resource */
     private $instance;
 
+    /** @var ?bool */
+    private $isStandaloneMode = null;
+
+    /** @var tao_helpers_form_FormElement */
+    private $element = null;
+
+    private $hasElement = false;
+
     public function withInstance(core_kernel_classes_Resource $instance): self
     {
         $this->instance = $instance;
+
+        return $this;
+    }
+
+    /**
+     * Used for tests
+     */
+    public function withStandaloneModel(bool $isStandaloneMode): self
+    {
+        $this->isStandaloneMode = $isStandaloneMode;
+
+        return $this;
+    }
+
+    /**
+     * Used for tests
+     */
+    public function withElement(?tao_helpers_form_FormElement $element): self
+    {
+        $this->hasElement = true;
+        $this->element = $element;
 
         return $this;
     }
@@ -75,11 +104,9 @@ class ElementMapFactory extends ConfigurableService
 
         $widgetUri   = $widgetResource->getUri();
         $propertyUri = $property->getUri();
+
         // Authoring widget is not used in standalone mode
-        if (
-            $widgetUri === Authoring::WIDGET_ID
-            && tao_helpers_Context::check('STANDALONE_MODE')
-        ) {
+        if ($widgetUri === Authoring::WIDGET_ID && $this->isStandaloneMode()) {
             return null;
         }
 
@@ -88,22 +115,14 @@ class ElementMapFactory extends ConfigurableService
             $widgetResource = new core_kernel_classes_Resource(GenerisAsyncFile::WIDGET_ID);
         }
 
-        $element = tao_helpers_form_FormFactory::getElementByWidget(
-            tao_helpers_Uri::encode($propertyUri),
-            $widgetResource
-        );
-
-        if (null === $element) {
+        $element = $this->getElement($propertyUri, $widgetResource);
+        if(null === $element) {
             return null;
         }
 
-        $isListsDependencyEnabled = $this->getFeatureFlagChecker()->isEnabled(
-            FeatureFlagCheckerInterface::FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED
-        );
-
         $parentProperty = null;
 
-        if ($isListsDependencyEnabled) {
+        if ($this->isListDependencyEnabled()) {
             $parentProperty = $this->getParentProperty($property);
 
             if ($parentProperty) {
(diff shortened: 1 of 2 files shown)


hectoras:
Note I think d8d54220f6580e16a00404c2c525a41376fed326 solves an issue, but it would be good for it to be double check by someone else just in case.

`if($widgetUri === AsyncFile::WIDGET_ID)` (that is `http://www.tao.lu/datatypes/WidgetDefinitions.rdf#AsyncFile`), then we replaced the resource with `$widgetResource = new core_kernel_classes_Resource(GenerisAsyncFile::WIDGET_ID)`. (that is, `http://www.tao.lu/datatypes/WidgetDefinitions.rdf#GenerisAsyncFile`). `$widgetUri` keept its value as `http://www.tao.lu/datatypes/WidgetDefinitions.rdf#AsyncFile`.

Then, the conditional `if ($element->getWidget() !== $widgetUri)` would compare the former value (`http://www.tao.lu/datatypes/WidgetDefinitions.rdf#AsyncFile`) with the URI from `GenerisAsyncFile::WIDGET_ID` (`http://www.tao.lu/datatypes/WidgetDefinitions.rdf#GenerisAsyncFile`, used when instantiating the new `$widgetResource`), so we'll always end up returning `null` for `AsyncFile::WIDGET_ID`.

The commit linked above ensures we compare the widget URI with the last URI used to instantiate it, which I think is the correct behaviour.

helpers/form/ElementMapFactory.php
```diff
@@ -56,13 +57,79 @@ class ElementMapFactory extends ConfigurableService
     /** @var core_kernel_classes_Resource */
     private $instance;
 
+    /** @var ?bool */
```
shpran, line 60: As I see, you are not adding `null` comment to other class properties, but they are also may be nullable - please, keep the same style.
```suggestion
    /** @var bool */
```

helpers/form/ElementMapFactory.php
```diff
@@ -56,13 +57,79 @@ class ElementMapFactory extends ConfigurableService
...
     }
 
+    public function withStandaloneMode(bool $isStandaloneMode): self
+    {
+        $this->isStandaloneMode = $isStandaloneMode;
+
+        return $this;
+    }
+
+    public function withElement(?tao_helpers_form_FormElement $element): self
```
shpran, line 97: Are you sure allowing `null` as an element is a good idea?

If you specify `null`, you'll still set `$this->hasElement` to `true`, but that's not correct.

helpers/form/ElementMapFactory.php
```diff
@@ -73,45 +140,35 @@ public function create(core_kernel_classes_Property $property): ?tao_helpers_for
             return null;
         }
 
-        $widgetUri   = $widgetResource->getUri();
-        $propertyUri = $property->getUri();
+        $widgetUri      = $widgetResource->getUri();
+        $propertyUri    = $property->getUri();
+        $expectedWidget = $widgetUri;
```
shpran, line 145:
```suggestion
        $widgetUri = $widgetResource->getUri();
        $propertyUri = $property->getUri();
        $expectedWidget = $widgetUri;
```

helpers/form/ElementMapFactory.php
```diff
@@ -73,45 +140,35 @@ public function create(core_kernel_classes_Property $property): ?tao_helpers_for
...
-        ) {
+        if ($widgetUri === Authoring::WIDGET_ID && $this->isStandaloneMode()) {
             return null;
         }
 
         // Horrible hack to fix file widget
         if ($widgetUri === AsyncFile::WIDGET_ID) {
-            $widgetResource = new core_kernel_classes_Resource(GenerisAsyncFile::WIDGET_ID);
+            $expectedWidget = GenerisAsyncFile::WIDGET_ID;
+            $widgetResource = new core_kernel_classes_Resource($expectedWidget);
```
shpran, line 155:
```suggestion
            $widgetResource = $property->getResource($expectedWidget);
```

helpers/form/ElementMapFactory.php
```diff
@@ -73,45 +140,35 @@ public function create(core_kernel_classes_Property $property): ?tao_helpers_for
...
-            $widgetResource
-        );
+        if ($this->hasElement) {
+            $element = $this->element;
+        } else {
+            $element = tao_helpers_form_FormFactory::getElementByWidget(
+                tao_helpers_Uri::encode($propertyUri),
+                $widgetResource
+            );
+        }
```
shpran, line 165: In this case, you don't need `$this->hasElement` property anymore.
```suggestion
        $element = $this->element ?? tao_helpers_form_FormFactory::getElementByWidget(
            tao_helpers_Uri::encode($propertyUri),
            $widgetResource
        );
```

helpers/form/ElementMapFactory.php
```diff
@@ -256,27 +337,63 @@ private function getListValues(
...
-        return $this->getValueCollectionService()->findAll(new ValueCollectionSearchInput($searchRequest));
+        return $this->getValueCollectionService()->findAll(
+            new ValueCollectionSearchInput($searchRequest)
+        );
+    }
+
+    private function isStandaloneMode(): bool
+    {
+        if ($this->isStandaloneMode === null) {
+            $this->isStandaloneMode = tao_helpers_Context::check('STANDALONE_MODE');
```
shpran, line 348: Is it possible to store this `STANDALONE_MODE` value somewhere?

helpers/form/ElementMapFactory.php
```diff
@@ -256,27 +337,63 @@ private function getListValues(
...
+            $this->valueCollectionService = $this->getContainer()->get(ValueCollectionService::class);
+        }
+
+        return $this->valueCollectionService;
     }
 
     private function getFeatureFlagChecker(): FeatureFlagCheckerInterface
     {
-        return $this->getContainer()->get(FeatureFlagChecker::class);
+        if ($this->featureFlagChecker == null) {
```
shpran, line 374:
```suggestion
        if ($this->featureFlagChecker === null) {
```

helpers/form/ElementMapFactory.php
```diff
@@ -256,27 +337,63 @@ private function getListValues(
...
+            $this->featureFlagChecker = $this->getContainer()->get(FeatureFlagChecker::class);
+        }
+
+        return $this->featureFlagChecker;
     }
 
     private function getLanguageClassSpecification(): LanguageClassSpecification
     {
-        return $this->getContainer()->get(LanguageClassSpecification::class);
+        if ($this->languageClassSpecification == null) {
```
shpran, line 383:
```suggestion
        if ($this->languageClassSpecification === null) {
```

helpers/form/ElementMapFactory.php
```diff
@@ -256,27 +337,63 @@ private function getListValues(
...
+            $this->languageClassSpecification = $this->getContainer()->get(LanguageClassSpecification::class);
+        }
+
+        return $this->languageClassSpecification;
     }
 
     private function getLanguageListElementSortService(): ListElementSorterInterface
     {
-        return $this->getContainer()->get(LanguageListElementSortService::class);
+        if ($this->languageClassSpecification == null) {
```
shpran, line 392:
```suggestion
        if ($this->languageClassSpecification === null) {
```

test/unit/helpers/form/ElementMapFactoryTest.php
```diff
@@ -0,0 +1,509 @@
...
+    {
+        return [
+            'Happy Path' => [
+                'expected' => $this->mockNamedElement(
+                    tao_helpers_form_elements_Authoring::WIDGET_ID,
+                    'property label'
+                ),
+                'expectedOptions' => [
+
+                ],
```
shpran, line 198: As I see it, in all cases the `expectOptions` element is empty - do we need it?
```suggestion
                'expectedOptions' => [],
```

test/unit/helpers/form/ElementMapFactoryTest.php
```diff
@@ -0,0 +1,509 @@
...
+                'validators' => [],
+            ],
+            'Happy Path with instance' => [
+                'expected' => $this->mockNamedElement(
+                    tao_helpers_form_elements_Authoring::WIDGET_ID,
+                    'property label'
+                ),
+                'expectedOptions' => [
+
+                ],
```
shpran, line 220:
```suggestion
                'expectedOptions' => [],
```

test/unit/helpers/form/ElementMapFactoryTest.php
```diff
@@ -0,0 +1,509 @@
...
+                ),
+            ],
+            'Happy Path with instance & listDependencyEnabled' => [
+                'expected' => $this->mockNamedElement(
+                    tao_helpers_form_elements_Authoring::WIDGET_ID,
+                    'property label'
+                ),
+                'expectedOptions' => [
+
+                ],
```
shpran, line 245:
```suggestion
                'expectedOptions' => [],
```

test/unit/helpers/form/ElementMapFactoryTest.php
```diff
@@ -0,0 +1,509 @@
...
+                ),
+            ],
+            'AsyncFile is implicitly converted to GenerisAsyncFile' => [
+                'expected' => $this->mockNamedElement(
+                    tao_helpers_form_elements_GenerisAsyncFile::WIDGET_ID,
+                    'property label'
+                ),
+                'expectedOptions' => [
+
+                ],
```
shpran, line 270:
```suggestion
                'expectedOptions' => [],
```

helpers/form/ElementMapFactory.php
hectoras: Yes, that's intentional. Calling `withElement` with `null` makes:

- `$this->element == null`
-  `$this->hasElement = true`

That allows testing the scenario that happens when `tao_helpers_form_FormFactory::getElementByWidget(...)` returns `null` (but still skipping the call to `getElementByWidget`).

https://github.com/oat-sa/tao-core/pull/3384/files#diff-40ca7eeede1916261bfa4c0f0ac0a5f2cf8c3cfd48146dc0113509e29615820aR340

helpers/form/ElementMapFactory.php
hectoras: See https://github.com/oat-sa/tao-core/pull/3384/files#r802472072, this is used to explicitly set the element to `null` in tests while also skipping the call to `getElementByWidget()`.

helpers/form/ElementMapFactory.php
```diff
@@ -56,13 +57,79 @@ class ElementMapFactory extends ConfigurableService
...
+
+        return $this;
+    }
+
+    public function withValueCollectionService(ValueCollectionService $service): self
+    {
+        $this->valueCollectionService = $service;
+
+        return $this;
+    }
```
gabrielfs7, line 131: You do not need to create those methods, since this is a configurableService and is 100% mockable in unit tests (by mocking container get - I can provide example).

helpers/form/ElementMapFactory.php
gabrielfs7: Why do we need `$this->hasElement`? I agree with @shpran, we should not allow null here.

helpers/form/ElementMapFactory.php
```diff
@@ -56,13 +57,79 @@ class ElementMapFactory extends ConfigurableService
...
     private $instance;
 
+    /** @var ?bool */
+    private $isStandaloneMode;
+
+    /** @var tao_helpers_form_FormElement */
+    private $element;
+
+    /**
+     * Used to allow setting the element explicitly to null in tests.
```
gabrielfs7, line 67: We do not need this attribute.

helpers/form/ElementMapFactory.php
```diff
@@ -256,27 +337,63 @@ private function getListValues(
...
+            $this->isStandaloneMode = tao_helpers_Context::check('STANDALONE_MODE');
+        }
+
+        return $this->isStandaloneMode;
+    }
+
+    private function getValidationRuleRegistry(): AbstractRegistry
+    {
+        if ($this->validationRuleRegistry === null) {
+            $this->validationRuleRegistry = ValidationRuleRegistry::getRegistry();
```
gabrielfs7, line 357: We might consider (if is safe) to add `ValidationRuleRegistry::getRegistry()` to the DI container using factory method and then just call it here. Then you would also **not** need a setter for this.

helpers/form/ElementMapFactory.php
hectoras: But then, when we are going to use the element, we have:

```
        $element = $this->element ?? tao_helpers_form_FormFactory::getElementByWidget(
            tao_helpers_Uri::encode($propertyUri),
            $widgetResource
        );
```

In that case we cannot test what happens when `getElementByWidget()` returns `null` because we cannot mock the result from that static call: We cannot make that return `null`, that static method itself calls `WidgetRegistry::getWidgetDefinition()` which is also static; that last one tries to get the widgets from some cache, but I think I've tried to change how that works from tests in the past and is messy (using reflection to modify `WidgetRegistry::$widgetCache` etc -- that would mean changing global state and can interfere with other tests as well).

`$this->hasElement` allows to mock the same behaviour we have when `tao_helpers_form_FormFactory::getElementByWidget()` returns `null` while preventing the call itself.

test/unit/helpers/form/ElementMapFactoryTest.php
hectoras: It is not needed _yet_: The idea is to use this variable to pass the options `ElementMapFactory::create()` should have set in the `tao_helpers_form_FormElement` instance it returns, but that is not being tested yet as part of this PR to prevent making it bigger.

helpers/form/ElementMapFactory.php
gabrielfs7: IMO It does not worth to do this adaptation in the class just to check this case. We are polluting the subject of test just because the clients are static. I would create a non static `tao_helpers_form_FormFactory->createElementByWidget()` and use it here.

helpers/form/ElementMapFactory.php
hectoras: Changed in 2fa724fe1605631ed7ac532d45084a763206e6ee, thanks!

PR #3395, opened 2022-02-07
feat: add pubsub integration for statistic file upload

gabrielfs7:
Add PubSub integration with statistics

https://oat-sa.atlassian.net/browse/ADF-938

### PRs

- [ ] https://github.com/oat-sa/extension-tao-delivery-rdf/pull/414


models/classes/Observer/GCP/PubSubClientFactory.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * @return \Google\Cloud\PubSub\PubSubClient
+     *
+     * @throws ErrorException
+     */
+    public function create(array $config = [])
+    {
+        if (!class_exists('Google\Cloud\PubSub\PubSubClient') || !$this->getPubSubFilePath()) {
```
hectoras, line 45: Suggestion: I think you may let PHP itself expand the FQN for the class, benefiting from the `use` clause you already have

```suggestion
        if (!class_exists(PubSubClient::class) || !$this->getPubSubFilePath()) {
```

shpran: In this case, PHP may fail if `PubSubClient` class does not exist.
I think we should keep it as a string instead of calling `::class`.

models/classes/Observer/GCP/PubSubObserver.php
```diff
@@ -0,0 +1,78 @@
...
+                ]
+            );
+
+        $this->logger->info(
+            sprintf(
+                'Pub/Sub observer updated for topic "%s", subject "%s", data "%s", messages "%s"',
+                $this->config[self::CONFIG_TOPIC],
+                get_class($subject),
+                substr((string)json_encode($subject), 0, 250) . '...',
+                var_export($messageIds, true)
```
hectoras, line 74: Suggestion: Since these are just message IDs [[ref](http://googleapis.github.io/google-cloud-php/#/docs/cloud-pubsub/v1.35.0/pubsub/topic?method=publish)], using `implode` may be a bit clearer.

```suggestion
                implode(', ', $messageIds)
```

models/classes/Observer/GCP/PubSubObserver.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public function __construct(PubSubClient $pubSubClient, LoggerInterface $logger, array $config = [])
+    {
+        $this->config = $config;
+        $this->pubSubClient = $pubSubClient;
+        $this->logger = $logger;
+    }
+
+    /**
+     * @inheritDoc
```
shpran, line 52: Maybe I'm wrong, but the right syntax for this is `{@inheritdoc }` 😅 

https://pear.php.net/reference/PhpDocumentor-latest/phpDocumentor/tutorial_tags.inlineinheritdoc.pkg.html

models/classes/Observer/Subject.php
```diff
@@ -0,0 +1,63 @@
...
+    public function withData(array $data): SubjectInterface
+    {
+        $this->data = $data;
+
+        return $this;
+    }
+
+    public function attach(SplObserver $observer)
+    {
+        $this->observers[get_class($observer)] = $observer;
```
hectoras, line 44: This means we are not allowing more than a single observer with the same object type, right?

(It that's not the case, I think we may use [spl_object_hash](https://www.php.net/manual/en/function.spl-object-hash.php) instead to generate the key)

models/classes/Observer/GCP/PubSubClientFactory.php
hectoras: It should not fail but just check if the class exists, so `::class` works like if we were copypasting the FQN here (I've double-checked just in case :D).

```
$ cat expand-class-fqn.php 
<?php

use Google\Cloud\PubSub\PubSubClient;

use Some\Random\ClassName;

if (!class_exists(PubSubClient::class)) {
        echo PubSubClient::class." does not exist\n";
}
if (!class_exists(ClassName::class)) {
        echo ClassName::class." does not exist\n";
}

$ php expand-class-fqn.php 
Google\Cloud\PubSub\PubSubClient does not exist
Some\Random\ClassName does not exist

```

models/classes/Observer/GCP/PubSubObserver.php
gabrielfs7: Only if I want to extend the description https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#41-making-inheritance-explicit-using-the-inheritdoc-tag 

If not, I can keep as it it

models/classes/StatisticalMetadata/Import/Observer/ObserverFactory.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->environmentVars = $environmentVars ?? $_ENV;
+        $this->pubSubClientFactory = $pubSubClientFactory;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function create(array $config = [])
+    {
+        if (class_exists('Google\Cloud\PubSub\PubSubClient') && $this->getPubSubTopic()) {
```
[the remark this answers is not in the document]
gabrielfs7: Yes, I can. I did not add, cause the output of `PubSubClient::class` will be `PubSubClient` and if for some reason, there is a library not using namespace and having this class, it might pass. 

Maybe I am over concerned about this. Wdyt?

models/classes/Observer/GCP/PubSubObserver.php
gabrielfs7: I did consider this, but I changed to var_export, cause if we return more info, I would have it logged, but I agree with you might not be necessary.

models/classes/Observer/GCP/PubSubClientFactory.php
gabrielfs7: https://github.com/oat-sa/tao-core/pull/3395#discussion_r802604349

gabrielfs7: Done :)

models/classes/Observer/Subject.php
gabrielfs7: Done :)

models/classes/StatisticalMetadata/Import/Processor/NotifyImportService.php
```diff
@@ -0,0 +1,150 @@
...
+    public function withAliases(array $aliases): self
+    {
+        $this->aliases = $aliases;
+
+        return $this;
+    }
+
+    /**
+     * @throws UserErrorException
+     */
```
shpran, line 89: We already have this inside current method

models/classes/StatisticalMetadata/Import/Processor/NotifyImportService.php
```diff
@@ -0,0 +1,150 @@
...
+
+            throw new UserErrorException(
+                __('Unable to sync statistical data. Please, contact the system administrator to for more details')
+            );
+        }
+    }
+
+    /**
+     * @throws Throwable
+     */
```
shpran, line 110: We already have this info inside method

test/unit/models/classes/Observer/GCP/PubSubClientFactoryTest.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        if (!class_exists(PubSubClient::class)) {
+            $this->markTestSkipped();
+        }
+
+        $this->expectException(ErrorException::class);
+
+        (new PubSubClientFactory([]))->create();
+    }
+}
```
shpran, line 47: Missed newline

test/unit/models/classes/Observer/GCP/PubSubObserverTest.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        if (!class_exists(PubSubClient::class)) {
+            $this->markTestSkipped();
+        }
+
+        $this->expectException(ErrorException::class);
+
+        (new PubSubObserver($this->pubSubClient, $this->logger))->update($this->createMock(SubjectInterface::class));
+    }
+}
```
shpran, line 111: Missed newline

shpran (approved):

AntonNeviarouski:
QA verified

models/classes/StatisticalMetadata/Import/Processor/NotifyImportService.php
gabrielfs7: This is for the client awareness.

models/classes/StatisticalMetadata/Import/Processor/NotifyImportService.php
gabrielfs7: This is for the client awareness.

PR #3405, opened 2022-02-16
fix/TR-2831_fix-remove-theme-by-id

btamas:
Related to https://oat-sa.atlassian.net/browse/TR-2831

Theme handler remove functionality did not work correctly, because the key is postfixed with an id as well (https://github.com/oat-sa/tao-core/blob/master/models/classes/theme/ThemeServiceAbstract.php#L133), like `bosa_selor_default_theme0`, `bosa_selor_default_theme1`.

### DIFF
--- a/models/classes/theme/ThemeService.php
+++ b/models/classes/theme/ThemeService.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2015 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ * Copyright (c) 2015-2022 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
  *
  */
 
@@ -102,10 +102,14 @@ class ThemeService extends ThemeServiceAbstract
             return false;
         }
 
-        $themes = $this->getOption(static::OPTION_AVAILABLE);
-        unset($themes[$themeId]);
+        $availableThemes = $this->getOption(static::OPTION_AVAILABLE);
+        foreach ($this->getAllThemes() as $key => $theme) {
+            if ($theme->getId() === $themeId) {
+                unset($availableThemes[$key]);
+            }
+        }
 
-        $this->setOption(static::OPTION_AVAILABLE, $themes);
+        $this->setOption(static::OPTION_AVAILABLE, $availableThemes);
 
         return true;
     }


siwane (approved):

models/classes/theme/ThemeService.php
```diff
@@ -102,10 +102,14 @@ public function removeThemeById($themeId)
...
         }
 
-        $themes = $this->getOption(static::OPTION_AVAILABLE);
-        unset($themes[$themeId]);
+        $availableThemes = $this->getOption(static::OPTION_AVAILABLE);
+        foreach ($this->getAllThemes() as $key => $theme) {
+            if ($theme->getId() === $themeId) {
+                unset($availableThemes[$key]);
+            }
+        }
```
gabrielfs7, line 110: As I saw the $key is always the themeId, so you can reduce your code to:

```suggestion
        $availableThemes = $this->getOption(static::OPTION_AVAILABLE);
        
        unset($availableThemes[$themeId]);
```

The unset will not throw any error in case the index does not exist.

models/classes/theme/ThemeService.php
```diff
@@ -102,10 +102,14 @@ public function removeThemeById($themeId)
             return false;
         }
 
-        $themes = $this->getOption(static::OPTION_AVAILABLE);
-        unset($themes[$themeId]);
+        $availableThemes = $this->getOption(static::OPTION_AVAILABLE);
+        foreach ($this->getAllThemes() as $key => $theme) {
+            if ($theme->getId() === $themeId) {
+                unset($availableThemes[$key]);
```
wazelin, line 108: No need to iterate further after the key has been identified.
```suggestion
                unset($availableThemes[$key]);

                break;
```

models/classes/theme/ThemeService.php
wazelin: That's almost exactly the code before the changes ;)

The PR description suggests the theme ID gets postfixed with an index or something.

models/classes/theme/ThemeService.php
wazelin: In fact, it would be a tad more readable if another method was introduced, fetching a key corresponding to a theme ID.
e.g.

```php
private function findThemeKey(string $id): ?string
{
    foreach ($this->getAllThemes() as $key => $theme) {
        if ($theme->getId() === $id) {
            return $key;
        }
    }

    return null;
}
```

That IMO gives a better understanding of what that loop's purpose is.

Then you can reduce your main unsetting logic to `unset($availableThemes[$this->findThemeKey($themeId)]);`

wazelin (changes requested):

models/classes/theme/ThemeService.php
btamas: The issue is exactly that. The `$key` is not the `$themeId`. A theme can be added multiple times and it is postfixed with an id: https://github.com/oat-sa/tao-core/blob/master/models/classes/theme/ThemeServiceAbstract.php#L133
Keys are `bosa_selor_default_theme0` and `bosa_selor_default_theme1` etc.
So, the idea here is to request the id from the theme class and if it is the same as the id what should be removed, remove it.

models/classes/theme/ThemeService.php
btamas: > That IMO gives a better understanding of what that loop's purpose is.
> 
> Then you can reduce your main unsetting logic to `unset($availableThemes[$this->findThemeKey($themeId)]);`

All should be removed, so it would not work.

gabrielfs7 (approved):

models/classes/theme/ThemeService.php
gabrielfs7: 🤦  True, I did not analyse it right. I've approved the PR :)

wazelin (approved):

PR #3406, opened 2022-02-16
fix: add regex validator for login and allow to fix broken login

shpran:
# [ADF-940](https://oat-sa.atlassian.net/browse/ADF-940)

## Companion PRs
* https://github.com/oat-sa/extension-tao-delivery/pull/540

## Changes
* Added regex validator for login field. The following characters no longer allowed: `"`, `'`, `\`, `/`, `>`, `<`
* Regex validator for login field is configured to be called during form initialization (preliminary validation) and after form submit
* Added possibility to fix existing logins, which contains invalid characters

### DIFF
--- a/actions/form/class.Users.php
+++ b/actions/form/class.Users.php
@@ -17,48 +17,32 @@
  *
  * Copyright (c) 2008-2010 (original work) Deutsche Institut für Internationale Pädagogische Forschung (under the project TAO-TRANSFER);
  *               2009-2012 (update and modification) Public Research Centre Henri Tudor (under the project TAO-SUSTAIN & TAO-DEV);
- *
+ *               2022 (original work) Open Assessment Technologies SA.
  */
 
-use oat\generis\model\GenerisRdf;
-use oat\generis\model\user\PasswordConstraintsService;
-use oat\tao\helpers\ApplicationHelper;
+use oat\generis\model\OntologyRdfs;
 use oat\generis\model\user\UserRdf;
-use oat\oatbox\user\UserLanguageServiceInterface;
+use Psr\Container\ContainerInterface;
+use oat\oatbox\service\ServiceManager;
+use oat\tao\helpers\ApplicationHelper;
 use oat\tao\model\controller\SignedFormInstance;
+use oat\oatbox\user\UserLanguageServiceInterface;
+use oat\generis\model\user\PasswordConstraintsService;
+use oat\tao\helpers\form\Feeder\SanitizerValidationFeeder;
+use oat\tao\helpers\form\Feeder\SanitizerValidationFeederInterface;
 
 /**
  * This container initialize the user edition form.
  *
- * @access public
  * @author Joel Bout, <joel.bout@tudor.lu>
- * @package tao
-
  */
 class tao_actions_form_Users extends SignedFormInstance
 {
-    // --- ASSOCIATIONS ---
-
-
-    // --- ATTRIBUTES ---
-
-    /**
-     * Short description of attribute user
-     *
-     * @access protected
-     * @var Resource
-     */
+    /** @var core_kernel_classes_Resource */
     protected $user;
 
-    /**
-     * Short description of attribute formName
-     *
-     * @access protected
-     * @var string
-     */
+    /** @var string */
     protected $formName = '';
-
-    // --- OPERATIONS ---
 
     /**
      * Short description of method __construct
@@ -107,7 +91,7 @@ class tao_actions_form_Users extends SignedFormInstance
             $options['excludedProperties'][] = UserRdf::PROPERTY_DEFLG;
         }
 
-        $options['topClazz'] = GenerisRdf::CLASS_GENERIS_USER;
+        $options['topClazz'] = UserRdf::CLASS_URI;
 
         parent::__construct($clazz, $this->user, $options);
     }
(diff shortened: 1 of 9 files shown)


actions/form/class.Users.php
```diff
@@ -257,4 +244,48 @@ protected function initElements()
             }
         }
     }
+
+    private function initLoginElement(): void
```
gabrielfs7, line 248: Is it also covering test-takers or only backoffice users?

Does it block script execution for old users/test-takers that had script as login (or other field)?

shpran: Oh, this is executed by logging in as TT

hectoras (approved):

actions/form/class.Users.php
```diff
@@ -257,4 +244,50 @@ protected function initElements()
...
+            'Regex',
+            [
+                'isPreValidationRequired' => true,
+                'format' => '/^[^"\'<>\\/]+$/',
+                'message' => 'Login must not contain the following characters: ", \', \, /, <, >',
+            ]
+        );
+
+        $element->feedInputValue();
+        $value = $element->getInputValue() ?? $element->getRawValue();
```
gabrielfs7, line 264: You could have a separate class to apply the validation based on the inputs:

```
class SanitizerValidationFeeder
{
    public function feed(Form $form, array $fields): void
    {
        foreach ($fields as $field) {
            $element = $this->form->getElement(tao_helpers_Uri::encode($field));
  
            /** @var tao_helpers_form_Validator $regexValidator */
            $regexValidator = tao_helpers_form_FormFactory::getValidator(
                'Regex',
                [
                    'isPreValidationRequired' => true,
                    'format' => '/^[^"\'<>\\/]+$/',
                    'message' => 'Field not contain the following characters: ", \', \, /, <, >',
                ]
            );
  
            $element->feedInputValue();
            $value = $element->getInputValue() ?? $element->getRawValue();

            if ($value !== null) {
                $element->addValidator($regexValidator);
            }
    }
}
```

And then you can add it here with the firtstaname, lastname, login, etc...

Wdyt?

helpers/form/Feeder/SanitizerValidationFeeder.php
```diff
@@ -0,0 +1,135 @@
...
+        return $element->getInputValue() ?? $element->getRawValue();
+    }
+
+    private function reset(): void
+    {
+        $this->validators = [];
+        $this->elements = [];
+    }
+
+    private function checkFormAvailability(): void
```
hectoras, line 124: Suggestion: Naming this sort of methods used for preconditions as `assertXXX` may make the caller code a bit more readable  (for example, it may be named `assertHasForm()` or `assertFormIsSet()` as you would do in tests). BTW this may be a matter of personal preference :)

helpers/form/Feeder/SanitizerValidationFeeder.php
```diff
@@ -0,0 +1,135 @@
...
+use tao_helpers_Uri;
+use RuntimeException;
+use tao_helpers_form_Form;
+use InvalidArgumentException;
+use tao_helpers_form_Validator;
+use tao_helpers_form_FormElement;
+use tao_helpers_form_elements_Textbox;
+use tao_helpers_form_elements_Textarea;
+
+class SanitizerValidationFeeder implements SanitizerValidationFeederInterface
```
gabrielfs7, line 34: - Please, do not forget to add tests 🙏

helpers/form/Feeder/SanitizerValidationFeederInterface.php
```diff
@@ -0,0 +1,47 @@
...
+     * @throws RuntimeException
+     */
+    public function addElement(tao_helpers_form_FormElement $element): self;
+
+    /**
+     * @throws RuntimeException
+     */
+    public function addElementByUri(string $elementUri): self;
+
+    public function feed(): void;
```
gabrielfs7, line 46: Should not the feed throw an exception in case the form or elements are not provided?

helpers/form/validators/class.Regex.php
```diff
@@ -16,19 +17,24 @@
...
 /**
- * Short description of class tao_helpers_form_validators_Regex
- *
- * @access public
  * @author Joel Bout, <joel.bout@tudor.lu>
- * @package tao
-
  */
-class tao_helpers_form_validators_Regex extends tao_helpers_form_Validator
+class tao_helpers_form_validators_Regex extends tao_helpers_form_Validator implements PreliminaryValidationInterface
```
gabrielfs7, line 30: Will it impact other forms using this validator?

helpers/form/Feeder/SanitizerValidationFeederInterface.php
shpran: Nope
We will call this feed for all the forms
But if no validators or/and elements provided - this will be safely skipped

Right now if we want to cover any form with this feeder - you need to add validators and elements - done 🙂

helpers/form/class.FormContainer.php
```diff
@@ -118,6 +128,8 @@ public function __construct(array $data = [], array $options = [])
                 $this->form->disable();
             }
 
+            $this->getSanitizerValidationFeeder()->feed();
```
gabrielfs7, line 131: We are impacting all the forms in the platform now. Are we sure we want to add this here?

helpers/form/validators/class.Regex.php
shpran: Nope.
I'm using a method from this interface to check if pre-validation required.
For this class I'm using default value (`false` - simple behavior).

gabrielfs7 (approved):
Thanks Andrei, for improving the platform! 🚀 

helpers/form/Feeder/SanitizerValidationFeeder.php
shpran: Thanks a lot for your suggestion 🙂 
Applied

PR #3420, opened 2022-02-18
Draft: Move logic common to all search index tasks into a parent class

hectoras:
**Associated Jira issue:** [ADF-568](https://oat-sa.atlassian.net/browse/ADF-568)

Result of an Exploratory Refactoring exercise: To be used as reference, but not to be merged.


gabrielfs7:
Closing just to avoid confusion in the review

PR #3442, opened 2022-03-03
Fix/TR-3634/Switch from hash keys to simple keys, use ttl for tokens

shpran:
# [TR-3634](https://oat-sa.atlassian.net/browse/TR-3634)

## Companion PRs
* https://github.com/oat-sa/extension-tao-bosa-selor/pull/117

## Changes
* Removed unused `TOKEN_NAME` constant
* Added/changed constants' visibility:
  * `const OPTION_PERSISTENCE` ===> `public const OPTION_PERSISTENCE`
  * `const TOKENS_STORAGE_KEY` ===> `private const TOKENS_STORAGE_KEY`
* Changed visibility for methods:
  * `protected function getPersistence()` ===> `private function getPersistence()`
  * `protected function getKey()` ===> `private function getKey()`
* Changed all hash function calls to simple key functions
* Added new option `ttl` to configure TTL value

### DIFF
--- a/models/classes/security/xsrf/TokenStore.php
+++ b/models/classes/security/xsrf/TokenStore.php
@@ -27,6 +27,8 @@ namespace oat\tao\model\security\xsrf;
  */
 interface TokenStore
 {
+    public const TOKEN_NAME = 'XSRF_TOKEN_NAME';
+
     public function getToken(string $tokenId): ?Token;
 
     public function setToken(string $tokenId, Token $token): void;
--- a/models/classes/security/xsrf/TokenStore.php
+++ b/models/classes/security/xsrf/TokenStore.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 (original work) Open Assessment Technologies SA ;
+ * Copyright (c) 2017-2022 (original work) Open Assessment Technologies SA.
  */
 
 namespace oat\tao\model\security\xsrf;
@@ -27,35 +27,14 @@ namespace oat\tao\model\security\xsrf;
  */
 interface TokenStore
 {
-    const TOKEN_NAME = 'XSRF_TOKEN_NAME';
-
-    /**
-     * @param string $tokenId
-     * @return Token|null
-     */
     public function getToken(string $tokenId): ?Token;
 
-    /**
-     * @param string $tokenId
-     * @param Token $token
-     */
     public function setToken(string $tokenId, Token $token): void;
 
-    /**
-     * @param string $tokenId
-     * @return bool
-     */
     public function hasToken(string $tokenId): bool;
 
-    /**
-     * @param string $tokenId
-     * @return bool
-     */
     public function removeToken(string $tokenId): bool;
 
-    /**
-     * @return void;
-     */
     public function clear(): void;
 
     /**
--- a/models/classes/security/xsrf/TokenStoreKeyValue.php
+++ b/models/classes/security/xsrf/TokenStoreKeyValue.php
@@ -15,16 +15,20 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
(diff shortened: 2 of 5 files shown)


models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -33,115 +37,138 @@
...
     {
         $tokens = [];
-        $tokensData = $this->getPersistence()->hGetAll($this->getKey());
-        if (!is_array($tokensData)) {
-            return $tokens;
-        }
 
-        foreach ($tokensData as $tokenData) {
+        foreach ($this->getKeys() as $key) {
+            $tokenData = $this->getPersistence()->get($key);
```
hectoras, line 100: There is no possibility to query for all keys at once here, right? (for example by making the Redis driver issue a `MGET` to fetch all keys from `getKeys()` in a single request etc)

shpran: I will try, but this might require a lot of changes

hectoras: Hmm... you are right, I'm rechecking this and probably it would need changes in Generis interfaces for cache etc
:grin: Feel free to keep it as it :slightly_smiling_face:

hectoras (approved):

models/classes/security/xsrf/TokenStore.php
```diff
@@ -27,35 +27,14 @@
  */
 interface TokenStore
 {
-    const TOKEN_NAME = 'XSRF_TOKEN_NAME';
```
gabrielfs7, line 30: Nobody was using it? Even customer extensions?

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -33,115 +37,138 @@
...
     }
 
-    /**
-     * {@inheritDoc}
-     */
     public function setToken(string $tokenId, Token $token): void
     {
-        $this->getPersistence()->hSet($this->getKey(), $tokenId, json_encode($token));
+        $this->getPersistence()->set(
+            $this->buildKey($tokenId),
```
gabrielfs7, line 65: - [ ] Missing update the unit tests

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -33,115 +37,138 @@
...
+     * @throws common_exception_Error
+     */
+    private function getPersistence(): common_persistence_AdvKeyValuePersistence
+    {
+        if (!isset($this->persistence)) {
+            $persistence = $this->getPersistenceManager()->getPersistenceById(
+                $this->getOption(self::OPTION_PERSISTENCE)
+            );
+
+            if (!$persistence instanceof common_persistence_AdvKeyValuePersistence) {
```
gabrielfs7, line 132: Are we sure all other customers are using this persistence type? Will it not break some environment?

models/classes/security/xsrf/TokenStore.php
shpran: I found only one (this) place
image

models/classes/security/xsrf/TokenStoreKeyValue.php
shpran: This is old code 🙂 
I just changed the place and method visibility

Previously it was protected and at the top of class
Also, this token store is not a default one - by default all customers use Session store

models/classes/security/xsrf/TokenStore.php
gabrielfs7: Maybe, can we keep and deprecate it to avoid breaking changes?

shpran: Sure
Then I will also rollback other visibility changes to be 110% sure

shpran: But I'm not sure about making this deprecated.
If someone wants to use it - he will see, that this one is deprecated, but there's nothing else to use.
I think I will keep this one without `deprecated` flag

models/classes/security/xsrf/TokenStoreKeyValue.php
gabrielfs7: omg you are right 🤦

gabrielfs7 (approved):

PR #3447, opened 2022-03-10
feat: RDF Object Mapper PoC

hectoras:
**Associated Jira issue:** [ADF-1174](https://oat-sa.atlassian.net/browse/ADF-1174)

[ADF-1174]: https://oat-sa.atlassian.net/browse/ADF-1174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


models/classes/RdfObjectMapper/Annotation/RdfResourceAttributeType.php
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA.
+ */
+
+namespace oat\tao\model\RdfObjectMapper\Annotation;
+
+class RdfResourceAttributeType
+{
+    public const URI = 1;
+    public const LABEL = 2;
+    public const COMMENT = 3;
```
hectoras, line 27: Having only `LITERAL` (for copying values as-is) and `RESOURCE` (for URIs pointing to other resources) may suffice.

The former values types here were matching exactly the fields used for the PoC (basically attributes currently found in `core_kernel_classes_Resource`; as well as having a `URI` type to hold resource URIs (for example, URIs used as _primary_ identifiers for resources).

This may be combined with PHP 7.4+ property type hints as well to validate things like conversions to integer etc.

PR #3451, opened 2022-03-14
fix/ADF-978: Improve performance of item loading for lists in the backend

hectoras:
**Associated Jira issue:** [ADF-978](https://oat-sa.atlassian.net/browse/ADF-978)

Note this removes the call to `enrichWithSelect()` from `count()` methods since the fields added with it are then replaced by `$query->select('count(t.id) AS c')`.

Related PRs:
- https://github.com/oat-sa/extension-tao-backoffice/pull/139

### DIFF
--- a/models/classes/Lists/Business/Domain/ValueCollection.php
+++ b/models/classes/Lists/Business/Domain/ValueCollection.php
@@ -88,19 +88,15 @@ class ValueCollection implements IteratorAggregate, JsonSerializable, Countable
                 continue;
             }
 
-            if (isset($visited[$value->getUri()])) {
-                $duplicationCandidate = $visited[$value->getUri()];
-            } else {
+            if (!isset($visited[$value->getUri()]) || $visited[$value->getUri()] === $value) {
                 $visited[$value->getUri()] = $value;
-                $duplicationCandidate = null;
+                continue;
             }
 
-            if ($duplicationCandidate !== null && $duplicationCandidate !== $value) {
-                $duplicates->addValue($value);
+            $duplicates->addValue($value);
 
-                if ($limit !== null && ++$counter >= $limit) {
-                    break;
-                }
+            if ($limit !== null && ++$counter >= $limit) {
+                break;
             }
         }
 
--- a/models/classes/Lists/Business/Domain/ValueCollection.php
+++ b/models/classes/Lists/Business/Domain/ValueCollection.php
@@ -81,13 +81,19 @@ class ValueCollection implements IteratorAggregate, JsonSerializable, Countable
     {
         $duplicates = new ValueCollection();
         $counter = 0;
+        $visited = [];
 
         foreach ($this->values as $value) {
             if (empty($value->getUri())) {
                 continue;
             }
 
-            $duplicationCandidate = $this->extractValueByUri($value->getUri());
+            if (isset($visited[$value->getUri()])) {
+                $duplicationCandidate = $visited[$value->getUri()];
+            } else {
+                $visited[$value->getUri()] = $value;
+                $duplicationCandidate = null;
+            }
 
             if ($duplicationCandidate !== null && $duplicationCandidate !== $value) {
                 $duplicates->addValue($value);
--- a/models/classes/Lists/Business/Domain/ValueCollection.php
+++ b/models/classes/Lists/Business/Domain/ValueCollection.php
@@ -95,7 +95,8 @@ class ValueCollection implements IteratorAggregate, JsonSerializable, Countable
 
             $duplicates->addValue($value);
 
-            if ($limit !== null && ++$counter >= $limit) {
+            $counter++;
+            if ($limit !== null && $counter >= $limit) {
                 break;
             }
         }
--- a/models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
+++ b/models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2020-2022 (original work) Open Assessment Technologies SA;
  *
  * @author Sergei Mikhailov <sergei.mikhailov@taotesting.com>
  */
@@ -29,7 +29,6 @@ use common_persistence_SqlPersistence as SqlPersistence;
 use core_kernel_classes_Class as KernelClass;
 use core_kernel_classes_Resource as KernelResource;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\Query\QueryBuilder;
 use oat\generis\model\OntologyRdf;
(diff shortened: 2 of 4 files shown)


models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -181,10 +182,12 @@ public function count(ValueCollectionSearchRequest $searchRequest): int
...
 
         $this->enrichQueryWithInitialCondition($query);
-        $this->enrichQueryWithSelect($searchRequest, $query);
         $this->enrichQueryWithValueCollectionSearchCondition($searchRequest, $query);
 
-        return $query->execute()->rowCount();
+        $result = $query->select('count(t.id) AS c')->execute();
+
+        $row = $result->fetch(FetchMode::NUMERIC);
+        return (int)$row[0] ?? 0;
```
gabrielfs7, line 190:
```suggestion
        return (int)($row[0] ?? 0);
```

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -397,13 +399,15 @@ private function getPersistence(): SqlPersistence
...
 
         $this->enrichWithInitialCondition($query);
-        $this->enrichWithSelect($searchRequest, $query);
         $this->enrichQueryWithValueCollectionSearchCondition($searchRequest, $query);
 
-        return $query->execute()->rowCount();
+        $result = $query->select('count(element.id) AS c')->execute();
+
+        $row = $result->fetch(FetchMode::NUMERIC);
+        return (int)$row[0] ?? 0;
```
gabrielfs7, line 410:
```suggestion
        return (int)($result->fetch(FetchMode::NUMERIC)[0] ?? 0);
```

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
gabrielfs7, line 190:
```suggestion
        return (int)($result->fetch(FetchMode::NUMERIC)[0] ?? 0);
```

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -397,13 +399,15 @@ private function getPersistence(): SqlPersistence
...
     public function count(ValueCollectionSearchRequest $searchRequest): int
     {
-        $query = $this->getPersistence()->getPlatForm()->getQueryBuilder();
+        $query = $this->getQueryBuilder();
 
         $this->enrichWithInitialCondition($query);
-        $this->enrichWithSelect($searchRequest, $query);
         $this->enrichQueryWithValueCollectionSearchCondition($searchRequest, $query);
 
-        return $query->execute()->rowCount();
```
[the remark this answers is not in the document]
hectoras: In this case I think we can't, because `rowCount` just returns the number of rows already returned by the `SELECT`, which means we've to wait for the result of the `SELECT` operation to come back to PHP (btw `execute()` & `rowCount()` are part of the underlying Doctrine DBAL library we use under the hood, we cannot modify its behaviour :) )

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
hectoras: This is working fine in my testing environment (which uses Postgres) and locally (using MySQL), so I think we are safe.

models/classes/Lists/Business/Domain/ValueCollection.php
```diff
@@ -81,13 +81,19 @@ public function getDuplicatedValues(int $limit = null): ValueCollection
...
         $counter = 0;
+        $visited = [];
 
         foreach ($this->values as $value) {
             if (empty($value->getUri())) {
                 continue;
             }
 
-            $duplicationCandidate = $this->extractValueByUri($value->getUri());
+            if (isset($visited[$value->getUri()])) {
```
gabrielfs7, line 91: Why do we need this change? I would avoid it, it may cause another regression.

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
```diff
@@ -181,10 +182,12 @@ public function count(ValueCollectionSearchRequest $searchRequest): int
         $query = $this->getQueryBuilder();
 
         $this->enrichQueryWithInitialCondition($query);
-        $this->enrichQueryWithSelect($searchRequest, $query);
         $this->enrichQueryWithValueCollectionSearchCondition($searchRequest, $query);
 
-        return $query->execute()->rowCount();
+        $result = $query->select('count(t.id) AS c')->execute();
```
gabrielfs7, line 187: - [ ] Lets not forget to check unit tests

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
```diff
@@ -397,13 +399,15 @@ private function getPersistence(): SqlPersistence
...
+        $result = $query->select('count(element.id) AS c')->execute();
+
+        $row = $result->fetch(FetchMode::NUMERIC);
```
gabrielfs7, line 409: - [ ] Lets not forget to check unit tests

models/classes/Lists/Business/Domain/ValueCollection.php
hectoras: This change is also done for performance reasons.

`extractValueByURI` iterates on all items that are already in the collection and returns the first one that matches the provided URI. It was called for every item to check which was the first one using that URI in the values array. Then, we compared the URI for the first one against the URI of the current one. Given we have `n` items in the collection, this performs `n` iterations, and each one, in turn, iterates over all items again. Therefore, the former algorithm is `O(n^2)`.

This change makes the method to store items in the collection in a hash table (an array) and performs lookups by URI. As as each array lookup is `O(1)`, this should run in `O(n)` and get a performance boost for big lists.

Note this still is run now for `RdsValueCollectionRepository`, as we fetch all list items when saving even if the frontend just sends the modified ones: That is needed in order to check if the URI provided for any of them is duplicated.

(BTW right now, in my local env, both this and the previous implementation seems to break right now due to the memory limit when used with lists with thousands of items (it reaches 130MB)).

models/classes/Lists/DataAccess/Repository/RdfValueCollectionRepository.php
hectoras: Added: b4591dbcc38f3434bdf71ff684b8064d9356aec4

models/classes/Lists/DataAccess/Repository/RdsValueCollectionRepository.php
hectoras: Added: b4591dbcc38f3434bdf71ff684b8064d9356aec4

models/classes/Lists/Business/Domain/ValueCollection.php
hectoras: Note this is not 100% true anymore, as now in the update use case we only populates existing items that are present in the request instead of the full list: [30ef65ac59b4780cb9eef86f7543e992561e208c](https://github.com/oat-sa/extension-tao-backoffice/pull/139/commits/08caead466fdc8d5cd291050c95df2bfbd8792f9)

By the way, this still is a performance improvement  if the full list is passed to `ValueCollectionService::persist()` (for example, this is used by `RdsValueCollectionRepository`).

gabrielfs7 (approved):

AntonNeviarouski:
QA verified

hectoras:
Released as [v50.10.4](https://github.com/oat-sa/tao-core/releases/tag/v50.10.4).

PR #3456, opened 2022-03-21
Fix/AUT-345/disable label for disabled input

bziondik:
Related to: https://oat-sa.atlassian.net/browse/AUT-345
Depends on: 
- [x] https://github.com/oat-sa/tao-core-ui-fe/pull/444

Fix label of the radio button has disabled state as input.

image


Silvia-loza (approved):
LGTM. Just version to be updated when released :blush:

marpesia (approved):
LGTM

PR #3464, opened 2022-03-31
Add a service to check feature visibility

meefox:
Related to https://oat-sa.atlassian.net/browse/AUT-2042

**Description:**
Add helper to check status (visible/hidden) for the feature using features helper.
It will return visible = true if missing from the configuration.

**Usage**
* helper needs to be injected as `'service/features'`
* then feature can be checked like `serviceFeatures.isVisible('item/multiColumn');`

**Changes:**
* Added helper
* Added test

**Requires:**
* It using `client_lib_config_registry.conf.php` as data source, with structure:
```
'helpers/features' => array(
    'visibility' =>  array(
        "item/multiColumn" => "show",
        "item/scrollableMultiColumn" => "show",
        "item/response/modalFeedback" => "hide",
        "item/interaction/*/shufflingChoices" => "hide",
        "item/customInteraction/*" => "hide",
        "item/customInteraction/audioPciInteraction" => "show",
        "test/item/timeLimits" => "hide",
        "test/item/itemSession/showFeedback" => "hide",
        "test/testPart/itemSession/*" => "hide",
        "delivery/multiEngine" => "hide"
    )
),
```

**Documentation:**
https://oat-sa.atlassian.net/wiki/spaces/~900676257/pages/1764229171/Using+features+helper

### DIFF
--- a/views/js/services/features.js
+++ b/views/js/services/features.js
@@ -43,20 +43,28 @@ define(['module'], function (module) {
         /**
          * Check is feature configured to be visible
          * based on client_lib_config_registry.conf.php
+         * possible match is exact match (item/feature) or wildcard match (item/*)
+         * using 2 wildcarcard symbols * is not supported
          * @param {String} featurePath path to feature supporting * ex('test/itemSession/*')
          * @returns {Boolean} true if feature is visible
          */
         isVisible: featurePath => {
             const regexp = buildRegexp(featurePath);
-            let targetKey;
-            let isPathExists = featuresKeys.some(featureKey => {
-                const isMatch = regexp.test(featureKey);
+            let targetKey = null;
 
-                if (isMatch) targetKey = featureKey;
-                return isMatch;
-            });
+            for (let i in featuresKeys) {
+                const exactMatch = featuresKeys[i] === featurePath;
 
-            return isPathExists && featuresVisibilityList[targetKey] === 'show';
+                if (exactMatch || regexp.test(featuresKeys[i])) {
+                    targetKey = featuresKeys[i];
+                }
+
+                if (exactMatch) {
+                    break;
+                }
+            }
+
+            return targetKey !== null && featuresVisibilityList[targetKey] === 'show';
         }
     };
 });
--- a/views/js/services/features.js
+++ b/views/js/services/features.js
@@ -32,10 +32,10 @@ define(['module', 'core/logger'], function(module, loggerFactory) {
         lookupPath = lookupPath.replace('*', '\\S+');
 
         try {
-            return RegExp(`^${lookupPath}$`);
+            return new RegExp(`^${lookupPath}$`);
         } catch (e) {
             logger.warn(`Lookup feature path ${lookupPath} was not found`);
-            return RegExp('^\0$');
+            return new RegExp('^\0$');
         }
     };
 
--- a/views/js/services/features.js
+++ b/views/js/services/features.js
@@ -41,11 +41,11 @@ define(['module', 'core/logger'], function(module, loggerFactory) {
 
     return {
         /**
-         * Check is feature configured
-         * based on client_lib_config_registry.conf.php to be visible
-         * and features that is not in the configuration are visible by default
-         * @param {String} featurePath full path to feature ex('test/itemSession/feature')
-         * @returns {Boolean} true if feature is visible (or missed from configuration)
+         * Check if the feature is configured to be visible
+         * based on the module configuration
+         * and features that are not in the configuration are visible by default
+         * @param {String} featurePath full path to feature ex('items/feature')
+         * @returns {Boolean} true if feature is visible (or missed from the configuration)
          */
         isVisible(featurePath = '') {
             let matchingPath = null;
--- a/views/js/services/features.js
+++ b/views/js/services/features.js
@@ -41,19 +41,20 @@ define(['module', 'core/logger'], function(module, loggerFactory) {
 
     return {
         /**
-         * Check is feature configured to be visible
-         * based on client_lib_config_registry.conf.php
+         * Check is feature configured
+         * based on client_lib_config_registry.conf.php to be visible
+         * and features that is not in the configuration are visible by default
          * @param {String} featurePath full path to feature ex('test/itemSession/feature')
-         * @returns {Boolean} true if feature is visible
+         * @returns {Boolean} true if feature is visible (or missed from configuration)
          */
         isVisible(featurePath = '') {
-            let targetKey = null;
+            let matchingPath = null;
 
             featuresKeys.some(path => {
                 const exactMatch = path === featurePath;
 
                 if (exactMatch || buildRegexp(path).test(featurePath)) {
-                    targetKey = path;
+                    matchingPath = path;
                 }
 
                 if (exactMatch) {
@@ -61,7 +62,7 @@ define(['module', 'core/logger'], function(module, loggerFactory) {
                 }
             });
 
-            return targetKey !== null && featuresVisibilityList[targetKey] === 'show';
+            return matchingPath === null || featuresVisibilityList[matchingPath] === 'show';
         }
     };
 });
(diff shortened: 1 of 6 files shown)


views/js/services/features.js
```diff
@@ -0,0 +1,70 @@
...
+         * possible match is exact match (item/feature) or wildcard match (item/*)
+         * using 2 wildcarcard symbols * is not supported
+         * @param {String} featurePath path to feature supporting * ex('test/itemSession/*')
+         * @returns {Boolean} true if feature is visible
+         */
+        isVisible: featurePath => {
+            const regexp = buildRegexp(featurePath);
+            let targetKey = null;
+
+            for (let i in featuresKeys) {
```
jsconan, line 55: **nitpick:** `for..in` loops over the keys, and therefore it is not suitable to be used on an array as it would also cover entries that are not array items. Please consider using `.forEach()` instead, or a look that respects the array structure. If you want an early return inside a `forEach()` you could also use `every()` or `some()`.

jsconan (changes requested):
Nice addition!

Regarding the implementation, there are a few things to look at, as you can see in the coment below.

Globally speaking, I wonder if we ever need to support the wildcard from the consumer API. For the definition, yes we need to wildcard the namespace. But not sure it is needed for the reading. Anyway, this is not a problem in the end.

Unfortunately, the unit tests don't pass in CLI.

views/js/services/features.js
```diff
@@ -0,0 +1,70 @@
...
+    const featuresVisibilityList = config.visibility || [];
+    const featuresKeys = Object.keys(featuresVisibilityList);
+
+    /**
+     * Build regexp from lookupPath converting '*' to '\S*'
+     * @param {String} lookupPath raw string of path to lookup
+     * @returns {RegExp} regexp to lookup in features list
+     */
+    const buildRegexp = lookupPath => {
+        lookupPath = lookupPath.replaceAll('*', '\\S*');
```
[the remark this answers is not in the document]
meefox: S+ better indeed.
Actually, if we want to support only one wild card. 
Just one simple .replace() may be sufficient enough in this case.

views/js/test/services/features/test.html
```diff
@@ -0,0 +1,50 @@
...
+<html>
+    <head>
+        <meta charset="utf-8">
+        <title>Test - services/features</title>
+        <base href="../../../../" />
+        <link rel="stylesheet" type="text/css" href="js/lib/qunit/qunit.css">
+        <script type="text/javascript" src="js/lib/qunit/qunit.js"></script>
+        <script type="text/javascript" src="js/lib/require.js"></script>
+        <script type="text/javascript" src="js/lib/qunit/qunit-parameterize.js"></script>
+        <script type="text/javascript" src="js/lib/blanket/blanket.min.js" data-cover-only="i18n.js"></script>
```
jsconan, line 11: **core:** This can be removed. `Blanket` is not compatible with ES6+. Beside this, the target is not accurate :) (`ata-cover-only="i18n.js` while the test module is `services/features`)

views/js/services/features.js
```diff
@@ -0,0 +1,68 @@
...
+     * @param {String} lookupPath raw string of path to lookup
+     * @returns {RegExp} regexp to lookup in features list
+     */
+    const buildRegexp = lookupPath => {
+        lookupPath = lookupPath.replace('*', '\\S+');
+
+        try {
+            return RegExp(`^${lookupPath}$`);
+        } catch (e) {
+            const logger = loggerFactory('services/features');
```
jsconan, line 36: **nitpick:** The logger needs to be created only once, at the module level.

jsconan (changes requested):
It looks better, and the unit tests don't fail in CLI 👍 

However, I found a possible issue with the fallback RegEx.
Plus some other comments

views/js/services/features.js
meefox: Error here considering as corner-case my intention not waste resource on something that not supposed to happen :)

views/js/services/features.js
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    return {
+        /**
+         * Check is feature configured to be visible
+         * based on client_lib_config_registry.conf.php
+         * @param {String} featurePath full path to feature ex('test/itemSession/feature')
+         * @returns {Boolean} true if feature is visible
+         */
+        isVisible: (featurePath = '') => {
```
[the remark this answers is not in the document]
jsconan: in this particular context yes, it behaves the same. However, an arrow function does not have the same lexical scope than a pure method.

views/js/test/services/features/test.js
```diff
@@ -0,0 +1,48 @@
...
+        {lookup: 'items/category/featureVisible', visible: true},
+        {lookup: 'items/category/featureHidden', visible: false},
+        {lookup: 'items/*/featureVisible', visible: true},
+        {lookup: 'items/*/featureHidden', visible: false},
+        {lookup: 'items/category/*/featureHidden', visible: false},
+        {lookup: 'items/category/subcategory/*', visible: true},
+        {lookup: 'items/*/featureHidden', visible: false},
+        {lookup: '*/featureVisible', visible: true},
+        {lookup: '*/subcategory/featureVisible', visible: false}
+    ];
```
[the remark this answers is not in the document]
meefox: Added module test. 
About JSON not sure, this data is tightened to this test case I would keep it at least until the test grew or data be needed anywhere else.

views/js/test/services/features/test.js
```diff
@@ -0,0 +1,52 @@
...
+        {lookup: 'items/*', visible: false},
+        {lookup: 'items/featureVisible', visible: true},
+        {lookup: 'items/featureHidden', visible: false},
+        {lookup: 'items/anyFeature', visible: false},
+        {lookup: 'items/category/*', visible: true},
+        {lookup: 'items/category/featureVisible', visible: true},
+        {lookup: 'items/category/featureHidden', visible: false},
+        {lookup: 'items/category/anyFeature', visible: true},
+        {lookup: 'items/category/subcategory/anyFeature', visible: true},
+        {lookup: 'items/category/subcategory/anyFeature', visible: true},
```
jsconan, line 40: **chore:** duplicated test case.

**thought:** I see tests to validate the lookup from a filled-up config. But I don't see test for incorrect entry nor for empty config

views/js/services/features.js
```diff
@@ -0,0 +1,67 @@
...
+     */
+    const buildRegexp = lookupPath => {
+        lookupPath = lookupPath.replace('*', '\\S+');
+
+        try {
+            return RegExp(`^${lookupPath}$`);
+        } catch (e) {
+            logger.warn(`Lookup feature path ${lookupPath} was not found`);
+            return RegExp('^\0$');
+        }
```
jsconan, line 39: **nitpick:** Something I didn't notice yet, the RegExp needs to be instantiated

**suggestion:**
```suggestion
        try {
            return new RegExp(`^${lookupPath}$`);
        } catch (e) {
            logger.warn(`Lookup feature path ${lookupPath} was not found`);
            return new RegExp('^\0$');
        }
```

jsconan (changes requested):
Very close to being approved :)

I have a few comments.

And a wondering: with this implementation we are considering a whitelist of features. This means if a feature is not registered yet, it will be considered hidden.

Looking at the ACs, it looks like it should rather be a blacklist:
>  `Show` is used as default value when a feature is not listed in the response

views/js/services/features.js
```diff
@@ -0,0 +1,67 @@
...
+        } catch (e) {
+            logger.warn(`Lookup feature path ${lookupPath} was not found`);
+            return new RegExp('^\0$');
+        }
+    };
+
+    return {
+        /**
+         * Check is feature configured to be visible
+         * based on client_lib_config_registry.conf.php
```
krampstudio, line 45:
```suggestion
         * Check if the feature is configured to be visible
         * based on the module configuration
```

krampstudio (approved):
I just have an optional suggestion.

Everything else looks good. 

- [ ] <s>New code is not subject to concurrency issues (if applicable)</s>

jsconan (approved):
LGTM. Nice work!

**Review Checklist**
- ~~[ ] New code is not subject to concurrency issues (if applicable)~~
- [x] Commits are atomic

PR #3463, opened 2022-03-30
fix: make properties label searchable by language

siwane:
**Description**
While editing a resource, property labels are not translated based on User Interface language.
This PR aims to fix it

**Ticket**
https://oat-sa.atlassian.net/browse/REL-665

PR #3484, opened 2022-04-13
Checking empty value of label to avoid fatal errors

alroniks:
The related task is [REL-665](https://oat-sa.atlassian.net/browse/REL-665)

It raises the issue on the Premium.

It fixes issues introduced [here](https://github.com/oat-sa/tao-core/pull/3463/files) and [here](https://github.com/oat-sa/tao-core/pull/3467/files)

**How to test?**

Create a user with a different than the default language interface language, log in with this user, and try to reach any type of object (or even rdf forms). For example, create a new user. You will get 500 error, which is unexpected normally.

### DIFF
--- a/helpers/form/ElementMapFactory.php
+++ b/helpers/form/ElementMapFactory.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 namespace oat\tao\helpers\form;
 
 use common_Logger;
+use oat\generis\model\OntologyAwareTrait;
 use tao_helpers_Uri;
 use tao_helpers_Context;
 use core_kernel_classes_Class;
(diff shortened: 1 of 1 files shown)


helpers/form/ElementMapFactory.php
```diff
@@ -286,15 +289,14 @@ private function getDescriptionFromTranslatedPropertyLabel(
...
-
-        if (empty(trim($propertyLabel))) {
+        $propertyLabel = current(
+            $property->getPropertyValues(
+                $this->getProperty(OntologyRdfs::RDFS_LABEL),
+                ['lg' => $language, 'one' => true]
+            )
+        );
+
+        if (false !== $propertyLabel && empty(trim($propertyLabel))) {
```
gabrielfs7, line 299: Maybe, to be more precise.

```suggestion
        if (is_string($propertyLabel) && empty(trim($propertyLabel))) {
```

gabrielfs7 (approved):

helpers/form/ElementMapFactory.php
alroniks: The issue is that `current` returns false if the array is empty.

bartlomiejmarszal (approved):

PR #3487, opened 2022-04-14
[Hotfix] Make properties label searchable by language

alroniks:
Backport of the following PRs:

https://github.com/oat-sa/tao-core/pull/3463
https://github.com/oat-sa/tao-core/pull/3467
https://github.com/oat-sa/tao-core/pull/3484

**Description**
While editing a resource, property labels are not translated based on User Interface language.
This PR aims to fix it

**How to test**
- Change user language
- Go to item tab and select Item
- Edit form should be translated (Label) if token exist in PO files

### DIFF
--- a/actions/form/class.Instance.php
+++ b/actions/form/class.Instance.php
@@ -83,6 +83,13 @@ class tao_actions_form_Instance extends tao_actions_form_Generis
         $instance = $this->getInstance();
         $guiOrderProperty = new core_kernel_classes_Property(TaoOntology::PROPERTY_GUI_ORDER);
 
+        // Guess language
+        try {
+            $language = $this->options['lang'] ?? common_session_SessionManager::getSession()->getInterfaceLanguage();
+        } catch (common_exception_Error $exception) {
+            $language = DEFAULT_LANG;
+        }
+
         //get the list of properties to set in the form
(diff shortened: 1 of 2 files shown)


actions/form/class.Instance.php
```diff
@@ -83,6 +83,13 @@ protected function initElements()
         $instance = $this->getInstance();
         $guiOrderProperty = new core_kernel_classes_Property(TaoOntology::PROPERTY_GUI_ORDER);
 
+        // Guess language
+        try {
+            $language = $this->options['lang'] ?? common_session_SessionManager::getSession()->getInterfaceLanguage();
```
peetya, line 88: Shouldn't we validate what is the value of `$this->options['lang']`? Like:
- is it a supported type or not
- is it a supported value or not
- is it null or empty string, etc...

Also, a test case would be great to cover this new guessing. WDYT?

peetya (changes requested):

actions/form/class.Instance.php
alroniks: It would be nice to do it in the normal development branch. I agree with your suggestion. But this specific one is only backport of previous fixes, which were already applied on current development and March release, but missed in April.

bartlomiejmarszal (approved):

PR #3491, opened 2022-04-20
[Fix] Clear token store on logout

shpran:
# [AUT-2124](https://oat-sa.atlassian.net/browse/AUT-2124)

## Changes
* Token Store is cleared when exit button is pressed
* Token Store is cleared when receiving a 403 error + logging out

## How to test
1. Configure `tao/security-xsrf-token.conf.php` to use `localStorage`:
   ```php
   <?php
   /**
    * Default config header created during install
    */

   return new oat\tao\model\security\xsrf\TokenService(array(
       'store' => new oat\tao\model\security\xsrf\TokenStoreSession(),
       'poolSize' => 10,
       'timeLimit' => 0,
       'validateTokens' => true,
       'clientStore' => 'localStorage'
   ));
   ```
2. Clear site data and refresh the page - you will be logged out
3. Log in
4. Create new class/item/test
5. Log out
6. Log it again
7. Try to create new class/item/test

### Previous behavior
`Error: 403 : Unable to process your request` pops-up, user log out.
### New behavior
No errors, class/item/test created and displayed under the tree.

### DIFF
--- a/views/js/controller/main.js
+++ b/views/js/controller/main.js
@@ -41,7 +41,28 @@ define([
     'layout/search',
     'layout/tree/loader',
     'layout/section-height',
-], function(module, $, _, context, router, helpers, uiForm, urlUtil, loggerFactory, feedback, generisRouter, sections, actionManager,versionWarning, loadingBar, nav, search, treeLoader, sectionHeight){
+    'core/tokenHandler'
+], function (
+    module,
+    $,
+    _,
+    context,
+    router,
+    helpers,
+    uiForm,
+    urlUtil,
+    loggerFactory,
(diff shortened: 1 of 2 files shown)


bartlomiejmarszal (commented):
Is this the only way to logout? What if Tao will forbid some access and logout user. Will user be able to create resource then?

jsconan (commented):
Not sure this is the best approach. What happens if the session is timed out?

If we need to make sure to start a session with an empty token store, there are better ways:
- clear the storage upon launching the session
- link the storage to the session id or to an id unique for a session

PR #3503, opened 2022-04-26
[Feature] Allow copying classes

shpran:
# [ADF-1028](https://oat-sa.atlassian.net/browse/ADF-1028)

## Companion PRs
* https://github.com/oat-sa/extension-tao-item/pull/577

## Tasks
- [x] Add "Copy To" button
- [x] Implement copying class
  - [x] Copy selected class
  - [x] Copy selected class properties
  - [x] Copy subclasses
  - [x] Copy properties of subclasses
  - [x] Copy instances
- [x] Process in the queue in the background
- [ ] Add unit tests


actions/class.RdfController.php
```diff
@@ -737,6 +741,80 @@ public function copyInstance()
         ], 412);
     }
 
+    /**
+     * @requiresRight uri READ
```
gabrielfs7, line 745: Maybe it is more clarifying `classUri`?

```suggestion
     * @requiresRight classUri READ
```

models/classes/resources/Service/InstanceMetadataCopier.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\resources\Service;
+
+use core_kernel_classes_Resource;
+use oat\tao\model\resources\Contract\InstanceMetadataCopierInterface;
+
+class InstanceMetadataCopier implements InstanceMetadataCopierInterface
```
gabrielfs7, line 30: What is the difference between properties and metadata? 

I think they are the same, but in the instance scope, we copy `values` and on class scope we copy `definition`.

Could we agree in a single nomenclature for them? What do you think?

https://github.com/oat-sa/tao-core/pull/3503/files#diff-20063cefde0308509e3fb3dac00d7219b396c66a2782a5f18ce2f0675e319957R30

models/classes/resources/Contract/ClassPropertyCopierInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\resources\Contract;
+
+use core_kernel_classes_Class;
+use core_kernel_classes_Property;
+
+interface ClassPropertyCopierInterface
```
gabrielfs7, line 30: Referring to https://github.com/oat-sa/tao-core/pull/3503/files#r863514730

models/classes/resources/Contract/InstanceContentCopierInterface.php
```diff
@@ -0,0 +1,35 @@
...
+namespace oat\tao\model\resources\Contract;
+
+use core_kernel_classes_Resource;
+
+interface InstanceContentCopierInterface
+{
+    public function copy(
+        core_kernel_classes_Resource $instance,
+        core_kernel_classes_Resource $destinationInstance
+    ): void;
```
hectoras, line 34: Returning `void` here seems inconsistent with other interfaces (for example, `ClassPropertyCopierInterface`). Could we return a reference to the instance here too? (and maybe add a PHPDoc comment clarifying that the returned instance is the source).

nitpick: Also, using `$from` and `$to` as parameter names is a bit more clear (they would be the same names used by the the standard [copy()](https://www.php.net/copy) function)

```suggestion
    public function copy(
        core_kernel_classes_Resource $from,
        core_kernel_classes_Resource $to
    ): core_kernel_classes_Resource;
```

gabrielfs7: IMO we should change the other interfaces to return `void` instead. Since `copy` is a **command** and it should not return anything. 

@hectoras the `core_kernel_classes_Resource $destinationInstance` is already the resource you want to return, so returning it is also redundant.

hectoras: I agree, my concern is more about consistency than having the destination returned (however, the return value would be a sort-of syntactic sugar for things like this: https://github.com/oat-sa/extension-tao-item/pull/577/files#r863586257).

In that regard, we'd also need to change [ClassCopier.php](https://github.com/oat-sa/tao-core/pull/3503/files#diff-116193cda220e3b93637a5ded7e849b7922572bc8836dfd01b5e57f978e54271) to not return the created class (btw I think the returned class is not used right now).

hectoras: Then, we may just change the parameter names to match the names from the standard [copy()](https://www.php.net/copy) function (suggestion).

```suggestion
    public function copy(
        core_kernel_classes_Resource $from,
        core_kernel_classes_Resource $to
    ): void;
```

actions/class.RdfController.php
```diff
@@ -737,6 +741,80 @@ public function copyInstance()
...
+     * @requiresRight destinationClassUri WRITE
+     */
+    public function copyClass(): void
```
gabrielfs7, line 748: Dont you think this method is doing more than a controller should do? I think besides the `$this->validateCsrf();` the rest should belong to an `application service` or `request handler`. Wdyt?

models/classes/resources/Contract/ClassPropertyCopierInterface.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    public function copy(
+        core_kernel_classes_Property $property,
+        core_kernel_classes_Class $destinationClass
+    ): core_kernel_classes_Property;
```
gabrielfs7, line 35: If possible:

```suggestion
    public function copy(
        core_kernel_classes_Property $property,
        core_kernel_classes_Class $destinationClass
    ): void;
```

models/classes/resources/Contract/InstanceContentCopierInterface.php
shpran: I think it's not a good idea to use `from`-`to` for all of them.
This will be ok for cases like:
* `copy class metadata to another class metadata`
* `copy instance metadata to another instance metadata`
* etc.

But in cases like `copy instance to class` or `copy class to class` sounds for me like `something`-`destination` or `something`-`to`.
Because in this case, we are copying `element` to the specific `destination`, not `from element`.

actions/class.RdfController.php
shpran: I just copied the logic from `copyInstance` method.
If I have time to refactor this part - I will do this 🙂

models/classes/resources/Contract/InstanceContentCopierInterface.php
shpran: Also, I think, that at least for `ClassCopier` and `InstanceCopier` it's better to return new instance.

Imagine, that I copied a class/instance to another class, which contains a billion of classes/instances.
If we return to the user just a destination class - he will be confused, because he needs to search for this class/instance in the list.

As for me, after copying, it's better to provide a link to the copied class/instance, so the user can open it immediately (or we can open this class/instance automatically).
For this, we need to return a copied class/instance.

hectoras (approved):

hectoras:
Released as [v50.11.0](https://github.com/oat-sa/tao-core/releases/tag/v50.11.0).

PR #3507, opened 2022-04-29
feat: TR-4948 add new feature feature flag for own graduated score

kilatib:
# [TR-3948](https://oat-sa.atlassian.net/browse/TR-3948)

# Added feature flag for correct migration apply on server


wazelin:
There's no need in adding all the feature flags here.
The flags relevant for a certain extension should be part of that extension.

Consider closing the PR.

kilatib:
FLAG migrated to extension repo

PR #3514, opened 2022-05-05
Fix/TR-3995/Max words restriction can be exceeded

jsconan:
2/4

Related to: https://oat-sa.atlassian.net/browse/TR-3995

Requires: 
 - [x] https://github.com/oat-sa/tao-core-sdk-fe/pull/145
 - [x] update the `package.json` file once the companion PR has been merged

The helper `util/strLimiter` is aimed at limiting a text to a maximum number of units, either chars or words. Unfortunately, so far it was exclusively expecting plain text and it was producing wrong results with HTML.

This pull request proposes to support either plain text or HTML text.

How to test:
- create a test having extended text interactions with a constraint for the max words and for the max length (a test sample is attached to the ticket).
- publish the test, then take it
- type some text in the extended text interaction, then copy it and paste it until the limit is reached.
- try to exhaust the limit by pasting again
- remove a few words, ideally less than the number in the clipboard, then paste again

Without the fix, the limit can be overflown.
With the fix, we cannot overflow the limit.


meefox (approved):

shaveko (approved):
Looks good

AntonNeviarouski:
QA verified

PR #3526, opened 2022-05-12
[Feature] Copy classes and instances metadata

shpran:
# [ADF-1030](https://oat-sa.atlassian.net/browse/ADF-1030)

## Require PRs
* https://github.com/oat-sa/generis/pull/1007

## Companion PRs
* https://github.com/oat-sa/extension-tao-item/pull/579
* https://github.com/oat-sa/extension-tao-backoffice/pull/144

## Changes
* Added the ability to open a resource directly from the task list
* Copier services DI registration was moved from `ResourceServiceProvider` to new `CopierServiceProvider`.
* Added the ability to copy class metadata (properties)
* Added the ability to copy instance metadata (property values)

## Tasks
- [x] Add unit tests
- [ ] ~~Use `redis` to map cloned properties to avoid memory issues~~ **Implemented another way without `redis`**
- [x] Update `oat-sa/generis` dependency


models/classes/resources/Service/InstanceMetadataCopier.php
```diff
@@ -24,15 +24,90 @@
...
+                continue;
+            }
+
+            $originalProperty = $instance->getProperty($originalPropertyUri);
+            $range = $originalProperty->getRange();
+            $propertyValues = $instance->getPropertyValuesCollection($originalProperty);
+
+            foreach ($propertyValues->getIterator() as $propertyValue) {
+                if ($range === null || $range->getUri() !== GenerisRdf::CLASS_GENERIS_FILE) {
+                    $destinationInstance->setPropertyValue($destinationProperty, $propertyValue);
```
gabrielfs7, line 93: Wont it set an old reference for lists (or non text properties) pointing to the original property instead of the cloned one? Did we check the database?

models/classes/resources/Service/ClassMetadataMapper.php
```diff
@@ -0,0 +1,45 @@
...
+ * @author Andrei Shapiro <andrei.shapiro@taotesting.com>
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\resources\Service;
+
+use core_kernel_classes_Property;
+
+// @TODO Use redis to to avoid memory issues
```
gabrielfs7, line 29: Do we plan to tackle this in this task?

models/classes/resources/Service/ClassMetadataMapper.php
```diff
@@ -0,0 +1,45 @@
...
+class ClassMetadataMapper
+{
+    private $map = [];
```
gabrielfs7, line 32: annotations?

models/classes/resources/Service/ClassMetadataMapper.php
shpran: I think yes, I will move this to redis in this task 🙂

shpran: I found a way how to avoid memory issues without redis 🙂

hectoras (approved):

models/classes/resources/Service/InstanceMetadataCopier.php
```diff
@@ -24,15 +24,137 @@
...
+    {
+        if (array_key_exists($property->getUri(), $this->sharedProperties)) {
+            return $property;
+        }
+
+        $originalPropertyUri = $this->classMetadataMapper->get($property);
+
+        return $originalPropertyUri === null
+            ? null
+            : $property->getProperty($originalPropertyUri);
```
hectoras, line 125: suggestion: The ternary operator here may make the code slightly harder to read, I would try to avoid it

```suggestion
        if ($originalPropertyUri === null) {
            return null;
        }
        
        return $property->getProperty($originalPropertyUri);
```

models/classes/resources/Service/ClassCopier.php
```diff
@@ -42,17 +42,22 @@ class ClassCopier implements ClassCopierInterface
...
+
     /** @var string[] */
     private $copiedClasses = [];
 
     public function __construct(
         RootClassesListServiceInterface $rootClassesListService,
         ClassMetadataCopierInterface $classMetadataCopier,
-        InstanceCopierInterface $instanceCopier
+        InstanceCopierInterface $instanceCopier,
+        ClassMetadataMapper $classMetadataMapper
```
gabrielfs7, line 55: Use the interface, if possible :) 

```suggestion
        ClassMetadataMapperInterface $classMetadataMapper
```

models/classes/resources/Service/ClassMetadataMapper.php
```diff
@@ -0,0 +1,56 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2022 (original work) Open Assessment Technologies SA.
+ *
+ * @author Andrei Shapiro <andrei.shapiro@taotesting.com>
+ */
+
+declare(strict_types=1);
+
+namespace oat\tao\model\resources\Service;
```
gabrielfs7, line 25: This is really specific for a `Copy` feature, isn't this namespace (or class name) too generic?

models/classes/resources/Service/InstanceMetadataCopier.php
```diff
@@ -24,15 +24,139 @@
...
+                || in_array($originalProperty->getUri(), $this->blacklistedProperties, true)
+            ) {
+                continue;
+            }
+
+            $propertyValues = $instance->getPropertyValuesCollection($originalProperty);
+
+            /** @var core_kernel_classes_Literal|core_kernel_classes_Resource|core_kernel_classes_Resource[] $propertyValue */
+            foreach ($propertyValues->getIterator() as $propertyValue) {
+                if ($this->isFileProperty($originalProperty)) {
```
gabrielfs7, line 104: Probably it will never reach this line, since we do not have File property working. How did you manage to test that?

gabrielfs7, line 104: Are you sure this verification is correct and necessary inside the loop?

```suggestion
                if ($this->isFileProperty($originalProperty)) {
```

models/classes/resources/Service/ClassMetadataMapper.php
shpran: `oat\tao\model\resources\Copier`?

models/classes/resources/Service/InstanceMetadataCopier.php
shpran: Idk
I just copied the logic we use to copy an instance.

shpran: We get the new property on each iteration here: https://github.com/oat-sa/tao-core/pull/3526/files/e2dc74f8f7bbaca07e04fb264e75c268ff48f587#diff-20817a35d80ea002b37258f3d9a4df176a55f96f0d46274b94fa78a6f81460caR91

And I cache them here: https://github.com/oat-sa/tao-core/pull/3526/files/e2dc74f8f7bbaca07e04fb264e75c268ff48f587#diff-20817a35d80ea002b37258f3d9a4df176a55f96f0d46274b94fa78a6f81460caR134

gabrielfs7: Seems we created a cache just to use this method inside the loop, no? But it can actually be called only once outside the loop and the result stored in a var that can be used inside the loop "if"?

gabrielfs7: I see 😢. I would at least remove the unnecessary cache and move this method outside the loop and in the loop have only:

```suggestion
                if ($isFileProperty) {
```

As I mentioned here https://github.com/oat-sa/tao-core/pull/3526/files/e2dc74f8f7bbaca07e04fb264e75c268ff48f587#r874410844

Wdyt?

models/classes/resources/Service/InstanceMetadataCopier.php
```diff
@@ -24,15 +24,139 @@
...
+                $originalProperty === null
+                || in_array($originalProperty->getUri(), $this->blacklistedProperties, true)
+            ) {
+                continue;
+            }
+
+            $propertyValues = $instance->getPropertyValuesCollection($originalProperty);
+
+            /** @var core_kernel_classes_Literal|core_kernel_classes_Resource|core_kernel_classes_Resource[] $propertyValue */
+            foreach ($propertyValues->getIterator() as $propertyValue) {
```
gabrielfs7, line 103:
```suggestion
            /** @var core_kernel_classes_Literal|core_kernel_classes_Resource|core_kernel_classes_Resource[] $propertyValue */
            foreach ($instance->getPropertyValuesCollection($originalProperty)->getIterator() as $propertyValue) {
```

models/classes/resources/Service/ClassMetadataMapper.php
gabrielfs7: Either this or rename the class `ClassMetadataCopyMapper` :)

test/unit/models/classes/resources/Service/ClassMetadataCopierTest.php
```diff
@@ -0,0 +1,133 @@
...
+            ->with(true)
+            ->willReturn(
+                [
+                    'classProperty1' => $classProperty1,
+                    'classProperty2' => $classProperty2,
+                ]
+            );
+
+
+
```
gabrielfs7, line 73:
```suggestion

```

gabrielfs7 (approved):

AntonNeviarouski:
QA verified

models/classes/resources/Service/ClassMetadataCopier.php
```diff
@@ -28,32 +28,49 @@
...
                 continue;
             }
 
-            $newPropertyResource = $property->duplicate();
-            $newProperty = $newPropertyResource->getProperty($newPropertyResource->getUri());
-            $newProperty->removePropertyValues($newProperty->getProperty(OntologyRdfs::RDFS_DOMAIN));
-            $newProperty->setDomain($destinationClass);
+            $newProperty = $this->copyProperty($property, $destinationClass);
+
+            $this->copiedProperties[] = $propertyUri;
```
hectoras, line 59: question: I missed this one on the first review and I'm probably a bit late since this has been already merged :sweat_smile: but, why is this a **class** variable?

I assume this is used to prevent copying the same property twice **for a given instance**, right (a given class)? If we call this two times for copying two different instances of the same class, I think it would prevent copying the properties for the later.

If that's the case, I think this should be a local variable instead (it would cause problems when copying recursively, and would also allow us to stop setting the _share_ flag for the service to `false` [here](https://github.com/oat-sa/tao-core/pull/3526/files#diff-a8e74ed08b0701b690705513f6726cd60153d0f0399f28f58cb8813bf203b64dR50)) so, overall, it would simplify things a little bit and prevent creating new copier instances for every class copied recursively.

I've not checked if that happens also with other copiers.

hectoras: Or is it to prevent copying properties from nested classes that have been already copied by previous calls?

PR #3541, opened 2022-05-30
Fix/oatsd 1881/collapsing sub lines

yukipastelcat:
# Pull request summary

Related to [OATSD-1881](https://oat-sa.atlassian.net/browse/OATSD-1881).

Updated `@oat-sa/tao-core-ui` dependency to 1.54.4 including the fix.


btamas (changes requested):
What are these new PHP files?

PR #3561, opened 2022-06-30
fix: check application cache directory during update

augustas:
When TAO instance is recreated and `taoUpdate.php` gets called, any migration trying to access DI container will fail with an error of missing cache directory. This change ensures the directory is present  

Related to : https://oat-sa.atlassian.net/browse/REL-783
  
#### Dependencies
 
Requires :
 - [ ] https://github.com/oat-sa/generis/pull/1013

### DIFF
--- a/install/class.Installator.php
+++ b/install/class.Installator.php
@@ -245,12 +245,8 @@ class tao_install_Installator
             }
 
             $setupFileCache = $this->getServiceManager()->get(SetupFileCache::class);
-            $cachePath = $file_path . 'generis' . DIRECTORY_SEPARATOR . 'cache';
-            if ($setupFileCache->createDirectory($cachePath)) {
-                $this->log('d', $cachePath . ' directory was created!');
-            } else {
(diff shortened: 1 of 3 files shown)


install/class.Installator.php
```diff
@@ -263,11 +266,7 @@ public function install(array $installData, callable $callback = null)
              * 5b - Create cache persistence
             */
             $this->log('d', 'Creating cache persistence..');
-            $persistenceManager->registerPersistence('cache', [
-                'driver' => 'phpfile'
-            ]);
-            $persistenceManager->getPersistenceById('cache')->purge();
-            $this->getServiceManager()->register(PersistenceManager::SERVICE_ID, $persistenceManager);
+            $setupFileCache->createPersistence();
```
gabrielfs7, line 269: Not sure that setFileCache should handle persistence config.

augustas: I think it is a part of tao cache setup https://github.com/oat-sa/generis/pull/1013#discussion_r911085028

gabrielfs7 (approved):

PR #3573, opened 2022-07-05
chore: remove PHP version restriction

gabrielfs7:
https://oat-sa.atlassian.net/browse/ADF-1097
https://oat-sa.atlassian.net/browse/ADF-1087
https://oat-sa.atlassian.net/browse/ADF-1105

## Goal 

Make the extension not dictating PHP version to easier our PHP maintenance, since the extension does not work in isolation and PHP version should be equally respected to all TAO extensions.


composer.json
```diff
@@ -56,31 +56,30 @@
...
     "require": {
-        "php": "^7.1",
         "ext-json": "*",
         "ext-mbstring": "*",
         "doctrine/dbal": "^2.10",
         "ezyang/htmlpurifier": "^4.9",
         "guzzlehttp/guzzle": "^6.5.7",
         "guzzlehttp/psr7": "^1.8.4",
-        "imsglobal/lti": "^3.0",
+        "oat-sa/imsglobal-lti": ">=4.0.0",
```
wazelin, line 65: This one doesn't seem to exist, or be available.

wazelin: Yes, but it's not on packagist.

wazelin: Added it https://packagist.org/packages/oat-sa/imsglobal-lti

wazelin: Please check with the repo owner if they can set up a webhook to update the packagist package automatically.

https://packagist.org/about#how-to-update-packages

wazelin: BTW, that library doesn't really allow running PHP 8.1.

https://github.com/oat-sa/imsglobal-lti/blob/master/composer.json#L15

>\>=7.4,<=8.1

I'm not sure it will allow any version of PHP 8.1 besides 8.1.0.

gabrielfs7: I was able to install all libs using PHP 8.1.7. I think `<=8.1` means `<8.2`

wazelin: But it's not
https://github.com/oat-sa/tao-core/runs/7394167529?check_suite_focus=true#step:2:575

Are you sure you didn't have the `php` version defined in the platform configuration of your root composer.json?

wazelin: I would've fixed that, but apparently I have no permissions in that newly added repo :/

wazelin: @shpran, would you be so kind to change that library's `php` version requirements to something like `^7.4 || ~8`?

PR #3347, opened 2022-01-21
fix: only execute initRendering when it is not in the authoring context

clagosarias:
https://oat-sa.atlassian.net/browse/ADF-916

### How to

- Go to Tests tab/Item tab
- Create a new test/item with dependant properties
- Click 'Authoring' button
- Actual result: Authoring page is opened, 2 errors appear in browser console.
- Expected result: Authoring page is opened, no errors appear in browser console.

### Description
This has been solved by only rendering the dependant properties when needed (not in authoring context)

PR #3587, opened 2022-08-09
Fix TR-4529 impossible to edit a Delivery form

wazelin:
# [TR-4529](https://oat-sa.atlassian.net/browse/TR-4529)

This PR aims to fix the condition that currently prevents a resource form initialization when a currently handled ajax request URI has the word `authoring` within it. Apparently, that was done to suppress some initialization errors on the Item and Test authoring screens.

Below, you can see a proof that the form initialization would resolve the issue.

https://user-images.githubusercontent.com/2943256/183646666-7d82b5f8-89d9-4ca6-bbd9-db5c11e0942c.mov

## Caused by
#3347 

## How to test

1. Have TAO 3.x running under a domain name with `authoring` in its name
2. Open a Delivery form
3. Save it

### DIFF
--- a/views/js/uiForm.js
+++ b/views/js/uiForm.js
@@ -101,20 +101,15 @@
             this.htmlEditors = {};
 
             $(document).ajaxComplete(function (event, request, settings) {
-                var testedUrl;
-
                 //initialize regarding the requested action
                 //async request waiting for html or not defined
                 if (settings.dataType !== 'html' && settings.dataType) {
                     return;
                 }
 
-                if (settings.url.indexOf('?') !== -1) {
-                    testedUrl = settings.url.substr(0, settings.url.indexOf('?'));
-                }
-                else {
-                    testedUrl = settings.url;
-                }
+                const testedUrl = settings.url.indexOf('?') === -1
+                    ? settings.url
+                    : settings.url.substr(0, settings.url.indexOf('?'));
 
                 /**
                  * Prevent manage-schema form initialization when the targeted url is related to authoring
--- a/views/js/uiForm.js
+++ b/views/js/uiForm.js
@@ -110,12 +110,13 @@
                 const testedUrl = settings.url.indexOf('?') === -1
                     ? settings.url
                     : settings.url.substr(0, settings.url.indexOf('?'));
+                const authoringRequestSuffix = 'authoring';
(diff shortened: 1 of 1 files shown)


uncleempty (approved):
LGTM

bziondik (approved):
LGTM

views/js/uiForm.js
```diff
@@ -101,33 +101,31 @@
...
+                }
 
-                    /**
-                     * Prevent manage-schema form initialization when the targeted url is related to authoring
-                     * associated action is "launchEditor"
-                    */
-                    if (!testedUrl.includes('authoring')) {
-                        self.initRendering();
-                    }
+                const testedUrl = settings.url.indexOf('?') === -1
```
kilatib, line 110:
```suggestion
                const testedUrl =  new URL(settings.url);
```

views/js/uiForm.js
```diff
@@ -101,33 +101,31 @@
...
+                    ? settings.url
+                    : settings.url.substr(0, settings.url.indexOf('?'));
+                const authoringRequestSuffix = 'authoring';
+
+                /**
+                 * Prevent manage-schema form initialization when the targeted url is related to authoring
+                 * associated action is "launchEditor"
+                */
+                if (testedUrl.indexOf(authoringRequestSuffix, testedUrl.length - authoringRequestSuffix.length) === -1) {
```
kilatib, line 119:
```suggestion
                if (!testedUrl.pathname.endsWith(authoringRequestSuffix)) {
```

kilatib (changes requested):

views/js/uiForm.js
kilatib: I'm not super familiar with JS version which we support so last suggestion with `endsWith` could be skiped

views/js/uiForm.js
wazelin: Right, but I will probably pass on that change, as it's irrelevant for the hotfix.

views/js/uiForm.js
wazelin: Although we don't support IE, I'd prefer not to break this piece of code's compatibility just for that.

kilatib (approved):

PR #3619, opened 2022-09-15
Not Allow to have Zero for MaxExecutions field

kilatib:
# [AUT-2566](https://oat-sa.atlassian.net/browse/AUT-2566)

## Description
When the user is editing any delivery, minimum value in Max. Executions field of Delivery properties can be set to 0, but the warning(when providing negative value) informs about the minimum value of the field(which is “1“).

## Aceptence criteria
1. Go to "Deliveries" page
2. Select a Delivery
3. Modify the “Max. Executions“ field in "Properties" form with a negative value
4. Click "Save" button
5. Modify the “Max. Executions“ field with “0“
6. Click "Save" button
7. Should see validation message.

## Development impact 
Looks like we skip numeric validation if number is zero. For keep amount of changes less and not broke any logic 
I add not skip zero amount and range validator is set

## Kitchen 
Kitchen : http://test-aut2566.playground.kitchen.it.taocloud.org:40371/
admin / admin

## Demo 
https://user-images.githubusercontent.com/2750628/190445891-4ce7265b-ea63-459f-9932-77d914bf4dd8.mov


helpers/form/validators/class.Numeric.php
```diff
@@ -44,7 +44,7 @@ public function evaluate($values)
         
         $rowValue = $values;
         $value = tao_helpers_Numeric::parseFloat($rowValue);
-        if (empty($rowValue)) {
+        if (empty($rowValue) && (!$this->hasOption('min') || $this->hasOption('max'))) {
```
poyuki, line 47: by ! you inverting only left `has` check or trying to invert both left and right check?

wazelin: If the latter was your intention, you may as well replace line 76 with `return true;`

kilatib: The reason was not leaving the validation if the value is 0 and any limit set, because we did not exclude that the max could be less then 0 or 0

wazelin (approved):

wazelin:
This looks reasonable, let's deploy and have it QAed before merging, though.

PR #3620, opened 2022-09-15
fix/ADF-1146: Distinguish which method requiresRight cached data belongs to

hectoras:
**Associated Jira issue:** [ADF-1146](https://oat-sa.atlassian.net/browse/ADF-1146)

This fix enables sharing the `@requiresRight` cache between `ActionDescription` instances while keeping track about which method the cached data belongs to, so it avoids using permissions stored previously for a different action in later calls.

There are also two small changes introduced while debugging to avoid errors when running TAO in PHP 8:

- A deprecated warning when calling `strlen()` with a `null` value.
- A warning when no HTTP Referer header is present that may end up polluting the HTML output if E_ALL is enabled

Related PR: https://github.com/oat-sa/extension-tao-item/pull/595 (it has been closed, as this fix in tao-core alone solves the issue)

### DIFF
--- a/actions/form/class.Instance.php
+++ b/actions/form/class.Instance.php
@@ -153,7 +153,7 @@ class tao_actions_form_Instance extends tao_actions_form_Generis
                     }
                 }
 
-                if ($this->isEmptyLabel($element)) { // Don't show empty labels
+                if ($this->isEmptyLabel($element)) {
                     continue;
                 }
 
--- a/actions/form/class.Instance.php
+++ b/actions/form/class.Instance.php
@@ -254,7 +254,6 @@ class tao_actions_form_Instance extends tao_actions_form_Generis
     private function isEmptyLabel($element): bool
     {
         return $element instanceof tao_helpers_form_elements_Label
-            && (($element->getRawValue() === null)
-                || (strlen($element->getRawValue()) === 0));
+            && empty($element->getRawValue());
     }
 }
--- a/actions/form/class.Instance.php
+++ b/actions/form/class.Instance.php
@@ -153,8 +153,7 @@ class tao_actions_form_Instance extends tao_actions_form_Generis
                     }
                 }
 
-                // don't show empty labels
-                if ($element instanceof tao_helpers_form_elements_Label && strlen($element->getRawValue()) === 0) {
+                if ($this->isEmptyLabel($element)) { // Don't show empty labels
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


models/classes/controllerMap/ActionDescription.php
```diff
@@ -34,8 +32,15 @@
...
-    private static $docBlock;
+    /**
+     * Data cache for DocBlock information used by getDocBlock().
+     *
+     * The method name the docblock is found in is used as the array index
+     * (i.e. it works as a "cache key")
+     *
+     * @var DocBlock[]
+     */
+    private static $docBlock = [];
```
gabrielfs7, line 43: Maybe, since it has a collection now it should be plural? Do we really need to explain this variable in the annotation? Isn't self-explainable in the code?

```suggestion
    /** @var DocBlock[] */
    private static $docBlocks = [];
```

actions/form/class.Instance.php
```diff
@@ -153,8 +153,7 @@ protected function initElements()
                     }
                 }
 
-                // don't show empty labels
-                if ($element instanceof tao_helpers_form_elements_Label && strlen($element->getRawValue()) === 0) {
+                if ($this->isEmptyLabel($element)) { // Don't show empty labels
```
gabrielfs7, line 156: The method name already explains the meaning, so no need for comments :) 

```suggestion
                if ($this->isEmptyLabel($element)) {
```

actions/form/class.Instance.php
```diff
@@ -251,4 +250,11 @@ private function isList(core_kernel_classes_Property $property): bool
             new core_kernel_classes_Class(TaoOntology::CLASS_URI_LIST)
         );
     }
+
+    private function isEmptyLabel($element): bool
+    {
+        return $element instanceof tao_helpers_form_elements_Label
+            && (($element->getRawValue() === null)
+                || (strlen($element->getRawValue()) === 0));
```
gabrielfs7, line 258: You can achieve same result with:

```suggestion
        return $element instanceof tao_helpers_form_elements_Label
            && empty($element->getRawValue());
```

hectoras: Changed, thanks! :)

actions/form/class.Instance.php
hectoras: Changed, thanks! :)

gabrielfs7 (approved):

shpran (approved):

hectoras:
Got two approvals, therefore I'm merging this PR in order to start the QA varidation.

PR #3624, opened 2022-09-20
fix: should not be null because in validation method it access like a…

kilatib:
# [TR-4622](https://oat-sa.atlassian.net/browse/TR-4622)

## Description 
The case defined here: 
https://github.com/oat-sa/extension-tao-deliver-connect/actions/runs/3090566408/jobs/4999531232

## development impact 
We call `validate()`  in constructor before properies defined it cost error of access to null as array


models/classes/auth/AbstractCredentials.php
```diff
@@ -55,11 +55,9 @@ public function __construct($properties = [])
      */
     protected function validate($properties)
```
wazelin, line 56: The argument is no-longer used.
Let's remote it.

Also, let's disable CI on PHP 7.2 and 7.3

wazelin (changes requested):

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,30 @@
+name: CI execute unit tests
```
wazelin, line 1: There is already a CI workflow file in this repo. No need to add another one.
https://github.com/oat-sa/tao-core/blob/develop/.github/workflows/continuous-integration.yaml#L17

wazelin (changes requested):

composer.json
```diff
@@ -83,6 +83,11 @@
         "league/openapi-psr7-validator": "0.9",
         "league/csv": "^9.6"
     },
+    "require-dev": {
+        "phpunit/phpunit": "^8.5",
+        "php-mock/php-mock": "dev-master",
+        "phpspec/prophecy": "~1.0"
+    },
```
wazelin, line 90: There's no need to add these.
```suggestion
```

wazelin (changes requested):

wazelin (approved):

PR #3650, opened 2022-10-07
fix: Fix warnings from login page under PHP 8

hectoras:
**Associated Jira issue:** [ADF-1105](https://oat-sa.atlassian.net/browse/ADF-1105)

This PR fixes some warnings found on a fresh installation running PHP8 when reaching the login page.

The errors trigger Javascript parsing errors, which in turn prevents the login form from being shown.


models/classes/session/Business/Domain/SessionCookieAttributeCollection.php
```diff
@@ -44,6 +45,7 @@ public function add(SessionCookieAttributeInterface $attribute): self
     /**
      * @return iterable|SessionCookieAttributeInterface[]
      */
+    #[ReturnTypeWillChange]
```
gabrielfs7, line 48: We're forward compatible with PHP 8, but only supporting PHP 7.4 feature for now :)

PR #3653, opened 2022-10-25
Hotfix/REL-879/downgrade minimal browser support

jsconan:
### Summary

The drop of IE support removed the transpilation to ES5. While this is fine for the modern browsers we support, it has an impact on the taoBooklet feature which is relying on wkhtmltopdf. This tool does not support ES6.

We need to temporarily downgrade the bundler to a version that generates ES5. 

### Details

Several libraries have been reverted:
- `@oat-sa/browserslist-config-tao` to version `0.3.1` (was `1.0.1`)
- `@oat-sa/grunt-tao-bundle` to version `0.6.1` (was `0.7.0`)
- `@oat-sa/tao-core-sdk` to version `1.19.2` (was `1.20.0`)

These downgrades fortunately only contain the change for the bundler. Event for `tao-core-sdk`.

### How to test
- checkout the branch
- play with TAO
- take test
- generate a booklet


bziondik (approved):
Checked locally only general flow - item/test authoring, delivery. Can't check booklets because can't install it on local env
- [x]  New code is not subject to concurrency issues (if applicable)
- [x]  Feature is working correctly on my local machine (if applicable)

PR #3655, opened 2022-11-04
Feature/ADF-1245/Add ES5 bundles

jsconan:
### Requires
- [x] https://github.com/oat-sa/grunt-tao-bundle/pull/36
- [x] update the `package*.json` files once the companion PR has been merged and released.

### Summary

Update the bundler so that we can now:
- generate standalone bundles that can be run on a final application without additional dependency
- change the transpilation targets so that we can generate ES5 bundles when needed

Add bundle tasks for generating an ES5 flavour of some bundles (vendor)

Add helpers for installing the AMD loader for either the standalone or the ES5-flavoured bundles.

### Details

The internal compiler has been updated, see the companion PR.
We can now have ES5 bundles aside from the regular ones. To properly use them, we need the full chain to declare dependencies to ES5. See the companion PRs for some examples.

### How to test
- checkout the branch
- install the dependencies: `cd tao/views/build && npm i`
- run bundle tasks: `cd tao/views/build && npx grunt taobundle`


helpers/Layout.php
```diff
@@ -175,6 +175,47 @@ public static function getAmdLoader($bundle = null, $controller = null, $params
...
+    public static function getAmdLoaderES5(string $bundle, string $controller, array $params = []): string
+    {
+        $configUrl = get_data('client_config_url');
+        $requireJsUrl = Template::js('lib/require.js', 'tao');
+        $bootstrapUrl = Template::js('loader/bootstrap.js', 'tao');
+
+        $loader = new AmdLoader($configUrl, $requireJsUrl, $bootstrapUrl);
+
+        return "<script src='" . Template::js('loader/vendor.es5.min.js', 'tao') . "'></script>\n" .
+            $loader->getBundleLoader($bundle, $controller, $params);
```
gabrielfs7, line 196: I see that the new code is close to the previews code `getAmdLoader` https://github.com/oat-sa/tao-core/pull/3655/files#diff-e7c3d126a29a1767a1509c8606959f79941db01563496665935b9785e84afad5R158

But it does not consider many aspects like `$bundleMode` or `common_session_SessionManager` (session check, etc). 

Just an idea, maybe instead of having 2 new methods, we could have a parameter `es5` to just handle the load in a different way. I am just afraid we could have some discrepancy while having the same logic in 3 places. 

Or maybe we could at least group the common logic in private methods and call them according to the other methods need.

helpers/Layout.php
```diff
@@ -175,6 +175,47 @@ public static function getAmdLoader($bundle = null, $controller = null, $params
...
+    /**
+     * Loads a standalone bundle.
+     * Loads the bundle no matter if this is production or debug mode.
+     *
+     * @param string $bundle the bundle URL
+     * @param string $controller the controller module id
+     * @param array $params additional parameters
+     * @return string the script tag
+     */
+    public static function getBundleLoader(string $bundle, string $controller, array $params = []): string
```
gabrielfs7, line 208: I think this method is not called in any place. I could not see it in this PR at least https://github.com/oat-sa/extension-tao-booklet/pull/128/files#diff-1b426c437d34a3d4e5e660102e185bc6261a66c93b98e80f7f0cdb427d8cec13R7

Should we still have this method here?

shaveko (approved):
Looks good to me

helpers/Layout.php
jsconan: Actually, it is added as support for the [standalone bundles](https://github.com/oat-sa/grunt-tao-bundle/pull/36/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R78). It was used in first place here: https://github.com/oat-sa/extension-tao-booklet/pull/128/commits/833b64def8931f2aa102ac5df144e30fa1be2c43

helpers/Layout.php
jsconan: Well, the intent behind it is to load the bundle in any mode. `getAmdLoader` will not load the bundles but directly the source code when the debug mode is active.

helpers/Layout.php
```diff
@@ -175,6 +175,47 @@ public static function getAmdLoader($bundle = null, $controller = null, $params
...
+    /**
+     * Create the AMD loader for the current context using ES5.
+     * Loads the bundle no matter if this is production or debug mode.
+     *
+     * @param string $bundle the bundle URL
+     * @param string $controller the controller module id
+     * @param array $params additional parameters
+     * @return string the script tag
+     */
+    public static function getAmdLoaderES5(string $bundle, string $controller, array $params = []): string
```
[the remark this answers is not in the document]
jsconan: done :)

helpers/Layout.php
jsconan: Anyway, I've refactored the API, splitting the AMD loader into smaller units. I hope the new shape is more rationale

helpers/Layout.php
jsconan: Anyway, I've refactored this part ;)

wazelin (approved):

helpers/Layout.php
gabrielfs7: thanks!

helpers/Layout.php
gabrielfs7: But the AMDLoader does not extends Layout and has its own method https://github.com/oat-sa/tao-core/blob/5dff0a1402acb3c4a96426cb0f2297135a449691/models/classes/layout/AmdLoader.php#L71

helpers/Layout.php
```diff
@@ -137,42 +139,100 @@ public static function renderIcon($icon, $defaultIcon)
...
+    ): string {
         if (\common_session_SessionManager::isAnonymous() && !$allowAnonymous) {
             $controller = 'controller/login';
             $bundle = Template::js('loader/login.min.js', 'tao');
         }
 
-        if ($bundleMode) {
-            return "<script src='" . Template::js('loader/vendor.min.js', 'tao') . "'></script>\n" .
-                    $loader->getBundleLoader($bundle, $controller, $params);
+        if (\tao_helpers_Mode::is('production')) {
```
gabrielfs7, line 231: We are as standard importing the FQN, just sharing with you :)

helpers/Layout.php
```diff
@@ -137,42 +139,100 @@ public static function renderIcon($icon, $defaultIcon)
...
+     * @param string $type the type of bundle, can be: '', 'es5', 'standalone' (default: '')
      * @return string the script tag
      */
-    public static function getAmdLoader($bundle = null, $controller = null, $params = null, $allowAnonymous = false)
-    {
+    public static function getBundleLoader(
+        string $bundle,
+        string $controller,
+        array $params = [],
+        string $type = ''
```
gabrielfs7, line 160: My suggestion was more about adding the $params['type'] instead of a new parameter, so we do not increase the method parameters and we do not break the method signature, but it is my opinion :) 

Thanks for the refactoring!

helpers/Layout.php
```diff
@@ -137,42 +139,100 @@ public static function renderIcon($icon, $defaultIcon)
                     $icon->getId()
                 );
 
-            case ''; // no source means an icon font is used
+            case '': // no source means an icon font is used
```
gabrielfs7, line 142: Good catch :)

gabrielfs7 (approved):

PR #3672, opened 2022-11-11
Fix/aut 2806/low risk fixes

bartlomiejmarszal:


views/package-lock.json
```diff
@@ -2657,9 +2657,9 @@
                     "dev": true
                 },
                 "socket.io-parser": {
-                    "version": "3.3.2",
```
andreluizmachado, line 2660: Are we sure that it's a good idea to change the package-lock directly?
is It a good practice?
what if we update the root dependency?
Screenshot 2023-01-25 at 10 14 54

bziondik (approved):
LGTM

views/package-lock.json
bartlomiejmarszal: I did not change that in .lock file. It was done along with npm update process.

pnal (approved):

andreluizmachado (approved):
LGTM

Karol-Stelmaczonek:
Closing, those changes are already in current develop.

PR #3698, opened 2022-12-15
fix/TR-4934/cookiePHP7.4

africz:
Not tested yet properly, not sure about this fix solve the issue or not

PR #3711, opened 2022-12-23
Properly set cookie by utilising new setcookie contract 

poyuki:
draft for https://github.com/oat-sa/tao-core/pull/3698

### DIFF
--- a/models/classes/session/Business/Contract/SessionCookieAttributeInterface.php
+++ b/models/classes/session/Business/Contract/SessionCookieAttributeInterface.php
@@ -27,4 +27,8 @@ namespace oat\tao\model\session\Business\Contract;
 interface SessionCookieAttributeInterface
 {
     public function __toString(): string;
+
+    public function getName(): string;
+
+    public function getValue();
 }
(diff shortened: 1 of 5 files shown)


models/classes/session/Business/Contract/SessionCookieAttributeInterface.php
```diff
@@ -27,4 +27,8 @@
 interface SessionCookieAttributeInterface
 {
     public function __toString(): string;
+
+    public function getName(): string;
+
+    public function getValue();
```
poyuki, line 33: that also is breaking change but i thing it shouldn't affect application for few instances of app i can confirm only 2 implementation of interface which you already adjusted

models/classes/session/Business/Domain/SessionCookieAttributeCollection.php
```diff
@@ -59,4 +59,13 @@ public function __toString(): string
...
     }
+
+    public function toCookieAttributesMap(): array
+    {
+        $resultMap = [];
+        foreach ($this->attributes as $attribute) {
+            $resultMap[$attribute->getName()] = $attribute->getValue();
+        }
+        return $resultMap;
+    }
```
poyuki, line 70: let's have possibility to cast it collection to proper assoc

PR #3725, opened 2023-01-20
[FIX] [PISA25-269] deprecated special method \tao_actions_RestTrait::returnSuccess

andreluizmachado:
Refs: https://github.com/oat-sa/extension-tao-itemqti/pull/2317/commits/5fcdd2664f6cee9ad8797df7fc96c463db394750


pnal:
@andreluizmachado could you add a short explanation why it is dangerous to use in a description?

pnal (approved):

PR #3746, opened 2023-02-21
fix: Create item or class from subclasses from search result items in left sidebar (Search by label) for TAO Enterprise INVALSI PHP 8.1

chinnu-focaloid:
Related to https://oat-sa.atlassian.net/browse/LSI-1825

**Description**: Fix 500 error when clicking on 'New class'/'New Item' on search result items in left sidebar (Search by label) for TAO Enterprise INVALSI PHP 8.1

Whenever a user tries to click on 'New class'/'New Item' on search result items if the item is under subclasses, there’s a 500 error message popping up related to Security Exception and missing signature of the parent class. 

image

This issue can be reproduced in the following environment
http://test-invalsi_php_8_1.playground.kitchen.it.taocloud.org:49721/
admin/admin

To test this fix or enable the left sidebar search, you will need to activate the resourceSelector. In the config file config/tao/client_lib_config_registry.conf.php, simply add:

```
        'layout/tree/loader' => array(
            'treeProvider' => 'resource-selector'
        ),
```
The fix has been deployed to the following environment
http://test-chinnu.playground.kitchen.it.taocloud.org:41411/tao/Main/login
admin/admin

PR #3750, opened 2023-02-28
Backport fix: Create item or class from subclasses from search result items in…

chinnu-focaloid:
… left sidebar (Search by label) for TAO Enterprise INVALSI PHP 8.1 (#3746)

* fix: Create item or class from subclasses from search result items in left sidebar


bziondik (approved):
FE part only

wazelin (approved):

PR #3751, opened 2023-03-01
feat: remove keys call to avoid locking keyset

gabrielfs7:
https://oat-sa.atlassian.net/browse/ADF-1328
https://oat-sa.atlassian.net/browse/ADF-1329

# Goal

- Improve REDIS performance by removing the `keys` call that locked the keyset and is not recommended to use on production (More details here https://oat-sa.atlassian.net/wiki/spaces/OAT/pages/1972731905/RFE-396+Token+Pool+Creation+Optimization#ISSUE-3---Redis-keys).

## Changes done

- Remove the `keys` usage.
- Use hash map to save tokens, saving storage and improving data retrieval from Redis.
- Create a command to delete expired tokens. It needs to be added to a CRON Job.

## How to test

Follow instructions https://oat-sa.atlassian.net/wiki/spaces/OAT/pages/1991311625/How+to+benchmark+TokenService+with+Redis

Also:

- Check the CSRF token validation will work by submitting items as TestTaker.
- ❗ The CSRF token validation inside the Authoring NEVER worked. It only logs a warning and nothing more.
- Run this command to check that tokens created after 120sec are removed `php index.php 'oat\tao\scripts\tools\Security\ClearCsrfTokenTool' -t 120`

## Benchmark results 

The benchmark shows that we have more Network, Ops/sec because of the new code comparing to the previous, but we are not blocking Redis operations for anything else since we removed the usage of `keys`:

Old implementation Vs New

Screenshot 2023-03-13 at 14 37 19

- Reduced Ops/sec from **2k/s** to **600/s**.
- Reduce input from 150kb to 50kb (3x less).

## Related PRs

- https://github.com/oat-sa/generis/pull/1030 (needed to check)
- https://github.com/oat-sa/tao-core/pull/3442/files (this PR caused the issue with `keys`)

## TODO 

- [x] Add unit tests
- [x] Fix code styles


models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -41,31 +40,29 @@ class TokenStoreKeyValue extends ConfigurableService implements TokenStore
...
-    private $persistence;
-
-    /** @var string */
-    private $keyPrefix;
+    private common_persistence_AdvKeyValuePersistence $persistence;
+    private string $keyPrefix;
 
     public function getToken(string $tokenId): ?Token
     {
-        if (!$this->hasToken($tokenId)) {
```
gabrielfs7, line 53: Method removed to avoid hitting Redis, instead, we just call `get` and in case the response is not a string, we create the token.

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -75,18 +72,18 @@ public function hasToken(string $tokenId): bool
 
     public function removeToken(string $tokenId): bool
     {
-        if (!$this->hasToken($tokenId)) {
```
gabrielfs7, line 78: Method removed to avoid hitting Redis, instead, we just call `del` and in case the response is not a `true`, we know that the token did not exist or was not removed.

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -41,31 +40,29 @@ class TokenStoreKeyValue extends ConfigurableService implements TokenStore
     public const OPTION_TTL = 'ttl';
 
     public const TOKENS_STORAGE_KEY = 'tao_tokens';
+    public const TOKENS_STORAGE_COLLECTION_KEY_SUFFIX = 'keys';
```
gabrielfs7, line 43: ❗ We control all keys per uses with a new entry in Redis, since `hSet`, `hSet`, etc do not support TTL.

hectoras (approved):
- [x] Target branch is correct

rgomeztao:
Overall looks very nice :)

For what i see, were taking care of the keys by creating a custom key that keeps all the keys from that pattern. The idea seems straightfoward, but i wonder if REDIS could provide something to support this approach :thinking: 

on the other hand, @hectoras and I executed the script on our grafana in order to check the concurrent operation in REDIS. on a normal hit basis with some time between executions, we see a graph similar that the one posted, BUT, if we execute the test file multiple times in a small amount of time, we get this

image

It seems that the effort multiplies for the next execution in the small amount of time. Im not sure if this is an issue :thinking:

rgomeztao (commented):
benchmark shows indicators of possible stress in multiple executions, needs confirmation that its not an issue :)

rgomeztao:
after discussing concerns with @gabrielfs7 all looks good, grrreat job!! :partying_face:

scripts/tools/Security/ClearCsrfTokenTool.php
```diff
@@ -0,0 +1,98 @@
...
+
+use oat\oatbox\extension\script\ScriptAction;
+use oat\oatbox\reporting\Report;
+use oat\tao\model\security\xsrf\TokenService;
+
+/**
+ * Clear all tokens created after 120 seconds ago:
+ *  php index.php 'oat\tao\scripts\tools\Security\ClearCsrfTokenTool' -t 120
+ */
+class ClearCsrfTokenTool extends ScriptAction
```
gabrielfs7, line 35: Added a new command to be executed via Cronjob to clear tokens every day based on some expiration time...

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -129,78 +119,84 @@ protected function getPersistence(): common_persistence_AdvKeyValuePersistence
...
     {
-        return $this->buildKey(self::TOKENS_STORAGE_COLLECTION_KEY_SUFFIX);
-    }
+        $persistence = $this->getPersistence();
 
-    private function buildKey(string $name): string
-    {
-        return sprintf('%s_%s', $this->getKeyPrefix(), $name);
-    }
+        /** @var common_persistence_PhpRedisDriver $driver */
```
hectoras, line 144: nitpick: The variable type for `$driver` is implicitly stated by the `$driver instanceof common_persistence_PhpRedisDriver` conditional below, the IDE should be able to guess the correct var type based on that.

models/classes/security/xsrf/TokenStoreKeyValue.php
```diff
@@ -129,34 +119,84 @@ protected function getPersistence(): common_persistence_AdvKeyValuePersistence
...
+        $pattern = sprintf('*%s*', self::TOKENS_STORAGE_KEY);
+        $countDeleted = 0;
 
-    private function buildKey(string $name): string
-    {
-        return sprintf('%s_%s', $this->getKeyPrefix(), $name);
+        if ($driver instanceof common_persistence_PhpRedisDriver) {
+            $iterator = null;
+
+            while ($iterator !== 0) {
```
hectoras, line 152: I think this `while` is always evaluating to `false`, is that intended?

hectoras: Forget about this one (ofc `null` will be `!== 0`), I'm still sleepy :sweat_smile:

scripts/tools/Security/ClearCsrfTokenTool.php
rgomeztao: so we go cronjob???? this def a significant improvement with last version :D, great job!

scripts/tools/Security/ClearCsrfTokenTool.php
```diff
@@ -0,0 +1,110 @@
...
+            return Report::createSuccess($message);
+        } catch (Throwable $exception) {
+            $message = sprintf(
+                'Error removing CSRF tokens: (%s) %s, trace %s',
+                get_class($exception),
+                $exception->getMessage(),
+                $exception->getTraceAsString()
+            );
+
+            $this->logInfo($message);
```
rgomeztao, line 100: suggestion: logError ;) its on the trait from the parent class 
nitpick: including the sprintf in the logInfo would reduce the halstead volume by 1. Up to you if you think keeping the $message variable helps the code being more readable :)

rgomeztao (approved):

scripts/tools/Security/ClearCsrfTokenTool.php
gabrielfs7: Yes, it was a typo. Fixed. Thanks!

PR #3752, opened 2023-03-01
fix: add retryTask for all exceptions

tikhanovichA:
For some cases related to remote publications, webhook can be failed because delivery is not ready yet. 
We can manage all those retries using `retryMax` option in the webhookRegistry.conf.php if it will be needed.

We need also backport it to v51.0.3

### DIFF
--- a/models/classes/webhooks/WebhookTaskServiceInterface.php
+++ b/models/classes/webhooks/WebhookTaskServiceInterface.php
@@ -39,6 +39,7 @@ interface WebhookTaskServiceInterface
     /**
      * Create and enqueue task for performing webhook
      * @param WebhookTaskParams $webhookTaskParams
+     * @param string $label
      */
-    public function createTask(WebhookTaskParams $webhookTaskParams);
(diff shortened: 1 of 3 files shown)


andreluizmachado:
Hey, can we also fix PHP CS mentioned on [CIs](https://github.com/oat-sa/tao-core/actions/runs/4304532634/jobs/7505695016)?

models/classes/webhooks/WebhookTaskServiceInterface.php
```diff
@@ -39,6 +39,7 @@ public function linkTaskToQueue($queueName);
     /**
```
pnal, line 39: `public const SERVICE_ID = 'tao/webhookTaskService';`
on line 30

pnal (changes requested):
Please fix a CI CS warning

wazelin (changes requested):
This looks alright-ish, but potentially dangerous.

Can a more specific exception be added instead to cover the case you're trying to here? If not, consider this approved once the CI is back to green again.

PR #3755, opened 2023-03-02
feat: Split ACL updates in subtasks

gabrielfs7:
https://github.com/oat-sa/extension-tao-dac-simple/pull/210/files

https://oat-sa.atlassian.net/browse/ADF-1320

- [x] Update unit tests

### DIFF
--- a/models/classes/listener/DataAccessControlChangedListener.php
+++ b/models/classes/listener/DataAccessControlChangedListener.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2020-2023 (original work) Open Assessment Technologies SA;
  *
  */
 
--- a/models/classes/listener/DataAccessControlChangedListener.php
+++ b/models/classes/listener/DataAccessControlChangedListener.php
@@ -49,7 +49,11 @@ class DataAccessControlChangedListener extends ConfigurableService
         /** @noinspection PhpUnhandledExceptionInspection */
         $resource = $this->getResource($event->getResourceId());
 
-        if ($resource->isClass() && !$event->isRecursive()) {
+        /**
+         * @TODO Verify if we already had a bug before when a class ACL is updated,
+         * but not marking `recursive` if it never applied
+         */
+        if ($resource->isClass() && !$event->isProcessNestedResources()) {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -73,7 +73,7 @@ public function processTask(TaskInterface $task)
             $report = Report::createInfo(__('Running task %s', $task->getId()));
             try {
                 $this->startUserSession($task);
-                $this->logInfo('Processing task ' . $task->getId(), $this->getLogContext());
+                $this->logInfo(sprintf('Processing task %s [%s]', $task->getLabel(), $task->getId()), $this->getLogContext());
```
gabrielfs7, line 76: - [ ] We should probably substr the label to avoid pass the DB field limit.

models/classes/listener/DataAccessControlChangedListener.php
```diff
@@ -49,7 +47,12 @@ public function handleEvent(DataAccessControlChangedEvent $event): void
         /** @noinspection PhpUnhandledExceptionInspection */
         $resource = $this->getResource($event->getResourceId());
 
-        if ($resource->isClass() && !$event->isRecursive()) {
+        /**
+         * @TODO Verify if we already had a bug before when a class ACL is updated,
+         * but not marking `recursive` if it never applied
+         */
+        if ($resource->isClass() && !$event->applyToNestedResources()) {
```
hectoras, line 54: Maybe here we need to extend the parameters passed to `UpdateDataAccessControlInIndex` so the listener (at AdvancedSearch) can run the ES update querying where the resource ID equals to this class BUT not updating the instances of the class themselves.

It looks like `IndexUpdater::updatePropertyValue` from AdvancedSearch updates the class AND all instances having the class as one of its types ([here](https://github.com/oat-sa/extension-tao-advanced-search/blob/master/model/SearchEngine/Driver/Elasticsearch/IndexUpdater.php#L233)).

models/classes/taskQueue/Worker/AbstractWorker.php
hectoras: Done cc18e34219968eb7930e37a215a885651a979e4c

rgomeztao (approved):
after a conversion with @hectoras and some clarifications, looks nice :), good job!

test/unit/model/listener/DataAccessControlChangedListenerTest.php
```diff
@@ -102,10 +115,12 @@ public function testHandleEventShouldCreateTaskSuccessfully(bool $isRecursive, b
...
             );
 
-        $this->sut->handleEvent(new DataAccessControlChangedEvent($documentUri, [], $isRecursive));
+        $this->sut->handleEvent(
+            new DataAccessControlChangedEvent($documentUri, [], $isRecursive, $isRecursive)
+        );
     }
 
-    public function provideSuccessfulCases(): array
+    public function handleEventSuccessfulCasesDataProvider(): array
```
gabrielfs7, line 123: Why apply a breaking change?

PR #3758, opened 2023-03-09
feat: Extend DataAccessControlChangedEvent to support updates skipping nested resources

hectoras:
**Associated Jira issue:** [ADF-1320](https://oat-sa.atlassian.net/browse/ADF-1320)

Changes in this PR introduce a new flag `applyToNestedResources` in `DataAccessControlChangedEvent` that, when is set, makes `DataAccessControlChangedListener` skip updates in search indexes for the class (previously that was the case when passing the recursive flag instead).

As part of this change, this PR updates unit tests to take into account that change and makes modified files to comply with PSR-12 (preventing validation errors in the pipelines).

Related PRs:
- https://github.com/oat-sa/extension-tao-dac-simple/pull/211
- https://github.com/oat-sa/extension-tao-advanced-search/pull/91

[ADF-1320]: https://oat-sa.atlassian.net/browse/ADF-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/models/classes/event/DataAccessControlChangedEvent.php
+++ b/models/classes/event/DataAccessControlChangedEvent.php
@@ -29,7 +29,7 @@ class DataAccessControlChangedEvent implements Event
     /** @var string */
     private $resourceId;
 
-    /** @var string */
+    /** @var ?string */
     private $rootResourceId;
 
     /** @var array */
@@ -45,16 +45,16 @@ class DataAccessControlChangedEvent implements Event
 
     public function __construct(
         string $resourceId,
-        string $rootResourceId, // @todo Update unit tests
         array $addRemove,
         bool $isRecursive = false,
-        bool $applyToNestedResources = false
+        bool $applyToNestedResources = false,
+        string $rootResourceId = null
     ) {
         $this->resourceId = $resourceId;
-        $this->rootResourceId = $rootResourceId;
         $this->addRemove = $addRemove;
         $this->isRecursive = $isRecursive;
         $this->applyToNestedResources = $applyToNestedResources;
+        $this->rootResourceId = $rootResourceId;
     }
 
     public function getName(): string
@@ -67,7 +67,7 @@ class DataAccessControlChangedEvent implements Event
         return $this->resourceId;
     }
 
-    public function getRrootResourceId(): string
+    public function getRootResourceId(): ?string
     {
         return $this->rootResourceId;
     }
--- a/models/classes/event/DataAccessControlChangedEvent.php
+++ b/models/classes/event/DataAccessControlChangedEvent.php
@@ -39,18 +39,18 @@ class DataAccessControlChangedEvent implements Event
     private $isRecursive;
 
     /** @var bool */
-    private $processNestedResources;
+    private $applyToNestedResources;
 
     public function __construct(
         string $resourceId,
         array $addRemove,
         bool $isRecursive = false,
-        bool $processNestedResources = false
+        bool $applyToNestedResources = false
     ) {
         $this->resourceId = $resourceId;
         $this->addRemove = $addRemove;
         $this->isRecursive = $isRecursive;
-        $this->processNestedResources = $processNestedResources;
+        $this->applyToNestedResources = $applyToNestedResources;
     }
 
     public function getName(): string
@@ -69,15 +69,15 @@ class DataAccessControlChangedEvent implements Event
     }
 
     /**
-     * @deprecated Use isProcessNestedResources cause processing recursively causes performance issues
+     * @deprecated Use applyToNestedResources because processing recursively causes performance issues
      */
     public function isRecursive(): bool
     {
         return $this->isRecursive;
     }
 
-    public function isProcessNestedResources(): bool
+    public function applyToNestedResources(): bool
     {
-        return $this->processNestedResources;
+        return $this->applyToNestedResources;
     }
 }
--- a/models/classes/event/DataAccessControlChangedEvent.php
+++ b/models/classes/event/DataAccessControlChangedEvent.php
@@ -38,8 +38,7 @@ class DataAccessControlChangedEvent implements Event
      */
     private $isRecursive;
 
-    /** @var bool */
-    private $applyToNestedResources;
+    private bool $applyToNestedResources;
 
     public function __construct(
         string $resourceId,
--- a/models/classes/event/DataAccessControlChangedEvent.php
+++ b/models/classes/event/DataAccessControlChangedEvent.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2020-2023 (original work) Open Assessment Technologies SA;
  */
 
 declare(strict_types=1);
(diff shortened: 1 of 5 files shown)


models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -73,14 +73,31 @@ public function processTask(TaskInterface $task)
...
+                $this->logInfo(
+                    sprintf(
+                        'Processing task %s [%s]',
+                        (strlen($task->getLabel()) > 255
+                            ? '...' . substr($task->getLabel(), -252)
+                            : $task->getLabel()),
+                        $task->getId()
+                    ),
+                    $this->getLogContext()
+                );
```
gabrielfs7, line 86: Is the field supporting max 255 chars? If yes, we also need to consider the whole message, including `Processing task...`

models/classes/listener/DataAccessControlChangedListener.php
```diff
@@ -49,7 +49,11 @@ public function handleEvent(DataAccessControlChangedEvent $event): void
         /** @noinspection PhpUnhandledExceptionInspection */
         $resource = $this->getResource($event->getResourceId());
 
-        if ($resource->isClass() && !$event->isRecursive()) {
+        /**
+         * @TODO Verify if we already had a bug before when a class ACL is updated,
+         * but not marking `recursive` if it never applied
+         */
+        if ($resource->isClass() && !$event->applyToNestedResources()) {
```
gabrielfs7, line 56: Some points:

- Shall it be also backward compatible with the `isRecursive`?
- If we verified that TODO, can we remove it from the comment?

```suggestion
        if ($resource->isClass() && !$event->applyToNestedResources() && !$event->isRecursive()) {
```

models/classes/taskQueue/Worker/AbstractWorker.php
hectoras: This goes to the log file, so I think there is no hard limit on the message size (and [PSR-3](https://www.php-fig.org/psr/psr-3/) does not seem to limit the line length). Locally, I get these fully recorded `tao.log` without data loss:

`[2023-03-09 16:29:50] UDEMO.INFO: Processing task Processing permissions for class QTI Interactions 2 [https://adf-1320-dac-simple-batches.docker.localhost/ontologies/tao.rdf#i63e4df0c5bf119618cd15b15aade3a81a] [https://adf-1320-dac-simple-batches.docker.localhost/ontologies/tao.rdf#i6409fb6e2af833814b2de5c7c514039db] {"PID":3814,"Iteration":10,"QueueName":"queue"} {"uid":"f1f5f313369be5f767fbebdb"}
`

I've checked the DB locally and this is never recorded in the DB as part of the task report (we only get the message provided in the `Report` instance generated by DAC Simple when the task finishes).

gabrielfs7: Ok, so now I did not get why the `255` check...

models/classes/taskQueue/Worker/AbstractWorker.php
```diff
@@ -73,14 +73,31 @@ public function processTask(TaskInterface $task)
...
+                    $task->getId(),
+                    TaskLogInterface::STATUS_RUNNING,
+                    TaskLogInterface::STATUS_DEQUEUED
+                );
 
                 // if the task is being executed by another worker, just return, no report needs to be saved
                 if (!$rowsTouched) {
-                    $this->logInfo('Task ' . $task->getId() . ' seems to be processed by another worker.', $this->getLogContext());
+                    $this->logInfo(
+                        sprintf('Task %s seems to be processed by another worker.', $task->getId()),
```
gabrielfs7, line 98: Optional: 

It would be nice to add the task label too, if we are refactoring this part. The task ID does not provide much context checking the logs. Wdyt?

models/classes/event/DataAccessControlChangedEvent.php
```diff
@@ -32,14 +32,25 @@ class DataAccessControlChangedEvent implements Event
...
+    /**
+     * @deprecated Use $applyToNestedResources cause processing recursively causes performance issues
+     * @var bool
+     */
     private $isRecursive;
 
-    public function __construct(string $resourceId, array $addRemove, bool $isRecursive = false)
-    {
+    /** @var bool */
+    private $applyToNestedResources;
```
gabrielfs7, line 42: PHP 7.4 

```suggestion
    private bool $applyToNestedResources;
```

models/classes/taskQueue/Worker/AbstractWorker.php
hectoras: 6865e3c14d5e949f1f1f0fa679818fcf2755d164

models/classes/taskQueue/Worker/AbstractWorker.php
hectoras: We did this change in the former PR, then copied it here (I'm reusing the same branch).

Initially, we just changed it to have the label in the logs because it helps debugging when lots of subtasks are triggered in a recursive operation. Then I think we thought this was logged to the DB so we added the substr call.

BTW this helps reading the logs for resources where the labels are too long, so personally I'd keep it (and probably some logging drivers/vendors will have restrictions in the line length).

models/classes/listener/DataAccessControlChangedListener.php
hectoras: Done: ee19b1543e01ec72d06c96992ea31148eb8cc9b2

rgomeztao (approved):
I dont see anything remarkable to comment apart of the current conversations. Tests run well in local :), nice job!

models/classes/taskQueue/Worker/AbstractWorker.php
hectoras: This was already done, marking the thread as Solved. Please do reopen it if you think there is something else pending related with this.

models/classes/listener/DataAccessControlChangedListener.php
```diff
@@ -49,7 +49,7 @@ public function handleEvent(DataAccessControlChangedEvent $event): void
         /** @noinspection PhpUnhandledExceptionInspection */
         $resource = $this->getResource($event->getResourceId());
 
-        if ($resource->isClass() && !$event->isRecursive()) {
+        if ($resource->isClass() && !$event->applyToNestedResources() && !$event->isRecursive()) {
```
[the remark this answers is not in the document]
hectoras: Please check 5296581c8e57bb77482322aa574ab84469550198 and latest changes from the related PRs. I think that, together with changes in the related PRs, solves the issue for the duplicated indexations.

I'm triggering the events as before from DAC Simple but I'm including an additional parameter (`$rootResourceId`) in the event form DAC Simple itself. That way, once we reach this listener, we can check that the event is not the one triggered for the root resource, so the ES update is only called for the root.

First we'll trigger all events from DAC Simple from each of its tasks/subtasks) as before. However, now all of them will carry also the URI of the "root" resource (the one associated with the former user request).

`DataAccessControlChangedListener` (here in Core) will receive all events, but it will filter out events for resources having `resource URI != root resource URI`. For the root one, an additional `UpdateDataAccessControlInIndex` task is created (previously it was triggered for all of them).

`UpdateDataAccessControlInIndex` should run a query to update the permissions in the ES index using only the URI for the root resource, but the query we already had there should match all resources contained in child classes as well, therefore updating the index for all resources having `root` in any previous level.

hectoras:
Tests passing

gabrielfs7 (approved):
Thanks @hectoras !

We tested together this on Héctor's computer covering existing previous cases. We need to deploy it later to a UDIR environment to make sure it will also work with their remote ACL.

hectoras:
Released as [v51.9.0](https://github.com/oat-sa/tao-core/releases/tag/v51.9.0).

PR #3780, opened 2023-04-11
fix: Change regex to accept CamelCase Migration Names

bartlomiejmarszal:
In the qtiItemPci extension according to camel case class name rules, we changed migration names that did not match to regex used to sort migrations. 

- Version202209131056091465QtiItemPci.php
- Version202209141533261465QtiItemPci.php
- Version202211031046021465QtiItemPci.php

This fix will match those migrations and order them in the correct way.

### DIFF
--- a/scripts/tools/migrations/TaoComparator.php
+++ b/scripts/tools/migrations/TaoComparator.php
@@ -27,6 +27,7 @@ use Doctrine\Migrations\Version\Version;
 use common_ext_ExtensionsManager as ExtensionsManager;
 use common_ext_Extension as Extension;
 use helpers_ExtensionHelper as ExtensionHelper;
+use League\Csv\Exception;
 
 /**
  * Class TaoComparator is used by a migration repository to sort migrations by extensions according to
@@ -36,6 +37,7 @@ use helpers_ExtensionHelper as ExtensionHelper;
  */
 class TaoComparator implements Comparator
 {
+    private const VERSION_REGEX = '/.*Version(\d+)_?(.*)$/';
     /** @var ExtensionsManager  */
     private $extensionsManager;
 
@@ -61,13 +63,12 @@ class TaoComparator implements Comparator
         $sortedExtensions = array_flip(array_keys($this->extensionHelper::sortByDependencies($merged)));
         $versionA = (string) $a;
         $versionB = (string) $b;
-        preg_match('/.*Version(\d+)_(.*)$/', $versionA, $matchesA);
-        preg_match('/.*Version(\d+)_(.*)$/', $versionB, $matchesB);
-        list($aClass, $aTime, $aExt) = $matchesA;
-        list($bClass, $bTime, $bExt) = $matchesB;
+        list($aClass, $aTime, $aExt) = $this->matchExtensionName($versionA);
+        list($bClass, $bTime, $bExt) = $this->matchExtensionName($versionB);
         if ($aExt === $bExt) {
             return ((int) $aTime) - ((int) $bTime);
         }
+
         return $sortedExtensions[$aExt] - $sortedExtensions[$bExt];
     }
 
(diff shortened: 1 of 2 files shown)


scripts/tools/migrations/TaoComparator.php
```diff
@@ -82,4 +83,12 @@ protected function getMissingExtensions()
         }
         return $result;
     }
+
+    private function matchExtensionName(string $versionA): array
```
Karol-Stelmaczonek, line 87: just $version

scripts/tools/migrations/TaoComparator.php
```diff
@@ -82,4 +83,12 @@ protected function getMissingExtensions()
...
+    {
+        preg_match(self::VERSION_REGEX, $versionA, $matches);
+        $extName = $matches[2];
```
Karol-Stelmaczonek, line 90: Now when you use indexes you need to check if $matches is populated and index exist

test/unit/scripts/tools/migrations/TaoComparatorTest.php
```diff
@@ -46,6 +46,7 @@ public function testCompare()
             $extensionBar4 = new Version('Version4_bar'),
             $extensionBar2 = new Version('Version2_bar'),
             $extensionBaz1 = new Version('Version1_baz'),
+            $illigalVersion = new Version('Version1Baz'),
```
Karol-Stelmaczonek, line 49: Why illegal? Its legit now after your change

scripts/tools/migrations/TaoComparator.php
pnal, line 90:
```suggestion
        $matches[2] = lcfirst($matches[2]);
```

scripts/tools/migrations/TaoComparator.php
```diff
@@ -82,4 +83,12 @@ protected function getMissingExtensions()
...
+        $matches[2] = strtolower($extName[0]) . substr($extName, 1);
```
pnal, line 91:
```suggestion
```

scripts/tools/migrations/TaoComparator.php
pnal: I would allow it to write warning in such cases or even more allow to break taoUpdate by throwing error (of course if this code is only used for taoUpdate)

test/unit/scripts/tools/migrations/TaoComparatorTest.php
Karol-Stelmaczonek: also you can add Version1baz for consistency

scripts/tools/migrations/TaoComparator.php
```diff
@@ -27,6 +27,7 @@
 use common_ext_ExtensionsManager as ExtensionsManager;
 use common_ext_Extension as Extension;
 use helpers_ExtensionHelper as ExtensionHelper;
+use League\Csv\Exception;
```
pnal, line 30: why it appeared?

scripts/tools/migrations/TaoComparator.php
Karol-Stelmaczonek: One more, it will handle the case when one version is with _ and other isn't?

bartlomiejmarszal: @Karol-Stelmaczonek It will accept both cases, yes. 
Unfortunately, there is no way to throw any warnings in here. I was trying but it will just fail un `usort` function and move forward. 
I created a separate ticket restricting migrations names to those automatically created.

jsconan (approved):
It works nicely. I don't have these noisy notices anymore. Good job!

I only added my 2cts, but nothing blocking actually.

**Review Checklist**
- [x] New code respects code style rules
- [x] New code respects best practices
- [x] Commits are atomic

jsconan (changes requested):
Wait, the base branch is not the correct one. It should be `develop`, not master!

scripts/tools/migrations/TaoComparator.php
```diff
@@ -36,6 +36,7 @@
  */
 class TaoComparator implements Comparator
 {
+    private const VERSION_REGEX = '/.*Version(\d+)_?(.*)$/';
```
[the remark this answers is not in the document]
jsconan: Yes, I know the `_` is now optional. But it is also processed by `(.*)`.

Ok, I see that it should not take place in the name either and should be kept outside of the capture group.
All good for me.

jsconan (approved):
Thank you for the change and the clarification.

All good for me!

PR #3783, opened 2023-04-12
fix: remove EnvironmentValueStorage

bartlomiejmarszal:
# How to reproduce

When creating a new image using a container built with environment variables we will not pass container environment variables defined in the new container. 

# Solution

We will not use EnvironmentValueStorage container to store environment variables but instead, we will access them directly.


models/classes/configurationMarkers/ConfigurationMarkers.php
```diff
@@ -85,8 +83,8 @@ private function walkReplaceMarkers(&$item): void
             return;
         }
 
-        $isSecretDefined = $this->secretsStorage->exist($matches[1] ?? '');
-        $this->printMatchNotification($isSecretDefined, $matches[1]);
+        $isSecretDefined = $_ENV[$matches[1]] ?? false;
```
andreluizmachado, line 86: instead of accessing global vars directly here, can't we inject it as a dependency ???

albertosgz (approved):
I agree with Andre's comment

models/classes/configurationMarkers/ConfigurationMarkers.php
gabrielfs7: Yes, I agree with @andreluizmachado . You can just do this:

```
public function __construct(
        SerializableSecretDtoFactory $serializableFactory,
        LoggerInterface $logger,
        array $envVars = null
    ) {
        $this->serializableFactory = $serializableFactory;
        $this->logger = $logger;
        $this->envVars = $envVars ?? $_ENV;
    }
```

And here 

```suggestion
        $isSecretDefined = $this->envVars[$matches[1]] ?? false;
```

bartlomiejmarszal: I'm a little confused here. The whole point of this change was to remove injecting env variables in __constructor. And now we want to inject it in a different place. 
What if I use `getenv` function instead of referring to a global variable?

gabrielfs7: The point is to remove it from the container definition, otherwise it is not dynamic anymore, but static.

pnal (changes requested):
agree with injecting env array as the class property

models/classes/configurationMarkers/ConfigurationMarkers.php
```diff
@@ -34,17 +34,16 @@ class ConfigurationMarkers
...
 
     public function __construct(
-        EnvironmentValueStorage $secretsStorage,
         SerializableSecretDtoFactory $serializableFactory,
         LoggerInterface $logger
     ) {
-        $this->secretsStorage = $secretsStorage;
         $this->serializableFactory = $serializableFactory;
         $this->logger = $logger;
+        $this->envVars = $_ENV;
```
gabrielfs7, line 46: We should still add it as constructor parameter to be able to mock it on tests and not couple it to this very class. Example: 

```
public function __construct(
        SerializableSecretDtoFactory $serializableFactory,
        LoggerInterface $logger,
        array $vars = null
    ) {
        $this->serializableFactory = $serializableFactory;
        $this->logger = $logger;
        $this->envVars = $vars ?? $_ENV;
```

P.s: You do not need to change anything in the Container provider ;)

Karol-Stelmaczonek (approved):

andreluizmachado (approved):

gabrielfs7 (approved):

PR #3814, opened 2023-05-15
Add workflow file

mike-ionut-mihai-sandu-tao:
Please review my changes

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -6,9 +6,7 @@ on:
   push: 
     branch: 
       - 'main'
-      - 'sonarqube-integration'
       - 'master'
-      - 'develop'
 
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar.yml
```diff
@@ -0,0 +1,29 @@
+name: Sonar_Analysis
+on:
+  pull_request: 
+    branches: 
+      - '**'
+  push: 
+    branch: 
+      - 'main'
```
jsconan, line 8: **thought:** I'd rather take the `develop` branch, in order to detect flaws the soonest as possible. Without considering that `main` is not yet the base branch, this branch is usually targeted by release pull requests, which are automated. We may never see the report.

gabrielfs7 (approved):
Approved with the condition that it will be merged back to develop, no tag will be released and the Solar Qube ruleset will be fixed on demand.

PR #3818, opened 2023-05-22
fix: fix empty credentials

SergiiTao:
image


taorafael (approved):
seems reasonably practical

mianchenia (approved):

uncleempty (approved):
Though I do agree with those changes as being reasonable to finalize the current ticket (as a kind of a smaller evil) and will approve PR, I would ask you to launch a small internal discussion tomorrow to see if we really want to introduce this type of change in the deprecated class.

PR #3839, opened 2023-06-15
Fix/solar 270/allow section visibility filter apply on get navigation elements by group

bartlomiejmarszal:
This change will allow disabling all navigation buttons using the feature flag with SectionVisibilityFilter config. 

How to use: 
- Create migration that will add desired section (perspective) to be disabled base on feature flag
- Define feature flag
- Clear cache

### DIFF
--- a/actions/class.Main.php
+++ b/actions/class.Main.php
@@ -407,7 +407,7 @@ class tao_actions_Main extends tao_actions_CommonModule
          * @var  Perspective $perspective
          */
         foreach (MenuService::getPerspectivesByGroup($groupId) as $i => $perspective) {
-            if (!$this->sectionVisibilityFilter->isVisible($perspective->getId())) {
+            if (!$this->getSectionVisibilityFilter()->isVisible($perspective->getId())) {
                 $this->logDebug(
                     sprintf(
                         "Perspective %s has been ignored base on Section Visibility Filter",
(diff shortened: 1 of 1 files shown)


actions/class.Main.php
```diff
@@ -401,7 +401,22 @@ public function index()
     private function getNavigationElementsByGroup($groupId)
     {
         $entries = [];
+
+        /**
+         * @var int $i
+         * @var  Perspective $perspective
+         */
         foreach (MenuService::getPerspectivesByGroup($groupId) as $i => $perspective) {
```
Karol-Stelmaczonek, line 409: variable $i is never used

actions/class.Main.php
```diff
@@ -401,7 +401,22 @@ public function index()
...
+            if (!$this->getSectionVisibilityFilter()->isVisible($perspective->getId())) {
```
Karol-Stelmaczonek, line 410: $perspective could be a null so will result in fatal if method is called on null
```suggestion
            if ($perspective && !$this->getSectionVisibilityFilter()->isVisible($perspective->getId())) {
```

Karol-Stelmaczonek (approved):

actions/class.Main.php
bartlomiejmarszal: 🤷‍♂️

PR #3857, opened 2023-07-25
fix: fix copy of copied classes

tikhanovichA:
https://oat-sa.atlassian.net/browse/ADF-1545

We need to fix the possibility of copying already copied classes.

**How to test:**
1. Go to items and create a new folder/class
2. Copy this class
3. Try to copy the folder you got in step 2

### DIFF
--- a/models/classes/resources/Service/ClassCopier.php
+++ b/models/classes/resources/Service/ClassCopier.php
@@ -181,9 +181,6 @@ class ClassCopier implements ClassCopierInterface, ResourceTransferInterface
 
     /**
      * Generates a unique identifier for a copy operation.
-     * @param string $from
-     * @param string $to
-     * @return string
      */
     private function generateCopyIdentifier(string $from, string $to): string
     {
--- a/models/classes/resources/Service/ClassCopier.php
+++ b/models/classes/resources/Service/ClassCopier.php
@@ -97,7 +97,7 @@ class ClassCopier implements ClassCopierInterface, ResourceTransferInterface
         core_kernel_classes_Class $class,
         core_kernel_classes_Class $destinationClass,
         bool $keepOriginalPermission = true,
-        string $copyIdentifier = '',
+        string $copyIdentifier = ''
     ): core_kernel_classes_Class {
         // Generate a unique identifier for the copy operation
         if (!$copyIdentifier) {
--- a/models/classes/resources/Service/ClassCopier.php
+++ b/models/classes/resources/Service/ClassCopier.php
@@ -81,8 +81,7 @@ class ClassCopier implements ClassCopierInterface, ResourceTransferInterface
     {
         $class = $this->ontology->getClass($command->getFrom());
         $destinationClass = $this->ontology->getClass($command->getTo());
-        $copyIdentifier = $this->generateCopyIdentifier($class->getUri(), $destinationClass->getUri());
-        $newClass = $this->doCopy($class, $destinationClass, $copyIdentifier, $command->keepOriginalAcl());
+        $newClass = $this->doCopy($class, $destinationClass, $command->keepOriginalAcl());
 
         return new ResourceTransferResult($newClass->getUri());
     }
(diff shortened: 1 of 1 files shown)


uncleempty (approved):
In general it's ok.
Just one question:
maybe it's better to create `$copyIdentifier` in the `doCopy` method instead of creating it higher and providing to each call of this method? Because as far as I can see all the needed args are also available inside. And this will make changes way less intrusive.

poyuki (approved):

models/classes/resources/Service/ClassCopier.php
```diff
@@ -90,15 +91,21 @@ public function copy(
         core_kernel_classes_Class $class,
         core_kernel_classes_Class $destinationClass
     ): core_kernel_classes_Class {
-        return $this->doCopy($class, $destinationClass);
+        $copyIdentifier = $this->generateCopyIdentifier($class->getUri(), $destinationClass->getUri());
+        return $this->doCopy($class, $destinationClass, $copyIdentifier);
```
shpran, line 95:
```suggestion
        $copyIdentifier = $this->generateCopyIdentifier($class->getUri(), $destinationClass->getUri());

        return $this->doCopy($class, $destinationClass, $copyIdentifier);
```

models/classes/resources/Service/ClassCopier.php
```diff
@@ -169,4 +176,15 @@ private function assertInSameRootClass(
 
         $this->assertionCompleted = true;
     }
+
+    /**
+     * Generates a unique identifier for a copy operation.
+     * @param string $from
+     * @param string $to
+     * @return string
```
shpran, line 184: We don't need this phpdoc since we already have this information in the method declaration.
```suggestion
```

shpran (approved):

models/classes/resources/Service/ClassCopier.php
```diff
@@ -91,16 +90,19 @@ public function copy(
...
+        return $this->doCopy($class, $destinationClass);
     }
 
     private function doCopy(
         core_kernel_classes_Class $class,
         core_kernel_classes_Class $destinationClass,
-        string $copyIdentifier,
-        bool $keepOriginalPermission = true
+        bool $keepOriginalPermission = true,
+        string $copyIdentifier = '',
```
poyuki, line 100: I suppose 7.4 didn't support such usage as test are failed

models/classes/resources/Service/ClassCopier.php
```diff
@@ -96,9 +96,18 @@ public function copy(
     private function doCopy(
         core_kernel_classes_Class $class,
         core_kernel_classes_Class $destinationClass,
-        bool $keepOriginalPermission = true
+        bool $keepOriginalPermission = true,
+        string $copyIdentifier = '',
```
shpran, line 100: I prefer to use `null` as default instead of empty string 🙂 
```suggestion
        ?string $copyIdentifier = null,
```

shpran (approved):

poyuki (approved):

hectoras:
TL;DR: Maybe this can be solved by simply declaring `ResourceTransferProxy` in the DI contanier as non-shared.

I think this happens because the class copier is instantiated once by a worker and then it "remembers" the IDs for the classes it copied for prior requests. And I think this happens because the service is instantiated once from the DI container and then reused for each task run by the worker, so wouldn't it be easier to just declare the services form the DI container as non-shared so their state is not remembered across invocations ([here](https://github.com/oat-sa/tao-core/blob/fix/ADF-1545/copy-of-copied-class/models/classes/resources/CopierServiceProvider.php))? 

Specifically, I think this may be happening because `ResourceTransferProxy` receives a `ClassCopierProxy` instance when it is constructed: `ClassCopierProxy` is non-shared, but `ResourceTransferProxy`  is, so it retains a reference to the previous version.

Note here the problem is caused by having a service that is statefull (`ClassCopier` having `$copiedClasses`) and we are trying to solve the issue by adding more state (which "operation" the copy belongs to). There is another problem (that was there before) related to memory usage: A copier instantiated by a worker will be increasing the worker memory footprint each time is invoked (by storing the copied classes and now also an additional operation ID per copy operation).

tikhanovichA:
Will be closed for solving it as @hectoras proposed.

PR #3860, opened 2023-07-31
fix: fix markers population in generis.conf

KirylHatalski:
Fix for markers placement in `generis.conf.php`.

Example of .env usage:
<pre>AWS_STACK_CONFIG_ORIGIN=tao-community.docker.localhost/
AWS_STACK_CONFIG_SOURCE=ontologies/tao.rdf
AWS_STACK_CONFIG_NAMESPACE=https://${AWS_STACK_CONFIG_ORIGIN}${AWS_STACK_CONFIG_SOURCE}</pre>

Example of `seed.json` usage:
<pre>"global": {
      "lang": "en-US",
      "mode": "debug",
      "instance_name": "tao-community",
      "namespace": "$ENV{AWS_STACK_CONFIG_NAMESPACE}",
      "url": "https://tao.docker.localhost/",
      "session_name": "tao_community",
      "timezone": "Europe/Luxembourg",
      "import_data": true
    },</pre>

### DIFF
--- a/install/class.Setup.php
+++ b/install/class.Setup.php
@@ -154,6 +154,13 @@ class tao_install_Setup implements Action
         }
 
         $global = $parameters['configuration']['global'];
+
+        $markers = new ConfigurationMarkers(
+            new SerializableSecretDtoFactory(),
+            $this->getLogger()
+        );
+
+        $global = $markers->replaceMarkers($global);
         $options['module_namespace'] = $global['namespace'];
         $options['instance_name'] = $global['instance_name'];
         $options['module_url'] = $global['url'];
@@ -228,10 +235,6 @@ class tao_install_Setup implements Action
         }
 
         //@TODO use $serviceManager->getContainer(ConfigurationMarkers::class) after refactoring taoSetup to use full DI
-        $markers = new ConfigurationMarkers(
-            new SerializableSecretDtoFactory(),
-            $this->getLogger()
-        );
         $parameters = $markers->replaceMarkers($parameters);
 
         foreach ($parameters['configuration'] as $extension => $configs) {
--- a/install/init.php
+++ b/install/init.php
@@ -38,11 +38,11 @@ if (function_exists("date_default_timezone_set")) {
 
 require_once($root . 'vendor/autoload.php');
 
+new DotEnvReader();
+
 if (tao_install_utils_System::isTAOInstalled()) {
     require_once($root . 'config/generis.conf.php');
 }
-
-new DotEnvReader();
 
 // Logger service initialization.
 $loggerService = new \oat\oatbox\log\LoggerService();
--- a/install/utils/class.ConfigWriter.php
+++ b/install/utils/class.ConfigWriter.php
@@ -123,6 +123,8 @@ class tao_install_utils_ConfigWriter
                     $content = preg_replace('/(\'' . $name . '\')(.*?)$/ms', '$1, ' . $val . ');', $content);
                 } elseif (is_numeric($val)) {
                     $content = preg_replace('/(\'' . $name . '\')(.*?)$/ms', '$1, ' . $val . ');', $content);
+                } elseif ((method_exists($val, '__toPhpCode'))) {
+                    $content = preg_replace('/(\'' . $name . '\')(.*?)$/ms', '$1, ' . $val->__toPhpCode() . ');', $content);
                 }
             }
             file_put_contents($this->file, $content);


serge-salamanka-1pt:
if it works for concatenated .env strings then is a great change

serge-salamanka-1pt (approved):

install/utils/class.ConfigWriter.php
```diff
@@ -123,6 +123,9 @@ public function writeConstants(array $constants)
                     $content = preg_replace('/(\'' . $name . '\')(.*?)$/ms', '$1, ' . $val . ');', $content);
                 } elseif (is_numeric($val)) {
                     $content = preg_replace('/(\'' . $name . '\')(.*?)$/ms', '$1, ' . $val . ');', $content);
+                } elseif ((method_exists($val, '__toPhpCode'))) {
+                    $content = preg_replace('/(\'' . $name . '\')(.*?)$/ms', '$1, '
+                        . $val->__toPhpCode() . ');', $content);
```
Karol-Stelmaczonek, line 128: Do we need to have the method called manually? I understand its being called somewhere else and we need this also here yes?

install/init.php
```diff
@@ -38,12 +38,12 @@
 
 require_once($root . 'vendor/autoload.php');
 
+new DotEnvReader();
```
Karol-Stelmaczonek, line 41: Do we need this change? Its intentional?

KirylHatalski: Yes, this part allows to work with .env variables in generis.conf space

install/class.Setup.php
```diff
@@ -154,6 +154,13 @@ public function __invoke($params)
         }
 
         $global = $parameters['configuration']['global'];
+
+        $markers = new ConfigurationMarkers(
+            new SerializableSecretDtoFactory(),
+            $this->getLogger()
```
bartlomiejmarszal, line 160: Can we move this into private method?

install/init.php
bartlomiejmarszal: Can we move it to private method?

install/utils/class.ConfigWriter.php
bartlomiejmarszal: I do not understand a purpose of this. Could you explain?

bartlomiejmarszal (commented):
**Review Checklist**
- [ ] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Commits messages are meaningful
- [ ] Commits are atomic

install/init.php
KirylHatalski: There are just raw script area, not sure that is possible

install/class.Setup.php
KirylHatalski: Done

install/utils/class.ConfigWriter.php
KirylHatalski: Any time when we generate generis.conf we going thru this step.
It take list of markers and inserting it in generis.conf constants.
Insertion is different for rather types of markers.
Marker which contains value from .env are serialisated and have their own type.
So i create a condition which catch this markers and do propreate insertion of them.

Karol-Stelmaczonek (approved):

install/class.Setup.php
```diff
@@ -154,6 +154,10 @@ public function __invoke($params)
         }
 
         $global = $parameters['configuration']['global'];
+
+        $markers = self::getConfigurationMarkers();
```
hectoras, line 158: nitpick: `getConfigurationMarkers` is not a static method, so you should use `$this` here

```suggestion
        $markers = $this->getConfigurationMarkers();
```

Karol-Stelmaczonek: Good catch Hector, but probably this method should be static so better to change the declaration to static.

hectoras: The signature for [__invoke](https://www.php.net/manual/en/language.oop5.magic.php#object.invoke) is dictated by PHP itself and it should never be static. It is triggered automatically by PHP when the object is used as a function and we are surely relying on that on the install scripts. In PHP 8 this will produce [a fatal error](https://php.watch/versions/8.0/non-static-static-call-fatal-error).

Regarding `getConfigurationMarkers()`, I'd prefer keeping it as non-static as the caller is also non-static.

KirylHatalski: Changed, thx for catch

PR #3870, opened 2023-08-10
fix: made script which can create required directories, container and warmup application cache, so we can initialize applicaton without installing or updating it.

vbyndych:
## Goal
Make sure the cache directory exists before creating a service container file. This way we can ensure no error happen during later steps.

Please check the linked PR for more context but in a nutshell:
> My idea is to create one more script similar to install or update which will take care of initializing the application during the new instance start-up. I decided not to put it inside the cache warmer itself because it is not his responsibility and in case we would change the pipeline later, we would still be able to use the application cache warmer without any modifications.

## Changelog
- fix: made script which can create required directories, container and warmup application cache, so we can initialize applicaton without installing or updating it.

## How to test / use
1. Install application
2. Remove everything inside  `data` directory
3. Run the command `php index.php oat\generis\scripts\tools\ContainerCacheWarmup`

## Linked PRs:
- https://github.com/oat-sa/generis/pull/1059

### DIFF
--- a/scripts/taoInit.php
+++ b/scripts/taoInit.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2023 (original work) Open Assessment Technologies SA;
+ */
+
+require_once dirname(__FILE__) . '/../includes/raw_start.php';
+
+use oat\generis\scripts\tools;
+use oat\oatbox\cache\SetupFileCache;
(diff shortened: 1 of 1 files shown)


gabrielfs7 (approved):
Thanks @vbyndych!

I am approving as I understand this code is needed for premium pipeline specific, but my `recommendation` is different:

- We already have a command `taoUpdate.php` that does (or should do) the same thing.
- Calling `taoUpdate.php` would produce the same result + generate also ACL, menu and other caches that we also need.
- Generating these other caches mentioned above would lead us to even more speed and avoiding them to be generate on the first real customer requests. 

So, I am approving, but I would definitely just use existing `taoUpdate` command instead as now we would have one more command to do part of the work.

augustas (approved):
Looks good 👍🏻

vbyndych:
@gabrielfs7 thank you so much for your advice. We have discussed it internally within the reliability team and have decided to give this script a try. There 2 main reasons for such decision:
1. Premium workflow is a bit different from enterprise one and we use the dedicated instance to update the platform, so taoUpdate runs only on one separate instance and only there. Of course, nobody will be heart if we will add taoUpdate to every web server, but here we come to the second reason.
2. I've done some tests with the latest version of tao premium and it seems that taoUpdate does not warm up enough cache to solve the cold start issue.  For example in my local environment Items page took only 45ms to load after taoInit and 9 seconds to load after taoUpdate. The reason behind it is that taoUpdate apparently loads the bare minimum of the cache (container, menu, languages) but omits parts that cause significant load (like ACL, controller annotations, etc.). Just to give you a taste, here are the cache misses for both of the situations:
[cache_misses_init.log](https://github.com/oat-sa/tao-core/files/12323612/cache_misses_init.log)
[cache_misses_update.log](https://github.com/oat-sa/tao-core/files/12323613/cache_misses_update.log)

That's why we want to give it a try and see if this new approach can bring some benefits and potentially could solve our client's performance issues.

PR #3891, opened 2023-09-26
Remove lodash usage

taopkorczak:
Related: https://oat-sa.atlassian.net/browse/ADF-1579

Remove lodash usage

### DIFF
--- a/views/js/layout/section.js
+++ b/views/js/layout/section.js
@@ -20,13 +20,11 @@
  */
 define([
     'jquery',
-    'lodash',
     'context',
     'util/url',
     'layout/generisRouter'
 ], function(
     $,
-    _,
     context,
     url,
     generisRouter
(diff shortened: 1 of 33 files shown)


views/js/layout/section.js
```diff
@@ -364,10 +362,10 @@ define([
                 $sectionPanel,
                 section;
 
-            if (!_.isObject(data)) {
+            if (typeof data !== 'object') {
```
Nicholass, line 365: suggestion: to add an explicit check: `data !== null && typeof data === 'object'` 

`typeof null` returns `object`, in the other hand `console.log(_.isObject(null));` returns `false`. I am not 100% aware is null check is performed before `create` method is called.

Nicholass (commented):
Great job! My only concern is `null` check for `isObject` substitution

PR #3899, opened 2023-10-06
chore: change version to 2023.11

pribi:
change version to 2023.11 LTS

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -356,9 +356,9 @@ return [
     ],
     'constants' => [
         #TAO version number
-        'TAO_VERSION' => '2023.11',
+        'TAO_VERSION' => '2023.11 LTS',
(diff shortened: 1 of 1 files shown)


manifest.php
```diff
@@ -356,9 +356,9 @@
     ],
     'constants' => [
         #TAO version number
-        'TAO_VERSION' => '2023.10',
+        'TAO_VERSION' => '2023.11',
```
andreluizmachado, line 359: can you confirm if we add LTS here?

pnal (approved):

manifest.php
pribi: I added LTS.

Karol-Stelmaczonek (approved):

PR #3913, opened 2023-10-19
fix: Set literal range for UpdatedBy resource

yaraslau-kavaliou:
Queries with `UpdatedBy` property produced the wrong query to neo4j because `UpdatedBy` range was set to `User` type.
E. g. - https://github.com/oat-sa/generis/blob/feat/REL-1198/list-neo4j/test/integration/model/persistence/starsql/ClassTest.php#L71
In order to fix rdf `UpdatedBy` range has been set to literal.

How to test:
1. Launch integration tests twice. No issue have to occur. (`php vendor/bin/phpunit -c phpunit.xml generis/test/integration/model/persistence/starsql/`)


models/ontology/tao.rdf
```diff
@@ -82,13 +82,14 @@
...
       <rdfs:label xml:lang="en-US"><![CDATA[Updated]]></rdfs:label>
       <rdfs:comment xml:lang="en-US"><![CDATA[Last updated date]]></rdfs:comment>
+      <rdfs:range rdf:resource="http://www.w3.org/2000/01/rdf-schema#Literal"/>
     </rdf:Description>
 
   <rdf:Description rdf:about="http://www.tao.lu/Ontologies/TAO.rdf#UpdatedBy">
       <rdfs:subClassOf rdf:resource="http://www.tao.lu/Ontologies/TAO.rdf#TAOObject"/>
       <rdfs:label xml:lang="en-US"><![CDATA[Author]]></rdfs:label>
       <rdfs:comment xml:lang="en-US"><![CDATA[User who authored the change]]></rdfs:comment>
-      <rdfs:range rdf:resource="http://www.tao.lu/Ontologies/TAO.rdf#User"/>
```
vbyndych, line 91: This was not a mistake, it was intentional to make the `UpdatedBy` field a relationship, so it would go to the user resource who made a change. But we have an issue with either detection mechanics ([isRelationship method](https://github.com/oat-sa/generis/blob/eb99b9078aa9ef19f3d0cee98d6790a567e8966a/core/kernel/classes/class.Property.php#L399)) or creation logic (https://github.com/oat-sa/generis/blob/eb99b9078aa9ef19f3d0cee98d6790a567e8966a/core/kernel/persistence/starsql/class.Class.php#L215), which despite the relation type sets it as a plain property.

So I suggest fixing it there rather than changing RDF.

vbyndych (changes requested):
@yaraslau-kavaliou You will need a migration to sync RDFs from the file to the database. See https://github.com/oat-sa/tao-core/blob/master/migrations/Version202307101518342234_tao.php as an example.

yaraslau-kavaliou:
> @yaraslau-kavaliou You will need a migration to sync RDFs from the file to the database. See https://github.com/oat-sa/tao-core/blob/master/migrations/Version202307101518342234_tao.php as an example.

Thanks, done

vbyndych (approved):

PR #3943, opened 2023-12-19
[FIX][AUT-3418] Get Country Code Language List from the file name instead of targetLanguage attribute

andreluizmachado:
Refs: https://oat-sa.atlassian.net/browse/AUT-3418

## Currently Behavior
All the RDF translations  from other languages were added to `en-US` language
## After the fix
RDF translations of other languages should be added to the proper language context and not on `en-US` language
## Demo

https://github.com/oat-sa/tao-core/assets/11900046/82d25a7f-96cc-4c9f-bc2b-387d24c9e714


bugalot (approved):
As far as I can foresee, this should do the trick @andreluizmachado !

mike-ionut-mihai-sandu-tao (approved):
lgtm, good job

PR #3950, opened 2024-01-15
Fix/rel 1418/fix phpdoc blocking cache warmup

augustas:
Fixing a phpdoc block that prevents cache warmup.
 
Related to : https://oat-sa.atlassian.net/browse/REL-985
 
Without the fix, phpdoc parser throws an error like:
```
Unexpected token "{", expected type at offset 0 on line 1
```
 
 - Run `php index.php 'oat\generis\scripts\tools\ApplicationCacheWarmup'`
 - Observer an error message `Cache warmup failed: Unexpected token "{", expected type at offset 0 on line 1`
  
#### How to test
 
To test, run the cache warmup command and get the `TAO cache warmed up!` report.

### DIFF
--- a/actions/class.File.php
+++ b/actions/class.File.php
@@ -80,7 +80,7 @@ class tao_actions_File extends tao_actions_CommonModule
 
     /**
      * Download a resource file content
-     * @param {String} uri Uri of the resource file
+     * @param String uri Uri of the resource file
      */
     public function downloadFile()
     {
--- a/actions/class.File.php
+++ b/actions/class.File.php
@@ -80,7 +80,7 @@ class tao_actions_File extends tao_actions_CommonModule
 
     /**
      * Download a resource file content
-     * @param String uri Uri of the resource file
+     * @param String id Uri of the resource file
      */
     public function downloadFile()
     {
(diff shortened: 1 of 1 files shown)


actions/class.File.php
```diff
@@ -80,7 +80,7 @@ protected function uploadFile($postedFile, $folder)
 
     /**
      * Download a resource file content
-     * @param {String} uri Uri of the resource file
+     * @param String id Uri of the resource file
```
vbyndych, line 83: **praise:** Thank you so much for fixing it, this has become a problem during neo4j testing.

**nitpick:** Could you change String to string so that it would be consistent with other type hints?
```suggestion
     * @param string id Uri of the resource file
```

vbyndych (approved):

yaraslau-kavaliou (approved):
- [ ] New code is covered by tests
- [ ]  Tests are running successfully on my local machine
- [x]  Feature is working correctly on my local machine
- [x]  Pull request's target is not master
- [x]  Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x]  Commits messages are meaningful
- [x]  Commits are atomic

PR #3978, opened 2024-03-12
fix google tag rendering

poyuki:
# [TR-6036](https://oat-sa.atlassian.net/browse/TR-6036)

## Summary
This PR aim to address issue with rendering of GA tag in some environments

## Development impact
- fixed usage of deprecated sort tag `<? ?>` (which at least on `uat.eu.premium.lab.taocloud.org` converted into html comment)
- used `$_ENV` instead of `getenv` (usage of which [discouraged](https://github.com/vlucas/phpdotenv?tab=readme-ov-file#putenv-and-getenv) by dotenv lib documentation )

[TR-6036]: https://oat-sa.atlassian.net/browse/TR-6036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helpers/Layout.php
+++ b/helpers/Layout.php
@@ -593,8 +593,8 @@ class Layout
      */
     public static function getAnalyticsCode(): void
     {
-        $gaTag = getenv('GA_TAG');
-        $environment = getenv('NODE_ENV') === 'production' ? 'Production' : 'Internal';
+        $gaTag = $_ENV['GA_TAG'];
+        $environment = $_ENV['NODE_ENV'] === 'production' ? 'Production' : 'Internal';
 
         if ($gaTag && method_exists(self::$templateClass, 'inc')) {
             call_user_func(
--- a/helpers/Layout.php
+++ b/helpers/Layout.php
@@ -593,8 +593,8 @@ class Layout
      */
     public static function printAnalyticsCode(): void
     {
-        $gaTag = $_ENV['GA_TAG'];
-        $environment = $_ENV['NODE_ENV'] === 'production' ? 'Production' : 'Internal';
+        $gaTag = $_ENV['GA_TAG'] ?? '';
+        $environment = isset($_ENV['NODE_ENV']) && $_ENV['NODE_ENV'] === 'production' ? 'Production' : 'Internal';
 
         if ($gaTag && method_exists(self::$templateClass, 'inc')) {
             call_user_func(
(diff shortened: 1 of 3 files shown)


kilatib (approved):
LGTM  now

helpers/Layout.php
```diff
@@ -593,8 +593,8 @@ public static function getThemeStylesheet($target)
      */
     public static function getAnalyticsCode(): void
     {
-        $gaTag = getenv('GA_TAG');
-        $environment = getenv('NODE_ENV') === 'production' ? 'Production' : 'Internal';
+        $gaTag = $_ENV['GA_TAG'];
+        $environment = $_ENV['NODE_ENV'] === 'production' ? 'Production' : 'Internal';
```
wazelin, line 597: We need to check if the keys are set in $_ENV now before accessing them to avoid undefined-key warnings.

wazelin (changes requested):

views/templates/layout.tpl
```diff
@@ -54,6 +54,6 @@ $hasVersionWarning = empty($_COOKIE['versionWarning'])
 <?=Layout::renderThemeTemplate(Theme::CONTEXT_BACKOFFICE, 'footer')?>
 
 <div class="loading-bar"></div>
-<? Layout::getAnalyticsCode(); ?>
+<?= Layout::getAnalyticsCode(); ?>
```
wazelin, line 57: This function returns `void`. 
Echoing its response out is pointless I believe.
```suggestion
<?php Layout::getAnalyticsCode(); ?>
```

wazelin (changes requested):

helpers/Layout.php
poyuki: actually `$_ENV['GA_TAG']` will be null if it not define as far as I can see

views/templates/layout.tpl
poyuki: fair enough

helpers/Layout.php
wazelin: Yes. But it will also raise a PHP warning :)

wazelin (approved):
Well done, @poyuki!

Let's merge and port to the March release:
1. On top of https://github.com/oat-sa/tao-core/releases/tag/v54.2.2
2. Then as part of TAO Community
3. Then include TAO Community to https://github.com/oat-sa/taocloud-ami-luoss/blob/release-2024.03.2/composer.json#L103

PR #3996, opened 2024-04-11
chore: crowdin ar-SA locale merged to ar-arb

augustas:
https://oat-sa.atlassian.net/browse/REL-1580


progys (approved):
thought: not sure why sonar fails as seems I can only see "main" branch and cannot switch to develop, but should not be related to current change. Otherwise looks fine for me.

PR #3999, opened 2024-04-15
chore: compiled language updates from crowdin

augustas:
https://oat-sa.atlassian.net/browse/REL-1580


locales/en-GB/messages_po.js
```diff
@@ -1 +1 @@
-{"Manage test takers":"Manage test-takers","Record and manage test takers.":"Record and manage test-takers.","Test takers library":"Test-takers library"}
\ No newline at end of file
+[]
```
progys, line 1: question: is it ok that original messages were removed with empty array?

progys (approved):

locales/en-GB/messages_po.js
augustas: I believe so, these labels are from test taker extension, and it is not translated there https://github.com/oat-sa/extension-tao-testtaker/blob/v8.12.2/locales/en-GB/messages.po#L46

PR #4006, opened 2024-04-25
Feature/AUT-3626/Feature flag to enable the Solar Design

jsconan:
### Summary

Add a feature flag to enable the Solar Design to the authoring.

### Details

Feature flag for activating the solar design system: `FEATURE_FLAG_SOLAR_DESIGN_ENABLED`

The feature flag will be activated with:
```sh
php index.php 'oat\tao\scripts\tools\FeatureFlag\FeatureFlagTool' -s FEATURE_FLAG_SOLAR_DESIGN_ENABLED -v true
```

Once activated, the CSS class `solar-design` is added to the body element of the page.

### How to test

- check out the branch: `git checkout -t origin/feature/AUT-3626/feature-flag-for-solar-design`
- log in to TAO backoffice
- open the browser dev tools, and check that the body element received the CSS class `solar-design`

### DIFF
--- a/models/classes/featureFlag/FeatureFlagCheckerInterface.php
+++ b/models/classes/featureFlag/FeatureFlagCheckerInterface.php
@@ -15,7 +15,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2020 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2020-2024 (original work) Open Assessment Technologies SA;
  */
 
 declare(strict_types=1);
@@ -29,6 +29,7 @@ interface FeatureFlagCheckerInterface
     public const FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED = 'FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED';
     public const FEATURE_FLAG_STATISTIC_METADATA_IMPORT = 'FEATURE_FLAG_STATISTIC_METADATA_IMPORT';
     public const FEATURE_FLAG_PASSWORD_CHANGE_DISABLED = 'FEATURE_FLAG_PASSWORD_CHANGE_DISABLED';
+    public const FEATURE_FLAG_SOLAR_DESIGN_ENABLED = 'FEATURE_FLAG_SOLAR_DESIGN_ENABLED';
 
     public function isEnabled(string $feature): bool;
 }
--- a/models/classes/theme/SolarDesignCheckerTrait.php
+++ b/models/classes/theme/SolarDesignCheckerTrait.php
@@ -27,10 +27,7 @@ trait SolarDesignCheckerTrait
     public function isSolarDesignEnabled(): bool
     {
         $themeData = $this->getThemeData();
-        if (array_key_exists(SolarDesignCheckerInterface::SOLAR_DESIGN_ENABLED, $themeData) &&
-            $themeData[SolarDesignCheckerInterface::SOLAR_DESIGN_ENABLED]) {
-            return true;
-        }
-        return false;
+
+        return !empty($themeData[SolarDesignCheckerInterface::SOLAR_DESIGN_ENABLED]);
     }
 }
--- a/models/classes/theme/SolarDesignCheckerTrait.php
+++ b/models/classes/theme/SolarDesignCheckerTrait.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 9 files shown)


models/classes/featureFlag/FeatureFlagCheckerInterface.php
```diff
@@ -29,6 +29,7 @@ interface FeatureFlagCheckerInterface
     public const FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED = 'FEATURE_FLAG_LISTS_DEPENDENCY_ENABLED';
     public const FEATURE_FLAG_STATISTIC_METADATA_IMPORT = 'FEATURE_FLAG_STATISTIC_METADATA_IMPORT';
     public const FEATURE_FLAG_PASSWORD_CHANGE_DISABLED = 'FEATURE_FLAG_PASSWORD_CHANGE_DISABLED';
+    public const FEATURE_FLAG_SOLAR_DESIGN_ENABLED = 'FEATURE_FLAG_SOLAR_DESIGN_ENABLED';
```
gabrielfs7, line 32: Can you please add it to the README.md https://github.com/oat-sa/tao-core?tab=readme-ov-file#environment-variables Thanks!

models/classes/theme/ThemeService.php
```diff
@@ -117,4 +120,21 @@ public function removeThemeById($themeId)
 
         return true;
     }
+
+    public function isSolarDesignEnabled(): bool
+    {
+        return $this->getFeatureFlagChecker()->isEnabled(
```
gabrielfs7, line 126: Probably it was already agree, but just in case I am asking. 

Will any theme have the `isSolarDesignEnabled` or would it be a config per theme? If not per theme, but for all themes, your code is good, but if we want it per theme, maybe the config should be part of the `theme` options and enabled by default via feature flag, but possible to disable via theme config.

models/classes/featureFlag/FeatureFlagCheckerInterface.php
jsconan: Thx, I missed this one.

hectoras (approved):

models/classes/theme/ThemeService.php
jsconan: Good point. Indeed, I was hesitating between the 2 approaches. Actually, the flag will force the Solar Design no matter the theme. Which means that we may have either the regular look and feel, or the new one, without changing the main theme. With the theme option, it means we'd need to develop one theme for each use case. Ok, I admit that today, we only have one with the portal, but I'm pretty sure tomorrow we'd been asked to also have it for some other enterprise instances... And have to duplicate the theme for only setting one value would be overkill. The Solar Design will be part of the default stylesheet and activated by a single top CSS class.

gabrielfs7: I liked your suggestion. Good idea :)

jsconan: Finally, I managed to have it :) 
A theme can now require the solar design without the need for the FF

gabrielfs7: Awesome! 🏆

models/classes/featureFlag/FeatureFlagCheckerInterface.php
gabrielfs7: Thanks!

models/classes/theme/SolarDesignCheckerTrait.php
```diff
@@ -0,0 +1,36 @@
...
+trait SolarDesignCheckerTrait
+{
+    public function isSolarDesignEnabled(): bool
+    {
+        $themeData = $this->getThemeData();
+        if (array_key_exists(SolarDesignCheckerInterface::SOLAR_DESIGN_ENABLED, $themeData) &&
+            $themeData[SolarDesignCheckerInterface::SOLAR_DESIGN_ENABLED]) {
+            return true;
+        }
+        return false;
```
gabrielfs7, line 34: chore: Can be simplified:

```suggestion
        $themeData = $this->getThemeData();

        return !empty($themeData[SolarDesignCheckerInterface::SOLAR_DESIGN_ENABLED]);
```

It should not generate an error if the index does not exits. 

**Personal opinion**: I would not create this trait (or interface). I think adding a new method `isSolarDesignEnabled` in the `ConfigurableTheme` should be enough.

That being said, we could them remove this from here https://github.com/oat-sa/tao-core/pull/4006/files#diff-a92025ec366e77d37bd57274d97219fcdb7de5aec058033d10c90f7a9f277c73R124

and let the only the `ConfigurableTheme` carry the business logic to decide if `SOLAR_DESIGN_ENABLED` or not. 

I am approving the PR as this is my personal view to have less code and a single point of maintenance.

gabrielfs7 (approved):
Thanks @jsconan 

models/classes/theme/SolarDesignCheckerTrait.php
jsconan: Thank you. Indeed the condition can be simplified and the `empty()` approach is even better as it relies on the implicit handling of values (less code = less bug)

Regarding the trait, the purpose is to remove the duplicates as the code is used in multiple places.

PR #4029, opened 2024-05-29
Fix/AUT-3626/Feature flag for solar design

jsconan:
### Summary

Fix a blocking issue breaking any page load when a theme service does not inherit from `oat\tao\model\theme\ThemeService`.

### Details

To check if the Solar Design is enabled, the theme classes implement a check with respect to a feature flag. However, this check was placed in a parent class that is not inherited by all theme services.

This check is now moved to the abstract class.

### How to test
- check out the branch: `git checkout -t origin/fix/AUT-3626/feature-flag-for-solar-design`
- log in to TAO backoffice
- the page should load
- check that the Solar Design applies

Note: the issue arose with the extension taoStyles which is using another theme service (PersistenceThemeService)

### DIFF
--- a/models/classes/theme/ThemeServiceAbstract.php
+++ b/models/classes/theme/ThemeServiceAbstract.php
@@ -228,7 +228,7 @@ abstract class ThemeServiceAbstract extends ConfigurableService implements Theme
             ->isEnabled(FeatureFlagCheckerInterface::FEATURE_FLAG_TAO_AS_A_TOOL);
     }
 
-    protected function getFeatureFlagChecker(): FeatureFlagCheckerInterface
+    private function getFeatureFlagChecker(): FeatureFlagCheckerInterface
     {
         return $this->getContainer()->get(FeatureFlagChecker::class);
     }
--- a/models/classes/theme/ThemeServiceAbstract.php
+++ b/models/classes/theme/ThemeServiceAbstract.php
@@ -233,7 +233,7 @@ abstract class ThemeServiceAbstract extends ConfigurableService implements Theme
         return $this->getContainer()->get(FeatureFlagChecker::class);
     }
 
-    protected function getContainer(): ContainerInterface
+    private function getContainer(): ContainerInterface
     {
         return $this->getServiceManager()->getContainer();
     }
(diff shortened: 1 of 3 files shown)


jsconan:
close and reopen to trigger the checks

models/classes/theme/ThemeServiceAbstract.php
```diff
@@ -216,9 +224,17 @@ protected function getThemeDetailsProviders()
...
     protected function isTaoAsToolEnabled(): bool
     {
-        return $this->getServiceManager()
-            ->getContainer()
-            ->get(FeatureFlagChecker::class)
+        return $this->getFeatureFlagChecker()
             ->isEnabled(FeatureFlagCheckerInterface::FEATURE_FLAG_TAO_AS_A_TOOL);
     }
+
+    protected function getFeatureFlagChecker(): FeatureFlagCheckerInterface
```
gabrielfs7, line 231: I recommend to apply least privilege principle unless we really need to use it on child classes. Otherwise, we must never use protected, so we can reduce the surface of any breaking change.

```suggestion
    private function getFeatureFlagChecker(): FeatureFlagCheckerInterface
```

models/classes/theme/ThemeServiceAbstract.php
```diff
@@ -216,9 +224,17 @@ protected function getThemeDetailsProviders()
...
+    {
+        return $this->getContainer()->get(FeatureFlagChecker::class);
+    }
+
+    protected function getContainer(): ContainerInterface
```
gabrielfs7, line 236: Same as here https://github.com/oat-sa/tao-core/pull/4029/files#r1618503189

```suggestion
    private function getContainer(): ContainerInterface
```

gabrielfs7 (approved):
Code seems straight forward. Did not check it locally. 

A least privilege principle practice must be applied in the code, but I am pre-approving as this is the only required change IMO.

models/classes/theme/ThemeServiceAbstract.php
jsconan: Correct. And I was sure you would say this 😄 
I decided to use protected instead of private because the other similar methods were doing it, and also because this is an abstract class with method that would be consumed by children

hectoras (approved):

PR #4033, opened 2024-06-03
Fix/ADF-1664/Solar design glitches (part 2)

jsconan:
### Summary

Fix another set of UI glitches in the backoffice when the Solar Design applies and a custom theme is already in place.

### Details

A few places were not following well the design:
- the tab selector when publishing a test (select local or remote)
- the buttons inside the item form (workflow buttons)
- the bottom bar for the look & feel (nice to have, but this would be hidden in the end)

The action bars in the lists and remote lists were hidden as they were considered empty.

### How to test

- check out the branch: `git checkout -t origin/fix/ADF-1664/solar-design-glitches-part2`
- log in to TAO backoffice
- look at the pages, compare with the UX mockups


shaveko (approved):
Looks nice.

The branch name in the PR description's `How to test` should be updated

PR #4036, opened 2024-06-05
fix: use TaoLtiSession to decide theme selection

bartlomiejmarszal:
When user access to Tao 3.x from Tao Portal Tao will have back button that will lead back to Tao Portal
When user login using Tao login page we will use regular Tao Theme

https://github.com/oat-sa/tao-core/assets/16231681/76775c8b-f490-4670-8038-5e1fcf868424

https://github.com/oat-sa/tao-core/assets/16231681/0f89e58f-1ec0-4f8c-a1d9-f667cc633e78


bartlomiejmarszal:
We cannot use taoLti reference

PR #4038, opened 2024-06-05
feat: Implementation of the content creation mode switch

Nevermind23:


actions/class.UserSettings.php
```diff
@@ -109,7 +109,8 @@ public function properties()
             ];
 
             if ($this->isSolarDesignEnabled()) {
-                $userSettingsData[GenerisRdf::PROPERTY_USER_INTERFACE_MODE] = $settingsForm->getValue('interface_mode');
+                $interfaceMode = $this->getResource($settingsForm->getValue('interface_mode'));
+                $userSettingsData[GenerisRdf::PROPERTY_USER_INTERFACE_MODE] = $interfaceMode->getUri();
```
hectoras, line 113: Quick note so we don't forget: This comes from the changes in generis [here](https://github.com/oat-sa/generis/pull/1118/files#diff-bed6cd04417e5baddf907dc0ce099c73a12c3ac7a0901a7b87153d50d0d8b82bR87), so we'll need to update [the composer.json here](https://github.com/oat-sa/tao-core/blob/feature/AUT-3626/content-creation-mode-switch/composer.json#L70) (in tao-core) to point to the new generis version once the changes are released.

For the time being, you may want to update it already to point to the new Generis branch instead, and then update it again to point to the tagged version once everything passes QA.

PR #4074, opened 2024-07-10
MoveTo Action for class fixed

bartlomiejmarszal:
Error was because class does not have any types as other resources. We had to take different strategy to get parent class


tikhanovichA (approved):
LGTM
tested locally

PR #3989, opened 2024-04-01
feat: added child indexation task for tests for grained task routing

pnal:
This PR aims to add additional task class for indexation of test resources, the task will do the same but the separate class will allow us to associate indexation tasks to queues more flexible, especially we can make a separate queue for indexing only tests resources to give them more priority.

How to test steps and a demo is in related adv. search extension PR

PR #4097, opened 2024-09-11
fix: properly determine resource indexable type to route index tasks

pnal:
This fixes https://github.com/oat-sa/tao-core/pull/3989
it was unable to detect if resource is a test if it's placed in a nested class

### DIFF
--- a/models/classes/resources/ResourceWatcher.php
+++ b/models/classes/resources/ResourceWatcher.php
@@ -144,9 +144,11 @@ class ResourceWatcher extends ConfigurableService
                 return;
             }
 
-            if ($this->hasResourceSupport($resource)) {
+            $rootIndexClass = $this->getResourceIndexType($resource);
+            if ($rootIndexClass) {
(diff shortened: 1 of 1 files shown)


models/classes/resources/ResourceWatcher.php
```diff
@@ -182,7 +187,7 @@ private function hasResourceSupport(core_kernel_classes_Resource $resource): boo
             $checkedResourceTypes[] = $class->getUri();
         }
 
-        return false;
+        return null;
     }
 
     private function getResourceTypes(core_kernel_classes_Resource $resource): array
```
gabrielfs7, line 193: @pnal I think we can replace this implementation here by calling `$resource->getParentClassIds()`, which will avoid loading lots of objects in memory. Wdyt?

pnal: I'm not sure this will work faster
We did not have an issue with this

tikhanovichA (approved):

PR #4104, opened 2024-09-25
fix: archive tasks using bulk update with predefined batch

tikhanovichA:
`'php error(1) in /var/www/html/tao/tao/models/classes/taskQueue/TaskLog/Entity/TaskLogEntity.php@120: Allowed memory size of 536870912 bytes exhausted`

During archiving a lot of tasks we may have the issue on the PHP side about memory size. For fixing that we will add updating by using batch size.

How to test:

1. Create a few tasks
2. Try to archive them


tikhanovichA:
https://github.com/oat-sa/tao-core/pull/4105

PR #4124, opened 2024-11-06
fix: unique identifiers

shpran:
Fix for unique identifiers + removing unnecessary classes


models/classes/Translation/Service/ResourceMetadataRetriever.php
```diff
@@ -0,0 +1,37 @@
...
+namespace oat\tao\model\Translation\Service;
+
+use core_kernel_classes_Resource;
+use oat\tao\model\TaoOntology;
+
+class ResourceMetadataRetriever
+{
+    public function getOriginalResourceUri(core_kernel_classes_Resource $resource): ?string
+    {
+        $property = $resource->getProperty(TaoOntology::PROPERTY_TRANSLATION_ORIGINAL_RESOURCE_URI);
```
gabrielfs7, line 32: I see the value of this class, but some points: 

- [ ] It will be the first time we are creating a class to shortcut ontology getProperty...
- [ ] If we keep calling `$resource->getProperty` all the time in a loop or other places, this is actually a performance issue. We would have to cache it.

models/classes/UniqueId/Service/ResourceUniqueIdRetriever.php
```diff
@@ -0,0 +1,49 @@
...
+class ResourceUniqueIdRetriever
+{
+    private FeatureFlagCheckerInterface $featureFlagChecker;
+
+    public function __construct(FeatureFlagCheckerInterface $featureFlagChecker)
+    {
+        $this->featureFlagChecker = $featureFlagChecker;
+    }
+
+    public function retrieve(core_kernel_classes_Resource $resource): ?string
```
gabrielfs7, line 38: I see the value of this class, but some points: 

- [ ] It will be the first time we are creating a class to shortcut ontology getProperty...
- [ ] If we keep calling `$resource->getProperty` all the time in a loop or other places, this is actually a performance issue. We would have to cache it.
- [ ] If this is a uniqueId retriever, shouldn't it contain the logic to get the unique id in any case? Calling the Proxy you created?

gabrielfs7 (approved):
Pre-approved as the expected behavior is working. 

Test and styling will be fixed in the integration branch

PR #4135, opened 2024-11-20
[Fix] Translation metadata not persisted in datastore

shpran:
# [ADF-1828](https://oat-sa.atlassian.net/browse/ADF-1828)

## Changes
* Translation lists type changed from System Object to List
* Translation lists marked as non-editable in RDF
* Languages list marked as non-editable in RDF
* Editable List Specification changed to check for editable property:
  * if property does not exists in the list properties - list will be editable by default
  * if property exists in the list properties - list will be editable/non-editable depending on setted value
* Now Ld Basic Triple encoder can encode Readonly properties

## Companion PRs
* https://github.com/oat-sa/extension-tao-test/pull/486

[ADF-1828]: https://oat-sa.atlassian.net/browse/ADF-1828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


migrations/Version202411200904142234_tao.php
```diff
@@ -0,0 +1,30 @@
...
+namespace oat\tao\migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
+use oat\tao\scripts\update\OntologyUpdater;
+
+/**
+ * phpcs:disable Squiz.Classes.ValidClassName
+ */
+final class Version202411200904142234_tao extends AbstractMigration
```
gabrielfs7, line 14: - [ ] @shpran This is will not be necessary. I've already removed it in a prior PR that I am also waiting for unit03 for vacancy. Lets remove it before merge, please

PR #4152, opened 2024-12-11
fix: root class must be retrieved if root class

gabrielfs7:
# Goal

- In case class is root class, use this to filter out translations

### DIFF
--- a/models/classes/GenerisTreeFactory.php
+++ b/models/classes/GenerisTreeFactory.php
@@ -364,7 +364,7 @@ class GenerisTreeFactory
         $mainClass = array_pop($parentClassIds);
 
         return in_array(
-            $mainClass,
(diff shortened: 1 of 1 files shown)


models/classes/GenerisTreeFactory.php
```diff
@@ -364,7 +364,7 @@ private function mustFilterTranslations(core_kernel_classes_Class $class): bool
         $mainClass = array_pop($parentClassIds);
 
         return in_array(
-            $mainClass,
+            $mainClass ?? $class->getUri(),
```
shpran, line 367: We can remove lines `363-364` and use the following code here:
```suggestion
            $class->getRootId() ?? $class->getUri(),
```

shpran (approved):

PR #4158, opened 2024-12-24
fix: update original test after sync metadata

gabrielfs7:
https://oat-sa.atlassian.net/browse/ADF-1855

# Goal 

Allow callbacks to be executed for original test after sync it

# Related PRs

- https://github.com/oat-sa/extension-tao-test/pull/495


shpran (approved):
LGTM

PR #4196, opened 2025-01-29
fix: update dependency with the changed labels

shaveko:
related to https://oat-sa.atlassian.net/browse/ADF-1865

### Description 

tao-core-ui dependency update 

!!!! Update to the release version before merging


jsconan (approved):
Quick code review only

PR #4198, opened 2025-01-30
feat: add container handling to ResourceRelationServiceProxy

bartlomiejmarszal:
This change will allow using containers and configurable services in `ResourceRelationServiceProxy`


Karol-Stelmaczonek (approved):

models/classes/resources/relation/service/ResourceRelationServiceProxy.php
```diff
@@ -65,6 +67,11 @@ public function findRelations(FindAllQuery $query): ResourceRelationCollection
             }
 
             foreach ($services as $serviceId) {
+                // Instead of throwing error when service does not exist, we just skip it
+                if (!$this->serviceExist($serviceId, $type)) {
```
gabrielfs7, line 71: For better performance, rather than creating a new method, I suggest the following: 

1 - create a variable `$container = $this->getServiceManager()->getContainer()` on the line 68.
2 - on the like 70, just to the following

```php
if (!$container->has($serviceId)) {
    continue;
}
```` 

In this way, you do not need to keep calling service manager and container inside the foreach.

File: models/classes/resources/relation/service/ResourceRelationServiceProxy.php [hunk 2]
```diff
@@ -80,7 +87,31 @@ public function findRelations(FindAllQuery $query): ResourceRelationCollection
 
     private function getResourceRelationService(string $serviceId): ResourceRelationServiceInterface
     {
-        return $this->getServiceLocator()->get($serviceId);
+        try {
+            return $this->getServiceManager()->get($serviceId);
+        } catch (ServiceNotFoundException $exception) {
+            return $this->getServiceManager()->getContainer()->get($serviceId);
+        }
```
gabrielfs7, line 94: No need for it, the container does it automatically for you :)

```suggestion
        return $this->getServiceManager()->getContainer()->get($serviceId);
```

models/classes/resources/relation/service/ResourceRelationServiceProxy.php
```diff
@@ -80,7 +87,31 @@ public function findRelations(FindAllQuery $query): ResourceRelationCollection
...
+    }
+
+    private function serviceExist(string $serviceId, string $type): bool
```
gabrielfs7, line 97: You do not need this method as per better performance, you can apply this logic here https://github.com/oat-sa/tao-core/pull/4198/files#r1944649031

models/classes/resources/relation/service/ResourceRelationServiceProxy.php
```diff
@@ -80,7 +87,31 @@ public function findRelations(FindAllQuery $query): ResourceRelationCollection
...
+    {
+        if (
+            $this->getServiceManager()->has($serviceId)
+            || $this->getServiceManager()->getContainer()->has($serviceId)
+        ) {
+            return true;
+        }
+
+        common_Logger::w(
```
gabrielfs7, line 106: If you still want to log it (in the new place I mentioned), please use the proper `LoggerService` as this static call is not recommended. Thank you!

models/classes/resources/relation/service/ResourceRelationServiceProxy.php
gabrielfs7: Btw, why do we need to do this check at all?

gabrielfs7 (changes requested):
Better thinking about this, the service existing check is not required at all. Either we add or remove those services via DI or existing migrations, so the services must always exist. 

If there is a corrupted installation that left a wrong service registered, the exception will be triggered, which is better than masking the error, but this is highly unlink to happen.

gabrielfs7 (approved):

tikhanovichA (approved):

KirylHatalski (approved):

PR #4200, opened 2025-02-04
feat: extend event logs

shpran:
# [ADF-1851](https://oat-sa.atlassian.net/browse/ADF-1851)

## Changes
* Added new event to cover translation actions:
  * Create
  * Update
  * Delete
* New event is triggered in related services to create/update/delete a translation
* New event will share the following information:
  * Action (create/update/delete)
  * Original resource URI
  * Translation resource URI
  * Locale code
  * Additional data, e.g. changed properties (old -> new)

[ADF-1851]: https://oat-sa.atlassian.net/browse/ADF-1851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/models/classes/Translation/Service/TranslationUpdateService.php
+++ b/models/classes/Translation/Service/TranslationUpdateService.php
@@ -24,8 +24,10 @@ namespace oat\tao\model\Translation\Service;
 
 use core_kernel_classes_Resource;
 use oat\generis\model\data\Ontology;
+use oat\oatbox\event\EventManager;
 use oat\tao\model\TaoOntology;
 use oat\tao\model\Translation\Command\UpdateTranslationCommand;
+use oat\tao\model\Translation\Event\TranslationActionEvent;
 use oat\tao\model\Translation\Exception\ResourceTranslationException;
 use Psr\Log\LoggerInterface;
 use Throwable;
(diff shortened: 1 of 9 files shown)


Nevermind23 (approved):

models/classes/Translation/Service/TranslationUpdateService.php
```diff
@@ -73,13 +78,30 @@ public function update(UpdateTranslationCommand $command): core_kernel_classes_R
...
+                TranslationActionEvent::ACTION_UPDATED,
+                $this->getOriginalResourceUri($instance),
+                $instance->getUri(),
+                $this->getLocaleCode($instance),
+                [
+                    $translationProgressProperty->getUri() => [
+                        'old' => $oldProgressUri,
+                        'new' => $command->getProgressUri(),
+                    ],
+                ]
```
gabrielfs7, line 102: On the current audit logs, there is no concept of `old` and `new`. This is something new we are adding now. I am concerned about consistent. Wdyt?

shpran: Why we cannot do it?
It's just for better visibillity on what was changed and it looks like this
image

PR #4208, opened 2025-02-17
feature: INF-283 integration

viktar-dzmitryieu-tao:
Ticket: https://oat-sa.atlassian.net/browse/INF-283

## What's Changed

- Updated the @oat-sa/tao-core-shared-libs with the new bundle of CKEditor which handles empty ruby tags when user loses focus on editor


tikhanovichA (approved):

Karol-Stelmaczonek:
Not needed

PR #4219, opened 2025-03-07
fix: reduce id collision

gabrielfs7:
# Goal 

Reduce ID collision

Going with 9 letters it has a 5.5 trilling uniqueness and a ~0.0092%  probability of collision for 1M records.

### DIFF
--- a/models/classes/IdentifierGenerator/Generator/NumericIdentifierGenerator.php
+++ b/models/classes/IdentifierGenerator/Generator/NumericIdentifierGenerator.php
@@ -26,11 +26,16 @@ class NumericIdentifierGenerator implements IdentifierGeneratorInterface
 {
     /**
      * This will return 9 digits numeric identifier base on time and random number
-     * i.e: 123456789
+     * i.e: ABCDEFGHI
      */
     public function generate(array $options = []): string
     {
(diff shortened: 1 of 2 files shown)


models/classes/IdentifierGenerator/Generator/NumericIdentifierGenerator.php
```diff
@@ -26,11 +26,16 @@ class NumericIdentifierGenerator implements IdentifierGeneratorInterface
 {
     /**
      * This will return 9 digits numeric identifier base on time and random number
-     * i.e: 123456789
+     * i.e: ABCDEFGHI
      */
     public function generate(array $options = []): string
```
shpran, line 31: This generator name is `NumericIdentifierGenerator`, so I think it's better to create new one and make possible to switch between them by some FF or configuration change.
Or rename this from `NumericIdentifierGenerator` to `UniqueIdentifierGenerator`.

gabrielfs7: Makes sense to create a new one indeed and update the installation / migrations scripts to use the new one

gabrielfs7:
Closed in favour of https://github.com/oat-sa/tao-core/pull/4277

PR #4235, opened 2025-04-16
fix: move translations together with a resource

pnal:
This aims to keep a test/item translations (hidden resources) together with origin resource when moving it to another class. This fixes the following case:
- Create item/test
- Create a translation
- Move the origin item/test to another class
- Remove the first class (it looks empty but actually it is not due to the translation implementation)
- Check that the translations for the moved items/resource is not lost anymore

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -1,8 +1,6 @@
 name: Continuous integration
 
 on:
-  push:
-    branches: [ develop ]
   pull_request:
     branches: [ develop ]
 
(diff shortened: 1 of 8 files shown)


tikhanovichA (approved):

.github/workflows/continuous-integration.yaml
```diff
@@ -2,7 +2,7 @@ name: Continuous integration
 
 on:
   push:
-    branches: [ master, develop ]
+    branches: [ develop ]
```
tikhanovichA, line 5: Let's delete this. We don't need to run it if push to develop

PR #4426, opened 2026-05-05
Feature/TR-6832/Grunt babel pretransform

oatymart:
https://oat-sa.atlassian.net/browse/TR-6832

### Part A - bundling

Installs https://github.com/oat-sa/grunt-tao-bundle/pull/75 and configures the babel pre-transform.

Added some artificial ES2018-ES2021 modern syntax examples to the JS code: optional chaining, nullish coalescing, logical assignment, try-catch without error param.

TODO:
- [x] revert those 4 examples after the review

To be validated on an environment with `DEBUG_MODE: false` set up, so the JS bundles are loaded.

Key commands:
`npx grunt taobundle`
`npx grunt [lowercaseextension]bundle`

---

### Part B - unit tests

To have qunit / Puppeteer understand the modern syntax (without using babel), I upgraded to [grunt-contrib-qunit 9](https://github.com/gruntjs/grunt-contrib-qunit/blob/main/CHANGELOG.md#v911) (Puppeteer 22.15.0). With AI, I adapted the chrome-bridge file which is injected to grunt-contrib-qunit, because it used an outdated API and the tests were hanging in the new Puppeteer.

Now, all tests are passing across 14 extensions I've tested (with very few minor fixes).

Key commands:
`npx grunt connect:test taotest`
`npx grunt connect:test:keepalive`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed QUnit test execution timing in asynchronous test scenarios.

* **Chores**
  * Updated build toolchain dependencies for improved stability.
  * Enhanced test infrastructure with improved QUnit event handling and error serialization.
  * Applied code quality improvements across helper functions and error handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/oat-sa/tao-core/pull/4426?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #4452, opened 2026-06-12
Fix/TR-6832/Remove modern syntax example

oatymart:
https://oat-sa.atlassian.net/browse/TR-6832

Follow-up to: https://github.com/oat-sa/tao-core/pull/4426

A tiny change was made last time in `features.js` to demonstrate bundling **extensions** with modern JS syntax in the code. The unfortunate side effect is it breaks `portableelement`. So it has to be reverted to avoid a bigger issue.

The `portableelement` [task definition](https://github.com/oat-sa/extension-tao-itemqti/blob/develop/views/build/grunt/portableelement.js#L44), which bypasses grunt-tao-bundle, will be modified later to make it compatible. This will happen in https://oat-sa.atlassian.net/browse/TR-6854.

**To test:** run `npx grunt portableelement -e=qtiItemPci -i=mathEntryInteraction -type=IMSPCI`

## Summary by CodeRabbit

* **Refactor**
  * Internal code improvements for better maintainability and modern syntax standards.

---

**Note:** This release contains no user-facing changes. Updates are focused on code quality improvements.


views/js/services/features.js
```diff
@@ -41,7 +41,7 @@ define(['module', 'core/logger'], function (module, loggerFactory) {
 
         try {
             return new RegExp(`^${lookupPath}$`);
-        } catch {
+        } catch (e) {
```
kilatib, line 44: If we add that variable, should we reuse it?

oatymart: This is just reverting to previous code state to avoid a new problem. The `e` was never used in the past.

kilatib (approved):
LGTM

PR #4502, opened 2026-08-11
fix: cleaning

tikhanovichA:
## What's Changed

- Updated class.TaoTranslate.php (what taoTranslate.php runs):
- Fixed the $de-AT warning — '${$language}' was PHP variable-variable syntax; it’s now '{$language}'.  Missing PO no longer aborts — err(..., true) became a non-fatal err(...) plus continue, so compileAll warns and moves on to the next locale.

Also clean extra tao folder after bad releasing


scripts/class.TaoTranslate.php
```diff
@@ -1502,10 +1502,11 @@ public function actionCompile()
                         "JavaScript compiled translations for extension '{$extension}' with '{$language}' written."
                     );
                 } else {
+                    // ponytail: missing locale dirs (e.g. empty/stub) — warn and skip so compileAll keeps going
```
bartlomiejmarszal, line 1505: Why is this here?

PR #4513, opened 2026-08-21
fix: add pluralization support

jebersole:
## Ticket:
https://oat-sa.atlassian.net/browse/AUT-4679

## What's Changed
Adds pluralization support to backoffice. This is to be used with the PR below to support Slovak translations.
image
image

## Dependencies PRs
- https://github.com/oat-sa/extension-tao-item/pull/773

## How it Works
For a Handlebars template call like:
{{__p "%d test" "%d tests" number}}
the flow now is:
1. Source code contains the plural marker
   - Example: [relatedTestsPopup.tpl (line 3)](/Users/je/oat/extension-tao-item/views/js/controller/items/tpl/relatedTestsPopup.tpl:3)
   - __p is one of the plural markers the extractor looks for.
2. taoTranslate.php -a update scans source files
   - This uses [class.SourceCodeExtractor.php (line 1)](/Users/je/oat/tao-core/helpers/translation/class.SourceCodeExtractor.php:1).
   - It finds __p(...), __.p(...), and template plural usages, then updates the extension PO files, for example:
     - extension-tao-item/locales/en-US/messages.po
     - extension-tao-item/locales/sk-SK/messages.po
   - In the PO file, that becomes a plural entry like:msgid "%d test"
     msgid_plural "%d tests"
     msgstr[0] "..."
     msgstr[1] "..."
3. PO files are converted into per-extension locale data
   - TAO writes messages_po.js per extension/locale.
   - That file contains:
     - normal translations
     - plural translations
     - the Plural-Forms header string
4. TranslationBundle.php merges extension translations into frontend bundles
   - This uses [TranslationBundle.php (line 1)](/Users/je/oat/tao-core/helpers/translation/TranslationBundle.php:1).
   - It reads all relevant messages_po.js files and produces the frontend locale bundle, especially:
     - messages.json
   - messages.json now contains:
     - translations
     - pluralForms
     - Example:
               ```
              {
                "translations": {
                  "Manage test takers": "Manage test-takers",
                  "%d test": {
                    "0": "%d test",
                    "1": "%d testy",
                    "2": "%d testov"
                  }
                },
                "pluralForms": "nplurals=3; plural=(n==1) ? 0 : (n>=2 && n<=4) ? 1 : 2;"
              }
              ```
5. The browser loads the generated JSON bundle
   - [views/js/i18n.js (line 1)](/Users/je/oat/tao-core/views/js/i18n.js:1) depends on json!i18ntr/messages.json.
   - So the browser receives:
     - translation strings
     - the raw pluralForms string from the PO header
6. __.p() builds and uses the plural rule at runtime
   - In [views/js/i18n.js (line 38)](/Users/je/oat/tao-core/views/js/i18n.js:38), __.p(...) calls pluralHelper.getPluralRule(i18nTr.pluralForms).
   - In [views/js/i18n/plural.js (line 63)](/Users/je/oat/tao-core/views/js/i18n/plural.js:63), createPluralRule(...) parses the Plural-Forms string and creates a runtime function named pluralRule.
   - Then __.p(...):
     - takes singular, plural, and count
     - runs pluralRule(count) to choose the plural index
     - looks up the translated plural form in translations
     - formats %d with the count
7. You see the translated text in the popup
   - So for Slovak, counts like 1, 3, 5 can resolve to different plural slots based on the PO rule.
   - That final resolved string is what appears in the backoffice UI.

## How To Test
After changing source strings/templates, the flow is:
1. Run taoTranslate.php -a update for the affected locale(s)
2. Review or adjust the PO translations if needed
3. Run GenerateTranslationBundles.php
4. Hard refresh the browser
5. Trigger the popup and verify the pluralized text changes with the count, for example a 7 inych

Commands:
docker exec nextgen_tao php tao/scripts/taoTranslate.php -a update -e taoItems -l en-US -r /var/www/html
docker exec nextgen_tao php tao/scripts/taoTranslate.php -a update -e taoItems -l sk-SK -r /var/www/html
docker exec nextgen_tao php tao/scripts/tools/GenerateTranslationBundles.php -v

## Summary by CodeRabbit

* **New Features**
  * Added plural translation support across PO, JSON, and JavaScript formats.
  * Added plural-aware JavaScript helpers with indexed variants, fallbacks, formatting, and Ruby-tag support.
  * Generated bundles now support JSON and AMD-compatible JavaScript output.
  * Improved extraction and parsing of singular, plural, contextual, and multiline translations.

* **Bug Fixes**
  * Preserved translation headers, contexts, metadata, and existing plural targets.
  * Improved handling of missing, sparse, legacy, and invalid translation data.
  * Prevented incomplete bundles when generation or file writing fails.

* **Tests**
  * Expanded coverage for plural translations, formatting, extraction, and bundle generation.


jebersole:
@coderabbitai plz approve if your comments have been addressed - @coderabbitai approve

views/js/i18n.js
```diff
@@ -1,8 +1,75 @@
...
+        }
+
+        if (translationsMap && typeof translationsMap === 'object') {
+            if (hasOwnTranslationIndex(translationsMap, pluralIndex)) {
+                return translationsMap[pluralIndex];
+            }
+
+            if (hasOwnTranslationIndex(translationsMap, 0)) {
+                return translationsMap[0];
+            }
```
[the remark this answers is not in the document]
jebersole: @coderabbitai verify this is now addressed, if so, @coderabbitai approve

helpers/translation/TranslationBundle.php
```diff
@@ -38,6 +40,14 @@
...
 class TranslationBundle
 {
+    private const JSON_BUNDLE_FILE = 'messages.json';
+    private const AMD_BUNDLE_FILE = 'messages.js';
+    private const DEFAULT_PLURAL_RULE = 'function (n) { return n === 1 ? 0 : 1; }';
+    private const GENERATED_PLURAL_RULE =
+        'function (n) { var index = Number(%s); if (!isFinite(index)) { return 0; } '
+        . 'index = Math.floor(index); if (index < 0) { return 0; } '
+        . 'if (index > %d) { return %d; } return index; }';
+
```
oatymart, line 42: I have some reservations about this approach - I don't get why we have to generate this JS code in PHP and serve it to `i18n.js` who will consume it.

In NextGen, the pluralisation was implemented entirely in the FE helper, and (I think) we are able to handle complex plural rules as seen [here](https://github.com/oat-sa/live-design-system/blob/develop/packages/core/i18n/test/language.spec.js#L137-L161).

But NextGen doesn't use any complex BE strings. Is this file needed to support pluralisations in the BE templates?

views/js/test/i18n/messages.json
```diff
@@ -2,11 +2,27 @@
...
+        "3": "%d many-days"
+      }
+    },
+    "%d ruby day": {
+      "_plural": "%d ruby days",
+      "_translations": {
+        "0": "{ruby}%d{/ruby} den",
+        "1": "{ruby}%d{/ruby} dni"
+      }
+    }
```
oatymart, line 20: Is the added structure based on some existing standard? Just to know if this has been explored.

helpers/translation/TranslationBundle.php
jebersole: @oatymart Yeah, but tao-core depends heavily on PHP parsing of .po files, and the FE is not the sole owner of it like in portal or deliver-fe. So there needs to be some kind of bridge between the two, and that's part of what this file is doing. I can see that portal can handle it in files like nextgen-stack/tao-portal/frontend/src/services/localeService.js. But tao-core doesn't do this and first handles it on the backend before sending it to the front. These js strings are a way to mimick portal's implementation, and are used to build the p11n rules function that gets embedded into the generated AMD locale bundle.

views/js/test/i18n/messages.json
jebersole: This ruby stuff is existing, e.g. tao-core/test/unit/models/classes/i18n/DefaultTranslationBundleProcessorTest.php - so the above is another test file in this vein.

oatymart: Thanks, I'm not concerned about the message keys or whether it is a test or not.
I meant the part that's inside with `_plural` and `_translations` - is it your invention or is it based on a standard?

views/js/i18n.js
```diff
@@ -40,21 +107,32 @@ define(['json!i18ntr/messages.json', 'core/format'], function(i18nTr, format){
...
-        if (arguments.length > 1) {
-            localized = format.apply(null, [localized].concat([].slice.call(arguments, 1)));
+        if (args.length > 0) {
+            localized = format(localized, ...args);
         }
 
         return convertRubyTags(localized);
     }
 
+    __.p = function __p(singular, plural, count, ...args){
```
wazelin, line 38: Cannot we extend the original `__` function with optional arguments?

views/js/test/i18n/messages.json
jebersole: Ah, yes, they are part of this PR so as to write changes into `messages_po.js`

views/js/i18n.js
jebersole: @wazelin do you mean import portal's version?

helpers/translation/class.POFile.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
Karol-Stelmaczonek, line 3: Please don't add random strict_types=1 to legacy code it can blow up with multiple types mismatch. Please remove it from other files too.

views/js/i18n.js
Karol-Stelmaczonek: Its problematic to have 2 separate translation functions. I also prefer adding argument for pluralization in `__` . Also you probably need to register new helper [here](https://github.com/oat-sa/tao-core-libs-fe/blob/master/src/handlebars/helpers.js) please check it.

helpers/translation/TranslationBundle.php
Karol-Stelmaczonek: I will agree with Martin on this the approach to generate JS via PHP is problematic at least. Why we can just emit JSON  (translations + plural forms) so the FE could consume it. We probably even use [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules). We only need a small category to gettext-index map because PO files are numeric. That lets us drop `p11nRules / messages.js` and keep `json!i18ntr/messages.json`.

Karol-Stelmaczonek:
Currently the helpers for translations doesn't have own unit test but I think its good idea to add those besides the new integration test. Even small one with cover for new function would be cool.

views/js/i18n.js
jebersole: @wazelin @Karol-Stelmaczonek Are you set on just using __()? It actually adds a ton of extra complexity, with loads of parsing.This method is already used with variable length params, including the number matching portal's __.p(). This would make it more bug-prone. It also would be a different system than portal's.

wazelin: > Its problematic to have 2 separate translation functions. I also prefer adding argument for pluralization in `__` . Also you probably need to register new helper [here](https://github.com/oat-sa/tao-core-libs-fe/blob/master/src/handlebars/helpers.js) please check it.

The helper point is valid, as we'll need this function to be available in handlebar templates too.
As for the second function, it seems we're already applying such a pattern [in here](https://github.com/oat-sa/live-design-system/blob/7ecf1281cb3daf688a6c18a25799f031dba1eb99/packages/core/i18n/language.js#L75-L89), so I'm OK with having two i18n functions.

helpers/translation/class.POFile.php
```diff
@@ -169,11 +171,20 @@ public function addTranslationUnit(tao_helpers_translation_TranslationUnit $tran
...
+            if ($this->isSameTranslationUnitIdentity($tu, $translationUnit)) {
+                if (
+                    $tu instanceof tao_helpers_translation_POTranslationUnit
+                    && $translationUnit instanceof tao_helpers_translation_POTranslationUnit
+                ) {
+                    $tu->setSourcePlural($translationUnit->getSourcePlural());
+                    if ($translationUnit->hasPluralTargets()) {
+                        $tu->setTargets($translationUnit->getTargets());
+                    } else {
+                        $tu->setTarget($translationUnit->getTarget());
```
oatymart, line 183: Should this `setTarget` call not also be guarded by the `!== ''` condition, like the lines below? Coderabbit thinks that you'll allow an empty translation to replace a filled one, if the msgid matches (for example across 2 extensions).

I don't know about `setTargets` because that value will be more complex to check for emptiness. Do what you think best there.

helpers/translation/class.POFileReader.php
```diff
@@ -62,26 +62,50 @@ public function read()
...
-            '(msgctxt\s+(?:"(?:[^"]|\\\\")*?"\s*)+)?' .
-            '(msgid\s+(?:"(?:[^"]|\\\\")*?"\s*)+)\s+' .
-            '(msgstr\s+(?:"(?:[^"]|\\\\")*?(?<!\\\)"\s*)+)/',
-            $fc,
-            $matches
-        );
-
-        preg_match('/sourceLanguage: (.*?)\\n/s', $fc, $sourceLanguage);
-        preg_match('/targetLanguage: (.*?)\\n/s', $fc, $targetLanguage);
+        $entries = preg_split("/(?:\r?\n){2,}/", trim($fc));
```
oatymart, line 65: Is there a guarantee that PO files contain a blank line between entries? I didn't notice any in TAO without it, but I feel like I've seen them in the wild. I feel like this is pretty low-risk, and we would fix it later by correcting the PO file, if this change in parsing strategy breaks something. (This is based on the concern that PO files are sometimes edited by external tooling.)

oatymart (commented):
It looks a lot better, and I've tested it as working. Still have a couple of minor questions, above.

Also as a general comment, when the refactoring is quite heavy like this one, it would be nice to add better documentation comments to each new function, describing what they do or how the tricky parts work.

helpers/translation/class.POFileReader.php
jebersole: yeah, I think it's safe to say they're newline-separated entries. If that changes, a lot of things would break, not just this

jebersole: ah, I see what you mean. Yes, there are such files, but not in this repo. These po files are written by tao-core/helpers/translation/class.POFileWriter.php, which explicitly adds those blank lines. It doesn't seem the PO spec requires them.

views/js/lib/handlebars/helpers.js
wazelin, line 1: Do we actually need this separate helper implementation?
Cannot we use https://github.com/oat-sa/tao-core-libs-fe/blob/master/src/handlebars/helpers.js?

views/templates/client_config.tpl
```diff
@@ -75,7 +75,7 @@ require.config({
     <?php foreach (get_data('extensionsAliases') as $name => $path) :?>
         '<?=$name?>'        : '<?=$path?>',
     <?php endforeach?>
-        'lib/handlebars/helpers' : '../node_modules/@oat-sa/tao-core-libs/dist/handlebars/helpers',
+        'lib/handlebars/helpers' : 'lib/handlebars/helpers',
```
oatymart: If this path is updated, maybe the one directly below it should follow the pattern.

jebersole:
This is deployed on the unit1 tenant1 content bank: https://construct1-oat-unit01.dev.gcp-eu.taocloud.org/

views/templates/client_config.tpl
wazelin: @jebersole, is the plan to restore this helpers path once the tao-core-libs gets released?

wazelin (approved):
Pre-approving on a condition that the handlebars template helpers will be restored, and use the ones provided by a library.

Well done!

views/templates/client_config.tpl
jebersole: @wazelin Yes, the handlebars files will be restored and we'll use the libs-fe import.

oatymart (approved):
Ok from my perspective, but please check all remaining comments.

wazelin (approved):
Looks ready to be merged to me.

wazelin:
@coderabbitai resolve

PR #4537, opened 2026-09-10
fix: restore Crowdin locale language for RDF PO sync after pluralization

tikhanovichA:
## Ticket:
1. POFileReader  has flipped the order of steps
Before: set targetLanguage on the file first, then addTranslationUnit()
After: addTranslationUnit() first, then set targetLanguage on the file.
addTranslationUnit() copies the file language onto each unit at add time

2. $triple->lg = $tu->getTargetLanguage(); this value is always set to en-US
Crowdin puts the translated text in the right folder (locales/hu-HU/…), but the PO metadata still says targetLanguage: en-US.

TAO already knew that. Old POFileReader had a workaround.
Now its:

1. Crowdin sets fake targetLanguage: en-US.
2. Env update runs OntologyUpdater -> RdfPack reads those POs.
3. After [PR #4513](https://github.com/oat-sa/tao-core/pull/4513), units are created before the folder-name fix runs -> each unit keeps default en-US.
4. RdfPack stores $tu->getTargetLanguage() -> DB gets Hungarian/Japanese/Chinese strings tagged as en-US.

## What's Changed
- Regression from #4513 / AUT-4679
- Crowdin `targetLanguage: en-US` + unit-before-language order
- Code fix + migration cleanup + sync
- Irreversible migration `down`

Before:
image

After:
`[notice] Deleted 4831 polluted en-US RDF literals and synced models
[warning] Migration oat\tao\migrations\Version202609101019002234_tao was executed but did not result in any SQL statements.
[notice] finished in 4328.1ms, used 44M memory, 1 migrations executed, 0 sql queries`
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PO translation imports to correctly apply languages from file headers and locale folders, including plural entries.
  * Ensured RDF translation data uses the configured language consistently.
  * Corrected existing RDF labels and comments that were incorrectly tagged as `en-US`.

* **Data Maintenance**
  * Added a database migration to clean up affected translation metadata in batches.

### DIFF
--- a/migrations/Version202609101019002234_tao.php
+++ b/migrations/Version202609101019002234_tao.php
@@ -1,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace oat\tao\migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use oat\generis\model\OntologyRdfs;
+use oat\oatbox\reporting\Report;
+use oat\tao\scripts\tools\migrations\AbstractMigration;
+use oat\tao\scripts\update\OntologyUpdater;
+
+/**
+ * Fix Crowdin RDF translations wrongly stored as en-US after pluralization PO reader regression.
+ *
+ * phpcs:disable Squiz.Classes.ValidClassName
+ */
+final class Version202609101019002234_tao extends AbstractMigration
+{
+    private const DELETE_CHUNK_SIZE = 500;
+
+    public function getDescription(): string
+    {
+        return 'Remove RDF label/comment triples wrongly tagged as en-US and re-sync ontology locales'
+            . ' (irreversible down)';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $deleted = $this->deleteWronglyTaggedEnUsRdfLiterals();
+        OntologyUpdater::syncModels();
+
+        $this->addReport(Report::createSuccess(
+            sprintf('Deleted %d polluted en-US RDF literals and synced models', $deleted)
+        ));
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addReport(Report::createInfo(
+            'Irreversible: polluted en-US RDF literals were deleted; re-sync after code fix if needed'
+        ));
+    }
+
+    private function deleteWronglyTaggedEnUsRdfLiterals(): int
+    {
+        $rows = $this->connection->fetchAllAssociative(
+            'SELECT id, subject, predicate, object, l_language
+             FROM statements
+             WHERE predicate IN (?, ?)
+               AND (l_language = ? OR l_language IS NULL OR l_language = ?)',
+            [
+                OntologyRdfs::RDFS_LABEL,
+                OntologyRdfs::RDFS_COMMENT,
+                'en-US',
+                '',
+            ]
+        );
+
+        $baselines = [];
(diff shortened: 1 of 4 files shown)


migrations/Version202609101019002234_tao.php
```diff
@@ -0,0 +1,95 @@
...
+            if ((string) ($row['l_language'] ?? '') !== 'en-US') {
+                continue;
+            }
+
+            $object = (string) $row['object'];
+            $key = $row['subject'] . "\0" . $row['predicate'];
+            $hasNonAscii = (bool) preg_match('/[^\x00-\x7F]/u', $object);
+            $differsFromBaseline = isset($baselines[$key]) && $baselines[$key] !== $object;
+
+            if ($hasNonAscii || $differsFromBaseline) {
```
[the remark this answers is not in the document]
tikhanovichA: We're keeping the current heuristic and an irreversible down() on purpose.

Why not exact locale-file matching
Pollution is wrong en-US RDF rdfs:label / rdfs:comment triples left by Crowdin POs that claimed targetLanguage: en-US while living under other locale folders. The reliable signals in the DB are: non-ASCII object text under en-US, or en-US text that differs from the empty-lang baseline for the same subject+predicate. Scanning installed locale PO files at migration time would be slower, environment-dependent (which extensions/locales are present), and easy to get out of sync with what was actually imported earlier. Delete + OntologyUpdater::syncModels() reimports from the fixed reader instead.

Why not snapshot + restore in down()
Restoring the deleted rows would put the bad en-US pollution back. The safe rollback path is: keep the PO reader / RdfPack fix and re-run ontology sync if needed—not reinsert wrongly tagged literals. So down() reports irreversible by design.

Deletion criteria (unchanged)
For rdfs:label / rdfs:comment only, delete when l_language = 'en-US' and either:

object contains non-ASCII, or
an empty-lang baseline exists for the same subject+predicate with a different object.
Paging by id is only a memory/perf change; behavior of what gets deleted stays the same.

kilatib (approved):
- [x] Changelog is updated according to changes (if applicable)

KirylHatalski (approved):

jebersole (approved):
Looks good! The root item is called Item now.

migrations/Version202609101019002234_tao.php
```diff
@@ -0,0 +1,136 @@
...
+                }
+            }
+
+            if ($idsToDelete === []) {
+                continue;
+            }
+
+            $placeholders = implode(',', array_fill(0, count($idsToDelete), '?'));
+            $this->connection->executeStatement(
+                "DELETE FROM statements WHERE id IN ({$placeholders})",
```
wazelin, line 105: _nitpick_
feel free to ignore, as these queries are anyway wrapped into a transaction by the migration mechanism.

This set of queries can be made into a single delete query instead if we collect all $idsToDelete, and then execute this query after the while loop.

Karol-Stelmaczonek (approved):

migrations/Version202609101019002234_tao.php
tikhanovichA: Thanks
I’d rather keep the deletes inside the loop though. We’re paging by id so we don’t load everything at once, and collecting all the ids first would kind of undo that.

PR #4540, opened 2026-09-10
fix: NYSED-40 load DotEnv values into getenv

bartlomiejmarszal:
## Summary

- update `DotEnvReader` to use `Dotenv::usePutenv()` so loaded variables are available via both `$_ENV` and `getenv()`
- keep constructor typed (`string $envFile = ''`) and add strict types declaration
- add unit test `DotEnvReaderTest` covering `$_ENV` + `getenv()` population from a temporary `.env` file

## Why

Runtime components that read environment variables through `getenv()` were not seeing values loaded by `DotEnvReader`.

## Validation

- added focused PHPUnit unit test: `test/unit/model/mvc/DotEnvReaderTest.php`


bartlomiejmarszal:
Scope clarification: this PR is intended to cover only DotEnvReader behavior.\n\nIntended files for review:\n- models/classes/mvc/DotEnvReader.php\n- test/unit/model/mvc/DotEnvReaderTest.php\n\nAny media/DirectorySearchQuery context is unrelated to this change scope.

PR #4549, opened 2026-09-16
fix: INV-1278 route class deletes correctly for resource-selector tree

tikhanovichA:
## Summary
- With `treeProvider = resource-selector`, deleting a class folder sent `uri` (class URI) and hit `deleteResource` instead of `deleteClass`.
- That path fails for subclasses (no `rdf:type`) and left the tree stale after a successful delete.
- `removeNode` now sends only `classUri` for classes; `RdfController::delete()` also routes a class `uri` to `deleteClass`.
- Classic jstree keeps working (class context already uses `classUri` only).
## Depends on
- Companion fix in **generis** https://github.com/oat-sa/generis/pull/1183 (guard empty types in `ResourceRepository`) recommended as a safety net.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resource deletion handling for items with and without a specific URI.
  - Synchronized the resource tree and class selector after deletions.
  - Cleared deleted default selections automatically.
  - Returned users to the root view when the currently open class folder was removed.
  - Improved refreshing and removal behavior following resource deletion events.
  - Preserved resource and tree context during selection and deletion operations.


actions/class.RdfController.php
```diff
@@ -1203,27 +1203,22 @@ public function delete()
             throw new common_exception_BadRequest('wrong request mode');
         }
 
+        // resource-selector sends uri for classes too; prefer deleteClass when uri is a class
```
Karol-Stelmaczonek: Why resource selector sends uri for classes too? This should be mitigated in resource selector so it will send classUri only for classes.

views/js/layout/actions/common.js
```diff
@@ -268,12 +268,22 @@ define([
...
+                id: actionContext.id,
+                signature: actionContext.signature
+            };
 
-            data.uri = uri.decode(actionContext.uri);
-            data.classUri = uri.decode(actionContext.classUri);
-            data.id = actionContext.id;
-            data.signature = actionContext.signature;
+            if (actionContext.type === 'class') {
+                data.classUri = uri.decode(actionContext.classUri || actionContext.uri);
```
Karol-Stelmaczonek: This blurs what true value sits in `uri` and will generate future problems.

Karol-Stelmaczonek (changes requested):

views/js/layout/actions/common.js
tikhanovichA: Sorry, what do you mean?

Karol-Stelmaczonek: Youre putting uri and uriClass in classUri. How to differente those values later if I want to?

Karol-Stelmaczonek (approved):

